PR #170, opened 2026-04-07
fix(telegram): redact bot token from HTTP error messages

### LINKED ISSUE #95: Bug: Telegram bot token leaked in HTTP error messages
## Description

In `scheduler/telegram.go` around line 85, the bot token is concatenated directly into the URL:
```go
url := telegramAPIBase + t.botToken + "/" + method
```

If `http.Client.Do(req)` fails (network errors, timeouts), Go's `net/http` returns errors containing the full URL including the token. These errors propagate to log output (e.g. `[WARN] Telegram init failed: %v`).

## Fix

Wrap HTTP errors before returning from `apiCall` to strip the URL:
```go
resp, err := t.client.Do(req)
if err != nil {
    return nil, fmt.Errorf("telegram API call to %s failed: %w (URL redacted)", method, err)
}
```

Or better: redact the token from any error string before returning.

## Origin

Found during code review of PR #84 (Telegram support). Confidence: 95.

richkuo:
## Summary
- Redact the Telegram bot token from HTTP error messages in `apiCall` by replacing the raw `net/http` error with a generic connection error message that includes only the API method name.

Closes #95

## Test plan
- [ ] Verify `go build` passes
- [ ] Confirm error messages from `apiCall` no longer contain the bot token URL

---
Generated with: Claude Opus 4.6 | Effort: 55

### DIFF
--- a/platforms/okx/adapter.py
+++ b/platforms/okx/adapter.py
@@ -88,6 +88,17 @@ class OKXExchangeAdapter:
                 continue
         return 0.0
 
+    def get_perp_price(self, symbol: str) -> float:
+        """Get current last price for a perpetual swap (e.g. 'BTC')."""
+        try:
+            ticker = self._exchange.fetch_ticker(f"{symbol}/USDT:USDT")
+            price = ticker.get("last") or 0
+            if price and price > 0:
+                return float(price)
+        except Exception:
+            pass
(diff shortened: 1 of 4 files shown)


richkuo:
@claude please resolve merge conflicts

PR #194, opened 2026-04-08
feat: add range_scalper strategy

richkuo:
## Summary

- Adds `range_scalper` strategy that detects low-volatility consolidation zones via Bollinger bandwidth + volume filter, then mean-reverts at band touches with short RSI confirmation
- Designed for short timeframes (1m–5m) where post-spike consolidation creates predictable oscillations — stays silent during trends
- Registered in spot, perps, and futures strategy registries with full test coverage (signal generation, trend suppression, edge cases)

## Test plan

- [x] Python syntax check passes for spot and futures strategies.py
- [x] Go build succeeds (`cd scheduler && go build .`)
- [x] Go tests pass (`cd scheduler && go test ./...`)
- [x] `TestRangeScalper` tests pass — signals in range-bound data, suppression during trends, all columns present
- [x] Edge case tests pass — empty DataFrame and single-row handling
- [ ] Smoke test with `go-trader init --json` to verify strategy appears in discovery

---
Generated with: Claude Opus 4.6 | Effort: auto

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -41,6 +41,9 @@ jobs:
       - name: Install Python dependencies
         run: uv sync --frozen --extra test
 
+      - name: Clean up lockfile changes before Claude checkout
+        run: git checkout -- uv.lock
+
       - name: Run Claude Code
         id: claude
         uses: anthropics/claude-code-action@v1


richkuo:
@claude fix the 2 issues

PR #230, opened 2026-04-12
fix: skip workflow when @claude appears only in code blocks or examples

richkuo:
## Summary

- Adds a `verify_invocation` step that strips code blocks/inline code spans from the trigger body, then checks `@claude` only appears at the start of a line (an actual invocation)
- Gates all subsequent job steps on `invoked == 'true'` so nothing runs on false positives
- Passes untrusted event bodies via `env:` variables (not inline `${{ }}` interpolation) to avoid command injection

## Root cause

The workflow's `contains(github.event.issue.body, '@claude')` condition fires whenever `@claude` appears anywhere in a body — including inside fenced code blocks or inline examples — causing unintended workflow runs.

---
Generated with: Claude Sonnet 4.6 | Effort: auto

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -25,26 +25,64 @@ jobs:
       id-token: write
       actions: read # Required for Claude to read CI results on PRs
     steps:
+      - name: Verify @claude is an actual invocation (not in a code block or example)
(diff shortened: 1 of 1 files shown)


richkuo:
@claude haiku

PR #256, opened 2026-04-13
fix: avoid double-counting shared wallets in portfolio risk

### LINKED ISSUE #243: Shared wallet strategies double-count balance in portfolio risk / peak value
### Bug Report

When multiple live strategies trade from the **same exchange account wallet** (e.g. two Hyperliquid perps strategies), the scheduler sums each strategy's `capital` independently to compute the total portfolio value and peak. This double (or triple) counts the same wallet balance, causing:
- Inflated `portfolioRisk.PeakValue`
- False or premature circuit breaker triggers
- Incorrect drawdown calculations

### Motivation

Shared wallets are common for platforms like Hyperliquid where multiple strategies trade perps from one account. The scheduler should detect this and use the real on-chain / API-reported balance instead of summing strategy capitals.

### Proposed Change

Introduce an `isSharedWalletStrats(cfg.Strategies)` helper (or platform-specific variant) that checks if multiple live strategies share the same account.

When detected, use the **actual exchange balance** to initialize `PortfolioRisk.PeakValue` and compute total portfolio value in the risk loop.

Example approach (opt-in via env var):
```go
if addr := os.Getenv("HYPERLIQUID_ACCOUNT_ADDRESS"); addr != "" {
    if bal, err := fetchHyperliquidBalance(addr); err == nil {
        totalPV = bal
    }
}
```

A generalized version could support any platform by adding a `WalletID` or `AccountID` strategy field, but Hyperliquid is the immediate pain point.

### Acceptance Criteria

- [ ] Portfolio peak no longer double-counts shared wallets
- [ ] Risk loop uses real balance for total portfolio value when shared wallet is detected
- [ ] Existing behavior unchanged for non-shared setups
- [ ] Tests cover shared-wallet scenario

richkuo:
Closes #243

When multiple live strategies share the same exchange account (e.g. two Hyperliquid perps strategies), the scheduler was summing each strategy's `PortfolioValue` independently when computing total portfolio value. For shared wallets this double-counts the same balance, inflating peak value and triggering false drawdown alarms.

## Summary
- New `scheduler/shared_wallet.go` with `SharedWalletKey`, `detectSharedWallets`, `fetchSharedWalletBalances`, `computeTotalPortfolioValue`, and `computeInitialPortfolioPeak`.
- Network I/O happens outside the state lock; total PV is computed under `mu.RLock()` with pre-fetched balances.
- Falls back to the per-strategy sum on fetch failure so the risk loop never sees a zero wallet.
- Legacy `capital_pct` init behavior is preserved for non-shared setups.

## Test plan
- [x] `go test ./...` passes
- [x] 13 new unit tests covering detection, total PV, fallback, peak init, and legacy paths

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -195,6 +195,11 @@ func reconcileHyperliquidPositions(stratState *StrategyState, sym string, positi
 // assigned to the strategy that opened them (via OwnerStrategyID).
 // Unowned on-chain positions are logged as warnings but not assigned.
 // Must be called WITHOUT holding any lock; acquires Lock internally.
+//
+// This is the self-contained entry point that fetches its own state. When the
+// scheduler has already fetched clearinghouseState earlier in the cycle (e.g.
+// for shared-wallet balance), use reconcileHyperliquidAccountPositions instead
+// to avoid a second round-trip to the HL API.
 func syncHyperliquidAccountPositions(hlStrategies []StrategyConfig, state *AppState, mu *sync.RWMutex, logMgr *LogManager) bool {
 	accountAddr := os.Getenv("HYPERLIQUID_ACCOUNT_ADDRESS")
 	if accountAddr == "" {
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude fix failing check

PR #257, opened 2026-04-13
Auto-clear latched portfolio kill switch on shared-wallet restart

### LINKED ISSUE #244: Kill switch remains latched forever on shared-wallet setups even after restart
### Bug Report

When the kill switch latches due to an incorrectly inflated `PortfolioRisk.PeakValue` (e.g. from double-counting a shared wallet), restarting the service does **not** clear it. The kill switch remains active indefinitely, blocking all new trades even though the underlying balance is fine.

### Motivation

A service restart should act as a safe reset opportunity. If the scheduler can validate the real on-chain balance on startup and determine the portfolio is above water, it should be allowed to unlatch the kill switch rather than staying dead forever.

### Proposed Change

In the startup initialization of `PortfolioRisk.PeakValue`, when a shared wallet is confirmed and the real balance is successfully fetched:

```go
if state.PortfolioRisk.KillSwitchActive {
    fmt.Printf("[INFO] Shared wallet: clearing kill switch (was latched at %s)\n", state.PortfolioRisk.KillSwitchAt.Format("2006-01-02 15:04 UTC"))
    state.PortfolioRisk.KillSwitchActive = false
    state.PortfolioRisk.KillSwitchAt = time.Time{}
}
```

This should be guarded by successful retrieval of the real balance, so a network failure on startup does not accidentally clear a legitimate kill switch.

### Acceptance Criteria

- [ ] Kill switch is cleared on startup when a valid real balance is fetched for a shared wallet
- [ ] Kill switch is preserved if balance fetch fails
- [ ] Non-shared-wallet setups are unaffected
- [ ] A log line is emitted whenever the kill switch is auto-cleared

richkuo:
Closes #244

## Summary

A latched `PortfolioRisk` kill switch — typically caused by a previously inflated `PeakValue` from shared-wallet double-counting — used to survive every restart, blocking all new trades indefinitely. Restart should instead be a safe reset opportunity when we can independently verify the wallet is reachable.

`ClearLatchedKillSwitchSharedWallet` runs once at startup; when at least one platform hosts a shared wallet (`capital_pct > 0` with > 1 strategy) **and** the real on-chain balance fetches successfully, the kill switch is unlatched and an `auto_reset` audit event is recorded. Network/config failures preserve the latch so legitimate kill switches are never lost.

Currently dispatches Hyperliquid via `fetchHyperliquidBalance`; other platforms return an error and keep the kill switch latched.

## Acceptance criteria

- [x] Kill switch cleared on startup when a valid real balance is fetched for a shared wallet
- [x] Kill switch preserved if balance fetch fails
- [x] Non-shared-wallet setups unaffected
- [x] Log line emitted whenever the kill switch is auto-cleared

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` passes (full suite)
- [x] New unit tests cover success, fetch-failure, no-shared-wallet, inactive-switch, multi-platform fallback, and the shared-wallet detector

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -67,6 +67,22 @@ func fetchHyperliquidBalance(accountAddress string) (float64, error) {
 	return val, nil
 }
 
+// defaultSharedWalletBalance dispatches a real on-chain balance lookup by
+// platform name for use with ClearLatchedKillSwitchSharedWallet (#244).
+// Returns an error for any platform that does not (yet) expose a real
+// balance endpoint, so callers preserve the kill switch on uncertainty.
+func defaultSharedWalletBalance(platform string) (float64, error) {
+	switch platform {
+	case "hyperliquid":
+		addr := os.Getenv("HYPERLIQUID_ACCOUNT_ADDRESS")
+		if addr == "" {
+			return 0, fmt.Errorf("HYPERLIQUID_ACCOUNT_ADDRESS not set")
+		}
+		return fetchHyperliquidBalance(addr)
+	}
+	return 0, fmt.Errorf("no shared-wallet balance fetcher for platform %q", platform)
+}
+
 // syncHyperliquidLiveCapital is a no-op kept for backward compatibility.
 // Capital is now managed per-strategy via config (Capital field) or capital_pct.
 // With multiple strategies on one account, overriding each strategy's capital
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

richkuo:
@claude fix all pr review suggestions

richkuo:
@claude

richkuo:
@claude

PR #260, opened 2026-04-14
Include perps in notional-exposure price-watch set (#245)

### LINKED ISSUE #245: Perps strategies excluded from price-watch symbol set in notional exposure check
### Bug Report

In the periodic notional-exposure check, the scheduler builds a list of price-watch symbols only for `spot` strategies. `perps` strategies are ignored when constructing `symbolSet`, which means their positions are not included in `PortfolioNotional` calculations.

### Motivation

Perps strategies (e.g. Hyperliquid perps) carry notional exposure just like spot. Excluding them from the price-watch list means the scheduler cannot correctly compute total portfolio notional value or enforce notional-based risk limits.

### Current Code (simplified)

```go
for _, sc := range cfg.Strategies {
    if sc.Type == "spot" && len(sc.Args) >= 2 {
        symbolSet[sc.Args[1]] = true
    }
}
```

### Proposed Change

Also include perps symbols, normalized to a standard symbol format (e.g. appending `/USDT` if missing):

```go
} else if sc.Type == "perps" && len(sc.Args) >= 2 {
    symbolSet[sc.Args[1] + "/USDT"] = true
}
```

A more robust generalization could use a platform-specific symbol normalizer so it works for non-USDT perps too.

### Acceptance Criteria

- [ ] Perps strategies are included in the price-watch symbol set
- [ ] Notional exposure calculation includes perps positions
- [ ] Spot behavior remains unchanged
- [ ] Test coverage added for perps notional calculation

richkuo:
Closes #245

## Summary

`PortfolioNotional` was silently dropping perps exposure because all three `symbolSet` builders (main loop, `--summary` handler, `/status` endpoint) only picked up `sc.Type == "spot"`. Naively adding the perps base asset (`"BTC"`) wouldn't work either -- `check_price.py` queries BinanceUS which requires the `"BTC/USDT"` form.

## Fix

- `collectPriceSymbols(strategies)` in `risk.go` returns the set of symbols to fetch (normalizing perps `"BTC"` → `"BTC/USDT"`) plus a mirror map from position-key → fetch-key.
- `mirrorPerpsPrices` back-fills the base-asset alias after `FetchPrices` returns. An existing price under the position key is preserved so a live exchange mid published by a strategy run still wins over a stale BinanceUS quote.
- Three call sites updated: `main.go:315` (main loop pre-fetch), `main.go:940` (`--summary`), and `server.go` (`StatusServer` gains a `priceMirror` field).

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` -- all existing tests + new: `TestPortfolioNotional_IncludesPerps`, `TestCollectPriceSymbols`, `TestCollectPriceSymbols_PerpsAlreadyNormalized`, `TestMirrorPerpsPrices`, `TestNewStatusServerExtractsSymbols` (extended).

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -311,17 +311,11 @@ func main() {
 			cycle, cycleStart.UTC().Format("2006-01-02 15:04:05 UTC"),
 			len(dueStrategies), len(cfg.Strategies))
 
-		// Collect symbols that need prices
-		symbolSet := make(map[string]bool)
-		for _, sc := range cfg.Strategies {
-			if sc.Type == "spot" && len(sc.Args) >= 2 {
-				symbolSet[sc.Args[1]] = true
-			}
-		}
-		symbols := make([]string, 0, len(symbolSet))
-		for s := range symbolSet {
-			symbols = append(symbols, s)
-		}
+		// Collect symbols that need prices. Spot strategies use the
+		// BinanceUS-formatted symbol directly; perps strategies use the base
+		// asset as their position key, so we fetch under a normalized
+		// "<base>/USDT" form and mirror the result back — #245.
+		symbols, perpsMirror := collectPriceSymbols(cfg.Strategies)
 
 		// Fetch current prices for portfolio valuation
 		prices := make(map[string]float64)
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

richkuo:
@claude

PR #265, opened 2026-04-14
fix: value perps positions as margin + PnL (#254)

### LINKED ISSUE #254: Bug: Perps positions valued like spot (multiplier=0) gives inflated portfolio values
## Problem

All hyperliquid perps positions have `multiplier=0` in the database, causing `PortfolioValue()` to treat them like spot positions — valuing the full notional instead of margin + leveraged PnL.

### Example

`hl-rmc-eth-live` on hyperliquid:
- Cash: $27.15
- Position: long 0.279 ETH @ $2,210.71 entry
- Current ETH: $2,201.10

**Current (wrong, multiplier=0):**
```
Value = $27.15 + 0.279 × $2,201.10 = $641.33
```

This treats the position as if the strategy *owns* 0.279 ETH outright (full notional). But it's a perps position — the strategy only has margin exposure, not the full dollar amount.

**Expected (correct, with proper leverage):**
```
Value = $27.15 + 0.279 × leverage × ($2,201.10 - $2,210.71) = $27.15 + PnL
```

### Root Cause

In `portfolio.go` line 42, the valuation branches on `pos.Multiplier > 0`:
```go
if pos.Multiplier > 0 {
    // Futures: PnL-based valuation
    total += pos.Quantity * pos.Multiplier * (price - pos.AvgCost)
} else if pos.Side == "long" {
    total += pos.Quantity * price  // ← perps fall here
}
```

All HL perps positions are stored with `multiplier=0`, so they hit the spot branch instead of the PnL branch.

Additionally, there is no config field to set the multiplier/leverage for a strategy or platform.

### Proposed Fix

**Option A — Real leverage from exchange:**
For live hyperliquid strategies, fetch the actual leverage from the Hyperliquid API. The `/info` endpoint with `clearinghouseState` returns per-position leverage. Store it as the position multiplier.

**Option B — Config-based leverage:**
Add a `leverage` or `multiplier` field to the strategy config:
```json
{
  "id": "hl-rmc-eth-live",
  "platform": "hyperliquid",
  "type": "perps",
  "leverage": 20,
  ...
}
```

When opening a perps position, use this config value as the multiplier.

**Option C — Hybrid:**
Use config as default, but override with real leverage from the exchange when available (live mode). Paper mode uses the config value.

### Changes Required

1. **`config.go`** — Add `Leverage float64` field to `StrategyConfig`
2. **Position creation** — Set `Multiplier` from config when opening perps positions
3. **HL position sync** — When reconciling on-chain positions, also sync the actual leverage from Hyperliquid API
4. **Paper mode** — Use config leverage value for paper positions
5. **`PortfolioValue()`** — Already handles `Multiplier > 0` correctly, no changes needed

### Impact

- Portfolio values for perps strategies will reflect actual margin + PnL, not inflated notionals
- Summary tables will show realistic strategy values
- Leaderboard rankings may shift significantly since perps PnL% will be calculated correctly

Reference only: LLM model used: MiniMax M2.7 · Reasoning effort: medium

richkuo:
Closes #254

## Summary

- All Hyperliquid perps positions were stored with `Multiplier=0`, so `PortfolioValue` hit the spot branch (`qty * price`) and reported inflated strategy values.
- New `ExecutePerpsSignal` with margin-based accounting: cash is only reduced by the fee, position is stamped `Multiplier=1` so valuation uses the PnL branch, and paper/live sizing uses a leveraged notional budget.
- New `StrategyConfig.Leverage` field (perps-only, [1,100], default 1) flows through HL + OKX perps dispatch and the init wizard.
- `fetchHyperliquidState` parses per-position leverage; `reconcileHyperliquidPositions` copies it into stored positions and migrates legacy `Multiplier=0` positions to `Multiplier=1`.

## Test plan

- [x] `go test ./...` passes
- [x] New unit tests cover `ExecutePerpsSignal` paper/live/close paths, the regression from the issue example, leverage validation, legacy position migration, and leverage parsing from `clearinghouseState`.
- [ ] Smoke test on a running scheduler with a real HL strategy

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -102,6 +102,7 @@ type StrategyConfig struct {
 	MaxDrawdownPct  float64                `json:"max_drawdown_pct"`
 	IntervalSeconds int                    `json:"interval_seconds,omitempty"` // per-strategy override (0 = use global)
 	HTFFilter       bool                   `json:"htf_filter,omitempty"`       // higher-timeframe trend filter
+	Leverage        float64                `json:"leverage,omitempty"`         // perps leverage multiplier (default 1 = no leverage); used for notional sizing and margin-based valuation (#254)
 	Params          map[string]interface{} `json:"params,omitempty"`           // custom strategy parameters passed to Python
 	ThetaHarvest    *ThetaHarvestConfig    `json:"theta_harvest,omitempty"`
 	FuturesConfig   *FuturesConfig         `json:"futures,omitempty"`
@@ -233,6 +234,13 @@ func LoadConfig(path string) (*Config, error) {
 			} else {
 				cfg.Strategies[i].MaxDrawdownPct = 60
 			}
+		}
+
+		// #254: Default leverage for perps strategies is 1x (no leverage)
+		// when unset. Spot/options/futures ignore Leverage; only perps uses it
+		// for notional sizing and setting Position.Multiplier.
+		if cfg.Strategies[i].Type == "perps" && cfg.Strategies[i].Leverage <= 0 {
+			cfg.Strategies[i].Leverage = 1
 		}
 
 		// #56: Default theta harvest for options strategies — sold options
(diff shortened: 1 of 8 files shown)


richkuo:
@claude

richkuo:
@claude fix the suggestions

richkuo:
@claude pr code review again

PR #271, opened 2026-04-16
fix(hl-sync): skip reconciliation for shared coins to prevent phantom circuit breakers

### LINKED ISSUE #258: Trade history drifts from exchange when multiple strategies share one account
### Problem
When multiple strategies trade the same asset on a shared exchange account, the internal trade history (`trades` table) drifts from what actually happened on the exchange. The `positions` table stays accurate (reconciled against exchange balance each cycle), but the trade log is unreliable for accounting.

### Observed on Hyperliquid live
- **Wallet fills (ground truth):** 37 ETH fills, net position = 0.315 ETH
- **hl-rmc-eth-live DB:** trades net to 0.460 ETH (should be 0.315)
- **hl-tema-eth-live DB:** trades net to 0.212 ETH (should be 0, it's flat)
- **Combined DB:** 0.672 ETH vs 0.315 actual — off by 2×
- Neither strategy shows realized PnL ($0 in DB vs $70.12 on wallet)

### Root causes
1. **Shared wallet, per-strategy accounting** — Multiple strategies share one exchange account/wallet. The exchange sees a single net position per asset; the bot logs trades per-strategy. When strategies overlap on the same coin, fills can't be cleanly attributed.
2. **Partial fills** — One bot "trade" may result in multiple exchange fills with different sizes/prices. The bot logs the intended order, not the actual fills.
3. **Interleaved orders** — When two strategies submit orders near-simultaneously, fills from one can be attributed to the other, or some fills get dropped entirely.
4. **Sells not always logged** — Some close trades appear in exchange fills but not in the DB trade history.

### Affects all platforms
This is not Hyperliquid-specific. Any platform where multiple strategies share credentials (Deribit, Binance, TopStep, Robinhood, OKX, IBKR) will have the same problem when two strategies trade the same asset.

### Expected behavior
- Trade history should reconcile with exchange fills.
- Each strategy's net position from its trade log should match its actual holdings.
- Realized PnL in DB should match exchange-reported PnL.

### Possible fixes
1. **Fetch fills post-trade** — After each order, query exchange fills API and log actual fills (not intended orders). Already have exchange tx IDs from #219.
2. **Position reconciliation on cycle** — Compare DB trade-derived position against exchange-reported position each cycle. Flag/log discrepancies.
3. **Per-strategy sub-accounts** — Where supported, use separate sub-accounts per strategy to avoid fill attribution issues.
4. **Trade log marked as "estimated"** — If exact attribution isn't possible, mark trade entries as estimated and track the reconciliation gap.

### Reproduction
1. Configure 2+ strategies on the same platform trading the same asset (e.g., `hl-tema-eth-live` + `hl-rmc-eth-live`)
2. Let them trade for a few days
3. Compare `trades` table per-strategy net position vs actual exchange position
4. Observe drift

richkuo:
## Summary

- Fixes phantom circuit breakers caused by shared-wallet position sync: when 2+ strategies trade the same coin (e.g. `hl-rmc-eth-live` + `hl-tema-eth-live`), one strategy selling previously caused the sync to remove the other strategy's virtual position, collapsing its portfolio value and tripping the drawdown kill switch
- Skips destructive reconciliation (qty changes, position removal) for shared coins; only non-destructive updates (multiplier migration #254, leverage sync) are applied
- Adds per-coin `ReconciliationGap` tracking on `AppState` (on-chain qty, virtual qty, delta, strategies) exposed via `/status` endpoint for drift visibility
- `reconcileHyperliquidAccountPositions` now takes an `allStrategies` param so shared-coin detection uses the full HL config, not just due strategies

## Test plan

- [ ] `TestAccountSyncSharedCoinSkipsReconciliation` — positions unchanged when coin is shared
- [ ] `TestAccountSyncSharedCoinNotRemovedWhenOnChainGone` — phantom circuit breaker scenario: position not deleted when on-chain is gone
- [ ] `TestAccountSyncSharedCoinMultiplierMigration` — multiplier migration and leverage sync still work for shared coins
- [ ] `TestAccountSyncMixedSharedAndNonShared` — BTC (sole owner) reconciles, ETH (shared) skips
- [ ] `TestAccountSyncSharedCoinGapClearedWhenNoLongerShared` — stale gaps cleaned up when coin returns to single owner
- [ ] All existing tests pass (`go test ./...`)

Closes #258

---
Generated with: Claude Opus 4.6 | Effort: high

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -256,7 +256,8 @@ func syncHyperliquidAccountPositions(hlStrategies []StrategyConfig, state *AppSt
 		return false
 	}
 
-	return reconcileHyperliquidAccountPositions(hlStrategies, state, mu, logMgr, positions)
+	// Self-contained entry: due and all are the same list.
+	return reconcileHyperliquidAccountPositions(hlStrategies, hlStrategies, state, mu, logMgr, positions)
 }
 
 // reconcileHyperliquidAccountPositions reconciles pre-fetched on-chain positions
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

PR #276, opened 2026-04-16
fix(perps): source HL/OKX marks from venue not BinanceUS spot

### LINKED ISSUE #262: Suppress [WARN] Skipping zero price noise for HL-only perps coins not listed on BinanceUS
## Background

Follow-up from PR #260. That PR added perps base assets (HL/OKX coin → `<base>/USDT`) to the `FetchPrices` set so `PortfolioNotional` can revalue perps at live marks. For perps coins that BinanceUS also lists (BTC, ETH, SOL, DOGE, …) this works perfectly.

## Problem

Hyperliquid lists many perps that BinanceUS does not quote — HYPE, kPEPE, kSHIB, PURR, WIF, POPCAT, etc. For those, `check_price.py` returns a zero/missing price, and `main.go:339` unconditionally logs:

```
[WARN] Skipping zero price for HYPE/USDT
```

This fires **every scheduler cycle** (every 5–60 min depending on tick config), for every HL-only symbol the user trades. It pollutes the log stream with what is effectively an expected condition, not a real warning. And it teaches operators to ignore `[WARN] Skipping zero price` lines in general — which is bad, because the same warning would also be the signal for a legitimate BinanceUS API outage on a covered pair.

## Not a regression

Before the fix, these symbols were never fetched, so the warning never fired. After the fix, they are fetched-and-skipped, which is a functional improvement (no actual exposure change — the fallback to `pos.AvgCost` is the same) but an operational regression in log hygiene.

## Options

1. **Categorize the symbol at collection time.** Have `collectPriceSymbols` return a separate \"best-effort\" set for symbols that originated from a `perps` strategy. In the main-loop fetch path, zero prices for best-effort symbols downgrade from `[WARN]` to `[DEBUG]` (or suppressed entirely).

2. **Cache known-missing symbols.** Record symbols that returned zero during the first N cycles and stop requesting them from `FetchPrices` entirely until a config reload. Simpler log-wise, but silently hides BinanceUS coverage changes.

3. **Per-platform mark fetcher.** Eventually push perps mark-price fetching into the platform adapter (HL/OKX already have WebSocket mark feeds); fetch HL marks via the HL SDK instead of BinanceUS spot. This is the \"right\" long-term fix — `/USDT` hardcode in `collectPriceSymbols` goes away — but it's a bigger change.

Recommendation: **(1)** as an immediate log-hygiene patch, plant a seed for **(3)** as the durable fix.

## Acceptance criteria (for option 1)

- [ ] `collectPriceSymbols` returns / exposes which symbols came from perps strategies (best-effort fetches).
- [ ] `main.go` fetch path does not emit `[WARN] Skipping zero price` for best-effort symbols.
- [ ] The warning still fires as today for spot symbols that unexpectedly come back zero.
- [ ] Regression test that verifies the categorization (can be a unit test on the helper, not on the log output).

## Related

- PR #260
- Issue #245

### LINKED ISSUE #263: PortfolioValue perps shorts pick up spot/perps basis drift as phantom PnL
## Background

Follow-up from PR #260. After that PR, `PortfolioValue` (via `computeTotalPortfolioValue` → portfolio-level kill switch) now reads perps positions at a live BinanceUS spot quote instead of `pos.AvgCost`. For longs, `total += pos.Quantity * price` produces a faithful mark-to-market notional. For shorts, `scheduler/portfolio.go:53` uses a different formula:

```go
// Short: profit = (avg_cost - current_price) * qty
total += pos.Quantity * (2*pos.AvgCost - price)
```

## Problem

Before PR #260, `price == pos.AvgCost` for perps shorts (via the fallback), so the short-side P&L contribution was hard-zeroed — the short was worth `2*AvgCost - AvgCost = AvgCost * Quantity`. After PR #260, `price` is the BinanceUS spot quote for the base asset, **not** the actual HL/OKX perps mark. Those two prices diverge (perps basis, funding, exchange-specific liquidity) — usually by <1 bp on majors, sometimes more on alts or during stress.

The divergence shows up in `PortfolioValue` as **phantom PnL on open perps shorts** that has nothing to do with the actual HL/OKX mark. Example:

- HL mark: ETH-PERP = 3200.10
- BinanceUS spot: ETH/USDT = 3199.85
- Short position: 2.0 ETH @ 3000 AvgCost

Pre-PR: short contributes `2 * (2*3000 - 3000) = 6000` to PortfolioValue.
Post-PR: short contributes `2 * (2*3000 - 3199.85) = 5600.30` (BinanceUS-based).
Truth (HL mark): should be `2 * (2*3000 - 3200.10) = 5599.80`.

So we went from "zero realism, frozen at entry" to "realism with basis noise", which is strictly better — but now kill-switch drawdown thresholds can be tripped (or masked) by a basis delta that has nothing to do with the actual perps position. And this flows into `state.PortfolioRisk.PeakValue`, which is sticky.

## The only acceptable fix

Source perps marks from the perps exchange itself. BinanceUS spot is the wrong oracle; the correct oracle is the venue the position lives on.

1. **Hyperliquid SDK**: `info.all_mids()` returns the mark (~mid) for every HL perp in a single round-trip. Tiny payload, authentication-free. `hyperliquid_balance.go` already touches the `Info` SDK for shared-wallet balance — reuse the same client.
2. **OKX (ccxt)**: `fetch_tickers(type='swap')` returns mark prices for all OKX perps in one call.
3. **Plumb into `prices` map**: new `fetchPerpsMarks` step that runs alongside `FetchPrices` in `main.go:322`, merges into `prices` under the position key (e.g., `prices["BTC"] = hl_mark`). `mirrorPerpsPrices` would still handle the legacy "fetch key == position key already" case if kept for belt-and-braces.

Once perps marks come from HL/OKX directly, the `/USDT` hardcode in `collectPriceSymbols` falls out, and HL-only coins (HYPE, kPEPE, PURR, …) are no longer routed through BinanceUS — closing the log-noise problem tracked in #262 as a side effect.

A "quick fix" that papers over the short-side formula (tolerance band, special-case AvgCost-fallback, etc.) is **not** acceptable: it leaves perps mark-pricing wired to the wrong oracle and accumulates technical debt. Either the mark source is correct or the bug isn't fixed.

## Acceptance criteria

- [ ] HL perps marks sourced from `info.all_mids()` (or equivalent), **not** BinanceUS spot.
- [ ] OKX perps marks sourced from OKX `fetch_tickers(swap)` via ccxt, **not** BinanceUS spot.
- [ ] `collectPriceSymbols` no longer emits `<base>/USDT` fetch keys for perps strategies — perps live in a separate marks path.
- [ ] `/USDT` hardcode in `collectPriceSymbols` is either removed entirely or scoped to a spot-only helper.
- [ ] `PortfolioValue` short-side formula uses the actual exchange mark — regression test covering an ETH short with a 10 bp basis delta asserts the HL mark wins.
- [ ] No `[WARN] Skipping zero price for <HL-only coin>/USDT` log lines for HL tokens not listed on BinanceUS (closes #262 as a side effect).
- [ ] No extra HTTP round-trips when neither HL nor OKX perps strategies are configured.
- [ ] `TestPortfolioNotional_IncludesPerps` and `TestPortfolioNotional_IncludesPerpsShort` still green — the exchange-native marks path composes with the existing notional tests.

## Closes on merge

- #263 (this issue)
- #262 (subsumed — the HL-only log noise vanishes when perps stop routing through BinanceUS)

## Related

- PR #260 (introduced the improvement-with-drift)
- Issue #245 (root cause for the perps-revaluation fix)

richkuo:
## Summary

- **Root cause**: `PortfolioValue` for perps positions was using BinanceUS spot as the price oracle. For shorts, `total += qty * (2*avgCost - spotPrice)` — when spot diverges from the perps mark by even a few basis points, the delta appears as phantom PnL and can trip or mask the portfolio-level kill switch.
- **Fix**: HL perps marks now come from `POST /info allMids` (new `fetchHyperliquidMids` in Go, auth-free, one round-trip for all HL coins). OKX perps marks come from a new `shared_scripts/fetch_okx_perps_marks.py` via ccxt `get_perp_price`. Both are merged into the `prices` map with preserve-existing semantics (strategy-published marks always win).
- **Side effect**: HL-only coins (HYPE, kPEPE, PURR, …) no longer hit BinanceUS, so the `[WARN] Skipping zero price for <coin>/USDT` log noise tracked in #262 disappears without any extra code.

## Files changed

| File | Change |
|------|--------|
| `scheduler/hyperliquid_marks.go` | New: `fetchHyperliquidMids` — POST `/info allMids`, filters to requested coins |
| `scheduler/hyperliquid_marks_test.go` | New: httptest-based unit tests |
| `shared_scripts/fetch_okx_perps_marks.py` | New: ccxt OKX swap mark fetcher |
| `scheduler/executor.go` | Add `FetchOKXPerpsMarks` |
| `scheduler/risk.go` | `collectPriceSymbols` spot-only; add `collectPerpsMarkSymbols` + `mergePerpsMarks`; remove `mirrorPerpsPrices` |
| `scheduler/main.go` | Wire HL/OKX perps rails in cycle loop + channel-summary path |
| `scheduler/server.go` | Wire perps rails in `/status` with throttled error logging |
| `scheduler/portfolio_test.go` | Regression: ETH short with 10 bp basis delta asserts HL mark wins |
| `scheduler/risk_test.go` | Update `TestCollectPriceSymbols`; add `TestCollectPerpsMarkSymbols`, `TestMergePerpsMarks`; remove obsolete mirror tests |
| `scheduler/server_test.go` | Update `TestNewStatusServerExtractsSymbols` for new field layout |

## Test plan

- [x] `go build .` — clean
- [x] `go test ./...` — all pass
- [x] `python3 -m py_compile shared_scripts/fetch_okx_perps_marks.py` — clean
- [x] `./go-trader init --json ...` smoke — exits 0
- [x] `TestPortfolioValueShort_UsesExchangeMarkNotSpotBasis` — pins 10 bp basis delta regression
- [x] `TestCollectPerpsMarkSymbols` — HL/OKX split, dedup, sort, filter
- [x] `TestMergePerpsMarks` — preserve-existing, skip-zero semantics
- [x] `TestFetchHyperliquidMids_*` — httptest stubs cover basic fetch, empty input, missing coin, HTTP error, invalid JSON, zero price

Closes #263
Closes #262

### DIFF
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -559,3 +559,28 @@ func FetchFuturesMarks(symbols []string) (map[string]float64, string, error) {
 	}
 	return marks, mode, nil
 }
+
+// FetchOKXPerpsMarks runs fetch_okx_perps_marks.py and returns a map of
+// base-coin→mark-price for OKX perpetual swaps. Routes through the OKX
+// adapter (ccxt fetch_ticker for USDT-margined swaps) because BinanceUS
+// does not quote OKX perps and the OKX swap mark diverges from the
+// BinanceUS spot quote (perps basis, funding, exchange-specific liquidity).
+// Fixes the phantom-PnL oracle for OKX perps shorts — see issue #263.
+//
+// Best-effort: callers treat an error or missing coin as a signal to fall
+// back to pos.AvgCost via the prices-miss path in PortfolioValue /
+// PortfolioNotional — same degradation semantics as FetchFuturesMarks.
+func FetchOKXPerpsMarks(coins []string) (map[string]float64, error) {
+	if len(coins) == 0 {
+		return map[string]float64{}, nil
+	}
+	stdout, stderr, err := RunPythonScript("shared_scripts/fetch_okx_perps_marks.py", coins)
+	if err != nil {
+		return nil, fmt.Errorf("okx perps marks fetch error: %w (stderr: %s)", err, string(stderr))
+	}
+	var marks map[string]float64
+	if err := json.Unmarshal(stdout, &marks); err != nil {
+		return nil, fmt.Errorf("parse okx perps marks: %w (stdout: %s)", err, string(stdout))
+	}
+	return marks, nil
+}
--- a/scheduler/hyperliquid_marks.go
+++ b/scheduler/hyperliquid_marks.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// fetchHyperliquidMids fetches the current mid prices for all HL perpetuals
+// in a single authentication-free round-trip and returns a coin→price map
+// filtered to the requested coins. Reuses hlMainnetURL from
+// hyperliquid_balance.go so tests can redirect to a stub server.
+//
+// The /info allMids response is a flat JSON object: {"BTC":"67500.50", ...}.
+// Each value is a numeric string. Coins not listed on HL are omitted from the
+// returned map — the caller falls back to pos.AvgCost via the prices-miss
+// path in PortfolioValue / PortfolioNotional (same graceful degradation as
+// fetch_futures_marks.py misses).
+//
+// This is the correct oracle for HL perps positions; BinanceUS spot is wrong
+// because spot/perps basis divergence (funding, liquidity, exchange-specific
+// pricing) shows up as phantom PnL in PortfolioValue — fixes issue #263.
+func fetchHyperliquidMids(coins []string) (map[string]float64, error) {
+	if len(coins) == 0 {
+		return map[string]float64{}, nil
+	}
+
+	payload := map[string]string{"type": "allMids"}
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("marshal allMids request: %w", err)
+	}
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	resp, err := client.Post(hlMainnetURL+"/info", "application/json", bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("http request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("http %d from %s/info allMids", resp.StatusCode, hlMainnetURL)
+	}
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("read allMids response: %w", err)
+	}
+
+	// Flat object: {"BTC": "67500.50", "ETH": "3200.10", ...}
+	var raw map[string]string
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return nil, fmt.Errorf("parse allMids response: %w", err)
+	}
+
+	want := make(map[string]bool, len(coins))
+	for _, c := range coins {
+		want[c] = true
+	}
+
+	marks := make(map[string]float64, len(coins))
+	for coin, priceStr := range raw {
+		if !want[coin] {
+			continue
+		}
+		p, err := strconv.ParseFloat(priceStr, 64)
+		if err != nil || p <= 0 {
+			continue
+		}
+		marks[coin] = p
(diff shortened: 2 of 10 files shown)


richkuo:
@claude

richkuo:
@claude fix all suggestions

PR #290, opened 2026-04-17
Persist trades immediately to SQLite (#289)

### LINKED ISSUE #289: Write trades to DB immediately instead of batching at end of cycle
### Problem

Trades are currently appended to in-memory `TradeHistory` and only persisted to the `trades` table at the end of each cycle via `SaveStateWithDB`. If go-trader crashes or is killed mid-cycle (e.g., `SIGKILL`, OOM, power loss), all in-memory trades for that cycle are lost permanently.

This also affects:
- Position state (also in-memory until cycle end)
- Circuit breaker force-close records
- PnL accuracy on recovery

### Current flow

1. Cycle starts
2. Strategies execute trades → appended to `s.TradeHistory` (in-memory)
3. Risk checks, position updates → all in-memory
4. `SaveStateWithDB` runs → writes everything to SQLite in one batch

### Proposed fix

Write each trade to the `trades` table immediately when executed (inside a transaction), before updating in-memory state. Keep the in-memory `TradeHistory` for fast access during the cycle, but treat the DB as the source of truth.

```go
func recordTrade(sdb *StateDB, trade Trade) error {
    _, err := sdb.db.Exec(`INSERT INTO trades (...) VALUES (?, ?, ?, ...)`,
        trade.StrategyID, trade.Timestamp, trade.Symbol, ...)
    return err
}
```

### Benefits

1. **Crash resilience** — trades survive mid-cycle crashes
2. **Audit trail** — even partial cycles have trade records
3. **Recovery** — easier to reconstruct state from DB after crash
4. **Consistency** — DB is the source of truth, not memory

### Considerations

- Need to handle the duplicate-insert guard (current code compares timestamps to skip already-written trades on restart)
- Position state changes should also be persisted more frequently (related to #288)

### Related

- #288 — Closed positions history table
- #258 — Trade history drift from shared wallet

richkuo:
## Summary
- Route every `TradeHistory` append through a new `RecordTrade` helper that writes the trade to SQLite immediately via a package-level `tradeRecorder` hook (wired to `StateDB.InsertTrade` at startup).
- Trades now survive mid-cycle crashes (SIGKILL, OOM, power loss) instead of waiting for the end-of-cycle `SaveStateWithDB` batch.
- Existing timestamp-based dedup inside `SaveState` already prevents double-insert on the cycle-end flush.

Closes #289

## Test plan
- [x] New tests in `scheduler/trade_persist_test.go` cover single-row insert, append+persist, no-recorder fallback, no-double-insert, mid-cycle crash survival, and `ExecuteSpotSignal` end-to-end.
- [x] `go test ./...` passes (full scheduler suite).
- [x] `go build ./...` clean.

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -183,6 +183,25 @@ func (sdb *StateDB) Close() error {
 	return sdb.db.Close()
 }
 
+// InsertTrade persists a single trade row immediately (#289). This is invoked
+// via the tradeRecorder hook the moment a trade is appended to TradeHistory,
+// so trades survive mid-cycle crashes even if SaveState never runs.
+func (sdb *StateDB) InsertTrade(strategyID string, trade Trade) error {
+	if sdb == nil || sdb.db == nil {
+		return fmt.Errorf("state db unavailable")
+	}
+	_, err := sdb.db.Exec(`INSERT INTO trades
+		(strategy_id, timestamp, symbol, side, quantity, price, value, trade_type, details, exchange_order_id, exchange_fee)
+		VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+		strategyID, formatTime(trade.Timestamp), trade.Symbol, trade.Side,
+		trade.Quantity, trade.Price, trade.Value, trade.TradeType, trade.Details,
+		trade.ExchangeOrderID, trade.ExchangeFee)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


richkuo:
@claude

richkuo:
@claude pr code review with the latest changes

richkuo:
@claude pr code review again with the most recent changes

PR #299, opened 2026-04-17
fix(perps): guard live HL orders against ExecutePerpsSignal no-ops (#298)

### LINKED ISSUE #298: Race condition in ExecutePerpsSignal causes trades to execute without database records
### Summary

A critical race condition in `ExecutePerpsSignal` (scheduler/portfolio.go) causes trades to be executed on Hyperliquid but never recorded in go-trader's database. This creates a dangerous gap between on-chain reality and go-trader's internal state, affecting risk calculations, drawdown monitoring, and portfolio valuation.

### Impact

- **0.716 ETH (~$1,670) in unrecorded positions** currently live on Hyperliquid
- Risk metrics, drawdown %, and portfolio value are all incorrect
- Circuit breakers may not trigger when they should (or trigger when they shouldn't)
- False sense of security from go-trader's reported position sizes

### Root Cause

In `ExecutePerpsSignal`, the position check logic runs AFTER the Hyperliquid order executes but BEFORE the Trade record is created:

```go
// ExecutePerpsSignal flow:
1. Signal = BUY
2. adapter.market_open(symbol, is_buy, size)  // ← EXECUTES ON-CHAIN
3. if pos.Side == "long" {                    // ← CHECKS STALE DATA
       logger.Info("Already long %s, skipping buy")  // ← WRONG!
       return tradesExecuted, nil  // ← NO TRADE RECORD CREATED
   }
4. Create Trade record  // ← NEVER REACHED
```

The `Already long` check sees stale position data (from before the Hyperliquid call), thinks the position already exists, and aborts without creating a Trade record — but the trade already executed on-chain.

### Evidence from Logs

**16:28:12 UTC — First missed trade:**
```
[INFO] Placing live buy ETH size=0.237992
[INFO] Live fill at $2334.00 qty=0.238000
[INFO] Already long ETH (qty=0.212000), skipping buy   ← Trade executed but not recorded
[INFO] Status: ... trades=0
```

**16:43:15 — hl-sync detects the gap:**
```
[WARN] hl-sync: shared coin ETH reconciliation gap: virtual=0.212000 on-chain=0.450000 delta=-0.238000
```

**17:57:51 — Second missed trade (same pattern):**
```
[INFO] Placing live buy ETH size=0.239253
[INFO] Live fill at $2321.60 qty=0.239000
[INFO] Already long ETH (qty=0.212000), skipping buy
[INFO] Status: ... trades=0
```

**18:12:54 — Gap accumulates:**
```
[WARN] hl-sync: gap: virtual=0.212000 on-chain=0.928000 delta=-0.716000
```

**20:14:17 — hl-tema-eth-live worked correctly:**
```
[INFO] BUY ETH: 0.236000 @ $2357.70 (1.0x, notional $556.42, fee $0.19)
[INFO] Status: ... trades=1
```

### Current State

- Hyperliquid clearinghouse shows: **0.928 ETH position** at $2,335.16 entry
- go-trader positions table shows: **0.212 ETH**
- Gap: **0.716 ETH** (exactly the 3 missed buys)
- Unrealized PnL: **-$1.73** (not tracked by go-trader)

### Proposed Fix

Option 1: Move the position check BEFORE executing on-chain:
```go
if signal == 1 { // Buy
    if alreadyLong {
        return 0, nil  // Check BEFORE market_open
    }
    result = adapter.market_open(...)  // Execute only if safe
    // Then create Trade record unconditionally
}
```

Option 2: Always create Trade record after successful market_open:
```go
result = adapter.market_open(...)
if result.Success {
    // Always record trade, regardless of subsequent checks
    trade := Trade{...}
    s.TradeHistory = append(s.TradeHistory, trade)
}
```

Option 3: Use the returned fill data to update position state before any checks:
```go
result = adapter.market_open(...)
if fill, ok := extractFill(result); ok {
    // Update position state immediately with fill data
    s.Positions[symbol] = &Position{...}
    // Then proceed with Trade record creation
}
```

### Related

- Issue #288 — Closed positions history table
- Issue #292 — Margin-based drawdown calculation
- PR #254 — Perps margin-based accounting

### Priority

**CRITICAL** — This affects live trading accuracy and risk management.

richkuo:
## Summary

- Adds `PerpsOrderSkipReason(signal, posSide)` — pure helper that mirrors the skip conditions in `ExecutePerpsSignal`; live order helpers must consult it before spawning the Python executor
- `runHyperliquidExecuteOrder` now calls the guard at entry and returns `ok=false` (triggers `liveExecFailed`) when the order would be a no-op
- Captures `hlPosSide` alongside `hlPosQty` in Phase 1 RLock and threads it through

## Root Cause

`runHyperliquidExecuteOrder` placed the live on-chain order unconditionally, then `ExecutePerpsSignal` hit the "already long, skipping buy" branch and returned `trades=0` — dropping the Trade record and leaving virtual state permanently behind actual exchange positions. 0.716 ETH (3 fills × ~0.238) accumulated this way.

## Test Plan

- [ ] `TestPerpsOrderSkipReason` — 8 subtests covering every branch of the guard
- [ ] `TestExecutePerpsSignalAlreadyLongIsInertNoOp` — pins that the skip path is side-effect-free
- [ ] `go test ./...` in `scheduler/` passes

Closes #298

---
Generated with: Claude Opus 4.7 | Effort: 35


richkuo:
@claude

richkuo:
Addressed review feedback:

**1. Doc tightened (3d55566)** — `PerpsOrderSkipReason` comment now explicitly scopes to side-based no-ops and calls out why `s.Cash < 1` is intentionally not mirrored (can't be derived from `(signal, posSide)` alone, and live callers already guard cash upstream before the order).

**2. Follow-up issue filed** — #300 tracks extending the guard to `runOKXExecuteOrder`, `runRobinhoodExecuteOrder`, and `runTopStepExecuteOrder` so the CLAUDE.md invariant doesn't rot.

Keeping the PR #299 scope tight to HL as suggested.

richkuo:
@claude

PR #301, opened 2026-04-17
Extend PerpsOrderSkipReason guard to OKX/Robinhood/TopStep live helpers

### LINKED ISSUE #300: Extend PerpsOrderSkipReason guard to OKX / Robinhood / TopStep live order helpers
## Context

[PR #299](https://github.com/richkuo/go-trader/pull/299) fixed a race condition where \`runHyperliquidExecuteOrder\` placed live on-chain orders before \`ExecutePerpsSignal\` checked whether the signal was a no-op — dropping 0.716 ETH worth of Trade records (see #298).

CLAUDE.md was updated with the invariant:
> Live order helpers (\`runHyperliquidExecuteOrder\`, \`runOKXExecuteOrder\`, \`runRobinhoodExecuteOrder\`, \`runTopStepExecuteOrder\`) must check the same skip conditions as the corresponding \`ExecuteXxxSignal\` BEFORE spawning the Python executor; otherwise on-chain fills land with no Trade record (#298).

But only the Hyperliquid helper was fixed. The sibling helpers still carry the same bug class.

## Scope

Extend the guard pattern to the remaining three live order helpers.

### OKX — \`runOKXExecuteOrder\` (scheduler/main.go:1878)
- **Perps path**: same bug as HL — \`ExecutePerpsSignal\` skip-on-already-long lands after live fill.
- **Spot path**: \`ExecuteSpotSignal\` has the same shape (\"Already long ... skipping buy\" at portfolio.go:375).
- Capture \`okxPosSide\` alongside \`okxPosQty\` in Phase 1 RLock and thread it through.

### Robinhood — \`runRobinhoodExecuteOrder\` (scheduler/main.go:~1731)
- Uses \`ExecuteSpotSignal\` for spot crypto.
- Capture \`rhPosSide\` alongside \`rhPosQty\`.

### TopStep — \`runTopStepExecuteOrder\` (scheduler/main.go:~1575)
- Uses \`ExecuteFuturesSignal\`, which has the same \"Already long %s, skipping buy\" branch (portfolio.go:~506).
- Capture \`tsPosSide\` alongside \`tsContracts\`.

## Proposed Approach

1. Add \`SpotOrderSkipReason(signal, posSide)\` and \`FuturesOrderSkipReason(signal, posSide)\` peers of the existing \`PerpsOrderSkipReason\`. For spot/futures the guard is:
   - \`signal == 1 && posSide == \"long\"\` → skip (already long)
   - \`signal == -1 && posSide != \"long\"\` → skip (no long to sell)
2. Thread \`posSide\` through each helper and call the guard at entry (same pattern as \`runHyperliquidExecuteOrder\` in #299).
3. Add regression tests for the new guards (mirror \`TestPerpsOrderSkipReason\`).

## Priority

**Medium-high** — same impact profile as #298 (silent state divergence between exchange and DB), but less field evidence since OKX / Robinhood / TopStep live trading is less active than HL currently. Fixing now prevents the bug from being realized in production.

## Related

- #298 — root bug report and fix for Hyperliquid
- PR #299 — introduced \`PerpsOrderSkipReason\` and the CLAUDE.md invariant

richkuo:
Closes #300

## Summary

Before this change only `runHyperliquidExecuteOrder` guarded against the #298 bug class (live fill lands but `ExecuteXxxSignal` returns 0, so no `Trade` is recorded and in-memory state drifts silently behind actual exchange holdings). The sibling live helpers carried the same shape of bug because they placed the Python order before the in-memory skip branches ran.

- Add `SpotOrderSkipReason` and `FuturesOrderSkipReason` peers of `PerpsOrderSkipReason` mirroring the side-based no-op branches in `ExecuteSpotSignal` / `ExecuteFuturesSignal`.
- Thread `posSide` captured under Phase 1 RLock into `runOKXExecuteOrder`, `runRobinhoodExecuteOrder`, `runTopStepExecuteOrder`. `Position.Quantity` is always positive, so the prior `posQty<=0` checks could not distinguish flat from short — a sell while short would fire live but no-op in memory.
- OKX dispatches on `sc.Type` (perps vs spot) to pick the correct guard.
- Regression tests `TestSpotOrderSkipReason` and `TestFuturesOrderSkipReason` mirror `TestPerpsOrderSkipReason`.
- CLAUDE.md invariant updated to reference all three guards.

## Test plan

- [x] `go build .` from `scheduler/`
- [x] `go test ./...` — all pass (12.7s)
- [ ] Smoke test with a live OKX/Robinhood/TopStep config before production rollout


richkuo:
@claude

Issue #304, opened 2026-04-17
Backtest: reporting bugs and maintenance cleanup


richkuo:
## Summary

Grab-bag of **medium** and **low** findings from the backtest review. Individually minor, collectively a maintenance burden and source of misleading metrics.

### M1 — Signal encoding assumption is fragile
- **Where:** `backtest/backtester.py:105, 117`.
- **What:** Uses `signal == 1` / `signal == -1`. `position.diff()` emits floats, other strategies emit ints. Works today but is not documented and will silently break if any strategy emits e.g. `+2` for \"double-size\".
- **Fix:** Cast signal to int or document the contract; assert the expected domain.

### M2 — No HTF filter support in backtest
- **Where:** `backtest/run_backtest.py` (entire CLI).
- **What:** Live scheduler passes `--htf-filter` per strategy via `shared_tools/htf_filter.py`. Backtest has no equivalent flag, so backtests of HTF-filtered strategies overstate trade count / returns vs. live.
- **Fix:** Add `--htf-filter` flag; apply `htf_trend_filter` with a platform-appropriate `fetch_fn`.

### M3 — Sharpe annualization hardcoded to `sqrt(365)`
- **Where:** `backtest/backtester.py:178-184`; `backtest/backtest_options.py:367` (correct only for daily).
- **What:** A 4h backtest has ~`365*6` equity points/year, but the code multiplies by `sqrt(365)`, overstating Sharpe by `sqrt(6) ≈ 2.45×`.
- **Fix:** Derive the factor from the candle timeframe: `{\"1d\":365, \"4h\":365*6, \"1h\":365*24, ...}`.

### M4 — Strangle leg can be silently skipped with `max_positions=1`
- **Where:** `backtest/backtest_options.py:200-236`.
- **What:** With `max_positions=1`, the call is opened but the put guard at line 220 prevents opening the put. No warning is logged; the strangle becomes a naked call.
- **Fix:** Either enforce `max_positions >= 2` for strangles or log a warning and document.

### M5 — Options annualized return uses equity-curve length as days
- **Where:** `backtest/backtest_options.py:349`.
- **What:** `days = len(self.equity_curve)`; with `check_interval=7` the equity curve samples weekly, so a one-year run reports `days=52` → `years=0.14` → wildly inflated annualized return.
- **Fix:** Compute elapsed calendar days between first and last equity-curve timestamps.

### L1 — No tests for any backtest module
- **Where:** Entire `backtest/` tree.
[...]

Issue #303, opened 2026-04-17
Backtest: parity gaps with live trading system


richkuo:
## Summary

Four **high-severity** findings show the backtester has drifted from the live system's strategy registry, fees, and options pipeline. Strategies that pass backtest will not behave as scored when deployed live.

### H1 — Only the spot registry is loaded
- **Where:** `backtest/run_backtest.py:13`, `backtest/backtester.py:14`, `backtest/optimizer.py:14`, `backtest/reporter.py:13` — all hardcode `sys.path.insert(0, ...'shared_strategies/spot')`.
- **What's missing:** `shared_strategies/futures/strategies.py` (perps/futures registry, includes `delta_neutral_funding` and futures variants of every spot strategy).
- **Also affected:** `optimizer.DEFAULT_PARAM_RANGES` (lines 187–236) covers only the original 10 spot strategies. Newer strategies (`supertrend`, `stoch_rsi`, `atr_breakout`, `amd_ifvg`, `order_blocks`, `vwap_reversion`, `chart_pattern`, `liquidity_sweeps`, `sweep_squeeze`, `psar`, `ichimoku`, `squeeze_momentum`, `pairs_spread`) are unsupported — `run_walk_forward` silently returns `None`.
- **Fix:** Add `--platform spot|futures` flag, load the correct registry conditionally. Backfill param ranges for every registered strategy.

### H2 — Commission hardcoded to 0.1% regardless of platform
- **Where:** `backtest/backtester.py:66` — `commission_pct: float = 0.001`.
- **What's wrong:** Live uses `CalculatePlatformSpotFee` with HL 0.035%, RH 0%, Luno 1%, BinanceUS 0.1%. Cross-platform strategy ranking is biased toward venues with lower actual fees because the backtest taxes them equally at BinanceUS rates.
- **Fix:** Accept `platform` in `Backtester.__init__`, map to the same rate table as `scheduler/fees.go`. Thread `--platform` through `run_backtest.py`.

### H3 — Walk-forward optimizer has no warmup buffer
- **Where:** `backtest/optimizer.py:83-88`.
- **What's wrong:** Each fold cuts `window_df = df.iloc[start_idx:end_idx]` and then further to `train_df`. On a 500-bar × 5-split run, each window is 100 bars. An SMA-80 from `DEFAULT_PARAM_RANGES` needs 80 bars of warmup — first ~80 signals are NaN, producing zero trades and `sharpe=0` that still \"wins\" if nothing else generates trades either.
[...]

PR #305, opened 2026-04-17
fix(backtest): critical correctness bugs — look-ahead, IV math (#302)

### LINKED ISSUE #302: Backtest: critical correctness bugs (look-ahead, IV math)
## Summary

A comprehensive review of `backtest/` surfaced three **critical** bugs that silently produce wrong results. Fixing these is a prerequisite to trusting any backtest output — including the numbers recorded in `backtest/THETA_HARVEST_RESULTS.md`.

### C1 — Look-ahead bias on `position.diff()` strategies
- **Where:** `backtest/backtester.py:96-128`
- **What:** Signals produced by `position.diff()` (sma_crossover, ema_crossover, triple_ema, macd, volume_weighted) are emitted on the same bar whose close drove the crossover. The backtester then fills at `row[\"close\"]` on that same bar (line 97), so the signal price equals the fill price.
- **Why it matters:** Live execution fires on the *next* bar's open (scheduler reads the last-closed bar's signal, then market-orders). Backtest is therefore structurally more profitable than live on every crossover strategy.
- **Fix:** Shift signals by 1 (`df[\"signal\"] = df[\"signal\"].shift(1).fillna(0)`) and fill at `row[\"open\"]` — or enforce this policy inside `Backtester.run()`.

### C2 — Historical volatility uses simple returns and a non-variance formula
- **Where:** `backtest/backtest_options.py:66-73` (also inherited by `backtest_theta.py`)
- **What:**
  ```python
  returns = [(closes[i] - closes[i-1]) / closes[i-1] for i in range(-window, 0)]
  variance = sum(r**2 for r in returns) / len(returns)
  ```
  Two compounded errors:
  1. Simple returns instead of log returns (`math.log(c[i]/c[i-1])`) — overstates vol for large crypto moves.
  2. `sum(r**2)` is the mean of squared returns, not variance — only correct if mean return is exactly zero, which is false over a 14-day BTC window.
- **Why it matters:** Inflated vol → Black-Scholes overprices every option → every \"sell premium\" decision looks more profitable than it is. The theta-harvest comparison in `THETA_HARVEST_RESULTS.md` is built on this input and should be considered invalid until C2 is fixed.
- **Fix:** Use log returns and subtract the mean before squaring (`(r - mean)**2`).

### C3 — `iv_rank` is actually an IV ratio, not a percentile rank
- **Where:** `backtest/backtest_options.py:76-85`
- **What:**
  ```python
  return min(max((recent_vol / max(hist_vol, 0.001)) * 50, 0), 100)
  ```
  True IV rank = `(current_IV - 52w_low) / (52w_high - 52w_low) * 100`. The current formula reaches 100 when recent_vol is just 2× hist_vol, not at a 52-week high.
- **Why it matters:** Live `VolMeanReversionStrategy` calls `adapter.get_iv_rank()` which uses a proper 60-day lookback percentile. The backtester triggers entries at entirely different moments than live, so conclusions do not transfer.
- **Fix:** Implement the percentile-based IV rank matching the adapter (or call the adapter's helper if data allows).

## Acceptance criteria

- [ ] C1: Signals shifted; new regression test asserts fill price equals the *next* bar's open for a synthetic crossover scenario.
- [ ] C2: Log returns + variance-around-mean; pytest compares vol output against `numpy.std(log_returns, ddof=0)` within tolerance.
- [ ] C3: IV rank implementation matches a percentile of a trailing window; pytest covers endpoints (0, 50, 100) and the \"flat vol\" case.
- [ ] `THETA_HARVEST_RESULTS.md` regenerated (or annotated as superseded) after fixes land.

## References

- Review context: CLAUDE.md \"Testing\" section — \"New functionality must include tests\"; bug fixes should include a regression test.
- Related issues being opened: parity/design (H-series), reporting & maintenance (M/L-series).

richkuo:
Closes #302.

Three critical backtest correctness bugs identified in the comprehensive backtest review. One commit per item for reviewability.

## Summary

- **C1 — Look-ahead bias** (0f09cd6): `Backtester.run` filled trades at the same bar's close that produced the signal, leaking future info into crossover strategies (sma/ema/triple_ema/macd/volume_weighted). Now shifts the signal column by +1 and fills at the next bar's open; falls back to close when only `close` is present. 4 regression tests in `backtest/tests/test_backtester_fills.py`.
- **C2 — Historical vol math** (eb47b3f): `calc_historical_vol` used simple returns and `sum(r**2)/n` (mean of squared returns, not variance). Replaced with log returns + population variance around the sample mean, matching `numpy.std(log_returns, ddof=0) * sqrt(365)`. `backtest_theta.py` inherits via its import. `THETA_HARVEST_RESULTS.md` annotated as superseded. 4 tests in `backtest/tests/test_options_vol_math.py`.
- **C3 — IV rank formula** (30acd53): `calc_iv_rank` returned `(recent/hist) * 50` — an IV ratio halved and clipped, not a percentile rank. Replaced with a true percentile: `(current - min) / (max - min) * 100` over a trailing 60-day distribution of rolling 14-day realised vol, matching `adapter.get_iv_rank()` shape used by the live `VolMeanReversionStrategy`. 6 tests in `backtest/tests/test_options_iv_rank.py`.

## Test plan

- [x] `uv run pytest` passes (326 tests; +14 new).
- [x] `python3 -m py_compile backtest/backtest_options.py backtest/backtest_theta.py backtest/backtester.py`.
- [x] Added `backtest/tests` to `testpaths` in `pyproject.toml`.
- [ ] Rerun `backtest_theta.py` against real data once this lands and update `THETA_HARVEST_RESULTS.md` with corrected figures (tracked in #304).

## Notes

- This PR is restricted to the three critical items in #302. Parity gaps (H-series) are in #303, reporting & cleanup (M/L-series) in #304.
- Acceptance criteria from #302 all addressed: shifted-signal regression test ✔, vol parity with numpy ✔, IV-rank percentile endpoints covered ✔, THETA_HARVEST_RESULTS annotated ✔.

---
Generated with: Claude Sonnet 4.6 | Effort: high

### DIFF
--- a/backtest/THETA_HARVEST_RESULTS.md
+++ b/backtest/THETA_HARVEST_RESULTS.md
@@ -1,12 +1,9 @@
 # Theta Harvesting Backtest Results
 
-> ⚠️ **SUPERSEDED — 2026-04-17.** These results were produced with the
-> pre-fix volatility and IV-rank math (issue #302). The old `calc_historical_vol`
-> used simple returns and `sum(r**2)/n` (not variance around the mean), and
-> `calc_iv_rank` returned an IV *ratio* rather than a percentile rank. Both
-> inflated Black-Scholes premiums and skewed entry timing, so the numbers
-> below should not be treated as live-comparable. Rerun after the #302 C2/C3
-> fixes ship before using any figure in this document.
+> ⚠️ **SUPERSEDED — 2026-04-17.** These figures were produced against a
+> pre-issue-#302 version of `calc_historical_vol` and `calc_iv_rank` that
+> inflated premiums and skewed entry timing. Rerun `backtest_theta.py`
+> against the current implementations before citing any number below.
 
 **Date:** 2026-02-11
 **Data:** Binance US, 2023-01-01 to 2026-02-11 (1,048 days)
--- a/backtest/THETA_HARVEST_RESULTS.md
+++ b/backtest/THETA_HARVEST_RESULTS.md
@@ -1,5 +1,13 @@
 # Theta Harvesting Backtest Results
 
+> ⚠️ **SUPERSEDED — 2026-04-17.** These results were produced with the
+> pre-fix volatility and IV-rank math (issue #302). The old `calc_historical_vol`
+> used simple returns and `sum(r**2)/n` (not variance around the mean), and
+> `calc_iv_rank` returned an IV *ratio* rather than a percentile rank. Both
+> inflated Black-Scholes premiums and skewed entry timing, so the numbers
+> below should not be treated as live-comparable. Rerun after the #302 C2/C3
+> fixes ship before using any figure in this document.
+
 **Date:** 2026-02-11
 **Data:** Binance US, 2023-01-01 to 2026-02-11 (1,048 days)
 **Strategy:** Vol Mean Reversion (sell strangles on high IV, buy straddles on low IV)
--- a/backtest/backtest_options.py
+++ b/backtest/backtest_options.py
@@ -98,7 +98,6 @@ def calc_iv_rank(closes: list, recent_window: int = 14,
     merely 2× historical vol rather than at a true lookback high. That
     triggered strangles at entirely different moments than live.
     """
-    # Need enough bars for a full rolling window plus lookback.
     if len(closes) < recent_window + lookback_days + 1:
         return 50.0
 
@@ -107,18 +106,12 @@ def calc_iv_rank(closes: list, recent_window: int = 14,
     ]
 
     def _rolling_vol(end_idx: int) -> float:
-        # Annualised realised vol of the ``recent_window`` log returns ending
-        # at index ``end_idx`` (inclusive). Uses variance around the sample
-        # mean — same formula family as calc_historical_vol.
         window = log_returns[end_idx - recent_window + 1: end_idx + 1]
         mean = sum(window) / len(window)
         variance = sum((r - mean) ** 2 for r in window) / len(window)
         return math.sqrt(variance * 365)
 
-    # Current realised vol — most recent full window.
     current = _rolling_vol(len(log_returns) - 1)
-
-    # Trailing rolling-vol distribution over the lookback window.
     history = [
         _rolling_vol(end)
         for end in range(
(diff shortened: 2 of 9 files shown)


richkuo:
Addressed review feedback in 18a729d (pushed).

**Important (all 3):**
- **Live-parity for C3**: `test_matches_live_okx_adapter_shape` in `test_options_iv_rank.py` now runs an inline reproduction of `platforms/okx/adapter.py` `get_vol_metrics` IV-rank logic against the backtest function on the same closes, asserting float equality. Parity is now enforced by CI, not English.
- **Bar-0 head guard**: replaced the incorrect "never fills" assertion with `test_signal_on_bar_zero_fills_on_bar_one` — a raw signal at index 0 is shifted to bar 1 and fills at bar 1's open. The original review suggestion was based on a misread of `shift(1)` semantics; the head boundary is genuinely that the signal moves, not that it disappears.
- **End-to-end golden file**: `test_backtester_end_to_end.py` runs the full `Backtester` with commission + slippage on a deterministic synthetic OHLC + SMA-crossover signals and pins `total_trades=6`, `final_capital=923.53`, `total_return_pct=-7.51`, `sharpe=-0.686`, `max_dd=-17.66`. A regression that fixes each unit test in isolation but drifts the integrated result will fail here.

**Suggestions:**
- Boundary tests at exact thresholds (`test_exact_minimum_length_computes_vol`, `test_one_below_minimum_returns_default`, `test_boundary_exact_minimum_computes_rank`, `test_boundary_one_below_minimum_returns_default`).
- Consecutive-signal guards (`test_buy_signal_while_long_is_ignored`, `test_sell_signal_while_flat_is_ignored`).
- Shares / final-cash assertions added to the look-ahead tests.
- Trimmed redundant WHAT-comments in `backtester.py:109,113` and `backtest_options.py:92,114,117`. Kept the WHY-docstrings on `calc_historical_vol` / `calc_iv_rank` — they document the specific math trap future contributors would otherwise "simplify" back.
- Trimmed narrating comments in test files.
- Dropped `C1/C2/C3` shorthand from test module docstrings (kept `#302`).
- Reworded `THETA_HARVEST_RESULTS.md` banner to not reference "after fixes ship".

**Test count**: 14 → 23 under `backtest/tests/`. Full suite 335 passing.

richkuo:
@claude

Issue #310, opened 2026-04-18
Leaderboard: remove per-product sections from PrecomputeLeaderboard


richkuo:
## Summary

When  is set (dedicated channel for full leaderboard posts), the  function currently generates 6 sections:

1. top10
2. bottom10
3. spot
4. perps
5. options
6. futures

The per-product sections (spot, perps, options, futures) add noise when the dedicated channel is meant to show a clean top/bottom aggregate view across all strategies.

## Request

Remove the per-product breakdown sections (spot, perps, options, futures) from  so that when posting to , only  and  are included.

**Files to modify:**
-  —  and 

**Behavior after change:**
-  posts → top10 + bottom10 only (aggregate across all strategies)
- Individual platform channels → unchanged (they still receive per-product summaries via )
-  → unchanged

Reference only: LLM model used: MiniMax-M2.7

richkuo:
@claude

PR #315, opened 2026-04-18
Make leaderboard compute on-demand instead of per-cycle

### LINKED ISSUE #313: Make PrecomputeLeaderboard on-demand instead of per-cycle
## Summary

Currently  runs at the end of every scheduler cycle (~5 min), even though the precomputed  is only used for:
- Daily  post at  (e.g. 09:00 UTC)
- Manual  flag invocations

The per-product breakdown (spot/perps/options/futures) was removed in #310, so the file now only holds  + . Computing this every 5 min is wasteful.

## Request

Move  to compute on-demand, right before posting. Specifically:

1. **In ** — call  (or inline compute) immediately before posting, so it always posts fresh data without a pre-compute cycle overhead

2. **In the  cycle logic** —  already computes fresh, so no change needed there

3. **In  flag** — the flag already triggers , so adding on-demand compute there is sufficient; remove the cycle-time pre-compute call

**Benefits:**
- Eliminates unnecessary computation on every cycle
- Data is always fresh at post time (no staleness concern)
- Simpler mental model: leaderboard data is computed exactly when needed

**Files to modify:**
-  — remove  call from end-of-cycle, add to  path
-  — possibly refactor to allow inline compute in 

richkuo:
Closes #313

## Summary

Previously `PrecomputeLeaderboard` ran at the end of every scheduler cycle (~5 min) and wrote `leaderboard.json`, but the data is only consumed by the daily cron post and the `--leaderboard` flag. The per-product breakdown was removed in #310, so every cycle was serializing just top10/bottom10 to disk for no benefit.

- Remove both per-cycle `PrecomputeLeaderboard` calls plus all on-disk `leaderboard.json` machinery (`LoadLeaderboard`, `leaderboardPath`, `LeaderboardData` type).
- Split compute from post: `BuildLeaderboardMessages(cfg, state, prices)` produces the messages; `postLeaderboardMessages` / `PostLeaderboard` send them. The cycle path builds under RLock and posts outside the lock; the `--leaderboard` flag fetches prices via the existing `fetchPricesForSummary` helper and posts directly.
- Update tests to exercise the in-memory API and drop the file-based fixtures.

## Test plan

- [x] `go build .` in `scheduler/`
- [x] `go test ./...`
- [ ] Verify `--leaderboard` flag posts fresh top10/bottom10 to Discord
- [ ] Verify daily auto-post still fires at configured UTC time


richkuo:
@claude

richkuo:
@claude sonnet create new github issues for the observations

PR #325, opened 2026-04-18
leaderboard: persist LastLeaderboardPostDate immediately after post

### LINKED ISSUE #319: leaderboard: LastLeaderboardPostDate not immediately persisted — crash can cause duplicate daily post
`LastLeaderboardPostDate` is set in memory under `mu.Lock()` (`main.go:964`) but only flushed to SQLite on the next cycle's `SaveStateWithDB`. A crash between the post and the next save causes the daily post to fire again on restart.

## Location

`scheduler/main.go:964` and `scheduler/db.go` (`SaveStateWithDB`)

## Notes

- Pre-existing behavior — this was true before PR #315 as well.
- Now that the leaderboard path is cleaner, worth tracking as a known gap.
- Fix: call `SaveStateWithDB` immediately after setting `LastLeaderboardPostDate`, or persist it in a separate targeted DB write.

Spotted during review of PR #315.

richkuo:
Previously `LastLeaderboardPostDate` was set in memory but only flushed to SQLite on the next cycle's `SaveStateWithDB`. A crash between the post and the next save could cause the daily leaderboard to post again on restart.

Fix: save state immediately inside the same `mu.Lock()` that sets the field.

Closes #319

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -965,8 +965,14 @@ func main() {
 			if err := postLeaderboardMessages(lbMessages, notifier); err != nil {
 				fmt.Printf("[WARN] Leaderboard auto-post failed: %v\n", err)
 			} else {
+				// Issue #319: persist LastLeaderboardPostDate immediately so a
+				// crash before the next cycle's SaveState cannot cause a duplicate
+				// daily post on restart.
 				mu.Lock()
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

PR #337, opened 2026-04-18
Flag perps shorts held under AllowShorts=false at startup (#336)

### LINKED ISSUE #336: perps state-vs-config gap: short position held under AllowShorts=false strategy
## Background

Carryover from PR #330 follow-up review. A perps strategy configured `AllowShorts=false` (legacy long-only — the default, used by `triple_ema`, `rsi_macd_combo`, all pre-#328 perps strategies) cannot itself open a short. But a short position can still end up in scheduler state via:

- **State migration** from a pre-#328 SQLite schema
- **Paper→live handoff** where paper state carried shorts
- **Operator manual edit** of `state.db`
- **Config change** — a strategy previously ran `AllowShorts=true`, opened a short, then got re-configured to `false` without closing first

## The desync

When a migrated short exists and the long-only strategy then emits `signal=1`:

| Layer | Action |
|-------|--------|
| `perpsLiveOrderSize` (no flip path because `!AllowShorts`) | fresh-open size, e.g. `budget/price = 0.475` |
| Exchange | buy of 0.475 lands against the 0.5 short → nets to **0.025 short remaining**, no long opened |
| `ExecutePerpsSignal` | sees short + signal=1 + fillQty=0.475 → closes entire 0.5 short virtually, opens 0.475 long |

Virtual state says `long 0.475`. Exchange state says `short 0.025`. Same class of desync as #298.

## Why it's not an executor fix

The executor would have to special-case a state the strategy isn't supposed to be in. The cleaner fix is upstream:

1. **Startup validation** — detect `position.Side == \"short\" && !strategy.AllowShorts` when loading state; either force-close, require explicit operator ack, or refuse to start.
2. **Migration path** — the pre-#328 schema had no concept of shorts, so a migration auditor could flag or force-close inherited shorts at the migration boundary.
3. **Config-change guard** — when `AllowShorts` transitions true→false, require a \"close all shorts\" step first (or auto-close on next cycle with loud logging).

## Repro

Manually seed `state.db` with a short position for a strategy that has `allow_shorts: false` in config, start the scheduler, wait for a `signal=1` cycle. Observe:
- Exchange order size doesn't match virtual close+open qty.
- Partial-fill on exchange vs clean flip in virtual state.
- Next `syncHyperliquidAccountPositions` will fight the scheduler about whether the position exists.

## Scope

- `scheduler/state.go` or `scheduler/config.go`: add a \"short-vs-AllowShorts\" check
- Decide policy: force-close, warn-and-stop, warn-and-continue
- Test seeding: write a `_test.go` that exercises the migration-path desync directly

## References

- PR #330: https://github.com/richkuo/go-trader/pull/330 (final review)
- Executor fee-accounting test that documents the gap: `TestExecutePerpsSignal_FlipDoesNotDoubleCountFee` (comment explicitly lists \"state migration, paper→live handoff, or a future adapter that opens shorts\" as the only ways to hit the branch)
- Sibling desync classes: #298, #300

richkuo:
Adds `ValidatePerpsAllowShortsConfig` startup check for the state-vs-config gap described in #336: a short position in state under a perps strategy configured with AllowShorts=false. The executor's fresh-open sizing path cannot reconcile this — a buy nets against the short on-exchange but flips virtually, leaving virtual state desynced. Policy is warn-and-continue with stderr log + owner DM, matching `ValidateState`. Includes regression tests for the gap case, the long-only baseline, and orphan state.

Close #336

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -87,6 +87,12 @@ func main() {
 		}
 	}
 
+	// #336: Detect perps shorts held under AllowShorts=false strategies. The
+	// executor can't reconcile this on its own — a fresh-open buy against an
+	// existing short nets on the exchange but flips virtually. Collect here,
+	// forward to owner DM once the notifier is wired below.
+	allowShortsWarnings := ValidatePerpsAllowShortsConfig(state, cfg)
+
 	// #42 / #243: Initialize portfolio peak from sum of capitals on first run.
 	// For strategies that share an exchange wallet (e.g. multiple Hyperliquid
 	// perps strategies on the same account), use the real on-exchange balance
@@ -191,6 +197,14 @@ func main() {
 	// to stderr via the nil-check in state.go.
 	if notifier.HasOwner() {
 		tradePersistWarn = func(msg string) {
+			notifier.SendOwnerDM("[state] " + msg)
+		}
+	}
+
+	// #336: Forward startup AllowShorts warnings to the owner so the desync is
+	// surfaced even when the operator isn't tailing stderr.
+	if len(allowShortsWarnings) > 0 && notifier.HasOwner() {
+		for _, msg := range allowShortsWarnings {
 			notifier.SendOwnerDM("[state] " + msg)
 		}
 	}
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

PR #330, opened 2026-04-18
Add triple_ema_bidir strategy for bidirectional perps trend trading

richkuo:
## Summary
- Add `triple_ema_bidir` in `shared_strategies/futures/strategies.py` — long on bullish stack, short on bearish stack, signal clamped to `{-1, 0, 1}`.
- Register `temab` short name and add to `defaultFuturesStrategies` + `defaultPerpsStrategies` in `scheduler/init.go`.
- Tests cover uptrend/downtrend/flat, direct flip clamping, custom params, registry, and edge cases.

## Why

The existing `triple_ema` is structurally long-only — `position ∈ {0, 1}` so `signal.diff()` emits `-1` only as a long-exit, never a short entry. `hl-tema-eth` sat at position=0 through ETH's 30m dump on 2026-04-18 even with `ema_short 2376 < ema_mid 2393 < ema_long 2394` fully bearish.

Live verification against Hyperliquid public `candles_snapshot` confirms the new strategy entered short at 2026-04-18 10:59 UTC (position=-1, signal=-1) and held through the latest bar — exactly the short `hl-tema-eth` would have missed.

## Test plan
- [x] `uv run pytest shared_strategies/` — 301 passed
- [x] `cd scheduler && go test ./...` — passed (`ok trading-scheduler 4.013s`)
- [x] `cd scheduler && go build .` — clean
- [x] `gofmt -l` — clean
- [x] Strategy discovery: `shared_strategies/futures/strategies.py --list-json` includes `triple_ema_bidir`
- [x] Live ETH 30m replay on Hyperliquid: position=-1 in last 2 bars, signal=-1 on entry bar, all signals ⊆ `{-1, 0, 1}`
- [x] End-to-end: `.venv/bin/python3 shared_scripts/check_hyperliquid.py triple_ema_bidir ETH 30m --mode=paper` returns valid JSON
- [x] Init wizard: `./go-trader init --json '{..."perpsStrategies":["triple_ema_bidir"]...}'` generates `hl-temab-eth` config correctly

## Config migration

Users wanting bidirectional behavior swap their strategy entry:

```json
{ "id": "hl-temab-eth", "strategy": "triple_ema_bidir", ... }
```

Existing `hl-tema-eth` (long-only `triple_ema`) keeps working unchanged — no breaking change.

Closes #328

---
Generated with: Claude Opus 4.7 | Effort: high

PR #342, opened 2026-04-19
Submit live HL closes when portfolio kill switch fires (#341)

### LINKED ISSUE #341: Portfolio kill switch fails to close shared-coin positions on Hyperliquid
## Bug Description

Portfolio kill switch fires but does **not** actually close any live Hyperliquid positions. `forceCloseAllPositions` only mutates virtual state (cash, `s.Positions`, trade history) — it never calls the HL exchange API. The shared-coin reconcile skip in `reconcileHyperliquidAccountPositions` makes the symptom worse for shared coins (e.g. ETH traded by `hl-tema-eth-live` + `hl-rmc-eth-live`) because the virtual position is empty too, so even the post-fire reconcile noise hides the on-chain leftover behind a "shared coin gap" warning.

### 1. `forceCloseAllPositions` is virtual-only (`scheduler/risk.go:549`)

```go
func forceCloseAllPositions(s *StrategyState, prices map[string]float64, logger *StrategyLogger) {
    for symbol, pos := range s.Positions {
        // ... compute pnl, update s.Cash
        RecordTrade(s, trade)         // local DB insert only
        recordClosedPosition(...)
        delete(s.Positions, symbol)   // virtual delete
    }
    // same for s.OptionPositions
}
```

`RecordTrade` (`state.go:40`) appends to `TradeHistory` and persists via `tradeRecorder` — there is no path from here into `runHyperliquidExecuteOrder` / `runOKXExecuteOrder` / `runRobinhoodExecuteOrder` / `runTopStepExecuteOrder`. The on-chain position is untouched on every platform, not just HL — HL is just where we noticed first because of the shared-wallet visibility.

The kill-switch firing block in `main.go:526-534` calls `forceCloseAllPositions` for every strategy and then sets `killSwitchFired = true`, which causes the rest of the cycle to skip the live-execution path (`if !killSwitchFired` at `main.go:611`). So no live close orders are ever attempted.

### 2. Shared-coin reconcile skip (`scheduler/hyperliquid_balance.go:319-321`)

```go
if sharedCoins[sym] {
    continue // handled below
}
```

The "handled below" block (lines 339-416) only:
- Applies non-destructive multiplier/leverage corrections
- Computes a reconciliation gap and stores it in `state.ReconciliationGaps`
- Logs `[WARN] hl-sync: shared coin %s reconciliation gap: ...`

It never assigns the on-chain qty to any strategy's `s.Positions[sym]`. So for shared coins, even a "correct" `forceCloseAllPositions` would have nothing to iterate.

## Failure Chain at Kill Switch Fire (corrected)

1. `fetchHyperliquidState` returns on-chain ETH=0.517 long.
2. `reconcileHyperliquidAccountPositions` skips ETH (shared coin) → virtual state stays empty for both `hl-tema-eth-live` and `hl-rmc-eth-live`.
3. `perpsMarginDrawdownInputs` computes 22.1% drawdown from on-chain marks but reads virtual positions = empty (or stale).
4. Portfolio kill switch triggers (drawdown > 20%).
5. `forceCloseAllPositions` iterates `s.Positions`. ETH not present → nothing to close. **Even if ETH were present**, only virtual state would be mutated — no HL API call would be made.
6. Real on-chain position survives untouched. Subsequent reconcile cycles re-emit the shared-coin gap warning.

## Evidence

- Kill switch fires at 04:37:45: `portfolio perps margin drawdown 22.1% exceeds limit 20.0%`
- `forceCloseAllPositions` invoked for both `hl-tema-eth-live` and `hl-rmc-eth-live`
- Cycle 5131 completes in 1.1s with **0 trades** (consistent with no exchange calls)
- On-chain check after kill switch: ETH `szi=0.517` still open
- Recurring log every ~5 minutes:
  ```
  [WARN] hl-sync: shared coin ETH reconciliation gap: virtual=0.000000 on-chain=0.517000
  ```

## Scope

This is broader than shared coins. **Every live position on every live platform (HL, OKX, Robinhood, TopStep) is currently abandoned by the portfolio kill switch.** Shared coins are the most acute symptom because reconcile cannot recover the orphaned position on the next cycle (it's intentionally ignored to avoid the #258 phantom drawdown). For non-shared coins, the on-chain position would also survive the kill switch, but the next reconcile would at least re-sync the virtual state to match — which then traps the strategy with a position it didn't open in this lifecycle (the comment at `hyperliquid_balance.go:235-236` explicitly says "If on-chain exists but NOT in this strategy's state, we skip it").

## Suggested Fix

Only one of the originally listed options actually fixes the underlying bug:

> **Call the exchange API directly from the kill switch.** When the portfolio kill switch fires, fetch current on-chain positions per platform and submit reduce-only close orders directly via the live execute helpers (or a new `forceCloseLive*` helper that mirrors `runXxxExecuteOrder`'s order-placement path without the per-strategy guards). Apply this to all live platforms, not just HL.

Options 1 (track shared positions at portfolio level) and 3 (assign shared coin to a primary strategy) are partial — they restore visibility but still rely on `forceCloseAllPositions` to actually liquidate, which it cannot do today.

### Implementation sketch

- Add `forceCloseLivePositions(cfg, state)` that, for each live strategy, dispatches by `sc.Platform` to a platform-specific live close (mirroring the existing `runXxxExecuteOrder` skip-reason pattern but unconditional).
- For HL specifically, fetch `clearinghouseState` once, then submit reduce-only market closes for every non-zero `assetPosition` belonging to a configured strategy. This avoids the shared-coin ownership ambiguity entirely — the kill switch closes whatever is on-chain.
- Only after live closes confirm should `forceCloseAllPositions` mutate virtual state (or have the next reconcile cycle do it via the existing "no longer on-chain, removing" branch at `hyperliquid_balance.go:223-234`).
- Surface failures (live close API error) prominently — kill switch should remain latched until all on-chain positions for live platforms are confirmed flat.

### Tests to add

- Unit: kill-switch path with a live HL strategy holding a non-shared coin → asserts the live-close helper was invoked (mock the HL execute path).
- Unit: kill-switch path with a shared coin where neither strategy's `Positions` has an entry → asserts the live close still fires for the on-chain qty.
- Regression: existing #258 phantom-drawdown scenario must still pass (kill switch must not be tripped by per-strategy reconciliation drift on shared coins).

Reference only: original LLM model used: MiniMax-M2.7. Revised by Claude Opus 4.7 after verifying claims against `scheduler/risk.go`, `scheduler/hyperliquid_balance.go`, `scheduler/state.go`, and `scheduler/main.go`.

richkuo:
## Summary

- Portfolio kill switch now submits reduce-only `market_close` to Hyperliquid for every non-zero on-chain position belonging to a configured live HL strategy. Previously `forceCloseAllPositions` only mutated virtual state and the on-chain position survived untouched.
- Latch-until-flat: virtual state is mutated only when **all** live closes succeed. Any error keeps `PortfolioRisk.KillSwitchActive=true` so the next cycle retries; per-coin errors surface in Discord + stderr.
- Closes on-chain qty directly (ignoring per-strategy ownership) so shared coins are liquidated even though `reconcileHyperliquidAccountPositions` deliberately leaves their virtual state empty (#258).
- Net-zero szi is treated as already-flat and never submitted — protects bidirectional perps (#330) where two strategies on the same wallet hold equal-and-opposite virtual exposure that nets to zero on-chain.

## Scope note

HL only this PR. OKX / Robinhood / TopStep have the same underlying bug but route order submission through subprocesses without a native Go on-chain position fetcher — adding them safely to the kill-switch hot path is its own design problem and would balloon this PR. Worth follow-up issues for each.

## Architecture

```
[CheckPortfolioRisk → !portfolioAllowed]
    ↓ (under Lock — set killSwitchFired, do NOT mutate virtual state)
[Unlock]
    ↓
[forceCloseHyperliquidLive(hlPositions, hlLiveAll, defaultHyperliquidLiveCloser)]
    ↓ subprocess: shared_scripts/close_hyperliquid_position.py per non-zero coin
    ↓ returns HyperliquidLiveCloseReport{ClosedCoins, AlreadyFlat, Errors}
    ↓
[if Errors empty AND hlStateFetched OK → mutate virtual state under Lock]
[else → keep latched, owner DM, retry next cycle]
```

`HyperliquidLiveCloser` is a function type so tests inject a fake without spawning Python.

## Test plan
- [x] `go test ./...` passes (full scheduler suite, 3.8s)
- [x] `uv run pytest shared_strategies/ shared_tools/ platforms/` passes (630 tests)
- [x] `python3 -m py_compile shared_scripts/close_hyperliquid_position.py`
- [x] `gofmt -l scheduler/*.go` clean
- [x] `go build` produces working binary
- [x] 7 new unit tests covering: non-shared coin happy path, shared-coin-with-empty-virtual (the #341 root cause), net-zero szi already-flat, short positions, partial close failure, unowned coin ignored, empty inputs no-op
- [ ] **Manual verification needed before merge**: trigger kill switch on a live HL paper-trading account with a small open position and confirm (a) the close order appears on HL, (b) virtual state stays intact if the close errors, (c) latched state retries on the following cycle.

## Files changed
- `shared_scripts/close_hyperliquid_position.py` (new) — wraps `HyperliquidExchangeAdapter.market_close`, JSON envelope, exit 1 on error
- `scheduler/executor.go` — `RunHyperliquidClose` + result types
- `scheduler/hyperliquid_balance.go` — `HyperliquidLiveCloser` type, `defaultHyperliquidLiveCloser`, `forceCloseHyperliquidLive`
- `scheduler/main.go` — kill-switch path rewired with latch-until-flat semantics
- `scheduler/hyperliquid_balance_test.go` — 7 new unit tests

Closes #341

---
Generated with: Claude Opus 4.7 | Effort: high


richkuo:
@claude

richkuo:
@claude review the latest changes, and see why the go CI check fails

richkuo:
@claude sonnet

PR #344, opened 2026-04-19
Guard initial_capital against silent overwrite in SaveState

### LINKED ISSUE #343: Initial capital values should never be recalculated without explicit user instruction
## Problem

During live trading operations, `initial_capital` / `EffectiveInitialCapital` values were modified multiple times by an operator agent while adjusting strategy state. These changes silently rewrote the PnL baseline, making historical performance numbers meaningless.

Specific incidents:
1. DB `initial_capital` overwritten from $505 to $632/$626 during state restoration
2. DB `initial_capital` overwritten again to $621.29 during position close
3. `config.json` `initial_capital` overwritten from $505 to $632/$626, then to $621.29, then back to $505

Each change shifted the PnL baseline, causing the Discord summary to show wildly different returns (+23%, +0%, -2.5%) for the same portfolio state.

## Expected Behavior

- `initial_capital` in both config.json and SQLite should be treated as **immutable baseline**
- Recalculating summaries, closing positions, restarting the trader, or adjusting cash values must **never** modify `initial_capital`
- The only valid way to change `initial_capital` is an explicit user command (e.g. a CLI flag or config edit)

## Suggested Fix

Add a guard in `SaveStateWithDB` (or wherever strategies are persisted) that preserves the existing `initial_capital` value unless an explicit override flag is passed:

```go
// Never overwrite initial_capital during normal state saves
if newInitialCapital > 0 && newInitialCapital != existingInitialCapital {
    if !explicitOverride {
        logger.Warn("initial_capital change blocked without explicit override (%.2f → %.2f)", existingInitialCapital, newInitialCapital)
        strategy.InitialCapital = existingInitialCapital
    }
}
```

richkuo:
Closes #343

Treats `initial_capital` as an immutable PnL baseline. `SaveState` now snapshots the existing per-strategy value before the DELETE+INSERT cycle and preserves it when the incoming value disagrees, logging a warning. The sanctioned way to change a baseline is the new `StateDB.SetInitialCapital` explicit-override method.

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -368,6 +368,14 @@ func (sdb *StateDB) SaveState(state *AppState) error {
 		}
 		existingInitCaps[id] = existing
 	}
+	// rows.Next() returns false on both exhaustion and mid-iteration error;
+	// without this Err() check a transient SQLite failure would yield a
+	// silently-incomplete snapshot and leave un-snapshotted strategies
+	// unprotected by the baseline guard for this save cycle.
+	if err := existingRows.Err(); err != nil {
+		existingRows.Close()
+		return fmt.Errorf("iterate existing initial_capital: %w", err)
+	}
 	existingRows.Close()
 
 	// 3. Delete all strategies (CASCADE deletes positions + option_positions).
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

richkuo:
@claude fix all of the suggested updates in the review

richkuo:
@claude sonnet

richkuo:
@claude fix all the suggestions

PR #391, opened 2026-04-22
feat: enable-by-default systemd install script + template unit

### LINKED ISSUE #390: Make installed systemd services enabled-by-default
## Problem

Ad hoc go-trader systemd services can be started successfully but not enabled, which means they do **not** come back after a reboot.

This happened with `go-trader-paper-testing.service`: it was running normally, the machine rebooted, and the service stayed down because the unit was installed but not enabled.

## Desired behavior

When a go-trader service is installed via the repo's setup flow, it should default to:
- `systemctl enable ...`
- optionally `systemctl start ...`

so the service survives reboots by default.

## Proposed fix

Add repo-supported service installation defaults:

1. Check in a systemd unit template or canonical service file(s), for example:
   - `systemd/go-trader.service`
   - `systemd/go-trader@.service`
   - or named examples like `systemd/go-trader-paper-testing.service`

2. Add an install/setup script that:
   - copies the unit into `/etc/systemd/system/`
   - runs `systemctl daemon-reload`
   - runs `systemctl enable <unit>`
   - optionally runs `systemctl start <unit>`

3. Document in `README.md` that installed go-trader services should be **enabled**, not just manually started.

## Why

This makes the expected behavior consistent across paper/live/testing instances and avoids silent downtime after reboot.

richkuo:
Closes #390

- `scripts/install-service.sh` copies a unit, runs `daemon-reload`, and `enable`s + starts it so services survive reboot
- `systemd/go-trader@.service` templated unit for ad hoc variants (e.g. `go-trader@paper-testing`)
- README now points at the installer and calls out that `start` without `enable` silently dies on reboot


richkuo:
@claude sonnet

richkuo:
@claude fix all review suggestions

PR #413, opened 2026-04-23
feat: warn when strategy interval_seconds exceeds top-level interval

### LINKED ISSUE #409: Validate strategy interval_seconds is not greater than top-level interval_seconds
## Enhancement

Currently, individual strategies can have `interval_seconds` set independently of the top-level `interval_seconds`. There is no validation to warn or error when a strategy interval is set to a value greater than the main loop interval.

This is confusing because the main loop controls when portfolio risk checks run, and having strategies run less frequently than the main loop can mislead operators about how often positions are being evaluated.

## Proposal

Add a validation warning (not a hard error) in config validation when any strategy has `interval_seconds` greater than the top-level `interval_seconds`:

```
[WARN] strategy "hl-tema-eth-live" interval_seconds=900 exceeds top-level interval_seconds=300. Strategy will only run every 3rd portfolio cycle.
```

This keeps flexibility (intentional slower strategies) while alerting operators to potential misconfiguration.

## Acceptance Criteria

- [ ] Config validation emits a warning when strategy interval > top-level interval
- [ ] Warning is clear about the ratio (e.g. "every 3rd portfolio cycle")
- [ ] Not a hard error — strategies can intentionally run less frequently
- [ ] Add to `ValidateConfig` or equivalent validation path

richkuo:
Closes #409

Adds a non-fatal [WARN] during ValidateConfig when a per-strategy interval_seconds is greater than cfg.IntervalSeconds, reporting the effective "every Nth portfolio cycle" cadence.

---
LLM: Claude Opus 4.7 (1M) | high | Tokens: 53 in / 14055 out | Cache: 2069627 read / 163371 written | Cost: $2.41

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -420,6 +420,41 @@ func ParseLeaderboardPostTime(s string) (int, int, bool) {
 	return h, m, true
 }
 
+// strategyIntervalExceedsGlobalWarning returns a [WARN] message when the
+// per-strategy interval exceeds the top-level interval (#409), or "" otherwise.
+func strategyIntervalExceedsGlobalWarning(sc StrategyConfig, globalInterval int) string {
+	if sc.IntervalSeconds <= 0 || globalInterval <= 0 || sc.IntervalSeconds <= globalInterval {
+		return ""
+	}
+	ratio := sc.IntervalSeconds / globalInterval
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #422, opened 2026-04-25
fix: repeat portfolio drawdown warning every cycle while in warn band

### LINKED ISSUE #420: Repeat portfolio drawdown warning every cycle while in warn band (not just once)
### Problem

The portfolio drawdown warning in \`CheckPortfolioRisk\` (\`scheduler/risk.go:453-481\`) fires once per breach via the \`WarningSent\` flag. If drawdown hovers in the warn zone (between \`max_drawdown_pct × warn_threshold_pct/100\` and \`max_drawdown_pct\`) for many cycles without recovering, the operator only sees a single Discord/Telegram alert. That's easy to miss — especially overnight or during a slow bleed — and by the time the hard kill switch fires there's been no repeated reminder to intervene.

### Proposed change

While equity or perps-margin drawdown exceeds the warning threshold, emit the warning notification on every cycle (not just the first). When drawdown recovers below the warn threshold, stop alerting (current recovery behavior is fine).

Concretely, remove the \`if !prs.WarningSent\` gate at \`risk.go:459\` so \`warning=true\` + \`reason\` are set on every cycle the condition is true. The \`WarningSent\` field can either be kept (for state reporting / \`/status\` visibility) or removed entirely if no longer load-bearing.

### Considerations

- **Notifier spam**: at a 30s–5min cycle cadence this could be noisy. Acceptable — warning zone is supposed to be urgent. If operators want quieter behavior we can add an optional \`warn_repeat_interval\` (e.g. every 15min) later, but default should be every cycle.
- **No config flag needed for v1**: keep it simple; this is the behavior users actually want per #410's rationale.
- **Kill switch path is unaffected** — that one already fires on every triggering cycle.

### Acceptance criteria

- \`CheckRisk\` returns \`warning=true\` on every cycle where equity or margin drawdown > warn threshold
- Existing warning-reason formatting is preserved
- Add/update a test in \`scheduler/risk_test.go\` asserting the warning fires across multiple consecutive calls without recovery
- Update the \`WarningSent\` docstring (or remove the field) to reflect the new behavior

richkuo:
## Summary
- Remove the `!WarningSent` rising-edge gate in `CheckPortfolioRisk` so the warn-band notification fires on every cycle until drawdown recovers (#420).
- Repurpose `WarningSent` as a "currently in warn band" state flag (kept for `/status` visibility, persisted in DB) — set every cycle the band is breached, cleared on recovery.
- Update existing test to assert warning fires on the second call; add `TestCheckPortfolioRisk_WarningRepeatsAcrossCycles` covering 5 consecutive in-band cycles.

## Test plan
- [x] `go test ./...` (full scheduler suite passes)
- [x] `go test -run 'CheckPortfolioRisk|Warning' ./...` (targeted)
- [x] `gofmt -l scheduler/risk.go scheduler/risk_test.go` clean
- [x] Recovery test (`TestCheckPortfolioRisk_WarningResetOnRecovery`) still passes — `WarningSent` clears below threshold and re-fires on re-cross
- [x] Kill-switch precedence test (`TestCheckPortfolioRisk_WarningNotAfterKillSwitch`) still passes — warning is suppressed when kill switch fires

Closes #420

---
LLM: Claude Opus 4.7 (1M) | medium

### DIFF
--- a/scheduler/risk.go
+++ b/scheduler/risk.go
@@ -213,13 +213,18 @@ type KillSwitchEvent struct {
 // reconstructable, while still exposing the margin signal for operators and
 // the kill switch. The kill switch fires on whichever signal breaches first.
 type PortfolioRiskState struct {
-	PeakValue                float64           `json:"peak_value"`
-	CurrentDrawdownPct       float64           `json:"current_drawdown_pct"`
-	CurrentMarginDrawdownPct float64           `json:"current_margin_drawdown_pct,omitempty"`
-	KillSwitchActive         bool              `json:"kill_switch_active"`
-	KillSwitchAt             time.Time         `json:"kill_switch_at,omitempty"`
-	WarningSent              bool              `json:"warning_sent,omitempty"`
-	Events                   []KillSwitchEvent `json:"events,omitempty"`
+	PeakValue                float64   `json:"peak_value"`
+	CurrentDrawdownPct       float64   `json:"current_drawdown_pct"`
+	CurrentMarginDrawdownPct float64   `json:"current_margin_drawdown_pct,omitempty"`
+	KillSwitchActive         bool      `json:"kill_switch_active"`
+	KillSwitchAt             time.Time `json:"kill_switch_at,omitempty"`
+	// WarningSent is true while drawdown is currently in the warn band
+	// (between warn threshold and kill-switch limit). It is set every cycle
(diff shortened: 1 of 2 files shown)


richkuo:
Superseded by #423, which contains the same fix plus margin-path repeat coverage and this PR's 5-cycle stability test merged in.

PR #423, opened 2026-04-25
[codex] Repeat portfolio drawdown warnings

### LINKED ISSUE #420: Repeat portfolio drawdown warning every cycle while in warn band (not just once)

richkuo:
## Summary

Portfolio drawdown warnings now repeat on every scheduler cycle while the portfolio remains in the warning band, instead of only notifying once per breach. This applies to both portfolio equity drawdown and the perps-margin drawdown signal introduced for leveraged positions.

The change keeps `PortfolioRiskState.WarningSent` because it is persisted in SQLite and exposed through status/state surfaces, but narrows its meaning to current warning-band visibility: `true` while either drawdown signal is above the warning threshold, `false` again after recovery below that threshold.

## Root Cause

`CheckPortfolioRisk` set `warning=true` only inside an `if !prs.WarningSent` guard. After the first warning, later cycles with the same ongoing warning-band drawdown updated the drawdown fields but returned `warning=false`, so `main.go` did not send Discord/Telegram reminders.

## Behavior

- Emits the existing warning reason every cycle while equity drawdown is above `max_drawdown_pct * warn_threshold_pct / 100`.
- Emits the existing warning reason every cycle while perps-margin drawdown is above the same warning threshold.
- Stops warning once both signals recover below the warning threshold.
- Leaves the hard portfolio kill-switch path unchanged.
- Does not add a config flag or notification interval for this first version.

## Tests

- Updated `TestCheckPortfolioRisk_WarningFires` to assert consecutive equity drawdown warning returns without recovery.
- Updated `TestCheckPortfolioRisk_MarginWarning` to assert consecutive perps-margin warning returns without recovery.
- Kept recovery coverage for `WarningSent` resetting below the warn threshold.

## Validation

- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go test ./...` from `scheduler/`

Closes #420

---
LLM: GPT-5 | medium

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -661,7 +661,12 @@ func main() {
 			// we snapshot before the call and restore if we're on a fallback
 			// cycle. Drawdown detection still runs against the frozen peak.
 			origPeak := state.PortfolioRisk.PeakValue
+			prevWarningSent := state.PortfolioRisk.WarningSent
 			portfolioAllowed, nb, portfolioWarning, portfolioReason := CheckPortfolioRisk(&state.PortfolioRisk, cfg.PortfolioRisk, totalPV, totalNotional, perpsLoss, perpsMargin)
+			// True only on the cycle that first enters the warn band; false on
+			// repeat cycles while still in band. Used to gate kill-switch event
+			// log appends — notifications still fire every cycle via portfolioWarning.
+			portfolioWarnBandEntered := portfolioWarning && !prevWarningSent
 			if usedPVFallback && state.PortfolioRisk.PeakValue > origPeak {
 				state.PortfolioRisk.PeakValue = origPeak
 			}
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

richkuo:
@claude sonnet

PR #424, opened 2026-04-25
fix: include version ldflag in manual update notification message

### LINKED ISSUE #419: Auto-upgrade build omits Version ldflag, causing Discord summaries to show (dev) instead of release tag
## Bug Description

The auto-upgrade flow in  builds the new binary without the  stamp. This means users who accept the DM upgrade prompt end up running a binary that reports , so Discord summary titles show  instead of the actual release tag (e.g. ).

## Location

  
Line: ~230  
Function: 

## Expected Behavior

The auto-upgrade build command should include the version stamp, matching the manual build instructions in README.md and SKILL.md:

Note: the  function already does this correctly (line 109):

But the **notification message** that tells users how to manually update omits the flag.

## Impact

- Users following the manual update instructions from Discord will build binaries without version stamps
- Discord summaries will show  instead of the actual version
- Makes it harder to confirm which version is running across multiple deployments

## Suggested Fix

Update  to include the  in the manual update command string, or better yet, extract the build command construction so both  and  share the same logic.

## Environment

- go-trader version: v0.35.0
- File:

richkuo:
The Discord notification for available updates showed a build command without the -ldflags stamp, so users following the manual instructions would produce a (dev) binary.

The `applyUpgrade` function already correctly stamped the version when performing the automated upgrade — this fixes the human-readable instructions in the Discord notification message.

Closes #419

---
LLM: Claude Sonnet 4.6 | high

---
LLM: Claude Sonnet 4.6 (1M) | high | Tokens: 31 in / 7554 out | Cache: 840184 read / 58607 written | Cost: $0.59

### DIFF
--- a/scheduler/updater.go
+++ b/scheduler/updater.go
@@ -227,7 +227,7 @@ func formatUpdateMessage(localHash, remoteHash, commitLog, newTag string, goChan
 	sb.WriteString("To update:\n```\n")
 	sb.WriteString("cd /path/to/go-trader && git pull --ff-only\n")
 	if goChanged {
-		sb.WriteString("cd scheduler && go build -o ../go-trader . && cd ..\n")
+		sb.WriteString("cd scheduler && go build -ldflags \"-X main.Version=$(git describe --tags --always)\" -o ../go-trader . && cd ..\n")
 	}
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

PR #432, opened 2026-04-27
fix: round sz to sz_decimals in HL market_close

### LINKED ISSUE #425: Bug: market_close passes unrounded sz causing float_to_wire rounding error
## Summary

When the circuit breaker kill switch tries to close a position, \`market_close\` passes the raw \`sz\` value directly to the Hyperliquid SDK without rounding to the asset's \`sz_decimals\`. This causes a \`float_to_wire causes rounding\` error from Hyperliquid, and the close fails every cycle in an infinite retry loop.

## Observed behavior

\`hl-tema-eth-live\` tripped its circuit breaker and has been stuck for hours trying to close 0.250965 ETH:

```
[CRITICAL] hl-circuit-close: strategy hl-tema-eth-live coin ETH sz=0.250965 failed: close failed: ('float_to_wire causes rounding', 0.2509645272613055)
```

The actual on-chain position size is \`0.2509645272613055\` but the stored/passed size is \`0.250965\`. Hyperliquid rejects it because the SDK's wire encoding can't represent the imprecise float.

## Root cause

\`market_open\` (adapter.py:236) rounds \`size\` to \`sz_decimals\` before calling the SDK:

\`\`\`python
sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3)
size = round(size, sz_decimals)
\`\`\`

But \`market_close\` (adapter.py:269) passes \`sz\` directly with no rounding:

\`\`\`python
return self._exchange.market_close(symbol, sz)
\`\`\`

## Fix

Apply the same \`sz_decimals\` rounding to \`market_close\` before passing \`sz\` to the SDK, matching what \`market_open\` already does.

richkuo:
## Summary

- `market_close` in the Hyperliquid adapter was passing `sz` directly to the SDK without rounding to the asset's `sz_decimals`, causing `float_to_wire causes rounding` rejections for per-strategy circuit-breaker closes
- Per-strategy CB sizing of a shared-wallet position produces fractional coin quantities with more precision than HL accepts (e.g. ETH `0.2509645272613055` with `sz_decimals=4`)
- The fix applies the same `sz_decimals` rounding already used by `market_open` and `place_stop_loss`, plus the `sz=None` full-close passthrough guard and zero-size `ValueError`

## Test plan

- [ ] `uv run pytest platforms/hyperliquid/test_adapter.py -v` — all 34 tests pass, including 3 new regression tests:
  - `test_market_close_partial_size_rounds_to_sz_decimals` (ETH `0.2509645272613055` → `0.251` at sz_decimals=4)
  - `test_market_close_full_close_passes_none_unchanged`
  - `test_market_close_partial_size_rounded_to_zero_raises`

Closes #425

---
LLM: Claude Sonnet 4.6 (1M) | medium

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -251,6 +251,14 @@ class HyperliquidExchangeAdapter:
             raise RuntimeError(
                 "market_close requires live mode (set HYPERLIQUID_SECRET_KEY)"
             )
+        if sz is not None:
+            # Round to asset's tick precision to avoid float_to_wire rounding error (#425)
+            if symbol not in self._info.asset_to_sz_decimals:
+                print(f"[WARN] sz_decimals not found for {symbol}, defaulting to 3", file=sys.stderr)
+            sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3)
+            sz = round(sz, sz_decimals)
+            if sz <= 0:
+                raise ValueError(f"Size rounded to zero for {symbol} (sz_decimals={sz_decimals})")
         return self._exchange.market_close(symbol, sz)
 
     def round_perps_trigger_px(self, symbol: str, px: float) -> float:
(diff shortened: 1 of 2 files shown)


richkuo:
@claude sonnet

PR #433, opened 2026-04-27
fix: HL CB drain partial fills + virtual decrement + leverage overwrite

### LINKED ISSUE #418: Per-strategy HL circuit-breaker drain ignores partial fills + leverage overwrite inflates next-cycle drawdown
## Bug

A per-strategy Hyperliquid circuit breaker fired and the on-chain reduce-only close only sold ~50% of the position, but the local virtual state was never decremented — virtual `pos.Quantity` stayed at 100% while on-chain held 50%. On the next cycle, `reconcileHyperliquidPositions` then copied the on-chain `leverage` (e.g. 20, exchange margin setting) into virtual state, which inflated the per-strategy drawdown denominator by ~10x and re-fired the circuit breaker.

The leverage overwrite (originally reported below) is real but downstream — it's the second-order amplifier. The root cause is the per-strategy CB drain.

### Root cause #1 — drain ignores fill size (`hyperliquid_balance.go:905-913`)

```go
partial := sz
_, err := closer(c.Symbol, &partial)   // result discarded
if err != nil { ... }
fmt.Printf("[INFO] hl-circuit-close: ... submitted reduce-only close sz=%.6f\n", ..., sz)
```

`HyperliquidCloseFill.TotalSz` carries the actual filled size from the SDK, but is never compared against requested `sz`. Any partial fill (market depth, slippage cap, `market_close` slippage param) is logged using the requested size, `allOK=true`, pending entry cleared — no retry next cycle. Operator-facing log says full close happened.

### Root cause #2 — drain never updates virtual position state

The drain only touches `RiskState.PendingCircuitCloses`. It does NOT:
- Decrement `statePos.Quantity` by what actually closed
- `RecordTrade` for the close fill
- Emit a Discord trade alert

For non-shared coins, `reconcileHyperliquidPositions` would catch the on-chain delta on the next cycle. But for shared-wallet coins (multi-strategy on the same coin), per-#258 the reconcile deliberately does NOT overwrite virtual quantities — so virtual state never catches up.

### Why exactly 50% — `computeHyperliquidCircuitCloseQty` (`hyperliquid_balance.go:664`)

For shared-wallet coins, the close size is `absSzi * (wFiring / sumW)`. Two strategies with equal `capital_pct` weights → close exactly 50% of on-chain. That's the intended weighted-share semantic for shared wallets, but combined with root cause #2, the firing strategy's virtual quantity stays at 100% even though only its 50% share was closed on-chain.

### Root cause #3 — on-chain leverage overwrite (the original report)

`reconcileHyperliquidPositions` at `hyperliquid_balance.go:266-269` (and the shared-coin path at line 432-440) unconditionally overwrites `statePos.Leverage` with the on-chain value:

```go
if onChainPos.Leverage > 0 && statePos.Leverage != onChainPos.Leverage {
    logger.Info(\"hl-sync: %s leverage %v → %v\", sym, statePos.Leverage, onChainPos.Leverage)
    statePos.Leverage = onChainPos.Leverage
    changed = true
}
```

`pos.Leverage` is only ever read in `perpsMarginDrawdownInputs` (`risk.go:1114`, `1128`) — the guard and the `notional / pos.Leverage` divisor. All sizing (`runHyperliquidExecuteOrder`, `runOKXExecuteOrder`, `perpsLiveOrderSize`, the entry path) uses `sc.Leverage` directly from config. So when on-chain leverage (HL exchange margin setting, e.g. 20) replaces config leverage (e.g. 2), drawdown denominator shrinks 10x → drawdown % inflated 10x → CB re-fires immediately.

This also propagates into `AggregatePerpsMarginInputs` (`risk.go:357`), which the **portfolio kill switch** uses (#296). Same on-chain leverage corruption affects portfolio-level kill-switch decisions, not just per-strategy.

### Cascade

1. CB fires for strategy at config leverage = 2
2. Drain submits weighted close (50% for shared-coin) or full close (sole-owner, but partial-filled)
3. HL fills only 50% on-chain
4. Drain ignores `TotalSz`, marks success, clears pending
5. Virtual `pos.Quantity` unchanged; if shared coin, never reconciles
6. Next cycle: hl-sync sees residual on-chain position, overwrites `statePos.Leverage` with on-chain (e.g. 20)
7. `perpsMarginDrawdownInputs` divides by 20 instead of 2
8. Drawdown % inflated ~10x → CB re-fires immediately, despite being well within configured risk limits

### Fix scope

**Fix 1 — verify fill in drain** (`hyperliquid_balance.go:905-913`):
- Compare `result.Close.Fill.TotalSz` against requested `sz`
- On under-fill: keep `allOK=false`, leave pending (or re-queue with `Size = sz - filled`) for next cycle
- Surface partial-fill warnings to Discord/operator

**Fix 2 — decrement virtual state on close** (`hyperliquid_balance.go` drain):
- After successful close, decrement `statePos.Quantity` by actual `Fill.TotalSz`
- `RecordTrade` for the close fill (sell side, with fee from `Fill.Fee`)
- Emit close trade alert
- Particularly load-bearing for shared-coin strategies where reconcile won't fix it later

**Fix 3 — config leverage for drawdown calc** (this issue's original scope):
- Pass `sc.Leverage` into `perpsMarginDrawdownInputs(s *StrategyState, configLeverage float64, prices map[string]float64)`
- Update both call sites: `risk.go:1203` (per-strategy) and `risk.go:362` (portfolio aggregate)
- Removes the `pos.Leverage` read entirely from risk math; hl-sync overwrite becomes harmless metadata
- Side-steps need to migrate corrupted `pos.Leverage` in existing `state.db`

### Acceptance Criteria

- [ ] Drain verifies actual `Fill.TotalSz` and treats under-fills as failure (pending preserved for retry)
- [ ] Drain decrements virtual `statePos.Quantity` by actual filled amount and records a Trade
- [ ] Per-strategy drawdown calc uses config leverage, not on-chain leverage
- [ ] `AggregatePerpsMarginInputs` (portfolio kill switch) also uses config leverage
- [ ] hl-sync still logs on-chain leverage for visibility
- [ ] Circuit breaker does not fire prematurely when on-chain leverage != config leverage
- [ ] Works correctly for strategies at 1x, 2x, 5x, 10x, 20x leverage
- [ ] No regression for strategies where config leverage matches on-chain
- [ ] No regression for sole-owner full-close path

### Tests

- Drain stub returns `Fill.TotalSz = 0.5 * requested` → assert `allOK=false`, pending NOT cleared, no virtual decrement
- Drain stub returns `Fill.TotalSz == requested` → assert virtual `pos.Quantity` decremented by `TotalSz`, Trade recorded, pending cleared
- Shared-coin scenario: two equal-weight strategies, one CB fires, drain fully fills weighted 50% → assert firing strategy's virtual quantity decremented (since shared-coin reconcile won't fix it)
- Stub `StrategyState` with `pos.Leverage = 20`, config `Leverage = 2` → `perpsMarginDrawdownInputs` divides by 2
- Aggregate variant with same setup → portfolio denominator uses config leverage

Reference only: LLM model used to update this issue: Claude Opus 4.7 | high

richkuo:
## Summary

Three related bugs in the per-strategy Hyperliquid circuit-breaker drain that combine into a self-reinforcing CB re-fire loop. See #418 for the full cascade analysis.

**RC1 — drain ignored partial fills**

`Fill.TotalSz` was never compared against the requested `sz`. A partial fill (slippage cap, market depth, `market_close` slippage param) was logged at the requested size, `allOK` stayed true, and pending was cleared. Fix: compare fill against requested with a 0.99× tolerance for HL lot-size rounding; on under-fill leave pending intact for retry and emit a `[CRITICAL]` log noting whether the SL was also cancelled and the residual is unprotected.

**RC2 — drain never updated virtual state**

Virtual `pos.Quantity` stayed at 100% after a weighted CB close while on-chain dropped to the strategy's share. For shared-wallet coins `reconcileHyperliquidPositions` deliberately does not overwrite quantities (#258), so the drift never self-healed. New helper `applyHyperliquidCircuitCloseFill` decrements `pos.Quantity` by `Fill.TotalSz`, books realized PnL net of `Fill.Fee` into `Cash`, calls `RecordTrade` (eager SQLite insert via `tradeRecorder` hook), and `recordClosedPosition` + delete on full close. `AvgCost` is preserved on partial close.

**RC3 — on-chain leverage overwrote drawdown denominator**

`reconcileHyperliquidPositions` unconditionally overwrote `statePos.Leverage` with the exchange's account-wide margin tier (e.g. HL 20×) even when the trader configured 2×. `perpsMarginDrawdownInputs` divided by `pos.Leverage`, so the corrupted value shrank margin 10× and inflated the drawdown ratio 10×, immediately re-firing the CB.

Belt-and-suspenders fix (both paths):
- **Read path**: `perpsMarginDrawdownInputs(s, configLeverage, prices)` now takes `sc.Leverage` directly — `pos.Leverage` is no longer read in risk math. `AggregatePerpsMarginInputs` gains a `configs []StrategyConfig` parameter so the portfolio kill switch uses the same source of truth.
- **Write path**: reconcile guards changed from `!= onChainPos.Leverage` to `== 0` in both the singleton and shared-coin paths. On-chain leverage now only seeds zero-value (legacy/uninitialised) positions — configured values are preserved against future margin-tier drift.

## Tests

10 new regression tests:

- Partial-fill: pending preserved, virtual quantity decremented by `TotalSz` (not requested size)
- Full-fill: position deleted, PnL booked in Cash, `ClosedPosition` row written
- Shared-coin variant: firing strategy's virtual quantity drops even though reconcile won't fix it
- `AvgCost` preserved on partial close
- Config leverage vs corrupted `pos.Leverage`: with `pos.Leverage=20`, `configLev=2` → margin = 1450 not 145
- Zero `configLeverage` → `(0, 0)` (falls back to peak-relative drawdown)
- `AggregatePerpsMarginInputs` reads from configs map
- Orphan strategy without config → contributes zero
- Reconcile preserves configured `pos.Leverage` against on-chain overwrite
- Reconcile seeds zero-value `pos.Leverage` from on-chain (legacy migration path preserved)

Closes #418

---
LLM: Claude Sonnet 4.6 | high

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -265,8 +265,17 @@ func reconcileHyperliquidPositions(stratState *StrategyState, sym string, positi
 			statePos.Multiplier = 1
 			changed = true
 		}
-		if onChainPos.Leverage > 0 && statePos.Leverage != onChainPos.Leverage {
-			logger.Info("hl-sync: %s leverage %v → %v", sym, statePos.Leverage, onChainPos.Leverage)
+		// #418: only seed leverage from on-chain when the virtual position has
+		// none yet (Leverage==0 → legacy/uninitialised). The entry path sets
+		// Leverage from sc.Leverage (config); the exchange's account-wide
+		// margin tier can differ (e.g. HL allows up to 20x while the trader
+		// sized at 2x) and unconditionally overwriting it inflates the
+		// perpsMarginDrawdownInputs denominator and can re-fire the circuit
+		// breaker spuriously. Defense in depth — risk math also reads
+		// sc.Leverage now, so this is belt-and-suspenders against any future
+		// consumer that reads pos.Leverage directly.
+		if onChainPos.Leverage > 0 && statePos.Leverage == 0 {
+			logger.Info("hl-sync: %s leverage init → %v (from on-chain, legacy/zero-value position)", sym, onChainPos.Leverage)
 			statePos.Leverage = onChainPos.Leverage
 			changed = true
 		}
@@ -436,12 +445,14 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 				pos.Multiplier = 1
 				changed = true
 			}
-			if onChainPos != nil && onChainPos.Leverage > 0 && pos.Leverage != onChainPos.Leverage {
+			// #418: same write-path guard as reconcileHyperliquidPositions —
+			// only seed leverage from on-chain when virtual is zero-value.
+			if onChainPos != nil && onChainPos.Leverage > 0 && pos.Leverage == 0 {
 				logger, err := logMgr.GetStrategyLogger(id)
 				if err != nil {
 					fmt.Printf("[ERROR] hl-sync: logger for %s: %v\n", id, err)
 				} else {
-					logger.Info("hl-sync: %s leverage %v → %v (shared coin)", coin, pos.Leverage, onChainPos.Leverage)
+					logger.Info("hl-sync: %s leverage init → %v (shared coin, from on-chain)", coin, onChainPos.Leverage)
 				}
 				pos.Leverage = onChainPos.Leverage
 				changed = true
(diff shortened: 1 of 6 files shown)


richkuo:
@claude

richkuo:
@claude fix all of the review observations

richkuo:
@claude

PR #437, opened 2026-04-27
[codex] Auto-reset confirmed-flat kill switch

### LINKED ISSUE #431: Bug: kill switch stays latched after force-close with no recovery path when discord.owner_id is unset
## Description

When the portfolio kill switch fires and successfully force-closes all positions, `portfolio_risk.kill_switch_active` stays at `1` indefinitely. If `discord.owner_id` is not configured, there is no in-process recovery path — the operator must restart the process or edit `state.db` to resume trading.

## Observed Behavior (live instance)

```
portfolio_risk row after force-close:
  peak_value=1261.87, current_drawdown_pct=3.63, current_margin_drawdown_pct=26.84, kill_switch_active=1
positions table: empty
```

Manual DB reset required to resume trading.

## Reproduction

1. Portfolio margin drawdown exceeds kill switch threshold (e.g., 20%).
2. Kill switch fires, `planKillSwitchClose` (`scheduler/main.go:788`) submits reduce-only closes; `plan.OnChainConfirmedFlat == true`; virtual state is zeroed at `scheduler/main.go:794–806`.
3. `kill_switch_active` stays at `1`. The stale `current_drawdown_pct` / `current_margin_drawdown_pct` are persisted as a frozen snapshot from the firing cycle.
4. On every subsequent cycle, `CheckPortfolioRisk` (`scheduler/risk.go:410`) early-returns `false, false, false, "kill switch is latched..."` — no recalculation, no re-trigger, just a held latch.
5. With `discord.owner_id` unset, `notifier.HasOwner()` is false, so the DM reset goroutine at `scheduler/main.go:857–881` never starts. Bot is stuck.

## Existing Recovery Paths (currently the only ways out)

- **DM reset** — `scheduler/main.go:857–881`. Requires `discord.owner_id` to be configured. Goroutine asks the owner via DM to reply `reset` (30-minute window).
- **Restart auto-clear** — `ClearLatchedKillSwitchSharedWallet` at `scheduler/risk.go:282`, called from `scheduler/main.go:149` on startup. Clears the latch and re-baselines `PeakValue` if shared-wallet balance can be fetched. `systemctl restart go-trader` recovers the instance — but only for shared-wallet setups (multiple strategies with `capital_pct > 0` on the same platform). Single-strategy or non-shared-wallet deployments get no benefit from restart.

Neither is acceptable as the *only* recovery for an operator who isn't running Discord.

## Not the Root Cause: PR #423

Earlier framing pointed at PR #423 (`cb0a543`) — incorrect. PR #423 only changed the `portfolioWarning` branch (`scheduler/risk.go:467–488`) so warn-band notifications repeat every cycle. That branch is unreachable while `KillSwitchActive` is true (the kill-switch early-return at `risk.go:410` runs first), so PR #423 cannot affect post-latch behavior.

Drawdown metrics not being zeroed post-close is a *consequence* of the early-return, not a regression — it has always behaved this way.

## Proposed Fix

The patch lives around the existing `OnChainConfirmedFlat` block in `main.go`, **not** in `risk.go` and **not** as a revert of PR #423. Two reasonable directions:

1. Auto-clear the latch when `plan.OnChainConfirmedFlat == true` AND no DM-capable owner is configured (`scheduler/main.go:794` is the natural site — virtual state is already being zeroed there). Re-baseline `PeakValue` to the post-close portfolio value, zero `CurrentDrawdownPct` / `CurrentMarginDrawdownPct`, and append an `auto_reset` `KillSwitchEvent`.
2. Add a `/status` POST endpoint or CLI subcommand to clear the latch, giving non-Discord operators a documented unlatch path that doesn't require restart or DB editing.

Option 1 is preferred — it removes the stuck-state automatically when on-chain flat is confirmed and Discord-based human-in-the-loop is unavailable.

## Acceptance

- After a kill-switch force-close on an instance with `discord.owner_id` unset, on the cycle following `OnChainConfirmedFlat == true`:
  - `kill_switch_active = 0`
  - `current_drawdown_pct = 0`, `current_margin_drawdown_pct = 0`
  - `peak_value` re-baselined to the post-close portfolio value
  - An `auto_reset` event appears in `kill_switch_events`
  - Bot resumes trading on the next cycle
- Existing DM-reset path is unchanged when `discord.owner_id` IS configured.
- Note: the fix should also zero `current_margin_drawdown_pct` explicitly — `ClearLatchedKillSwitchSharedWallet` currently only zeroes `current_drawdown_pct` (both fields recalculate on the next tick once unlatched, but zeroing both at clear-time keeps the DB clean for operators inspecting it directly).

Reference only: LLM model used: GLM 5.1 (original report); review and rewrite by Claude Sonnet 4.6.

richkuo:
## Summary

Fixes #431 by adding a non-Discord recovery path for the portfolio kill switch after automated force-close confirms all managed live venues are flat.

## What changed

- Auto-clear the portfolio kill-switch latch when `plan.OnChainConfirmedFlat == true` and no DM-capable owner is configured.
- Re-baseline `PortfolioRisk.PeakValue` to the post-close portfolio value, zero both drawdown fields, clear warning state, and append an `auto_reset` audit event.
- Preserve the existing owner DM reset flow when an owner is configured.
- Also zero `CurrentMarginDrawdownPct` in the existing shared-wallet startup auto-clear helper so direct DB inspection stays clean.

## Validation

### DIFF
--- a/scheduler/kill_switch_close.go
+++ b/scheduler/kill_switch_close.go
@@ -178,6 +178,17 @@ type KillSwitchClosePlan struct {
 	LogLines []string
 }
 
+func (p KillSwitchClosePlan) CanAutoResetWithoutOwner() bool {
+	return p.OnChainConfirmedFlat && !p.OKXSpotPresent && !p.RHOptionsPresent
+}
+
+const killSwitchManualResetLine = "Virtual state cleared. Manual reset required."
+const killSwitchAutoResetLine = "Virtual state cleared. Kill switch auto-reset; trading will resume next cycle."
+
+func formatKillSwitchAutoResetMessage(msg string) string {
+	return strings.Replace(msg, killSwitchManualResetLine, killSwitchAutoResetLine, 1)
+}
+
 // HLStateFetcher re-fetches Hyperliquid on-chain positions for the kill-switch
 // opportunistic-fetch path. Exposed as a function type so tests can stub the
 // HTTP call. The default wraps fetchHyperliquidState.
@@ -496,7 +507,7 @@ func formatKillSwitchMessage(hlAddr string, plan KillSwitchClosePlan, portfolioR
 			parts = append(parts, gapNotes...)
 		}
 		summary := strings.Join(parts, "; ")
-		return fmt.Sprintf("%s\n%s\n%s. Virtual state cleared. Manual reset required.", header, portfolioReason, summary)
+		return fmt.Sprintf("%s\n%s\n%s. %s", header, portfolioReason, summary, killSwitchManualResetLine)
 	}
 
 	var segments []string
--- a/scheduler/kill_switch_close_test.go
+++ b/scheduler/kill_switch_close_test.go
@@ -129,6 +129,9 @@ func TestPlanKillSwitchClose_HappyPath(t *testing.T) {
 	if !plan.OnChainConfirmedFlat {
 		t.Fatalf("expected ConfirmedFlat, got plan=%+v", plan)
 	}
+	if !plan.CanAutoResetWithoutOwner() {
+		t.Fatal("expected happy-path confirmed-flat plan to allow no-owner auto-reset")
+	}
 	if len(plan.CloseReport.ClosedCoins) != 1 || plan.CloseReport.ClosedCoins[0] != "ETH" {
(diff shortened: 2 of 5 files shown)


richkuo:
@claude

richkuo:
@claude

PR #440, opened 2026-04-28
fix: make workflow LLM footer authoritative over Claude's self-written footer

richkuo:
## Summary

- Claude proactively appends an `LLM: <model> | <effort>` footer to its own comments (it learned the convention from `CLAUDE.md`), but it has no access to the `--effort` flag it was invoked with — so it guesses, producing mismatches (e.g. "medium" when the run used "high").
- The previous idempotency guard in the "Append LLM footer" step (`grep -q "LLM:" → exit 0`) preserved Claude's wrong guess instead of overwriting it.

## Changes

- **`strip_llm_footer.py`** (new): stdin→stdout utility that strips any trailing `(---\n)?LLM: …` block. Regex is end-of-string anchored so mid-body `LLM:` tokens (e.g. in code blocks) are preserved.
- **`rewrite_create_pr_link.py`**: calls `strip_llm_footer` on the prefilled `body=` query param before appending the authoritative footer. The Create-PR link's pre-filled PR description is now also corrected.
- **`claude.yml`**: replaced the `grep -q "LLM:" → exit 0` guard with `BODY=$(printf '%s' "$BODY" | python3 .github/scripts/strip_llm_footer.py)`. The workflow is now always authoritative.
- **Tests**: 8 new tests in `test_strip_llm_footer.py`; `test_rewrite_create_pr_link.py` gains a stale-footer regression case and a true idempotency test (replaces the old "documents non-idempotency" caveat test).

## Test plan

- [ ] `python3 .github/scripts/test_strip_llm_footer.py` — 8 tests pass
- [ ] `python3 .github/scripts/test_rewrite_create_pr_link.py` — 7 tests pass (including new stale-footer + idempotency cases)
- [ ] Trigger `@claude` on a PR and confirm the comment ends with the workflow-resolved effort level even when Claude's response body already contains a different `LLM:` footer

---
LLM: Claude Sonnet 4.6 | high

### DIFF
--- a/.github/scripts/rewrite_create_pr_link.py
+++ b/.github/scripts/rewrite_create_pr_link.py
@@ -5,17 +5,16 @@ PR body ends with our model/effort footer instead of the default
 Reads the comment body from $BODY_IN and the footer text from $FOOTER_TEXT.
 Prints the rewritten comment body to stdout.
 
-Not idempotent on its own: if a Create-PR link's `body=` already ends with a
-`LLM: ...` footer (not the default Claude Code attribution), the
-else-branch in `rewrite` will append a second footer. The caller in
-.github/workflows/claude.yml guards against re-runs with a
-`grep -q "LLM:"` check before invoking this script.
+Idempotent: any existing LLM footer in the prefilled body= param is stripped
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

PR #448, opened 2026-04-28
[codex] fix perps drawdown circuit breaker

### LINKED ISSUE #447: fix: perps drawdown CB silent failures — wrong denominator + TotalTrades guard
## Summary

Two stacked bugs prevent the per-strategy drawdown circuit breaker from firing on a leveraged perps position with unrealized loss exceeding `max_drawdown_pct`.

---

## Bug 1 — `perpsMarginDrawdownInputs` returns `(0, 0)`, causing wrong denominator

`CheckRisk` (`risk.go:1268`) is designed to compute drawdown as:

```
unrealizedLoss / (notional / configLeverage)
```

i.e. loss against the deployed margin (capital at risk before leverage). With leverage=20 and notional≈$1,084, margin≈$54. A $7 unrealized loss → **13% drawdown** against that margin.

However, `perpsMarginDrawdownInputs` (`risk.go:1179`) silently returns `(0, 0)` when `pos.Multiplier <= 0`, causing `CheckRisk` to fall back to peak-relative drawdown:

```
unrealizedLoss / PeakValue   →   $7 / $496 = 1.4%
```

This is a ~10× understatement of the real drawdown. The CB does not fire.

**Root cause:** `pos.Multiplier` is 0 on the position (likely set before `ExecutePerpsSignal` began writing `Multiplier: 1` at `portfolio.go:540`, or for a position that bypassed that path). The live HL reconciler fixes this for live strategies (`hyperliquid_balance.go:265`), but paper strategies have no equivalent migration.

**Fix:** Add a startup migration in `LoadStateWithDB` (or the state-load path) that sets `Multiplier = 1` for any perps position where `Multiplier <= 0`, mirroring the live reconciler logic.

---

## Bug 2 — `TotalTrades > 0` guard blocks the CB on first open position

Even if Bug 1 were fixed and drawdown correctly showed 13%, the CB still would not fire:

```go
// risk.go:1291
if r.TotalTrades > 0 && r.CurrentDrawdownPct > r.MaxDrawdownPct {
```

`TotalTrades` is only incremented by `RecordTradeResult` (`risk.go:1324`), which is called when a position is **closed**. A strategy that has opened but never closed a position has `TotalTrades = 0`, so `0 > 0` is always false — the CB is permanently disabled regardless of how deep the unrealized loss grows.

**Fix:** Remove the `r.TotalTrades > 0` guard from the drawdown check. The consecutive-losses CB below it (`r.ConsecutiveLosses >= 5`) already requires completed trades and is unaffected.

```go
// Before
if r.TotalTrades > 0 && r.CurrentDrawdownPct > r.MaxDrawdownPct {

// After
if r.CurrentDrawdownPct > r.MaxDrawdownPct {
```

---

## Reproduction (both bugs combined)

1. Configure a perps strategy with `max_drawdown_pct: 10`, `leverage: 20`
2. Open a position — this is the first trade, so `TotalTrades = 0`
3. Position has `pos.Multiplier = 0` (old state or paper mode without migration)
4. Let ETH drop ~5% from entry → margin drawdown = ~13%, peak-relative drawdown = ~1.4%
5. Observe: status reports ~1.4%, CB does not fire

## Impact

- A perps strategy with a first open position and `Multiplier = 0` has **zero drawdown protection** — neither the wrong denominator nor the `TotalTrades` guard will trip the CB before liquidation
- Portfolio kill switch is the only remaining guard, but its default thresholds (25% max / 15% warn) are much looser than a tightly-configured per-strategy limit

---
LLM: Claude Sonnet 4.6 | high

richkuo:
## Summary

- migrate legacy perps positions with `Multiplier <= 0` to `Multiplier = 1` during SQLite state load
- allow max-drawdown circuit breakers to fire before the first closed trade
- add focused regressions for the state-load migration and first-open-position drawdown CB behavior

## Root Cause

Legacy/paper perps positions could retain `Multiplier = 0`, causing perps margin drawdown to fall back to peak-relative drawdown. Separately, the max-drawdown check required `TotalTrades > 0`, so a first open position could breach its configured drawdown threshold without tripping the per-strategy circuit breaker.

Closes #447

## Validation

- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'TestLoadStateWithDB_MigratesLegacyPerpsMultiplier|TestCheckRisk_PerpsDrawdownFiresBeforeAnyClosedTrades' .`
- `git diff --check`

LLM: GPT-5 Codex | extra high

### DIFF
--- a/scheduler/risk.go
+++ b/scheduler/risk.go
@@ -1288,7 +1288,7 @@ func CheckRisk(sc *StrategyConfig, s *StrategyState, portfolioValue float64, pri
 		} else {
 			r.CurrentDrawdownPct = 0
 		}
-		if r.TotalTrades > 0 && r.CurrentDrawdownPct > r.MaxDrawdownPct {
+		if r.CurrentDrawdownPct > r.MaxDrawdownPct {
 			r.CircuitBreaker = true
 			r.CircuitBreakerUntil = now.Add(24 * time.Hour)
 			setHyperliquidCircuitBreakerPending(sc, s, assist)
--- a/scheduler/risk_test.go
+++ b/scheduler/risk_test.go
@@ -1558,6 +1558,43 @@ func TestCheckRisk_PerpsMarginDrawdown_FiresEarly(t *testing.T) {
 	}
 }
 
+func TestCheckRisk_PerpsDrawdownFiresBeforeAnyClosedTrades(t *testing.T) {
+	s := &StrategyState{
+		ID:   "hl-first-trade",
+		Type: "perps",
+		Cash: 500,
+		RiskState: RiskState{
+			PeakValue:      500,
+			MaxDrawdownPct: 10,
+			TotalTrades:    0,
+			DailyPnLDate:   todayUTC(),
+		},
+		Positions: map[string]*Position{
+			"ETH": {Symbol: "ETH", Quantity: 1, AvgCost: 100, Side: "long", Multiplier: 1, Leverage: 20},
+		},
+		OptionPositions: make(map[string]*OptionPosition),
+		TradeHistory:    []Trade{},
+	}
+	prices := map[string]float64{"ETH": 80}
+	pv := PortfolioValue(s, prices)
+	sc := &StrategyConfig{ID: "hl-first-trade", Platform: "hyperliquid", Type: "perps", Leverage: 20}
+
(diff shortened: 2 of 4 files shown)


richkuo:
@claude

PR #458, opened 2026-04-28
fix: auto-calculate separator length from header format string

### LINKED ISSUE #457: Tech debt: separator dashes should be auto-calculated from format string width
## Problem

Separator line length is hardcoded as a magic number (`strings.Repeat("-", 84)` / `strings.Repeat("-", 75)`) in `writeCatTablePartial`. When column widths change, someone has to manually recount and update the number — easy to get wrong (just happened with the #445 follow-up).

## Fix

Auto-calculate separator length by formatting the header row and measuring its length (minus trailing newline), so the dashes always match the table width regardless of format changes.

```go
header := fmt.Sprintf("%-16s %7s ...", "Strategy", "Init", ...)
sep := strings.Repeat("-", len(strings.TrimSuffix(header, "\n")))
```

richkuo:
## Summary

- Auto-calculate separator dash length from the formatted header row instead of hardcoding magic numbers
- Fix label truncation threshold (`len(label) > 14` → `len(label) > 16`) to match the 16-char Strategy column width
- Fix separator lengths (were 72/64, should be 84/75) — now auto-derived so they can't drift again

## Testing

- `go test ./scheduler -run Discord -v` — PASS
- Verified auto-calculated lengths match expected: wallet=84, no-wallet=75

Closes #457

---
Effort: ~2 hours across multiple iterations — initial column width changes, label truncation bug fix, separator miscalculation fix, auto-calculation refactor, and rebase onto main with conflict resolution

### DIFF
--- a/scheduler/discord.go
+++ b/scheduler/discord.go
@@ -773,7 +773,7 @@ func writeCatTablePartial(sb *strings.Builder, bots []botInfo, showWalletPct, in
 	}
 	sb.WriteString("\n```\n")
 	if showWalletPct {
-	header := fmt.Sprintf("%-16s%9s %6s %6s %8s%5s %8s%5s %4s %4s %5s", "Strategy", "Init", "Value", "PnL", "PnL%", "DD", "Wallet%", "Tf", "Int", "#T", "W/L")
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

richkuo:
@claude fix failing test

PR #459, opened 2026-04-28
[codex] fix shared-wallet fallback kill switch

### LINKED ISSUE #452: Bug: Portfolio kill switch uses max(PV) instead of sum on shared-wallet balance fetch failure
## Summary

When the Hyperliquid wallet balance fetch fails, `computeTotalPortfolioValue` falls back to `max(member PV)` instead of summing member strategy values. This produces a falsely low portfolio value that can trigger the kill switch incorrectly. Two related defects surfaced while investigating — both are documented below since they compound the impact.

## Reproduction

1. Two live HL strategies sharing a wallet (e.g. `hl-rmc-eth-live` + `hl-tema-eth-live`)
1. HL `clearinghouseState` fetch fails (network error, API blip, etc.)
1. `computeTotalPortfolioValue` enters the shared-wallet fallback path

## Actual Behavior

Fallback uses `maxPV` of member strategies:
- `hl-rmc-eth-live` PV = $567.86
- `hl-tema-eth-live` PV = $647.70
- Fallback: `max($567.86, $647.70)` = **$647.70**
- Peak (set when balance was available): $1280.98
- Equity DD = ($1280.98 - $647.70) / $1280.98 = **49.4%** → kill switch fires

## Expected Behavior

Fallback should SUM member PVs:
- Sum = $567.86 + $647.70 = **$1215.56**
- Equity DD = ($1280.98 - $1215.56) / $1280.98 = **5.1%** → no trigger

## Root Cause

`shared_wallet.go:243-262` (fallback branch inside `computeTotalPortfolioValue`):

```go
// Current (wrong for fallback):
maxPV := 0.0
for _, id := range ids {
    s, ok := state.Strategies[id]
    if !ok { continue }
    if pv := PortfolioValue(s, prices); pv > maxPV {
        maxPV = pv
    }
}
total += maxPV
```

The `maxPV` approach was designed to avoid double-counting when the real wallet balance IS available (correct for that case). But each strategy maintains its own per-strategy virtual `s.Cash` slice (verified at `portfolio.go:212` and the only `s.Cash` mutation paths in `hyperliquid_balance.go`), so when the real balance is NOT available (fallback path), summing per-strategy PVs is the correct approximation.

The existing peak-freeze (`main.go:742-744`) only prevents `PortfolioRisk.PeakValue` from ratcheting *up* during fallback cycles — it does NOT prevent the kill switch from firing on a falsely low fallback PV against an already-set peak. That's why this incident triggered.

## Fix

Change fallback from `maxPV` to `sumPV`:

```go
sumPV := 0.0
for _, id := range ids {
    s, ok := state.Strategies[id]
    if !ok { continue }
    sumPV += PortfolioValue(s, prices)
}
total += sumPV
```

Note: #243 (the original issue this logic addressed) was about double-counting when real balance IS fetched — that path is unaffected since `total += bal` (real balance) is used instead.

## Evidence

From `kill_switch_events` on 2026-04-24T00:33:26Z:

```
triggered | equity | 49.4% | value=$647.61, peak=$1280.98
```

$647.61 matches exactly `hl-tema-eth-live`'s standalone PV, confirming `max(PV)` was used instead of `sum`.

---

## Related defect: `forceCloseHyperliquidLive` doesn't re-fetch to verify on-chain state

Rich confirmed the on-chain HL position WAS actually closed by the kill switch in this incident, but `ConfirmedFlat()` returned false — so `forceCloseAllPositions` (gated at `main.go:848` by `plan.OnChainConfirmedFlat`) never ran and virtual state still showed the position as open.

Root cause: `HyperliquidLiveCloseReport.ConfirmedFlat()` (`hyperliquid_balance.go:539-541`) is just `len(r.Errors) == 0`. `forceCloseHyperliquidLive` (`hyperliquid_balance.go:586`) does NOT re-fetch positions after closing to verify. If the close subprocess returns an error after the order already filled on-chain (e.g. timeout, connection drop post-submit), the error is recorded, `ConfirmedFlat()` returns false, and virtual state is preserved despite the actual position being flat.

This creates state divergence: on-chain flat, virtual state still has the position open. Combined with the maxPV bug above, the next cycle's fallback PV is computed against now-stale virtual positions, perpetuating the latched-retry loop.

**Fix options:**
1. After every successful close call, re-fetch HL positions and treat coin as confirmed flat if `szi == 0`, regardless of subprocess error
1. Add a verification pass before returning the report — if `ClosedCoins` is non-empty AND a follow-up `FetchHyperliquidPositions` shows no remaining configured exposure, flip `Errors` entries to `AlreadyFlat`

## Downstream symptom: false-kill latched-retry loop

When the maxPV fallback falsely triggers the kill switch, the LATCHED retry loop cannot resolve itself even though the portfolio is healthy. Each subsequent cycle re-enters the fallback path (assuming the API blip persists for several cycles), recomputes maxPV against virtual positions that may now be stale (per the close-verification bug above), and re-fires.

This is a symptom, not a separate defect — fixing maxPV→sumPV plus the re-fetch verification should eliminate the loop. Tracking here so we don't lose sight of the operator-experience side.

---

## Note on a previously-flagged concern (resolved by code reading, not a bug)

An earlier draft claimed the Discord notification implies the position was successfully closed while virtual state still shows it open. This was a misread: `formatKillSwitchMessage` (`kill_switch_close.go:586`) already produces `**PORTFOLIO KILL SWITCH (LATCHED, RETRYING)** ... Virtual state preserved. Next cycle will retry.` for the un-confirmed-flat path. The message is correct — no fix needed on the notification side.

Reference only: original report drafted by GLM 5.1; validated and consolidated by Claude Opus 4.7 (1M) | high

richkuo:
## Summary

- Change shared-wallet portfolio fallback from `max(member PV)` to `sum(member PV)` when the real wallet balance fetch is missing.
- Keep the fallback flag so portfolio peaks are still not ratcheted upward during balance-fetch failure cycles.
- Add Hyperliquid post-close verification after close subprocess errors, clearing the close error when a follow-up state fetch proves the coin is flat on-chain.

## Root Cause

The real-balance shared-wallet path correctly contributes one fetched account value to avoid double-counting, but the fetch-failure fallback reused a `max(member PV)` approximation. In shared-wallet HL setups, each strategy carries its own virtual cash/position slice, so fallback should sum those slices; using the max could understate equity and trigger a false portfolio kill switch.

A related kill-switch close path could also stay latched after an HL close actually filled on-chain if the close subprocess returned an error after submission. The planner now verifies HL state after such errors before deciding whether virtual state must remain preserved.

Closes #452

## Validation

LLM: GPT-5 Codex | medium

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -527,7 +527,9 @@ type HyperliquidLiveCloseReport struct {
 	// pre-filters szi≠0, so this branch should not fire in production) AND the
 	// adapter-side already_flat envelope flag, which IS production-reachable
 	// when the eventual-consistency window between the Go-side fetch and the
-	// SDK submit lets a position close out from under us (#350).
+	// SDK submit lets a position close out from under us (#350), or when a
+	// post-close verification fetch proves a coin is flat after the close
+	// subprocess returned an error (#452).
 	AlreadyFlat []string
 	// Errors is non-nil so coin-keyed writes don't panic; len() works on nil maps too.
 	Errors map[string]error
--- a/scheduler/kill_switch_close.go
+++ b/scheduler/kill_switch_close.go
@@ -201,6 +201,34 @@ func defaultHLStateFetcher(addr string) ([]HLPosition, error) {
 	return pos, err
 }
 
+// clearVerifiedFlatHLErrors removes close errors for coins that a follow-up
+// clearinghouseState fetch proves are now flat. This handles the post-submit
+// failure window where the reduce-only close filled on-chain, but the Python
+// subprocess still returned an error before Go saw a confirmed result (#452).
+func clearVerifiedFlatHLErrors(report *HyperliquidLiveCloseReport, positions []HLPosition) []string {
+	if report == nil || len(report.Errors) == 0 {
+		return nil
+	}
+
+	open := make(map[string]bool)
+	for _, p := range positions {
+		if p.Size != 0 {
+			open[p.Coin] = true
+		}
+	}
+
+	var verified []string
+	for _, coin := range report.SortedErrorCoins() {
+		if open[coin] {
+			continue
+		}
+		delete(report.Errors, coin)
+		report.AlreadyFlat = append(report.AlreadyFlat, coin)
+		verified = append(verified, coin)
+	}
+	return verified
+}
+
 // planKillSwitchClose runs the kill-switch close logic without touching any
 // mutable state — no locks, no virtual state mutation, no Discord delivery.
 // The caller applies mutations based on the returned plan.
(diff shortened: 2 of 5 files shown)


richkuo:
@claude

PR #469, opened 2026-04-28
Fix Hyperliquid kill-switch fill attribution for shared coins

### LINKED ISSUE #466: Fix fill PnL mispricing in shared-coin heterogeneous CapitalPct setups (kill-switch)
## Background

Identified in PR #464 (Fix Hyperliquid kill-switch to record real close fills) as a pre-existing issue, not introduced by that PR.

## Problem

When multiple strategies share a coin on Hyperliquid and their `CapitalPct` ratios diverge
from their actual virtual position ratios, the per-strategy fill share from
`hyperliquidKillSwitchFillShare` over- or under-shoots the true close quantity.

- **Over-share**: `applyHyperliquidCircuitCloseFill` caps `qtyClosed` at `pos.Quantity`
  (`scheduler/hyperliquid_balance.go:1205-1208`), so excess is silently dropped — no peer
  redistribution.
- **Under-share**: the virtual position residual is then closed at mark price by
  `forceCloseAllPositions` (`scheduler/risk.go:1075`), mispricing that sliver of PnL.
  The fail-closed branch in `hyperliquidKillSwitchFillShare` even acknowledges this
  fallback in its own comment.

The root cause is that `hyperliquidKillSwitchFillShare` uses capital-weight fractions
(`hlStrategyCapitalWeights` reads `CapitalPct`/`Capital`), which don't necessarily match
the runtime virtual quantity fractions when strategies have been running for different
durations or have diverged from their initial capital allocations (e.g. one strategy
compounded gains, another took losses).

## Affected code

- `scheduler/hyperliquid_balance.go` — `hyperliquidKillSwitchFillShare`,
  `applyHyperliquidKillSwitchCloseFill`, `applyHyperliquidCircuitCloseFill` (qty cap)
- `scheduler/risk.go` — `forceCloseKillSwitchPositions` → `forceCloseAllPositions`
  (residual mark-price close path)

Note: `computeHyperliquidCircuitCloseQty` is **not** affected — it sizes a per-strategy
CB *order* before submission, separate from the kill-switch fill-split path.

## Suggested fix

Split proportionally by virtual `pos.Quantity` instead of `CapitalPct` when computing
per-strategy fill shares.

Caveat: this is not a one-line weight-formula swap. `hyperliquidKillSwitchFillShare`
currently sees only `[]StrategyConfig` for peers — to weight by virtual quantity it
would need a peer-state snapshot taken before the kill-switch close runs (with
appropriate locking), or the caller would need to compute and pass the per-peer
quantities in.

## Impact

The accuracy hit is bounded by `(mark − fill) × residual_qty`. Since mark is snapshotted
in the same kill-switch cycle as the fill, the spread on liquid HL coins is typically
small — the per-event dollar error is usually cents, not material PnL leakage.

Higher when positions have drifted significantly *and* the mark-vs-fill spread is wide
at close time. Low priority — bookkeeping precision rather than meaningful capital
loss — but worth fixing for clean per-strategy PnL attribution.

Ref: PR #464, review observation 2

richkuo:
## Summary
- Snapshot Hyperliquid virtual quantities before kill-switch flattening and use those quantities to split shared-coin fills.
- Thread the snapshot through the confirmed-flat close path so sequential state mutation does not distort peer ratios.
- Add regression coverage for shared-coin closes where configured capital percentages diverge from runtime position sizes.

## Testing
- Added and updated Go unit tests for the Hyperliquid kill-switch close path.
- Ran `go test ./...` in `scheduler` and the focused Hyperliquid kill-switch test subset.
- Verified formatting and diff hygiene with `gofmt` and `git diff --check`.

Closes #466

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -700,6 +700,39 @@ func hlStrategyCapitalWeights(peers []StrategyConfig) []float64 {
 	return out
 }
 
+type hlVirtualQuantitySnapshot map[string]map[string]float64
+
+// snapshotHyperliquidVirtualQuantities captures the per-strategy virtual
+// quantities that exist before a portfolio kill-switch close mutates state.
+func snapshotHyperliquidVirtualQuantities(strategies map[string]*StrategyState, hlLiveAll []StrategyConfig) hlVirtualQuantitySnapshot {
+	if len(strategies) == 0 || len(hlLiveAll) == 0 {
+		return nil
+	}
+	out := make(hlVirtualQuantitySnapshot)
+	for _, sc := range hlLiveAll {
+		coin := hyperliquidSymbol(sc.Args)
+		if coin == "" {
+			continue
+		}
+		ss := strategies[sc.ID]
+		if ss == nil {
+			continue
+		}
+		pos := ss.Positions[coin]
+		if pos == nil || pos.Quantity <= 0 {
+			continue
+		}
+		if out[coin] == nil {
+			out[coin] = make(map[string]float64)
+		}
+		out[coin][sc.ID] = pos.Quantity
+	}
+	if len(out) == 0 {
+		return nil
+	}
+	return out
+}
+
 // computeHyperliquidCircuitCloseQty returns the unsigned coin quantity for a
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

PR #475, opened 2026-04-29
Add per-position trade IDs for lifetime stats

### LINKED ISSUE #471: trades table: partial closes inflate #T and corrupt W/L — add position_id for per-position aggregation
### Problem

Each row in `trades` is a single fill. Close legs are flagged with `is_close=1` and carry `realized_pnl`. `LifetimeTradeStatsAll` (state.go / db.go) computes:

- `#T`  = `COUNT(*) WHERE is_close=1`
- `W/L` = wins counted on `realized_pnl > 0`, losses on `realized_pnl < 0` (breakevens are neither — see db.go:1199-1206)

When a single position closes in N partial fills, both metrics treat it as N independent round trips.

**Example:** open 0.5 BTC → close 0.25 (PnL +\$10) → close 0.25 (PnL -\$3). Net: 1 winning round trip, +\$7. Currently reported: 2 round trips, 1W/1L.

### Proposed fix

Add `position_id TEXT` to `trades`. Every open leg generates a **freshly unique** id (UUID or timestamp+nonce — not derived from contract or symbol); subsequent partial closes for that position carry the same id. Then:

- `#T`  = count of distinct `position_id`s among close rows, scoped per strategy (with legacy-NULL handling — see SQL caveat below)
- `W/L` = wins on `SUM(realized_pnl) > 0`, losses on `SUM(realized_pnl) < 0`, grouped by `(strategy_id, position_id)`. Net-zero positions are neither (matches today's per-leg behavior).

### SQL caveat (legacy NULL rows + per-strategy aggregation)

A naive `COUNT(DISTINCT position_id)` and `GROUP BY position_id` will silently corrupt data:

- `COUNT(DISTINCT col)` skips NULLs in SQLite, so legacy NULL close rows would count as **0** instead of one-per-leg.
- `GROUP BY col` collapses all NULL rows into a **single group**, so their `realized_pnl` would sum together as one synthetic round trip in W/L.
- `LifetimeTradeStatsAll` returns stats **per strategy** — grouping must include `strategy_id` or attribution is lost.

The implementation must:

1. Use a fallback key per legacy row, e.g.:

   ```sql
   CASE
     WHEN position_id IS NULL OR position_id = ''
     THEN 'legacy:' || rowid
     ELSE position_id
   END
   ```

2. Group by **both** `strategy_id` and the fallback-key expression (inner aggregation), then outer-aggregate per strategy. Note `<` (strict) for losses to match existing breakeven semantics:

   ```sql
   SELECT strategy_id,
          COUNT(*) AS round_trips,
          SUM(CASE WHEN net_pnl > 0 THEN 1 ELSE 0 END) AS wins,
          SUM(CASE WHEN net_pnl < 0 THEN 1 ELSE 0 END) AS losses
   FROM (
     SELECT strategy_id,
            CASE WHEN position_id IS NULL OR position_id = ''
                 THEN 'legacy:' || rowid
                 ELSE position_id END AS pkey,
            SUM(realized_pnl) AS net_pnl
     FROM trades
     WHERE is_close = 1
     GROUP BY strategy_id, pkey
   )
   GROUP BY strategy_id;
   ```

This gives every legacy row its own synthetic group (preserving today's per-leg counting for historical data) and ensures two strategies that happen to share an id can't bleed into each other.

### Options-specific caveat (do NOT reuse `OptionPosition.ID`)

`OptionPosition.ID` is built deterministically from contract details at `options.go:125`:

```go
posID := fmt.Sprintf(\"%s-%s-%s-%.0f-%s\",
    result.Underlying, action.OptionType, action.Action, action.Strike, action.Expiry)
```

This means two separate buys of the same option contract at different times produce the **same** `posID`. If reused as `trades.position_id`, those distinct round trips would silently collapse into one in `#T` and net into a single W/L outcome.

For options, generate a fresh per-open `position_id` (e.g. `posID + \"-\" + timestamp` or a UUID) and persist it on `OptionPosition` separately from the contract-derived display ID. Same rule applies anywhere else a position struct already has a non-unique `ID` field.

### Position persistence (positions / option_positions)

Open positions are persisted across restarts via the `positions` and `option_positions` tables (db.go:63-75 and db.go:120-139). Neither has a column for the trade-position id today. If we generate the id in memory only, a restart between open and close drops it — and the post-restart partial closes write `position_id = ''` to the trades table, falling back to per-leg legacy behavior for that position.

Required:

- Add `position_id TEXT NOT NULL DEFAULT ''` to **both** `positions` and `option_positions` (idempotent ALTER TABLE migrations).
- Update `Position` and `OptionPosition` Go structs and the corresponding save/load paths to round-trip the column.
- Note: the existing option_positions `id` column is the contract-derived display ID — keep it, add the new trade-position-id alongside.

### SaveState fallback insert path (db.go:666)

There are **two** `INSERT INTO trades` paths in db.go:

1. `InsertTrade` (eager, called by `RecordTrade`).
2. SaveState's append-only flush at db.go:666 — re-inserts any `TradeHistory` rows whose eager `InsertTrade` failed earlier in the cycle.

Both must be updated to include `position_id`. Skipping the fallback path would leave a quiet bug: trades that hit the retry path land with NULL `position_id` even though the in-memory row has it set.

### Trade SELECT paths must round-trip `position_id`

Inserting the column without updating the read paths leaves in-memory and audit views inconsistent. Two trade-row SELECTs need their column lists extended:

1. **db.go:1095** — `LoadState` per-strategy trade history loader. After restart, in-memory `Trade` rows would have empty `position_id` even though the column has data. Loaded rows are marked `persisted=true`, so this is primarily a consistency/auditability issue rather than the SaveState retry path.
2. **db.go:1270** — `QueryTrades` (audit/history endpoint backing `/history`). Without `position_id` in the SELECT and the response struct, operator-facing trade audits never see the id.

Required:

- Add `PositionID string` field to the `Trade` struct.
- Extend SELECT lists at db.go:1095 and db.go:1270 to include `position_id`.
- For legacy rows, use `COALESCE(position_id, '') AS position_id` in SELECTs or scan via `sql.NullString`; do **not** scan nullable `position_id` directly into a Go `string`.
- Extend row scanners at both call sites to populate the new field.
- Decide whether `/history` HTTP response should expose `position_id` (recommend yes for consistency).

`backfillTradeCloseFlags` (db.go:308, only reads `rowid, details`) and the TradingView export query (db.go:1314, out of scope) do **not** need updating.

### Scope

- Schema: add `position_id TEXT` to `trades`, `positions`, and `option_positions` (idempotent migrations); index `(strategy_id, position_id)` on `trades`.
- `Position` struct: add `TradePositionID` (UUID or unique string); generate at open.
- `OptionPosition` struct: add `TradePositionID` separate from the contract-derived `ID`; generate at open.
- `Trade` struct: add `PositionID string`.
- Position save/load (positions + option_positions): round-trip the new column.
- `InsertTrade` / `RecordTrade`: accept and persist `position_id`.
- SaveState flush insert (db.go:666): include `position_id` in the column list.
- LoadState trade loader (db.go:1095) and `QueryTrades` (db.go:1270): include `position_id` in SELECT and scan it into `Trade.PositionID`, handling legacy NULLs safely via `COALESCE` or `sql.NullString`.
- `TradeHistory` in-memory struct: carry `position_id` so the SaveState fallback path has it available.
- `LifetimeTradeStatsAll`: rewrite using the (strategy_id, fallback-key) grouping, `<` for losses.
- All open paths populate `position_id`: spot, perps (incl. bidirectional flips), futures, options.
- Perps flip semantics: a flip retires the old `position_id` and starts a new one (flip = close + reopen).
- Backfill: leave legacy rows with NULL/empty `position_id`; the fallback-key expression handles them.

### Acceptance

- New unit test: 1 open + 2 partial closes → `#T`=1, W/L computed on net realized PnL.
- New unit test: legacy rows with NULL/empty `position_id` count one-per-leg (matches pre-fix behavior).
- New unit test: same option contract opened, closed, then opened+closed again → `#T`=2 (not 1).
- New unit test: two strategies with overlapping `position_id` values cannot bleed counts into each other.
- New unit test: breakeven close (`realized_pnl == 0`) counts as neither win nor loss.
- New unit test: position persists `position_id` across save/load round trip (both `positions` and `option_positions`).
- New unit test: SaveState flush path writes `position_id` correctly when `InsertTrade` is bypassed.
- New unit test: LoadState reads `position_id` back into `Trade.PositionID` (round-trip) without NULL-scan errors on legacy rows.
- New unit test: `QueryTrades` returns `position_id` for both new and legacy rows without NULL-scan errors.
- Existing tests still pass.

### Out of scope

- TradingView export changes (separate issue if export needs per-position grouping).
- Per-position PnL display in Discord summaries / `/status` (separate enhancement).
- Removal of vestigial `RiskState.TotalTrades` / `WinningTrades` / `LosingTrades` counters — tracked in #472. Note: after this issue ships, those in-memory counters will still be incremented per close leg, so the `FormatCategorySummary` fallback branch (`discord.go:441-447`) remains buggy for partial closes until #472 lands. Acceptable because the fallback only fires when DB has no rows for a strategy.

### Follow-up

- #472 — remove the vestigial `RiskState` counters once this lands and runs cleanly for a cycle.

richkuo:
### Summary
- Add `position_id` persistence across `trades`, `positions`, and `option_positions`.
- Thread stable per-position IDs through open, close, assignment, stop-loss, and circuit-breaker trade paths.
- Rewrite lifetime trade stats to aggregate by position and preserve legacy NULL/empty rows.
- Extend load/query paths and add regression coverage for partial closes, legacy rows, and option reopen behavior.

### Testing
- `go test ./...` in `scheduler`
- Focused regression tests for lifetime stats, round-trip persistence, and trade history queries

Closes #471

LLM: GPT-5 | extra high

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -63,6 +63,7 @@ CREATE TABLE IF NOT EXISTS strategies (
 CREATE TABLE IF NOT EXISTS positions (
     strategy_id TEXT NOT NULL REFERENCES strategies(id) ON DELETE CASCADE,
     symbol TEXT NOT NULL,
+    position_id TEXT NOT NULL DEFAULT '',
     quantity REAL NOT NULL,
     avg_cost REAL NOT NULL,
     side TEXT NOT NULL,
@@ -120,6 +121,7 @@ CREATE INDEX IF NOT EXISTS idx_closed_opt_closed_at ON closed_option_positions(c
 CREATE TABLE IF NOT EXISTS option_positions (
     strategy_id TEXT NOT NULL REFERENCES strategies(id) ON DELETE CASCADE,
     id TEXT NOT NULL,
+    position_id TEXT NOT NULL DEFAULT '',
     underlying TEXT NOT NULL,
     option_type TEXT NOT NULL,
     strike REAL NOT NULL,
@@ -143,6 +145,7 @@ CREATE TABLE IF NOT EXISTS trades (
     strategy_id TEXT NOT NULL,
     timestamp TEXT NOT NULL,
     symbol TEXT NOT NULL,
+    position_id TEXT NOT NULL DEFAULT '',
     side TEXT NOT NULL,
     quantity REAL NOT NULL,
     price REAL NOT NULL,
@@ -158,8 +161,9 @@ CREATE TABLE IF NOT EXISTS trades (
 CREATE INDEX IF NOT EXISTS idx_trades_strategy ON trades(strategy_id);
 CREATE INDEX IF NOT EXISTS idx_trades_symbol ON trades(symbol);
 CREATE INDEX IF NOT EXISTS idx_trades_timestamp ON trades(timestamp DESC);
--- idx_trades_close (#455) is created in migrateSchema, not here, so legacy
--- DBs add the is_close column before the index references it.
+-- idx_trades_close (#455) and idx_trades_strategy_position (#471) are created
+-- in migrateSchema, not here, so legacy DBs add columns before indexes
+-- reference them.
 
 CREATE TABLE IF NOT EXISTS portfolio_risk (
     id INTEGER PRIMARY KEY CHECK (id = 1),
@@ -251,6 +255,11 @@ func (sdb *StateDB) migrateSchema() error {
 		"ALTER TABLE trades ADD COLUMN is_close INTEGER NOT NULL DEFAULT 0",
 		"ALTER TABLE trades ADD COLUMN realized_pnl REAL NOT NULL DEFAULT 0",
 		"CREATE INDEX IF NOT EXISTS idx_trades_close ON trades(strategy_id, is_close)",
+		// Per-position trade grouping (#471).
+		"ALTER TABLE trades ADD COLUMN position_id TEXT NOT NULL DEFAULT ''",
+		"ALTER TABLE positions ADD COLUMN position_id TEXT NOT NULL DEFAULT ''",
+		"ALTER TABLE option_positions ADD COLUMN position_id TEXT NOT NULL DEFAULT ''",
+		"CREATE INDEX IF NOT EXISTS idx_trades_strategy_position ON trades(strategy_id, position_id)",
 	}
 	for _, ddl := range migrations {
 		if _, err := sdb.db.Exec(ddl); err != nil {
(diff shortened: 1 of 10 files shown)


richkuo:
@claude

richkuo:
@claude

PR #476, opened 2026-04-29
Fix summary cadence wall-clock throttling

### LINKED ISSUE #474: fix: summary cadence ignores wall-clock time, posts every micro-cycle
## Problem

After #417 (`b4235d2`), `schedulerDelay` became variable (1s..longest-strategy-interval) instead of fixed at the config tick interval. The main loop now wakes up many times per `intervalSeconds` window — `state.CycleCount++` (`main.go:415`) increments once per loop iteration, so cycle count grows much faster than wall-clock / `intervalSeconds`.

`ShouldPostSummary` (`config.go:154-197`) still uses cycle-count math: `cyclesBetween = duration / intervalSeconds`. That assumes one cycle == `intervalSeconds` wall-clock seconds, which no longer holds.

Worked example with `freq="5m"`, `intervalSeconds=300`, 15 staggered strategies on 300s:
- `cyclesBetween = 300/300 = 1`
- `(cycle-1) % 1 == 0` for every cycle
- `chRan` gates posting on "channel has a due strategy" → fires ~once every 20s
- Result: ~15 posts per 5min instead of 1. With `freq="1h"` (`cyclesBetween=12`) the throttle becomes ~4 min instead of 1h.

`b4235d2` already converted `lastAutoUpdateCheck` to wall-clock and acknowledged the cycle-vs-wallclock decoupling (`main.go:397-404`). `ShouldPostSummary` is the last remaining cycle-count consumer of this kind — `LeaderboardPostTime`, heartbeat auto-update, and `daily` auto-update are all already wall-clock.

## Root Cause

Commit `b4235d2` (#417) — "Speed up drawdown warning checks" changed the scheduler delay from fixed-interval to variable, decoupling cycle count from wall-clock time.

## Fix

Replace cycle-count gating with wall-clock tracking, mirroring the pattern `lastRun` already uses (`main.go:351`):

- Add `lastSummaryPost map[string]time.Time` keyed by channel key.
- Helper signature becomes pure and time-injectable:
  ```go
  ShouldPostSummary(freq string, continuous, hasTrades bool, lastPost, now time.Time) bool
  ```
  `cycle` and `intervalSeconds` drop out of the gating helper. `cycle` is still passed separately to `FormatCategorySummary` for the display header — leave that alone.
- Delete `summaryCyclesBetween` (`config.go:165-197`); it becomes dead code.

### Behavior spec (tightened)

| `freq` value | continuous | Behavior |
|---|---|---|
| empty (legacy) | true | Post when `chRan` (every channel-run, current behavior) |
| empty (legacy) | false | Wall-clock 1h (matches today's "12 cycles × 300s" intent) |
| `every` / `per_check` / `always` (`dur==0`) | any | Post when `chRan` |
| `hourly` / `daily` / explicit duration (`dur>0`) | any | `time.Since(lastPost) >= dur` |
| invalid | any | Fall back to legacy (continuous=chRan, non-continuous=1h) |

The "5m default for continuous channels" idea from the original draft is dropped — silently changing UX for users with empty config is wrong.

### Trade-forced posts must reset the window

`hasTrades=true` posts AND advances `lastPost[chKey]` to `now`. Otherwise a trade alert followed seconds later by a cadence-driven post defeats the throttle. Existing `TestShouldPostSummary_TradesForcePost` only checks the post fires; no test covers the window reset.

### Restart safety: persist `lastSummaryPost`

Add `LastSummaryPost map[string]time.Time` to `AppState` (`state.go:75` block) and the SQLite serializer (`db.go`). In-memory-only would cause a post-blast on every restart — acceptable for occasional restarts, noisy under auto-updates / service flapping. Cost is tiny; restart safety is meaningful.

### SIGHUP hot reload

`cfg.SummaryFrequency[chKey]` is read live, so the wall-clock approach picks up reloads naturally. **Do not clear `lastSummaryPost` in `applyHotReloadConfig`** — channel keys are stable, and clearing would cause a post-blast on every SIGHUP.

### Affected code

- `scheduler/config.go` — rewrite `ShouldPostSummary`, delete `summaryCyclesBetween`
- `scheduler/main.go` — declare `lastSummaryPost` near `lastRun` (`:351`); call new signature at `:1396`; update `lastSummaryPost[chKey] = now` after a successful post (including `hasTrades=true`)
- `scheduler/state.go` + `scheduler/db.go` — persist `LastSummaryPost` in `AppState` + SQLite
- `scheduler/config_reload.go` — confirm `lastSummaryPost` is NOT cleared
- `scheduler/summary_frequency_test.go` — full rewrite (cycle-based today; needs `now`/`lastPost` injection)

### Test plan

- `hasTrades=true` posts AND advances `lastPost` (post at T=0, no trade at T=10s should suppress; trade at T=10s should post and reset).
- `dur > 0`: post fires only when `now - lastPost >= dur`. Test cycle: post at T=0, suppress at T=dur-1s, post at T=dur, suppress immediately after, post at T=2*dur.
- `every` / legacy-continuous: posts on every call regardless of elapsed time.
- Legacy empty + non-continuous: 1h wall-clock cadence (regression-protect the migration).
- Hot-reload mid-window: change freq from `5m` → `1h`; existing `lastPost` continues to gate against the new duration.
- Invalid freq falls back to legacy (existing test, updated for new signature).
- Round-trip persistence: save state with non-empty `LastSummaryPost`, reload, verify timestamps survive.

## Severity / scope

Medium severity: operational noise (Discord/Telegram spam, possible rate-limit pressure on busy configs), no capital impact. Worth fixing soon, not a hotfix.

Effort: ~half-day. ~50 LOC in `config.go` + `main.go`, ~15 LOC for state persistence in `state.go`/`db.go`, full rewrite of `summary_frequency_test.go`. No public API changes outside `scheduler/`.

Reference only: LLM model used: Claude Opus 4.7 (1M) | xhigh

richkuo:
## Summary
- Replace cycle-count summary throttling with wall-clock per-channel timestamps so variable scheduler wakeups do not spam summaries.
- Persist `LastSummaryPost` in SQLite state so restarts and auto-updates keep cadence windows intact.
- Rewrite summary cadence tests for wall-clock behavior, trade-forced window resets, persistence, migration, and hot reload preservation.

## Test plan
- [x] `cd scheduler && /opt/homebrew/bin/gofmt -w config.go main.go state.go db.go summary_frequency_test.go db_test.go config_reload_test.go`
- [x] `cd scheduler && /opt/homebrew/bin/go test ./...`
- [x] `cd scheduler && /opt/homebrew/bin/go build .`

Closes #474

## Metadata
LLM: GPT-5.5 | planning effort: high | build effort: high

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -107,13 +107,13 @@ type Config struct {
 	Correlation          *CorrelationConfig         `json:"correlation,omitempty"`
 	Platforms            map[string]*PlatformConfig `json:"platforms,omitempty"`
 	LeaderboardSummaries []LeaderboardSummaryConfig `json:"leaderboard_summaries,omitempty"` // #308 — configurable per-channel leaderboards
-	SummaryFrequency     map[string]string          `json:"summary_frequency,omitempty"`     // #30 — per-channel summary cadence; keys match Discord/Telegram channel keys (e.g. "spot", "options", "hyperliquid"). Values: Go duration ("30m", "2h"), alias ("hourly", "every"/"per_check"/"always"), or empty for legacy default (continuous: every cycle; spot: hourly)
+	SummaryFrequency     map[string]string          `json:"summary_frequency,omitempty"`     // #30 — per-channel summary cadence; keys match Discord/Telegram channel keys (e.g. "spot", "options", "hyperliquid"). Values: Go duration ("30m", "2h"), alias ("hourly", "every"/"per_check"/"always"), or empty for legacy default (continuous: every channel run; spot: hourly)
 	RiskFreeRate         *float64                   `json:"risk_free_rate,omitempty"`        // #397 — annualized risk-free rate used in Sharpe-ratio calculations (e.g. 0.02 for 2%). Nil/missing falls back to DefaultAnnualRiskFreeRate; an explicit 0 is respected so backtest comparisons can pin to a 0% benchmark.
 	TradingViewExport    TradingViewExportConfig    `json:"tradingview_export,omitempty"`    // #3 — optional symbol overrides for TradingView portfolio CSV exports
 }
 
 // ParseSummaryFrequency converts a summary_frequency value to a duration.
-// Returns -1 to mean "use legacy default", 0 to mean "every cycle", or a
+// Returns -1 to mean "use legacy default", 0 to mean "every channel run", or a
 // positive duration when caller should post every duration. An unrecognized
 // value returns a non-nil error.
 func ParseSummaryFrequency(s string) (time.Duration, error) {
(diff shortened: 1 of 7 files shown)


richkuo:
@claude

PR #483, opened 2026-04-29
Add open/close strategy composition support

### LINKED ISSUE #480: Split strategies into open + close concepts (decoupled entry/exit)
## Problem

A strategy today is a single Python function (`shared_strategies/registry.py`) that emits an integer `signal` (-1/0/1) from OHLCV. The same function handles **both** entry and exit — `signal=1` opens long, `signal=-1` closes long (and optionally opens short). This conflates two genuinely independent decisions:

- *When to enter*
- *When to exit*

The coupling means you can't reuse exit logic across entry strategies. Other algo frameworks (backtrader, freqtrade) treat these as orthogonal — this issue proposes the same split here.

## Proposed model

- **Open strategy**: OHLCV → `{action: \"long\" | \"short\" | \"none\"}`. No exit logic.
- **Close strategy**: `(position, market_state, params)` → `{close_fraction: 0..1}`. No entry logic.
- A configured trading strategy = one open strategy + an array of close strategies.

```jsonc
{
  \"id\": \"hl-btc-example\",
  \"open_strategy\": \"sma_crossover\",
  \"close_strategies\": [\"close_strategy_a\", \"close_strategy_b\"],
  \"disable_implicit_close\": false
}
```

### Composition rule

When multiple close strategies fire the same cycle: **max `close_fraction` wins.** Order-independent, no priority surprises. (Alternatives — priority-order or sum-capped — rejected as adding complexity without a current use case.)

### Ordering within a cycle

Close layer runs before open layer (close long → then maybe open short on the now-flat position). This already matches how `ExecutePerpsSignal` / `ExecuteSpotSignal` sequence their close-then-flip path.

## Back-compat — zero config migration

Every existing monolithic strategy auto-becomes `open_strategy: <self>, close_strategies: [<self>]`:
- Open layer reads `signal=1` as \"open long\", `signal=-1` as \"open short\" (gated by `AllowShorts`).
- Close layer reads `signal=-1` as \"close long\", `signal=1` as \"close short\".
- The Python function runs **once per cycle**; both layers consume the same result.

Existing configs work bit-for-bit. New configs opt into explicit `open_strategy` / `close_strategies` fields when they want decoupled exits.

A `disable_implicit_close: true` flag opt-out covers the case where a user wants explicit-close-only behavior (no signal-reversal exits).

## Scope (what this issue covers)

1. **Close-strategy interface**
   - Python signature + JSON return shape
   - Where close-strategy implementations live (proposed: `shared_strategies/exits/`)
   - Registry mechanism (parallel to entry registry, or kind-tagged in the same one)
2. **Config schema** (`scheduler/config.go`)
   - `OpenStrategy string` (defaults to existing `Script` / strategy name for back-compat)
   - `CloseStrategies []string`
   - `DisableImplicitClose bool`
   - `ValidateConfig`: open strategy must exist; close strategies must exist; reject unknown names
3. **Go dispatch** (`scheduler/main.go`)
   - Per-cycle: run open eval, then iterate close strategies on each open position
   - Apply max-wins composition
   - Wire results into existing `ExecuteSpotSignal` / `ExecutePerpsSignal` / `ExecuteFuturesSignal` paths
4. **Check scripts** — extend `shared_scripts/check_*.py` to invoke close strategies (or a new dedicated script)
5. **Backtester parity** (`backtest/backtester.py`) — same composition logic in simulation, otherwise optimization is meaningless
6. **Tests**
   - Pure helpers: max-wins composition, ordering (close-before-open), implicit-self-as-close mapping
   - Back-compat: a config with no `open_strategy` / `close_strategies` produces identical behavior to today
7. **Docs** — CLAUDE.md update covering the open/close split and the back-compat mapping

## Out of scope (separate issues)

- Specific close strategy implementations — each gets its own issue/PR
- Rewriting existing entry strategies into pure \"open strategies\" — they continue to work via the implicit mapping; can be cleaned up later if desired
- Per-strategy config knobs for individual close strategies — handled by each close strategy's own params block

## Sharp edges to call out during implementation

- **Per-cycle cost**: today is one Python invocation per strategy. New model adds N close evals per open position. Either batch into one subprocess call or accept the overhead.
- **`Position.Quantity` mutability**: partial closes from a close strategy mean qty shrinks mid-life. Audit drawdown calc, leaderboard, fee attribution to confirm nothing assumes it's immutable post-open.
- **Round-trip grouping (#471)**: every partial close must share the original `PositionID` so `#T` / `W/L` collapse correctly.
- **Catastrophic flip path** (`triple_ema_bidir`) needs to play nicely with explicit close strategies — flips should cancel pending exit-state from the close layer.

## Migration plan

Ship in stages:
1. Interface + config schema + back-compat mapping (no behavior change)
2. First close strategy implementation (separate PR/issue)
3. Backtester parity
4. Optional: rewrite a handful of strategies into explicit open+close form to validate the pattern

richkuo:
## Summary
- Add open/close strategy fields to strategy config and validate them for supported strategy types.
- Thread composed open/close decisions through scheduler checks and shared Python signal checks.
- Extend the backtester to consume `open_action` and max-wins `close_fraction` outputs for split entry/exit logic, including short opens.
- Address review feedback by deferring strategy-name registry lookup to Python, letting check scripts own implicit close fallback, suppressing close-while-flat opens, and trimming paper-mode snapshots.
- Add Go and Python regression coverage for config validation, composition semantics, and backtester parity.

## Testing
- `python3 -m py_compile shared_tools/strategy_composition.py shared_scripts/check_strategy.py shared_scripts/check_hyperliquid.py shared_scripts/check_okx.py shared_scripts/check_robinhood.py shared_scripts/check_topstep.py backtest/backtester.py`
- `/Users/richardkuo/Work/openclaw/go-trader/.venv/bin/python3 -m pytest shared_tools/test_strategy_composition.py backtest/tests/test_open_close_backtester.py -q`
- `/Users/richardkuo/Work/openclaw/go-trader/.venv/bin/python3 -m pytest backtest/tests -q`
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler build -o ../go-trader .`

Closes #480

LLM: GPT-5 | extra high


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude fix the two optional follow ups

richkuo:
@claude sonnet

richkuo:
@claude sonnet fix the minor issue

PR #488, opened 2026-04-29
feat(hl): default HL perps to isolated margin mode

### LINKED ISSUE #486: HL: set isolated margin mode by default
## Problem

go-trader never explicitly sets the margin mode when placing Hyperliquid orders. Whatever mode is configured on the account (cross or isolated) is used silently. Cross margin is the HL default for new accounts, which means a single runaway strategy can drain equity from unrelated positions before the per-strategy drawdown check fires.

## Proposed Change

Before placing a live order, call HL's `update_leverage` API with `is_cross=False` to enforce isolated margin for the position. This aligns on-chain margin mode with go-trader's per-strategy risk model.

**Where to implement:**
- `shared_scripts/check_hyperliquid.py` (or the `--execute` path) — call `exchange.update_leverage(leverage, coin, is_cross=False)` before the order
- Add a `margin_mode` config field to `StrategyConfig` (`"isolated"` | `"cross"`, default `"isolated"`) so operators can opt into cross if desired
- `scheduler/config.go` — new field + validation
- `scheduler/init.go` — wizard prompt (or silent default)

## Why isolated by default

- Hard loss cap per position — a liquidation cannot spill into other strategies' margin
- Matches the intent of per-strategy `max_drawdown_pct` and circuit-breaker logic already in the codebase
- Explicit liq price visible in HL UI (currently shows N/A for cross positions)

## Notes

- Changing margin mode mid-position requires closing first — implementation should only set mode when opening a new position (qty == 0 before order)
- `update_leverage` sets both leverage and cross/isolated in one call on HL

---
LLM: Claude Sonnet 4.6 | high

richkuo:
Closes #486

HL perps strategies now default to isolated margin mode. The scheduler calls `update_leverage(leverage, coin, is_cross=...)` before opening from flat, enforcing a hard per-position loss cap that matches go-trader's per-strategy risk model. Cross remains opt-in via `margin_mode: "cross"` in config.

- New `StrategyConfig.MarginMode` field (HL perps only, `isolated` | `cross`)
- Default isolated applied in both `LoadConfig` and `generateConfig`
- `update_leverage` called only on fresh opens (`posQty == 0`) since HL rejects mode changes on open positions
- Fail-closed: if `update_leverage` raises, the order is aborted
- Hot-reload mirrors `Leverage` envelope — change allowed when flat, rejected when positions open

## Operator note
Run **one HL perps strategy per coin per account**. HL aggregates positions per coin, so a second strategy on the same coin will hit `update_leverage` rejection (HL won't change margin mode on an open position) and fail-close the order. Multi-strategy-per-coin support tracked in #491.

Tests: 4 Go + 6 Python, all passing.

---
LLM: Claude Opus 4.7 (1M) | high | Tokens: 114 in / 37004 out | Cache: 8253415 read / 172397 written | Cost: $6.13


richkuo:
@claude resolve merge conflicts then review

PR #492, opened 2026-04-30
feat(hl): allow multiple HL perps strategies on the same coin

### LINKED ISSUE #491: Allow multiple HL perps strategies on the same coin (margin mode + leverage)
Follow-up from #488 review.

## Problem
Hyperliquid aggregates positions per coin per account, but go-trader tracks them per strategy. When two HL perps strategies share a coin on the same account:

1. Strategy A opens ETH (sets isolated@5x via `update_leverage`).
2. Strategy B (also ETH) sees a fresh signal. Locally `posQty == 0`, so #488's logic calls `update_leverage` again.
3. HL rejects the call because there's already an open ETH position — HL won't let you change margin mode on an open position.
4. Fail-closed kicks in and B's order is aborted.

This makes "two strategies, same coin, same account" effectively unsupported, even when both strategies agree on mode and leverage.

## Proposed fix

**1. Skip `update_leverage` when HL already matches the desired state.**
Query `info.user_state()` (or reuse existing `hlLiveAll` snapshot) before calling `update_leverage`. If `(coin, leverage, margin_mode)` on HL already matches what the strategy wants, skip the call. Strategy B then opens onto the shared on-chain position; per-strategy bookkeeping in SQLite keeps the legs separated.

**2. Config-load validation for peer strategies.**
In `LoadConfig`, group HL perps strategies by `(account, coin)` and require all peers to share `MarginMode` and `Leverage`. Fail fast at startup rather than at first trade. Mirror the same check in `validateHotReloadCompatible`.

## Known limitations even after this fix
- Strategies still share on-chain margin, leverage, and stop-loss OIDs.
- Close fills already split by virtual-qty in `hyperliquidKillSwitchFillShare`, but per-trade stop-loss OIDs (`Position.StopLossOID`) can collide if both strategies place reduce-only triggers.
- Two strategies cannot hold *opposite* directions on the same coin (HL netting).

For full isolation (independent direction, leverage, margin), HL subaccounts are the only correct path — tracked separately if/when we need it.

## Acceptance
- [ ] `update_leverage` skipped when current HL state matches desired
- [ ] Peer-strategy config validation at `LoadConfig` and hot-reload
- [ ] Test: two HL perps strategies on ETH with matching mode/leverage both open successfully
- [ ] Test: mismatched peer config fails at `LoadConfig` with clear error
- [ ] Stop-loss OID collision behavior documented or guarded

richkuo:
Closes #491.

## Summary
- Skip `update_leverage` when HL already matches the desired (margin_mode, leverage), so a peer strategy on a shared coin can open onto the existing on-chain position instead of fail-closing.
- Validate HL perps peers (same coin) share `margin_mode` / `leverage` at `LoadConfig` and on SIGHUP hot reload.
- At most one peer may carry `stop_loss_pct` — avoids OID collision on reduce-only triggers.

## Test plan
- [x] `go test ./...`
- [x] `uv run pytest shared_scripts/test_check_hyperliquid.py`

---
LLM: Claude Opus 4.7 (1M) | high | Tokens: 95 in / 33083 out | Cache: 6654441 read / 180681 written | Cost: $5.28

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -336,7 +336,9 @@ class HyperliquidExchangeAdapter:
 
         HL's SDK takes both fields in one call, so callers always pass both.
         Fails closed: HL rejects this when there is an open position on the
-        coin, so the scheduler must only invoke it when opening from flat.
+        coin, so the scheduler must only invoke it when opening from flat —
+        or when ``get_position_leverage`` confirms the on-chain state already
+        matches the desired (mode, leverage) pair (#491).
         """
         if not self._exchange:
             raise RuntimeError(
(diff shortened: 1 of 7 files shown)


richkuo:
@claude

richkuo:
@claude fix all suggestions

richkuo:
@claude sonnet

PR #502, opened 2026-04-30
Add HL trailing stop-loss support

### LINKED ISSUE #501: HL perps: synthetic trailing stop-loss
## Problem

Today, `stop_loss_pct` (and `stop_loss_margin_pct`) places a single reduce-only trigger at a fixed price computed from entry. Once the position moves favorably, the stop never advances — profits give back to the original SL on a reversal.

Hyperliquid does not offer a native trailing stop at the API level (confirmed against `hyperliquid` Python SDK `TriggerOrderType` and the official exchange-endpoint docs — only `{triggerPx, isMarket, tpsl}` is accepted, no trailing delta). So a trailing stop has to be synthesized by the scheduler: track a high-water mark, and re-issue the trigger when price advances enough to warrant raising the stop.

## Proposed Change

Add `trailing_stop_pct *float64` to `StrategyConfig` (HL perps only, same scope as `stop_loss_pct`).

Semantics:

- Long position: stop trails at `high_water_px × (1 − trailing_stop_pct/100)`. `high_water_px` initialized to `avg_cost` at entry, then ratcheted up on any cycle where mark > current high-water.
- Short position: stop trails at `low_water_px × (1 + trailing_stop_pct/100)`. Ratcheted down only.
- Trail price never moves *against* the position — only ever in the direction of profit.

Mutually exclusive with `stop_loss_pct` and `stop_loss_margin_pct` on the same strategy. `EffectiveStopLossPct(sc)` should return the trailing pct if set.

## State

Add to `Position`:

- `StopLossHighWaterPx float64` — best mark seen while position open (high for long, low for short).
- Existing `StopLossOID` / `StopLossTriggerPx` reused.

Persisted in SQLite `positions` table (new column, idempotent migration via `PRAGMA table_info` — same pattern as the SL columns).

## Cycle logic (main.go / hyperliquid_balance.go)

Per HL perps strategy with `trailing_stop_pct > 0` and an open position, after the mark fetch:

1. Compute new candidate high-water (`max(highWater, mark)` for long, `min` for short).
2. Compute new candidate trigger price from candidate high-water + trailing pct.
3. If `|new_trigger − current_trigger| / current_trigger >= min_move_pct` (default 0.5%, configurable later), enqueue a cancel-and-replace: cancel `StopLossOID`, place a new SL trigger at the new price, update `StopLossOID` / `StopLossTriggerPx` / `StopLossHighWaterPx` under Lock.
4. Otherwise update only `StopLossHighWaterPx` (free, no order churn).

Use existing `place_stop_loss` / `cancel_trigger_order` paths in `check_hyperliquid.py`. A new `--mode=update-stop-loss` (or a dedicated `update_stop_loss.py` entry-point) would let Go re-issue the trigger without spawning a full strategy check.

## Constraints / risks

- **HL 1000 open-trigger cap** (per #481). The min-move threshold is critical to avoid burning OID slots — needs to be tuned conservatively. Default 0.5% means at most ~200 trail updates per 100% move, comfortably under the cap for any one strategy.
- **Cycle gap risk** — between cancel and re-place, the position is briefly unprotected. Order matters: place new SL first if HL allows two reduce-only triggers on the same coin, otherwise accept the gap (stop_loss has always been best-effort).
- **Multi-strategy peers** (#491) — if multiple strategies share a coin, only one peer can own a positive SL. Trailing inherits the same constraint; validation in `hyperliquidPeerStrategyErrors` should treat `trailing_stop_pct > 0` the same as `EffectiveStopLossPct > 0`.
- **Mode/leverage changes** — trailing pct change while position open: allow (it's a price level, no exchange state change). Test in `validateHotReloadStateCompatible`.

## Init wizard

Prompt for trailing during live HL perps setup as a third option alongside `stop_loss_pct` / `stop_loss_margin_pct`. Explain the high-water-mark mechanics with one example.

## Tests

- Pure helper for the trail computation (`computeTrailingStopUpdate(side, mark, highWater, trailingPct, minMovePct, currentTrigger)` → returns new high-water + new trigger or zero deltas).
- Peer-validation regression: trailing on one peer + fixed SL on another should fail.
- Hot-reload: changing `trailing_stop_pct` mid-position should be accepted; switching from fixed → trailing while position open should be rejected (or accepted with documented semantics).
- HL stub closer test confirming cancel-then-place ordering + state mutations under Lock.

## Out of scope

- ATR-based trailing (separate issue).
- Time-based ratchet (e.g., move SL to breakeven after N hours).
- Spot / non-HL platforms — none of them currently support reduce-only triggers via our adapters.

---

richkuo:
## Summary
- Add synthetic Hyperliquid perps trailing stop-loss config, persistence, and scheduler ratcheting.
- Add a dedicated stop-loss update mode that cancels and replaces HL trigger orders.
- Cover trailing math, validation, hot reload, and Python update behavior with focused tests.

Closes #501.

## Test plan
- go test ./... (from scheduler)
- uv run --with pytest pytest shared_scripts/test_check_hyperliquid.py platforms/hyperliquid/test_adapter.py

## LLM
- GPT-5.5, high effort


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

PR #511, opened 2026-04-30
Split shared strategies into open and close registries

richkuo:
## Summary
- Move shared open strategies under `shared_strategies/open/` and update callers/discovery paths.
- Add a formal `shared_strategies/close/` registry with `tp_at_pct`, `tiered_tp_pct`, and `tiered_tp_atr` evaluators.
- Route open/close composition through the close registry first while preserving open-registry fallback compatibility.

Closes #509

## Test plan
- `uv run --extra test pytest`
- `go test ./...` from `scheduler/`
- `uv run python -m py_compile shared_scripts/check_strategy.py shared_scripts/check_hyperliquid.py shared_scripts/check_robinhood.py shared_scripts/check_topstep.py shared_scripts/check_okx.py shared_tools/strategy_composition.py shared_strategies/close/registry.py shared_strategies/close/tp_at_pct.py shared_strategies/close/tiered_tp_pct.py shared_strategies/close/tiered_tp_atr.py`

## LLM Metadata
- LLM: GPT-5.5
- Effort level: high

Made with [Cursor](https://cursor.com)

PR #513, opened 2026-04-30
Fix close registry import resolution

richkuo:
## Summary
- Load the close strategy registry via an explicit loader to avoid `registry.py` sys.path collisions.
- Validate explicit close strategy names against close evaluators before falling back to open strategies.
- Update strategy docs and add an open/close check-script smoke test.

## Test plan
- `uv run --extra test pytest`
- `uv run --extra test pytest shared_scripts/test_check_strategy_open_close.py shared_scripts/test_check_hyperliquid.py shared_scripts/test_check_execute_fees.py`
- `go test ./...` from `scheduler/`
- `uv run python -m py_compile shared_tools/close_registry_loader.py shared_tools/strategy_composition.py shared_scripts/check_strategy.py shared_scripts/check_hyperliquid.py shared_scripts/check_robinhood.py shared_scripts/check_topstep.py shared_scripts/check_okx.py`


richkuo:
Closing because the review-fix commit was pushed to the existing PR #511 instead.

PR #515, opened 2026-04-30
Prevent per-strategy Hyperliquid circuit breakers from closing shared coins

### LINKED ISSUE #512: bug: per-strategy circuit breaker closes shared-wallet positions from other strategies
## Bug

When a per-strategy circuit breaker fires (e.g. 5 consecutive losses), it calls `forceCloseAllPositions` which closes positions on the exchange. But if multiple strategies share the same wallet/coin on Hyperliquid, the exchange-side close wipes **all** positions for that coin — not just the triggering strategy's.

This leaves other strategies with phantom virtual positions (local state shows open, exchange shows closed).

## Example

- `hl-tema-eth-live` and `hl-rmc-eth-live` both trade ETH on the same Hyperliquid wallet
- `hl-tema-eth-live` hits 5 consecutive losses → circuit breaker fires
- `forceCloseAllPositions` market-sells all ETH on the exchange
- `hl-rmc-eth-live` still thinks it has a 0.48 ETH long position
- HL sync detects reconciliation gap but only warns, doesn't heal

## Expected Behavior

Per-strategy circuit breaker (consecutive losses) should:
- **Stop the triggering strategy from opening new trades** (pause trading)
- **NOT close positions on the exchange** if other strategies share the same coin/wallet
- Only the triggering strategy is paused; other strategies continue unaffected

The force-close should only happen when the strategy is the sole owner of the position, or when it's a portfolio-level kill switch (not per-strategy CB).

## Acceptance Criteria
- [ ] Per-strategy circuit breaker does not close exchange positions when other strategies share the coin
- [ ] Per-strategy circuit breaker only pauses the triggering strategy
- [ ] If strategy is sole owner of the position, force-close is still appropriate
- [ ] Portfolio-level kill switch behavior unchanged (it should close everything)

richkuo:
## Summary
- Changed per-strategy Hyperliquid circuit-breaker handling so shared-coin wallets are paused without submitting exchange-side close orders.
- Sole-owner Hyperliquid circuit breakers still enqueue and execute force-closes as before.
- Preserved portfolio kill-switch behavior and updated the main loop so shared-coin detection still works even when the HL position fetch is unavailable.
- Added and updated regression tests for shared-coin pause behavior, sole-owner force-close behavior, and pending-close recovery.

## Testing
- `go build` in `scheduler` passed after the change.

Closes #512

LLM: Codex (GPT-5) | medium

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -660,47 +660,6 @@ func hlLiveStrategiesForCoin(coin string, hlLiveAll []StrategyConfig) []Strategy
 	return out
 }
 
-func hlStrategyCapitalWeight(sc StrategyConfig) float64 {
-	if sc.CapitalPct > 0 {
-		return sc.CapitalPct
-	}
-	if sc.Capital > 0 {
-		return sc.Capital
-	}
-	return 1.0
-}
-
-// hlStrategyCapitalWeights returns per-peer weights for proportional close
-// sizing on a shared coin. When peers mix units (one declares CapitalPct as a
-// fraction, another declares raw Capital in dollars), their sum is nonsensical
-// (e.g. 0.5 + 1000 ≈ 1000.5) and the CapitalPct-only peer's share collapses to
-// ~0, producing a no-op close. Detect the mismatch and fall back to equal
-// weights (1.0 each) so the firing strategy still gets a meaningful share.
-// When all peers use the same field, behavior matches hlStrategyCapitalWeight
-// (#356 review).
-func hlStrategyCapitalWeights(peers []StrategyConfig) []float64 {
-	hasPct := false
-	hasAbs := false
-	for _, p := range peers {
-		switch {
-		case p.CapitalPct > 0:
-			hasPct = true
-		case p.Capital > 0:
-			hasAbs = true
-		}
-	}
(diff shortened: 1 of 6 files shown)


richkuo:
**LGTM** — Hyperliquid shared-coin handling is now consistent end-to-end: `computeHyperliquidCircuitCloseQty` refuses partial sizing when multiple live peers trade the same coin, `setHyperliquidCircuitBreakerPending` bails the same way, `shouldForceCloseAllPositionsOnCircuitBreaker` stops virtual-only CB closes that would disagree with the real wallet, `main.go` fills `riskAssist.HLLiveAll` even when the HL position fetch fails so that shared-coin detection still works (`scheduler/main.go`), and `runPendingHyperliquidCircuitCloses` drops stale HL pending rows and avoids stuck-CB churn for shared coins instead of firing weighted closes (`scheduler/hyperliquid_balance.go`); sole-owner flows and regression coverage (`scheduler/risk_test.go`, `scheduler/hyperliquid_balance_test.go`) match that intent and `go test ./...` in `scheduler/` passes on the PR branch.

**Optional (non-blocking):** Consider a notifier or louder-than-`[INFO]` path when clearing shared-coin pending or when a CB latches without an auto-close path, so ops do not rely only on logs.

PR #520, opened 2026-05-01
fix(perps): add margin_per_trade_usd field; remove hardcoded 0.95 buffer

### LINKED ISSUE #518: fix: sizing_leverage budget is treated as notional but should account for exchange leverage
## Bug

The position sizing formula is:

```
budget = cash × sizing_leverage × 0.95
size = budget / price
```

`budget` is treated as the **notional** value of the position. But when calculating margin drawdown, the code does:

```
margin = notional / leverage
```

So with `sizing_leverage=0.1` and `leverage=20`:
- budget = $560 × 0.1 × 0.95 = **$53.20** (treated as notional)
- size = $53.20 / $2257 = **0.024 ETH** (tiny position)
- margin = $53.20 / 20 = **$2.66**

But the user expected `sizing_leverage=0.1` to mean "use $53 as **margin**", which at 20x leverage should be:
- margin = **$53**
- notional = $53 × 20 = **$1,060**
- size = $1,060 / $2257 = **0.47 ETH**

## The Problem

`sizing_leverage` controls the **notional** budget, not the **margin** budget. But users naturally think in terms of "how much margin do I want to risk per trade." When leverage is high (e.g. 20x), the two interpretations diverge massively.

## Proposed Fix

Change the sizing formula to think in terms of **margin per trade**, not notional:

```
margin_per_trade = cash × sizing_leverage × 0.95
notional = margin_per_trade × leverage
size = notional / price
```

This way `sizing_leverage=0.1` means "risk 10% of cash as margin" which at 20x gives a proper-sized position.

Alternatively, add a `margin_per_trade_usd` field that directly specifies the margin amount (as mentioned in #497).

## Also: Remove the 0.95 Buffer

The `× 0.95` safety buffer is hardcoded and opaque. It silently reduces position size by 5% without the user's control. Remove it — if users want a buffer, they can set a lower `sizing_leverage` or `margin_per_trade_usd` value explicitly.

## Acceptance Criteria
- [ ] Sizing formula accounts for exchange leverage so the budget represents margin, not notional
- [ ] Or: add `margin_per_trade_usd` config field
- [ ] Remove hardcoded `× 0.95` buffer from sizing formula
- [ ] Backward compatible for existing configs that use `sizing_leverage` without `leverage`
- [ ] Update docs/comments to clarify the sizing model

richkuo:
Closes #518

## Summary
- Add opt-in `margin_per_trade_usd` (perps-only); when set, notional = min(margin, cash) × exchange_leverage, fixing the case where `sizing_leverage=0.1` + `leverage=20` produced a tiny position
- Add `PerpsOpenNotional` + `ComputePerpsOpenNotional` helpers so paper/live paths share one formula
- Remove hardcoded × 0.95 safety buffer across spot/perps/futures/options
- Validate `margin_per_trade_usd` (perps-only, positive); wire into hot-reload

## Test plan
- [x] `go test ./...` passes
- [x] New coverage: `TestPerpsOpenNotional`, `TestExecutePerpsSignalMarginPerTradeUSDOverridesSizingLeverage`, `TestLoadConfigMarginPerTradeUSD*`
- [x] Updated existing perps/spot sizing tests for signature + 0.95 removal

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 129 in / 65608 out | Cache: 14151395 read / 244695 written | Cost: $10.25


richkuo:
@claude

PR #527, opened 2026-05-01
feat(close): add tiered_tp_atr_live close strategy

### LINKED ISSUE #523: feat(close): add tiered_tp_atr_live — variable ATR updated per tick
## Motivation

`tiered_tp_atr` uses `entry_atr` (captured once when the position is opened) for the life of the position. In volatile markets, a stale ATR can cause take-profit tiers to fire too early or too late.

This feature request adds a new close strategy that recomputes the ATR on every tick from live OHLCV data, so TP tiers always reflect current market volatility.

### New file
`shared_strategies/close/tiered_tp_atr_live.py`

### Strategy name
`tiered_tp_atr_live`

### Behavior
- On each `evaluate()` call, fetch the current ATR from `market` (key: `"atr"` or `"live_atr"`)
- Compute `atr_profit = profit_distance / current_atr` instead of using `entry_atr`
- If `current_atr <= 0`, fall back to `entry_atr` (for compatibility during warm-up candle gaps)
- Otherwise identical logic to `tiered_tp_atr` — same `tiers` param, same close_fraction behavior

### Param: `atr_source`
- `"live"` (default) — use ATR from market data each tick
- `"entry"` — use `entry_atr` from position (mirrors existing `tiered_tp_atr`)
- Allows users to toggle without changing strategy name

### Backwards compatibility
`entry_atr` path is identical to the existing `tiered_tp_atr`; no behavior changes for existing users.

### Implementation notes
- The ATR value is expected in `market["atr"]` (float, already computed by the caller from the current candle)
- No separate ATR calculation added inside this evaluator — keeps it stateless and fast
- Registry registration in `shared_strategies/close/registry.py`

## References
- Existing `tiered_tp_atr`: `shared_strategies/close/tiered_tp_atr.py`
- ATR stamping pattern: `stampEntryATRIfOpened` in Go runtime

## Priority
Normal

---

richkuo:
Adds `tiered_tp_atr_live`, a new close evaluator that recomputes ATR per tick from market data so take-profit tiers track current volatility instead of the entry-time ATR snapshot used by `tiered_tp_atr`.

- Reads `market["atr"]` (falls back to `market["live_atr"]`) each evaluation
- `atr_source` param (`live` default | `entry`) toggles live vs. snapshot without renaming the strategy
- Falls back to `position.entry_atr` when live ATR is missing/zero, so warm-up candle gaps don't block closes
- Reuses `tiered_tp_atr`'s tier parser for identical tier semantics

Closes #523

---

### DIFF
--- a/shared_scripts/check_hyperliquid.py
+++ b/shared_scripts/check_hyperliquid.py
@@ -55,7 +55,7 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'platforms', 'h
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'shared_strategies', 'open', 'futures'))
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'shared_tools'))
 
-from atr import ensure_atr_indicator
+from atr import ensure_atr_indicator, latest_atr
 
 
 def _make_dataframe(candles):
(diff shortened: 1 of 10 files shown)


richkuo:
@claude

richkuo:
@claude sonnet

PR #533, opened 2026-05-01
Paper-mode trailing stop on HL perps (#532)

### LINKED ISSUE #532: Support trailing stop loss in paper mode (Hyperliquid)
Currently `trailing_stop_pct` and `trailing_stop_atr_mult` only run for live execution paths. Paper mode skips trailing-stop handling.

**Proposal:** Support trailing stop loss simulation in **paper mode across all appropriate platforms** — i.e., every platform/type where trailing-stop configuration is supported and position-side mark-price evaluation is available (currently Hyperliquid perps, and future platforms as support is added).

Implementation should **reuse existing trailing-stop semantics/helpers** (side-aware high/low-water handling, trailing distance from configured effective trailing %, min-move debounce behavior) so paper mode behavior matches live behavior as closely as possible.

Use in-memory position state fields (`StopLossHighWaterPx`, `StopLossTriggerPx`) to track trailing state and evaluate breaches on each cycle; when breached, trigger a synthetic close using the same `recordPerpsStopLossClose` helper that the live trailing-stop immediate-fill path uses (`scheduler/main.go:1303`).

### Multi-strategy / partial-stop requirement
When multiple strategies share the same underlying platform position (or otherwise represent overlapping exposure), trailing-stop simulation must support **partial stop-loss closes** at strategy scope rather than naively flattening the full shared position. Behavior should align with existing partial-close semantics (`close_fraction` / strategy-owned quantity accounting) so one strategy's trailing breach does not accidentally close peer strategies' residual exposure.

No exchange orders needed — purely price-based evaluation in scheduler state.

richkuo:
## Summary
- Paper-mode HL perps now run the same trailing-stop loop as live: the scheduler advances `Position.StopLossHighWaterPx` per cycle, replaces `Position.StopLossTriggerPx` when the favorable move clears the min-move debounce, and books a synthetic close via `recordPerpsStopLossClose` when the mark crosses the trigger.
- New pure helpers in `scheduler/hyperliquid_trailing_stop.go`: `trailingStopBreached(side, mark, currentTrigger)` and `runHyperliquidTrailingStopPaper(sc, side, pos, mark, highWater, currentTrigger)`. Both reuse `effectiveTrailingStopPct` and `computeTrailingStopUpdate` so paper behavior matches live semantics exactly (side-aware high-water, trailing %, min-move debounce, ATR-mult support).
- Each strategy's `StrategyState.Positions` is isolated, so peer strategies on the same coin retain their own virtual exposure when one strategy breaches — satisfies the multi-strategy / partial-stop requirement without any cross-strategy logic.
- Missing-EntryATR alert (`atrMultMissingEntryATR`) ungated from the `hyperliquidIsLive` check since paper mode now runs the same trailing loop and benefits from the same diagnostic.

## Test plan
- [x] `go test ./...` — full suite passes
- [x] New pure-helper tests cover breach (long & short), trigger ratchet, min-move no-op, first-cycle bootstrap from `AvgCost`, no-trailing/non-HL/zero-mark guards, and strategy-state isolation
- [x] `go build` succeeds
- [ ] Operator smoke test on a paper HL perps strategy with `trailing_stop_pct` configured: open a position, watch `StopLossHighWaterPx` ratchet on favorable moves, confirm synthetic close at the trigger price when mark drops past it

Closes #532

---
LLM: Claude Opus 4.7 | high


richkuo:
@claude

PR #546, opened 2026-05-02
Go scheduler: RegimeConfig, AllowedRegimes gating, /status regime field

richkuo:
Closes #541

## Summary

- `config.go`: `RegimeConfig{Enabled, Period, ADXThreshold}`, `Config.Regime *RegimeConfig`, `StrategyConfig.AllowedRegimes []string`; `LoadConfig` defaults (period=14, threshold=20); `ValidateConfig` rejects unknown labels and invalid bounds
- `config_migration.go`: bump `CurrentConfigVersion` 10 → 11 (no data migration; empty defaults preserve legacy behavior exactly)
- `strategy_composition.go`: `Regime string` added to `StrategyDecisionFields` — one change propagates to all 5 result structs
- `regime.go`: `regimeAllowsEntry(allowed, current)` pure helper — allows when list is empty, current is empty, or current matches
- `main.go`: regime gate at 6 dispatch sites; `stratState.Regime` updated each cycle
- `state.go`: `StrategyState.Regime string` persisted with state
- `server.go`: `StratStatus.Regime` exposed in `/status` JSON
- `config_reload.go`: `AllowedRegimes` is soft-reloadable (mirrors `CloseStrategies`); `cfg.Regime` change requires restart (mirrors `Correlation`)

## Behavior

When `allowed_regimes` is empty or unset, behavior is bit-for-bit identical to before — all entries permitted. When set, entries are skipped and logged when `result.Regime` is not in the list. Existing positions are always managed by close paths regardless.

`regime: null` from `check_options.py` becomes `""` in the result struct, which `regimeAllowsEntry` treats as "always allow" — so options strategies are unaffected.

## Test plan

17 new tests in `regime_test.go`:
- `ValidateConfig` accepts valid/nil regimes, rejects unknown labels and invalid period/threshold
- `regimeAllowsEntry` empty list, matching, non-matching, empty-current cases
- `StrategyDecisionFields.Regime` JSON round-trip and omitempty
- `CurrentConfigVersion == 11`
- hot-reload: `AllowedRegimes` accepted, `Regime` config rejected

Full suite: `go -C scheduler test ./...` passes.

---
LLM: Claude Sonnet 4.6 (1M) | high

PR #556, opened 2026-05-02
ValidateConfig: reject allowed_regimes on options + warn when regime disabled

### LINKED ISSUE #553: Regime gate: options strategy gap + ValidateConfig silent no-op for allowed_regimes
Two related gaps surfaced in review of PR #546.

## 1. Options gate is absent, not just empty-string

`OptionsResult` (`scheduler/options.go:49`) does not embed `StrategyDecisionFields`, so the regime gate (`regimeAllowsEntry`) is never called at the `case "options":` dispatch site (`scheduler/main.go:~1268`).

The PR body describes this as `"regime: null becomes '' which regimeAllowsEntry treats as always allow"` — but the actual mechanism is that the gate call is **absent entirely**, not that it's called with an empty string. Runtime outcome is identical (options entries always pass through), but the description is misleading, and there's a concrete validator gap:

`ValidateConfig` does not warn when an operator sets `allowed_regimes` on an options strategy. The config passes validation, the gate silently no-ops, and the operator has no signal that the configuration is ineffective.

**Fix:**
- Either add a `ValidateConfig` error/warning when `allowed_regimes` is non-empty on a strategy with `type == "options"` (since the gate is unimplemented for options), or
- Wire options into the gate properly once `check_options.py` emits a regime field (#544).

## 2. AllowedRegimes is a no-op at runtime until check scripts are merged

No check script currently calls `latest_regime()` or emits `"regime"` in its JSON output on `main` — that wiring is in PR #545 (not yet merged). The Go scaffolding in PR #546 is correct and backward-compatible (`empty result.Regime → always allow`), but an operator who configures `allowed_regimes` on main today will see no gating effect and no warning.

**Fix:**
- Add a startup or validation log message when `cfg.Regime.Enabled == false` and any strategy has `allowed_regimes` set, warning that regime gating requires the check scripts to emit the `regime` field.
- Or simply document the dependency and ensure #545 merges before #546 is used in production.

Relates to PR #546, #545, #544, #550.

richkuo:
## Summary

Addresses two silent no-op gaps in regime gating flagged in #553 (review of PR #546):

1. **Hard error for `allowed_regimes` on options strategies** — the regime gate is absent at the `case "options":` dispatch site in `main.go`. `ValidateConfig` now rejects non-empty `allowed_regimes` on `type=options` strategies so operators get a startup error rather than a silent config that does nothing. This will be removed once the gate is wired for the multi-position options model.

2. **`[WARN]` when `regime.enabled=false` but `allowed_regimes` is set** — when regime detection is disabled, the gate cannot read a regime label from the check script output and becomes a no-op. A per-strategy `[WARN]` is now printed at startup for each affected strategy so operators know to either enable regime detection or remove the field.

Five new tests in `regime_test.go`:
1. `TestValidateConfig_AllowedRegimes_RejectsOnOptions`
2. `TestValidateConfig_AllowedRegimes_AcceptsEmptyOnOptions`
3. `TestValidateConfig_AllowedRegimes_AcceptsOnSpotWithRegimeEnabled`
4. `TestValidateConfig_AllowedRegimes_WarnsWhenRegimeDisabled`
5. `TestValidateConfig_AllowedRegimes_NoWarnWhenRegimeEnabled`

Plus `minimalOptionsConfig()` helper mirroring `minimalSpotConfig()`.

Closes #553

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -866,6 +866,12 @@ func ValidateConfig(cfg *Config) error {
 				errs = append(errs, fmt.Sprintf("%s: allowed_regimes[%d] unknown label %q (valid: trending_up, trending_down, ranging)", prefix, j, label))
 			}
 		}
+		// The regime gate is not wired at the options dispatch site (#553), so
+		// allowed_regimes is a silent no-op for options strategies. Reject it
+		// here until the gate is properly implemented for the multi-position model.
+		if sc.Type == "options" && len(sc.AllowedRegimes) > 0 {
+			errs = append(errs, fmt.Sprintf("%s: allowed_regimes is not enforced for type=options (gate not wired at options dispatch; see issue #553)", prefix))
+		}
 
 		// Live-mode futures require TopStep API credentials.
 		if sc.Type == "futures" {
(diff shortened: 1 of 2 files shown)


richkuo:
@claude sonnet

PR #567, opened 2026-05-04
fix: reconcile shared-coin HL positions on stop-loss or external close (#565)

### LINKED ISSUE #565: bug: shared-coin HL positions not reconciled when closed by stop loss
## Problem

When a shared-coin position (e.g. ETH traded by both `hl-tema-eth-live` and `hl-rmc-eth-live`) is closed on-chain by a stop-loss trigger, the go-trader local state still shows the position as open. The bot never detects the close.

## Root Cause

In `hyperliquid_balance.go`, `reconcileHyperliquidAccountPositions` skips shared coins for quantity changes and position removal:

```go
if sharedCoins[sym] {
    continue // handled below
}
```

The shared-coin block (line 398+) only does non-destructive updates (multiplier migration, leverage sync) and records reconciliation gaps for display — it never removes closed positions from state.

## Impact

- Stale positions show in status/logs indefinitely
- Portfolio value calculation is wrong
- Bot keeps passing stale position data to check scripts
- Manual restart does not fix it

## Reproduction

1. Have 2+ strategies trading the same coin (e.g. ETH)
2. Position gets stop-lossed on-chain
3. Bot never reconciles the close, position stays in local state

richkuo:
## Problem

When a shared-coin HL position (multiple strategies on the same wallet coin, e.g. `hl-tema-eth-live` + `hl-rmc-eth-live`) was closed on-chain by a stop-loss trigger or manually, the scheduler's local state never reconciled the close. Virtual positions stayed open indefinitely, causing wrong portfolio values, stale data passed to check scripts, and no recovery path short of hand-editing `state.db`.

Root cause: `reconcileHyperliquidAccountPositions` deliberately skipped destructive updates for shared coins (#258/#515) to avoid phantom CB triggers from intermediate on-chain states. That protection is correct for ambiguous gaps — but unambiguous closes were left unhandled.

## Solution

Two detectors added inside the shared-coin reconciliation loop, after the existing signed-qty computation:

**Detector 1 — full flat**: `|onChainQty| < 1e-6` and `|virtualQty| > 0` → the aggregate position is gone. Closes all peers: SL owner via `recordPerpsStopLossClose` at `StopLossTriggerPx` (reason `hl_sync_stop_loss`), non-owners via `recordClosedPosition` (reason `hl_sync_external`, close price 0).

**Detector 2 — SL owner partial close**: Non-zero on-chain qty with a gap, exactly one peer holds a `StopLossOID`, and the on-chain residual matches the signed sum of all non-owner peers' virtual qty (within `1e-6`). HL trigger orders are sized to the owner's qty at arm time, so when the trigger fires the peers' portion stays on-chain unchanged. Closes the owner only via `recordPerpsStopLossClose`.

All other mismatches (ambiguous gaps) continue to fall through to gap-only recording — the #258/#515 protection is fully preserved.

## Tests

1. `TestReconcileSharedCoin_OwnerStopLossFired_ClosesOwnerOnly` — Detector 2, long
2. `TestReconcileSharedCoin_OwnerStopLossFired_Short` — Detector 2, short
3. `TestReconcileSharedCoin_AllPositionsClosedExternally` — Detector 1, owner + peer
4. `TestReconcileSharedCoin_GapWithoutSLOwner_LeavesPositionsAlone` — regression guard for #258
5. `TestReconcileSharedCoin_ResidualMismatch_LeavesPositionsAlone` — residual mismatch, no close
6. `TestAccountSyncSharedCoinClosedWhenOnChainGone` — updated from the old `NotRemoved` test to assert new correct behavior (tema's stale virtual is closed via `hl_sync_external` when on-chain is flat)

Closes #565

---

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -467,6 +467,99 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 		}
 		delta := virtualQty - onChainQty
 
+		// Detect and reconcile unambiguous shared-coin closes (#565).
+		//
+		// Detector 1 — full external close: on-chain is flat but virtual is not.
+		// Covers stop-loss sweep of the aggregate position, manual close on HL UI,
+		// and kill-switch closes that finish between scheduler cycles.
+		//
+		// Detector 2 — SL owner partial close: exactly one peer holds a resting
+		// trigger (StopLossOID) and the on-chain residual matches the signed sum of
+		// all non-owner peers' virtual qty. HL trigger orders are sized to the
+		// owner's qty at arm time, so when the trigger fires the non-owner peers'
+		// portion remains on-chain untouched.
+		//
+		// All other qty mismatches (ambiguous gaps that #258/#515 protect) fall
+		// through to the gap-recording block unchanged.
+		if math.Abs(onChainQty) < 1e-6 && math.Abs(virtualQty) > 1e-6 {
+			// Detector 1: everything gone on-chain — close all peers.
+			for _, id := range stratIDs {
+				ss := state.Strategies[id]
+				if ss == nil {
+					continue
+				}
+				pos := ss.Positions[coin]
+				if pos == nil {
+					continue
+				}
+				logger, logErr := logMgr.GetStrategyLogger(id)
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #570, opened 2026-05-04
fix: backfill EntryATR for existing positions missing it

### LINKED ISSUE #568: fix: backfill EntryATR for existing positions that were opened without it
## Problem

\`stampEntryATRIfOpened\` only stamps \`EntryATR\` when \`trades > 0\` (a new trade opened this cycle). Positions opened before the ATR-stamping feature was added, or opened by manual intervention, have \`EntryATR == 0\` forever.

This causes:
- \`tiered_tp_atr\` to noop (no take-profit)
- \`trailing_stop_atr_mult\` to noop (no trailing stop)
- \`stop_loss_atr_mult\` to noop (no stop-loss)
- The WARN message: \`tiered_tp_atr configured but Position.EntryATR is 0\`

## Fix

Split \`stampEntryATRIfOpened\` into two paths rather than removing the \`trades <= 0\` guard entirely:

1. **New trade opened this cycle (\`trades > 0\`):** stamp as today (existing behavior, unchanged).
2. **Existing position with \`EntryATR == 0\` (\`trades <= 0\`):** also attempt to stamp using the current cycle's ATR as a backfill.

Both paths share the same ATR extraction (\`indicatorFloat\`) and plausibility guard (\`atr > AvgCost*0.5\`). The existing \`pos.EntryATR != 0\` check already prevents overwriting a valid value, so backfilling is safe.

```go
// Current (broken for existing positions):
func stampEntryATRIfOpened(s *StrategyState, symbol string, indicators map[string]interface{}, trades int) {
    if trades <= 0 || s == nil {  // <-- blocks backfill for old positions
        return
    }
    ...
    if !exists || pos == nil || pos.EntryATR != 0 {
        return
    }
```

The fix allows the function to run when \`pos.EntryATR == 0\` regardless of \`trades\`. The backfilled ATR won't match the exact entry-bar ATR, but a close approximation unblocks ATR-dependent features for affected positions — far better than permanent noop.

richkuo:
## Summary

- Removes the `trades <= 0` guard from `stampEntryATRIfOpened` so it runs every scheduler cycle, not only when a new trade opens.
- Positions that pre-date ATR-stamping (#525) or were opened via manual exchange intervention now get `EntryATR` backfilled on the next cycle, unblocking `tiered_tp_atr`, `trailing_stop_atr_mult`, and `stop_loss_atr_mult`.
- Drops the now-unused `trades` parameter from the function signature and all 5 call sites.

## Test plan

- Updated `TestStampEntryATRIfOpened`: flipped the `trades == 0` no-op case to assert backfill behavior; added regression case for "atr already set and no new trade" (must not overwrite).
- `go test ./...` passes.

Closes #568

---
LLM: Claude Sonnet 4.6 (1M) | medium

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -1936,7 +1936,7 @@ func executeSpotResult(sc StrategyConfig, s *StrategyState, db *StateDB, result 
 		logger.Error("Trade execution failed: %v", err)
 		return 0, ""
 	}
-	stampEntryATRIfOpened(s, result.Symbol, result.Indicators, trades)
+	stampEntryATRIfOpened(s, result.Symbol, result.Indicators)
 	if pos, ok := s.Positions[result.Symbol]; ok {
 		stampOpenTradeFromPosition(s, db, result.Symbol, pos)
 	}
@@ -1948,8 +1948,8 @@ func executeSpotResult(sc StrategyConfig, s *StrategyState, db *StateDB, result 
 	return trades, detail
 }
 
-func stampEntryATRIfOpened(s *StrategyState, symbol string, indicators map[string]interface{}, trades int) {
-	if trades <= 0 || s == nil {
+func stampEntryATRIfOpened(s *StrategyState, symbol string, indicators map[string]interface{}) {
+	if s == nil {
 		return
 	}
 	atr, ok := indicatorFloat(indicators, "atr")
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

richkuo:
@claude sonnet implement the optional review suggestion

PR #577, opened 2026-05-04
fix: include type=manual strategies in HL on-chain reconciler

### LINKED ISSUE #576: bug: reconciler does not remove positions closed on exchange
## Problem

When a position opened via `type=manual` is closed on the Hyperliquid exchange (SL trigger fires, TP fills via HL trigger order, or manual close in the HL UI) while the scheduler is running, the ghost position is never removed from bot state.

The root cause is that `hlLiveDue` / `hlLiveAll` are built with `sc.Type == "perps"` (`main.go:601/607`), so `type=manual` strategies are excluded from `reconcileHyperliquidAccountPositions` entirely. This gap was introduced when `type=manual` was added (#569) without updating the live-HL strategy filter.

Note: regular `type=perps` strategies are **not** affected — `reconcileHyperliquidPositions` already handles the `onChainPos == nil && statePos != nil` case correctly for those.

## Reproduction

1. Open a position via `go-trader manual-open`
2. Close the position directly on the exchange (let SL/TP trigger fire, or close in HL UI)
3. The bot continues reporting the position as open indefinitely

## Expected Behavior

The reconciler should detect that the on-chain position has gone to size 0 and:
1. Record a close trade with the realized PnL (via `recordPerpsStopLossClose` if an OID is set, otherwise `recordClosedPosition` with `hl_sync_external`)
2. Remove the position from state
3. Update strategy cash

## Fix

Include `type=manual` HL strategies in the `hlLiveDue` / `hlLiveAll` builds (or add a dedicated reconciliation pass for manual strategies alongside the existing `type=perps` one).

## Workaround

Manual SQLite intervention:
```sql
DELETE FROM positions WHERE strategy_id = 'manual-eth';
UPDATE strategies SET cash = <correct_value> WHERE id = 'manual-eth';
```
Then restart the bot.

richkuo:
## Problem

`type=manual` HL strategies were excluded from `reconcileHyperliquidAccountPositions` because `hlLiveDue`/`hlLiveAll` filtered on `sc.Type == "perps"` only. When a manual position was closed on-chain (SL trigger fire, TP fill, or HL UI close), the bot retained the ghost position indefinitely.

## Fix

Add `isHLLiveReconcilable` predicate (`perps|manual` + `platform=hyperliquid` + live args) and build dedicated `hlReconcileAll`/`hlReconcileDue` lists for the reconciler call site. The existing perps-only `hlLiveAll`/`hlLiveDue` are intentionally unchanged — they feed kill-switch close, trailing-stop arming, and risk math which are perps-specific.

The reconciler itself (`reconcileHyperliquidPositions`) is type-agnostic and already handles the `onChainPos == nil && statePos != nil` case correctly.

## Tests

1. `TestReconcileManualPositionExternalClose` — manual position flat on-chain → removed with `close_reason=hl_sync_external`
2. `TestReconcileManualPositionSLFired` — manual position with resting OID flat on-chain → removed with `close_reason=stop_loss` at trigger price
3. `TestIsHLLiveReconcilable` — unit-tests the new predicate across 6 cases (perps/manual × live/paper, spot, non-HL)

Closes #576

---

### DIFF
--- a/scheduler/hyperliquid_balance_test.go
+++ b/scheduler/hyperliquid_balance_test.go
@@ -2579,3 +2579,82 @@ func TestRunPendingHyperliquidCircuitCloses_TenthFailureNotifies(t *testing.T) {
 		t.Errorf("expected 1 DM on failure #10 (every-10th cadence), got %d", len(dmMsgs))
 	}
 }
+
+// TestReconcileManualPositionExternalClose verifies that a type=manual HL strategy
+// whose on-chain position is flat gets its virtual position removed with
+// close_reason="hl_sync_external". Regression for #576.
+func TestReconcileManualPositionExternalClose(t *testing.T) {
+	state := &AppState{
+		Strategies: map[string]*StrategyState{
+			"manual-eth": {
+				ID: "manual-eth", Cash: 100,
+				Positions: map[string]*Position{
+					"ETH": {Symbol: "ETH", Quantity: 0.5, AvgCost: 2000, Side: "long",
+						Multiplier: 1, Leverage: 5, OwnerStrategyID: "manual-eth"},
+				},
+			},
+		},
+	}
+	sc := StrategyConfig{
+		ID: "manual-eth", Platform: "hyperliquid", Type: "manual",
+		Symbol: "ETH", Timeframe: "1h", Leverage: 5,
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

PR #586, opened 2026-05-05
fix: credit peer cash on detected HL external close (#584)

### LINKED ISSUE #584: hl-live summary diverges from real HL account balance after external closes
## Problem

The hl-live summary (go-trader-live, channel `1488455770772213771`) falls out of sync with the real Hyperliquid exchange account balance because `s.Cash` is not corrected when positions are closed externally (stop-loss sweep, manual close on HL UI, or kill-switch close between scheduler cycles).

### Root cause

`reconcileHyperliquidAccountPositions` (hyperliquid_balance.go) has two "unambiguous close" detectors added in #565/#567:

1. **Full flat** (on-chain qty ≈ 0, virtual qty ≠ 0): all peers are closed
2. **SL owner partial close**: one strategy's stop-loss fired and closed only their portion

For Detector 1 non-SL-owner peers, the code calls:
```go
recordClosedPosition(ss, pos, 0, 0, "hl_sync_external", now)
delete(ss.Positions, coin)
```

`recordClosedPosition` records **0 PnL** (third argument = 0) because the close price is unknown. Crucially, **`s.Cash` is never updated** — the realized PnL is lost to the virtual state.

The SL owner path correctly calls `recordPerpsStopLossClose` which does `s.Cash += pnl`, so only the SL owner's cash is accurate.

### Effect

- `PortfolioValue(s, prices)` for affected non-owner strategies = stale cash + 0 position value
- `channelValue["hyperliquid"]` in the summary = sum of stale per-strategy PVs
- **Real HL `accountValue`** (fetched via clearinghouseState) = correct
- The summary's TOTAL diverges from the real HL balance after any external close of a shared-wallet position

## Proposed Solution

**Option A (correct cash):** When an external close is detected and the close price is unknown, record the close with `closePrice=0` but **also fetch the on-chain close price** (via fill history or a separate query) and credit `s.Cash` with the realized PnL. This requires an additional HL API call in the reconcile path.

**Option B (use real balance):** For hl-live shared-wallet strategies, bypass per-strategy `PortfolioValue` in the summary total and substitute the real `accountValue` fetched from clearinghouseState — matching how `computeTotalPortfolioValue` already handles shared wallets elsewhere. The per-strategy breakdown still shows virtual PnL; only the TOTAL row uses the real balance.

**Option C (suppress stale PnL display):** If the close price is unknown, mark the strategy's PnL as "?" or exclude it from the TOTAL rather than showing a wrong number. Least preferred as it hides information rather than fixing the root cause.

### Recommended

**Option B** — lowest risk, already partially implemented via `computeTotalPortfolioValue`. The TOTAL row in `FormatCategorySummary` should use the real shared-wallet balance for hl-live, not the sum of potentially-stale per-strategy PVs.

## Implementation Notes

- Changes are in `scheduler/hyperliquid_balance.go` and `scheduler/discord.go` (for FormatCategorySummary TOTAL row)
- The `fetchHyperliquidBalance` / `fetchHyperliquidState` functions already exist and are used by `computeTotalPortfolioValue`
- The fix needs to thread the real shared-wallet balance into the TOTAL row computation in `FormatCategorySummary`

Reference only: LLM model used: minimax-m2.7

richkuo:
## Summary

- New helper \`recordPerpsExternalClose\` (\`portfolio.go\`) mirrors \`recordPerpsStopLossClose\` but takes an arbitrary close price; computes PnL, deducts the platform fee, credits \`s.Cash\`, and records the trade + ClosedPosition.
- \`reconcileHyperliquidAccountPositions\` now takes a \`prices map[string]float64\`. In the Detector 1 non-SL-owner branch it uses \`prices[coin]\` as the close price via the new helper; falls back to the legacy zero-PnL \`recordClosedPosition\` only when no mark is available.
- \`main.go\` passes the already-fetched HL marks at the call site. \`syncHyperliquidAccountPositions\` (no prefetched prices) passes \`nil\` to preserve legacy behavior.
- 2 new regression tests in \`hyperliquid_balance_test.go\`:
  1. credit-cash path: peer with mark price has \`s.Cash\` credited with mark-based PnL
  2. fallback path: nil prices → zero-PnL preserved

## Why

Per #584, the hl-live summary diverged from the real HL account balance after any external close (manual UI close, kill-switch close, stop sweep) of a shared-wallet position because the non-SL-owner peer's \`s.Cash\` was never updated. Using the current mark as the close price is the most robust fix without an extra API call — it corrects \`s.Cash\` at the source so per-strategy \`PortfolioValue\`, the summary TOTAL, and \`CheckRisk\`'s drawdown math all become accurate.

## Test plan

- [x] \`go -C scheduler test ./...\` green
- [x] \`TestReconcileSharedCoin_AllPositionsClosedExternally_CreditsPeerCash\` (new)
- [x] \`TestReconcileSharedCoin_AllPositionsClosedExternally_NoMarkPrice_FallsBack\` (new)
- [x] All existing reconcile tests pass with \`nil\` prices (backwards-compatible)

Closes #584

---

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -510,7 +510,12 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 				} else if mark, ok := prices[coin]; ok && mark > 0 && recordPerpsExternalClose(ss, coin, mark, "hl_sync_external", logger) {
 					// #584: credit s.Cash with mark-based PnL so the per-strategy
 					// PortfolioValue (and the summary TOTAL) match the real HL
-					// account after an external close.
+					// account after an external close. The mark is fetched at
+					// cycle start, so cp.RealizedPnL is an *approximation* — it
+					// will drift from the true on-chain fill price (which can be
+					// minutes earlier). Do not treat the resulting Trade /
+					// ClosedPosition rows as authoritative for tax or reporting;
+					// they exist to keep cash bookkeeping in sync.
 					changed = true
 				} else {
 					// No mark price available — fall back to recording the
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -329,8 +329,10 @@ func syncHyperliquidAccountPositions(hlStrategies []StrategyConfig, state *AppSt
 		return false
 	}
 
-	// Self-contained entry: due and all are the same list.
-	return reconcileHyperliquidAccountPositions(hlStrategies, hlStrategies, state, mu, logMgr, positions)
+	// Self-contained entry: due and all are the same list. Prices are
+	// unavailable in this path (caller did not pre-fetch); external-close
+	// PnL falls back to zero (legacy behavior pre-#584).
+	return reconcileHyperliquidAccountPositions(hlStrategies, hlStrategies, state, mu, logMgr, positions, nil)
 }
 
 // reconcileHyperliquidAccountPositions reconciles pre-fetched on-chain positions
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

richkuo:
@claude fix optional suggestions

richkuo:
@claude sonnet

richkuo:
@claude sonnet fix the optional review fix

richkuo:
@claude sonnet

PR #587, opened 2026-05-05
fix: query userFills for real exchange fee on HL fills

### LINKED ISSUE #585: live HL trades record exchange_fee=0, modeled fee causes portfolio drift
## Problem

All live Hyperliquid close trades record `exchange_fee = 0` in the trades table. The bot calculates its own modeled fee via `CalculatePlatformSpotFee` and bakes it into `realized_pnl`, but this modeled fee doesn't match Hyperliquid's actual fill fee.

Over time, the per-trade fee discrepancy compounds into portfolio drift between virtual cash and the real on-chain `accountValue`.

## Evidence

`hl-tema-eth-live` on go-trader-live: 28 out of 29 close trades have `exchange_fee = 0.0000`. The only one with a non-zero fee is a manually inserted record. Meanwhile, the HL fills API returns `fee` on every fill.

After reconciling a missed close trade, virtual portfolio total ($1,263.43) still drifted $3.89 from the on-chain balance ($1,259.54) — entirely from accumulated fee miscalculation across 50+ trades.

## Root Cause

`check_hyperliquid.py` extracts `fee` from the order placement response (`filled.get("fee")`), but the HL order placement response does **not** include `fee` in the `filled` status object. The real fee is only available via the separate `userFills` API endpoint (`POST /info {"type":"userFills","user":"..."}`).

The Go plumbing is already correct — `HyperliquidFill.Fee`, `executionFee(modeledFee, fillFee, useFillFee)`, and the `fillFee` parameter in `ExecutePerpsSignalWithLeverage` all handle the fee properly when it's provided. The problem is upstream: the fee is never populated because the wrong API response is being read.

### Primary: query userFills post-fill

1. **`check_hyperliquid.py`** — after placing an order and getting the fill, query the `userFills` API (filter by OID or timestamp) to fetch the real `fee` and `closedPnl` from HL, and include them in the script output
2. **`platforms/hyperliquid/adapter.py`** — expose a fills lookup method so the close script can also use it
3. The Go side needs no changes — it already reads `fill.Fee` from the execution result and passes it through to `executionFee()` with `useFillFee=true` for close legs

### Other affected paths (must fix or acknowledge)

4. **`close_hyperliquid_position.py:166-168`** — same bug. Kill-switch closes go through `adapter.market_close()` and read `filled.get("fee")` from the same response shape — also returns None. Without fixing this, kill-switch fills will continue to record `exchange_fee=0`.
5. **Manual close path (`main.go:1547-1559`)** — reads `fill.Fee` directly from `RunHyperliquidExecute` for `type=manual` strategies. Same upstream source, same 0 result. Will keep drifting unless `check_hyperliquid.py` is fixed (which item 1 covers) — but must be included in testing.
6. **SL-trigger and trailing-stop closes** — never see a fill response at all. Booked by the on-chain reconciler (`syncHyperliquidAccountPositions` / `recordPerpsStopLossClose`) after the trigger fires. Those paths use `CalculatePlatformSpotFee` with no `useFillFee` opt-in. To get real fees there, the reconciler would need to query `userFills` retroactively when it detects a closed position. **Not in scope of this issue** but same drift source — worth a follow-up.

### Implementation notes

7. **Race condition:** `userFills` is indexer-backed and may not show the fill immediately after `market_open` returns. Fix needs retry-with-backoff or query-by-OID-with-poll, otherwise the fix will sometimes still record `fee=0`.
8. **Match by OID, sum across partial fills:** The `userFills` lookup should match by OID rather than timestamp where possible — partial fills produce multiple rows, and a single market order can fragment across price levels. The implementation needs to sum `fee` across all fills sharing the OID rather than reading just the first row.
9. **Historical drift won't self-heal.** All existing rows in `trades` with `exchange_fee=0` stay that way. A one-shot backfill script (query `userFills` history, match by OID, update `trades.exchange_fee` and `realized_pnl`) is needed as a follow-up to recover the accumulated drift.

richkuo:
## Summary
- Adds `HyperliquidExchangeAdapter.lookup_fill_fee_by_oid` that queries HL's `userFills` indexer endpoint, sums `fee` and `closedPnl` across partial fills sharing the same OID, and retries with backoff to absorb indexer lag (default 4 attempts × 0.5s).
- Wires the lookup into both `check_hyperliquid.py` (signal-driven opens/closes/flips, manual closes) and `close_hyperliquid_position.py` (kill-switch). Both capture `since_ms` immediately before submitting the order to bound the lookup window.
- Lookup failures are non-fatal — the Go side already falls back to the modeled `CalculatePlatformSpotFee` when `fillFee=0` (`useFillFee && fillFee > 0` guard in `executionFee`/`exchangeFeeForTrade`).

## Why
HL's `placeOrder` response omits `fee` from `statuses[0]["filled"]`; the only authoritative source is the `userFills` API. Live HL trades were recording `exchange_fee=0` for ~28/29 close legs (#585 evidence) and using the modeled fee in PnL, causing the virtual cash balance to drift ~$3.89 from on-chain accountValue across 50+ trades on `hl-tema-eth-live`.

## Out of scope (follow-ups)
- **SL-trigger / trailing-stop closes** — booked by the on-chain reconciler after the trigger fires; would need a retroactive `userFills` query keyed off the closed-position timestamp rather than an OID owned by the scheduler. Same drift source but architecturally different.
- **Historical backfill** of existing `trades.exchange_fee=0` rows; new code only fixes go-forward writes.

## Test plan
- [x] `platforms/hyperliquid/test_adapter.py` — 7 new tests covering missing-address short-circuit, fee+pnl aggregation across partial fills, string-OID coercion, retry-until-indexed, retry-budget exhaustion, exception swallowing, non-list-response handling.
- [x] Full Python suite (`platforms/`, `shared_tools/`, `shared_strategies/`) — 729 passed.
- [x] `go test ./...` from `scheduler/` — passes.
- [x] `go build .` from `scheduler/` — clean.
- [ ] Manual smoke on go-trader-live: submit a small HL trade, confirm `trades.exchange_fee` matches the on-chain `userFills` fee for that OID and that wallet drift stops accumulating.

Closes #585

---

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -32,6 +32,24 @@ except ImportError:
     _SDK_AVAILABLE = False
 
 
+def _safe_float(v) -> float:
+    if v is None:
+        return 0.0
+    try:
+        return float(v)
+    except (TypeError, ValueError):
+        return 0.0
+
+
+def _safe_int(v) -> int:
+    if v is None:
+        return 0
+    try:
+        return int(v)
+    except (TypeError, ValueError):
+        return 0
+
+
 def _round_perps_px(px: float, sz_decimals: int) -> float:
     """Round a perps price to HL's tick rules.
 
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

PR #599, opened 2026-05-05
fix: reject two manual HL strategies on the same coin (#595)

### LINKED ISSUE #595: Guard manual full closes before wallet-wide Hyperliquid market_close
## Summary
Manual full-close paths now pass `closeFullPosition=true` into `RunHyperliquidExecute`, which makes `check_hyperliquid.py` call `adapter.market_close(sz=None)`. That flattens the entire Hyperliquid wallet position for the coin, not only the strategy's virtual quantity.

## Risk
The perps strategy path uses `shouldCloseFullPosition` to avoid this when peers share the coin, but manual paths do not. Config validation blocks manual-vs-live-perps sharing, but it does not block two manual strategies on the same coin. A full close for one manual strategy can flatten the peer's on-chain exposure while queuing/bookkeeping only the first strategy's `closeQty`, leaving scheduler state and cash wrong.

Impacted paths:
- `scheduler/manual.go:336-339`
- `scheduler/main.go:1534-1537`

## Suggested fix
Either keep manual full closes sized, or add a sole-peer guard that includes `type=manual` strategies as well as live perps strategies. If `market_close(sz=None)` is used, book the actual returned fill quantity rather than the precomputed `closeQty`.

## Review context
Found during recent-commits review; severity P1 because this can create real on-chain/virtual divergence for shared manual HL coins.

richkuo:
## Summary
- Manual full-close paths (`manual-close` CLI, scheduler close-eval loop) call `RunHyperliquidExecute(..., closeFullPosition=true)`, which makes `check_hyperliquid.py` invoke `adapter.market_close(sz=None)` and flattens the entire wallet position for the coin.
- The perps path's `shouldCloseFullPosition` sole-peer guard never runs for manual closes, and `hlLiveAll` is perps-only — so even routing through that helper would not see manual peers.
- Config validation already blocks `manual ↔ perps` on the same coin; this extends the same symbol-conflict check to `manual ↔ manual`. The structural fix (manual-aware peer guard at the call sites) is left for a follow-up; this closes the actual exploit window.

## Test plan
- [x] `go test ./...` (full scheduler suite) passes
- [x] New `TestValidateConfigManualSymbolConflict` covers manual+manual (rejected), manual+perps (rejected — regression coverage for existing rule), and manual+manual on different coins (allowed)

Closes #595

---

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -974,12 +974,29 @@ func ValidateConfig(cfg *Config) error {
 			// Fix #6: manual strategies cannot share a coin with a live perps strategy —
 			// the scheduler's close-eval loop snapshots pos under RLock without owning
 			// the full mutex, creating a TOCTOU window if a perps peer mutates the same position.
+			//
+			// #595: also reject two manual strategies on the same coin. Manual full-close
+			// paths pass closeFullPosition=true unconditionally, which calls
+			// adapter.market_close(sz=None) and flattens the entire wallet position. The
+			// perps path's shouldCloseFullPosition sole-peer guard does not run for manual
+			// closes, so without this check a full close on one manual strategy would
+			// silently flatten a peer manual strategy's on-chain exposure.
 			if sc.Symbol != "" {
 				for _, other := range cfg.Strategies {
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #600, opened 2026-05-05
fix: narrow HL coin-size fill fallback to one logical close (#596)

### LINKED ISSUE #596: Avoid overcounting Hyperliquid coin-size fill-fee fallback
## Summary
`lookupHyperliquidFillByCoinSize` scans the full reconciler lookup window and sums every fill matching the same coin and approximate size.

## Risk
The reconciler window is currently 24h. If the account had another same-sized fill on the same coin inside that window, the fallback can include unrelated fees and closed PnL, overcharging the reconciled close. The function comment says newest-first / first match, but the implementation does not sort or stop after a logical close group.

Impacted path:
- `scheduler/hyperliquid_fills.go:138-148`

## Suggested fix
Narrow or group the fallback match. Possible approaches:
- sort newest-first and select one logical close group instead of summing all matches
- include fill direction/reduce-only semantics where available
- tighten the time window around the detected reconcile event
- keep OID-keyed lookup as the preferred path and make coin-size fallback conservative when multiple candidates exist

## Review context
Found during recent-commits review; severity P2 because it can silently distort reconciler-booked fees/cash while still clearing positions.

richkuo:
## Summary
- `lookupHyperliquidFillByCoinSize` previously summed every coin+size match across the 24h reconciler window, so two unrelated same-size closes on the same coin would conflate fees and closedPnl onto whichever close was being reconciled — silently distorting `Trade.ExchangeFee` and the cash credit booked in `recordPerpsExternalClose`.
- Sort fetched fills newest-first by `Time`, anchor on the first matching record, and aggregate fee/closedPnl across only that record's OID group (handles partial fills sharing one OID without admitting unrelated same-size closes). When the anchor lacks a usable OID, return only that single record.
- Fixed the misleading "Newest-first scanning so the first match is the most recent fill" comment that did not match the previous implementation (no sort, no break-on-first).

## Test plan
- [x] `go test ./scheduler/...` (full suite, 21s, all green)
- [x] New regression: `TestLookupHyperliquidFillByCoinSize_PicksNewestGroupNotSumOfWindow` — two unrelated 0.1 BTC closes (OIDs 100 + 200) → result anchors on OID 200 only, fee=0.50 not 0.90.
- [x] New regression: `TestLookupHyperliquidFillByCoinSize_AggregatesPartialFillsByAnchorOID` — newest match's OID has a partial sibling; older same-size fill on a different OID is ignored.
- [x] Existing tests preserved: `TestLookupHyperliquidFillByCoinSize_MatchesByCoinAndSize`, `TestLookupHyperliquidReconcileFillFee_OIDFirstFallsBackToCoinSize`, `TestReconcileHyperliquidPositions_ExternalCloseUsesFillFee`, `TestReconcileHyperliquidAccountPositions_DetectorOneUsesFillFee`.

Closes #596

---

### DIFF
--- a/scheduler/hyperliquid_fills.go
+++ b/scheduler/hyperliquid_fills.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #602, opened 2026-05-05
[codex] Widen HL fee backfill lookup window

### LINKED ISSUE #597: Backfill HL fees should fetch userFills before earliest DB trade timestamp
## Summary
`go-trader backfill hl-fees` fetches Hyperliquid `userFills` starting exactly at the earliest DB trade timestamp for the selected strategies.

## Risk
The DB trade timestamp is recorded after the Python order call returns, while the exchange fill timestamp is earlier. For the first trade in a selected strategy/range, the matching fill can land just before `earliest`, so the OID is not returned by `userFillsByTime`. That row is then skipped/unmatched and cash replay keeps the modeled fee.

Impacted path:
- `scheduler/backfill_hl_fees.go:495-496`

## Suggested fix
Subtract a safety buffer before calling `runFetchHLUserFills`, for example several minutes or one scheduler-cycle-sized window. Keep the displayed report clear about the actual query lower bound.

A regression test could seed an earliest trade timestamp that is later than its corresponding mocked fill and assert the fill is still available to the planner.

## Review context
Found during recent-commits review; severity P2 because it can make historical fee backfills incomplete while appearing to run successfully.

richkuo:
## Summary
- query Hyperliquid `userFills` from 10 minutes before the earliest DB trade timestamp during `backfill hl-fees`
- show both the query start and earliest trade timestamp in the command output
- add regression tests for the lookback helper and near-epoch clamp

## Root Cause
Trade rows are timestamped after the Python order call returns, while Hyperliquid fill timestamps come from the earlier exchange fill. Starting `userFillsByTime` exactly at the first DB trade timestamp can miss that first order's OID.

## Validation
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'Test(Backfill|PlanBackfill|ApplyBackfill)' -count=1`

Closes #597

### DIFF
--- a/scheduler/backfill_hl_fees.go
+++ b/scheduler/backfill_hl_fees.go
@@ -33,6 +33,24 @@ type HLUserFillsResult struct {
 	Error          string                   `json:"error"`
 }
 
+// backfillHLUserFillsLookback widens the userFills query before the first DB
+// trade timestamp. Trade rows are recorded after the Python order call returns,
+// while HL's fill timestamp is earlier, so an exact lower bound can miss the
+// first order in a targeted backfill (#597).
+const backfillHLUserFillsLookback = 10 * time.Minute
+
+func backfillUserFillsStartTime(earliestTrade time.Time) time.Time {
+	if earliestTrade.IsZero() {
+		return time.Time{}
+	}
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #603, opened 2026-05-05
[codex] Harden Hyperliquid fee lookup shape

### LINKED ISSUE #598: Harden Hyperliquid execute fee lookup shape and tests
## Summary
The new `check_hyperliquid.py` fee lookup assumes any truthy `lookup_fill_fee_by_oid()` result behaves like a plain dict.

## Risk
Existing execute tests use a bare `MagicMock` adapter. `lookup_fill_fee_by_oid()` is truthy and `.get()` returns another `MagicMock`, which is then written into `fill["fee"]` and fails JSON serialization. Production currently returns a dict, but the test failure exposes a brittle boundary: malformed/truthy SDK or mock responses can leak non-JSON values into the subprocess contract.

Impacted path:
- `shared_scripts/check_hyperliquid.py:454-458`

Observed failure:
- `shared_scripts/test_check_hyperliquid.py::TestFillExtraction::test_fill_with_oid_and_fee`
- `shared_scripts/test_check_hyperliquid.py::TestFillExtraction::test_fill_with_oid_no_fee`

## Suggested fix
Either update the existing test mocks to return `{}` for `lookup_fill_fee_by_oid`, or harden the production code to only consume lookup results when they are dict-like with numeric `fee` / `closed_pnl` values. The latter gives a stronger subprocess JSON contract.

## Review context
Found during recent-commits review; severity P3 because production behavior is likely fine with the current adapter, but focused Python tests fail and the boundary is easy to harden.

richkuo:
## Summary
- Harden Hyperliquid execute fee lookup before copying userFills data into the subprocess JSON contract.
- Apply the same shape and numeric-value guard to the Hyperliquid close script.
- Add regression tests for numeric lookup payloads, truthy non-mapping responses, and malformed lookup values.

## Root Cause
The execute path treated any truthy `lookup_fill_fee_by_oid()` result as dict-like. A bare `MagicMock` is truthy and can return more mocks from `.get()`, which can leak non-JSON values into stdout.

## Validation
- `uv run --extra test pytest shared_scripts/test_check_hyperliquid.py shared_scripts/test_close_hyperliquid_position.py`
- `uv run --extra test pytest shared_scripts/test_check_execute_fees.py`
- `git diff --check`

Closes #598

### DIFF
--- a/shared_scripts/check_hyperliquid.py
+++ b/shared_scripts/check_hyperliquid.py
@@ -27,6 +27,7 @@ import json
 import math
 import time
 import traceback
+from collections.abc import Mapping
 from datetime import datetime, timezone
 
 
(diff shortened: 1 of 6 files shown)


richkuo:
@claude

richkuo:
@claude sonnet fix suggested issues

richkuo:
@claude sonnet

PR #604, opened 2026-05-05
[codex] Add HL shared-coin protection orders

### LINKED ISSUE #601: shared-coin HL strategies have no on-chain SL/TP orders — bot crash = unprotected positions
## Problem

When a live Hyperliquid position is opened by a shared-coin strategy (e.g. `hl-tema-eth-live`), **no orders are placed on-chain** — no SL, no TP1, no TP2.

`normalizeHyperliquidPeerStopLosses` (config.go:665-701) coerces omitted `stop_loss_*` to `StopLossPct=0` for every peer when multiple strategies share a coin, which disables both the auto-SL and the #562 `stop_loss_atr_mult=1.0` default. TP1/TP2 are computed each cycle by close evaluators (`tiered_tp_atr_live`) and executed via `market_close` — there are no resting orders.

If the bot crashes or misses a cycle, the position is unprotected.

## Current Behavior

1. Position opens → `slPct=0` passed to `RunHyperliquidExecute` (main.go:2394, 2431); no SL trigger placed.
2. Post-open ATR-mult arming (main.go:1438) is gated on `StopLossATRMult > 0`, which stays nil for normalized peers.
3. TP1/TP2 evaluated in-process each cycle by `tiered_tp_atr_live`; closes via market order.

## Proposed Fix

Place SL, TP1, and TP2 as on-chain reduce-only orders **sized to the strategy's virtual quantity**, attributed by OID, and tracked entirely on our side. Each strategy places its own three orders independently — HL allows many reduce-only orders against the same coin, and per-strategy sizing means each set only closes that strategy's portion of the pooled position. No coordination with peer strategies needed at the exchange level.

### Order placement (at open)

After the open fill is confirmed and `EntryATR` is stamped:

1. **SL** — reduce-only trigger order at `avg_cost ± stop_loss_atr_mult * entry_atr`, qty = full `position.quantity`.
2. **TP1** — reduce-only limit order at `avg_cost ± tp1_mult * entry_atr`, qty = `floor(position.quantity * tp1_fraction, sz_decimals)`.
3. **TP2** — reduce-only limit order at `avg_cost ± tp2_mult * entry_atr`, qty = remainder.

OIDs persisted on `Position`: `StopLossOID`, `TP1OID`, `TP2OID`.

### Lifecycle

- **TP1 fills** → detect via OID poll or fill event → reduce `Position.Quantity` by TP1 size, leave TP2 in place (it was sized for the remainder up front), no cancel needed.
- **TP2 fills** → close out the position, cancel SL.
- **SL fills** → close out the position, cancel both TPs.
- **External close detected** by reconciler → cancel all three OIDs, then run existing close-recording flow.
- **Each cycle** → verify all three OIDs are still on the book; re-place any missing ones (covers post-restart recovery and HL trigger eviction).
- **Partial close from a different code path** (e.g. CB drain, manual-close) → cancel + re-place all three at the new sized quantity.

### Sizing isolation

Each strategy's reduce-only orders are sized to *its* virtual quantity, so when TP1 for Strategy A fires it reduces only A's portion of the pooled position. Strategy B's quantity and orders are untouched. Bookkeeping continues to live in SQLite (`Position.Quantity`, `InitialQuantity`, OID columns) — HL is just the trigger venue.

### Relationship to prior peer-aware fixes

This proposal is compatible with #494, #515, and #592/#593 because all three were defending against *unsized* close actions affecting the pooled position. Reduce-only orders sized per-strategy don't have that problem:

- **#494** can stop coercing peer `stop_loss_*` to 0; per-strategy sizing makes the race a non-issue.
- **#515** shared-CB drain can issue per-strategy sized closes too (or just cancel the OIDs and let the SL ride).
- **#592/#593** `shouldCloseFullPosition` is unaffected — it already uses sized closes when peers exist.

### Considerations

- **HL 1000-trigger cap** (referenced near hyperliquid_trailing_stop.go #481): 3 orders × N open positions × M peers. Need to monitor headroom and degrade gracefully (cancel TP2, keep SL+TP1) if approaching the cap.
- **Reduce-only validity with opposite-direction peers:** if Strategy A is long and Strategy B goes short until pooled is short, A's reduce-only sell may be rejected at trigger time. Confirm HL's behavior here — reduce-only orders that can't reduce typically just don't fill rather than cancel.
- **Restart recovery:** at startup, fetch `openOrders` by user, match against `Position.{StopLossOID,TP1OID,TP2OID}`. Missing OIDs → re-place. Stale OIDs (no matching position) → cancel.
- **OID poll vs. websocket:** existing reconciler uses `userFills` polling. Same path can detect TP1/TP2/SL fills by OID.
- **Cancel-on-failure:** if open succeeds but any of SL/TP1/TP2 placement fails, cancel the partial set and notifier-alert — don't leave the position with mismatched protection.
- **Backwards compatibility:** existing `tiered_tp_atr_live` close evaluator becomes a fallback for non-HL platforms and for HL when on-chain placement fails.

### Out of scope

- Sub-account isolation for full per-strategy independence (different leverage, margin, direction) remains the right answer for those cases — this issue is specifically about exchange-side protection for same-direction shared-coin strategies.
- Options on-chain TP/SL (different platform mechanics).

richkuo:
## Summary

- Add per-strategy Hyperliquid protection syncing for shared-coin perps positions: fixed ATR stop-loss plus tiered ATR TP1/TP2 reduce-only orders sized to the strategy virtual position.
- Persist TP1/TP2 OIDs on positions, migrate SQLite state, and verify/re-place missing protection on live idle cycles and immediately after live opens.
- Stop normalizing shared-coin peers out of default ATR stop protection now that protection orders are sized per strategy.
- Cancel TP OIDs alongside SL OIDs during full/flip closes, circuit closes, and kill-switch flattening so stale protection does not linger on-chain.

## Root Cause

Shared-coin Hyperliquid strategies were normalized to opt out of default SL ownership to avoid unsized trigger races on the pooled exchange position. That left peer strategies without on-chain protection, and tiered TP exits only existed as in-process close evaluators.

## Validation

- `uv run --extra test pytest platforms/hyperliquid/test_adapter.py shared_scripts/test_check_hyperliquid.py`

Closes #601 

LLM: GPT-5.5 | high

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -427,13 +427,17 @@ class HyperliquidExchangeAdapter:
         )
 
     def open_order_oids(self, symbol: str | None = None) -> set[int]:
-        """Return currently open order OIDs, optionally filtered by coin (#601)."""
+        """Return currently open order OIDs, optionally filtered by coin (#601).
+
+        Raises whatever the underlying SDK raises — callers in
+        check_hyperliquid.py treat a raise as "open-orders fetch failed,
+        defer placement decisions" so wrapping it in try/except here would
+        silently coerce uncertainty into "no open orders" and produce the
+        over-place hazard we're trying to avoid.
+        """
         if not self._account_address:
             return set()
-        try:
-            orders = self._info.open_orders(self._account_address)
-        except Exception:
-            raise
+        orders = self._info.open_orders(self._account_address)
         out: set[int] = set()
         for order in orders or []:
             if not isinstance(order, dict):
@@ -445,13 +449,27 @@ class HyperliquidExchangeAdapter:
                 out.add(oid)
         return out
 
-    def cancel_trigger_order(self, symbol: str, oid: int) -> dict:
-        """Cancel a resting trigger order by OID (#412)."""
+    def cancel_order_by_oid(self, symbol: str, oid: int) -> dict:
+        """Cancel any resting order (trigger or limit) by OID.
+
+        HL's cancel endpoint is order-type-agnostic — it accepts the OID and
+        figures out the underlying order kind from the book. Trigger orders
+        (stop-loss, take-profit-trigger) and limit orders (reduce-only TP
+        limits placed via place_take_profit_limit) are both cancellable
+        through this single primitive (#604 review #4).
+        """
         if not self._exchange:
             raise RuntimeError(
-                "cancel_trigger_order requires live mode (set HYPERLIQUID_SECRET_KEY)"
+                "cancel_order_by_oid requires live mode (set HYPERLIQUID_SECRET_KEY)"
             )
         return self._exchange.cancel(symbol, int(oid))
+
+    # Backwards-compatible alias. The original name implied only trigger
+    # orders were supported; in practice HL's cancel works for any order
+    # type. New code should call cancel_order_by_oid; existing callers can
+    # keep using cancel_trigger_order without a rename churn.
+    def cancel_trigger_order(self, symbol: str, oid: int) -> dict:
+        return self.cancel_order_by_oid(symbol, oid)
 
     def update_leverage(self, leverage: int, symbol: str, is_cross: bool) -> dict:
         """Set leverage and margin mode (cross/isolated) for ``symbol`` (#486).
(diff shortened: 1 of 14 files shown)


richkuo:
@claude

richkuo:
@claude fix all pr suggestions

richkuo:
@claude

richkuo:
@claude re-review

PR #606, opened 2026-05-05
Make top-level ATR stop default configurable

### LINKED ISSUE #605: make default_stop_loss_atr_mult a top-level config field instead of hardcoded constant
## Problem

`DefaultStopLossATRMult` is hardcoded as a Go constant (`1.0`) in `config.go:324`. There's no way to change the default without recompiling. Additionally, shared-coin strategies have their stop fields normalized to `0` by `normalizeHyperliquidPeerStopLosses`, so they can't benefit from the default at all.

## Proposed Fix

1. Add a top-level config field (e.g., `default_stop_loss_atr_mult`) with a default value of `1.0`
2. Both sole-owner and shared-coin strategies use this config value as their default `stop_loss_atr_mult`
3. For shared-coin strategies, the first peer per coin gets the default; the rest get `0` (only one SL owner per coin — existing peer normalization logic)
4. Individual strategies can still override with their own `stop_loss_atr_mult`, `stop_loss_pct`, etc.
5. User can change the config value at any time without recompiling

### Config example

```json
{
  "default_stop_loss_atr_mult": 1.0,
  "strategies": [...]
}
```

Setting `"default_stop_loss_atr_mult": 1.5` would apply 1.5× ATR stop-loss to all strategies that don't explicitly override their stop fields.

richkuo:
## Summary
- Added `default_stop_loss_atr_mult` as a top-level config field with a `1.0` fallback.
- Updated HL perps/manual defaulting so omitted fixed ATR stops use the configured top-level default.
- Changed same-coin peer normalization so the first configured omitted peer can own the default stop, while later peers are normalized to `0`.
- Bumped config version and updated migration, reload handling, example config, and repo guidance docs.

## Testing
- Added unit coverage for the new top-level default, peer ownership behavior, and migration output.

Closes #605

---
LLM: GPT-5.5 | high


richkuo:
@claude

richkuo:
@claude fix merge conflicts, and then double check the pr to see if any of the issues are still valid, if they are, fix them

PR #617, opened 2026-05-06
Add Detector 3 for Hyperliquid TP partial-fill reconciliation

### LINKED ISSUE #609: Add reconciler Detector 3 for TP partial-fill virtual/on-chain qty drift
## Background

PR #604 added per-strategy on-chain TP1/TP2 reduce-only orders for shared-coin HL perps. When a TP tier fires, `run_sync_protection` in `platforms/hyperliquid/adapter.py` correctly detects it via `TPFilledExternally` (a per-tier slice as of #615) and zeroes the corresponding `pos.TPOIDs[idx]`. However, the existing reconciler in `scheduler/hyperliquid_balance.go` has no path to handle this state.

## Problem

`reconcileHyperliquidAccountPositions` has two shared-coin detectors:

1. **Detector 1** (line 521): fires when on-chain qty ≈ 0 (full external close)
2. **Detector 2** (line 582): fires when on-chain residual matches the non-SL-owner peer sum (SL trigger fired on owner)

Neither matches a TP partial-fill scenario: after a TP tier fills, virtual qty is still the original (e.g. 1.0) while on-chain qty is half (e.g. 0.5). The gap persists until TP2/SL also fires. During that window:

- **PnL** is understated (unrealized gain on the TP leg is not booked)
- **Per-strategy drawdown** reads stale virtual qty in `perpsMarginDrawdownInputs`
- **Kill-switch margin math** over-counts deployed margin

`reduce-only` bounds the blast radius on the exchange side, but the cash and virtual-qty drift in scheduler state is real.

## Secondary effect: SL is sized against stale virtual qty

`applyHyperliquidProtectionSync` zeroes the filled tier's entry in `pos.TPOIDs` but leaves `pos.Quantity` at the original virtual qty. The next `syncHyperliquidProtection` call builds its plan from that stale qty, so it places the SL sized for 1.0 ETH when only 0.5 ETH is on-chain.

On-chain this is bounded — the SL is reduce-only, so HL will only fill up to the available 0.5 ETH. But in a shared-coin scenario it matters: if Strategy A and Strategy B each had 0.5 ETH virtual and A's TP1 fired, A's next SL placement is for 0.5 ETH when only 0.25 ETH of that remains A's on-chain portion. The reduce-only cap prevents an over-close, but the sizes are wrong relative to the shared position.

Note: as of #606, peer normalization now assigns the default `stop_loss_atr_mult` (1.0× ATR) to the first omitted peer in a shared-coin group, so at least one peer will have an active SL in a typical two-strategy setup. This makes the SL sizing drift scenario above more likely in practice rather than a theoretical edge case.

Decrementing `pos.Quantity` to match on-chain (the Detector 3 fix) also corrects the SL sizing automatically on the next `syncHyperliquidProtection` cycle.

## Proposed Fix

Add a **Detector 3** that fires when:

```
0 < onChainQty < virtualQty  (for longs; analogous for shorts)
```

The detector should be **tier-agnostic**: it does not need to know whether TP1, TP2, or any future tier caused the drop. It simply observes that virtual qty exceeds on-chain qty, books the difference as a partial close (trade record + cash credit via `bookPerpsClose`), and decrements `pos.Quantity` to match on-chain. The corrected virtual qty then causes the next `syncHyperliquidProtection` cycle to place a correctly-sized SL automatically.

N-tier TP support (#612) is complete as of #615 — `pos.TPOIDs` is now a variable-length slice. Detector 3 does not need to reference specific tiers.

This can be a follow-up PR — the on-chain reduce-only protection from #604 bounds the risk; the cash-bookkeeping and SL-sizing corrections are a separate correctness improvement.

## Files to touch

- `scheduler/hyperliquid_balance.go` — add Detector 3 block after line 581
- `scheduler/hyperliquid_balance_test.go` (or equivalent) — regression test for TP-fired partial-close state

richkuo:
## Summary
- Adds a third shared-coin reconciler path for Hyperliquid TP partial fills when on-chain quantity is a same-direction subset of virtual quantity.
- Books the partial close into trade history and cash, then reduces the remaining virtual position so protection sync resizes SL/TP orders from the true residual.
- Covers both long and short cases with regression tests for PnL, position quantity, and gap normalization.

## Testing
- Added focused unit coverage for long and short TP partial-fill reconciliation scenarios.

Closes #609

LLM: GPT-5.5 | high | Harness: OpenAI Codex

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -749,6 +749,9 @@ func hyperliquidHasClearedTPTier(sc StrategyConfig, pos *Position, closeQty floa
 	if len(tiers) == 0 {
 		return false
 	}
+	if len(pos.TPOIDs) < len(tiers) {
+		return false
+	}
 	tpOIDs := tpOIDsForTierCount(pos.TPOIDs, len(tiers))
 	hasCleared := false
 	hasActive := false
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -400,6 +400,10 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 		}
 		coinStrategies[sym] = append(coinStrategies[sym], sc.ID)
 	}
+	strategyByID := make(map[string]StrategyConfig, len(allStrategies))
+	for _, sc := range allStrategies {
+		strategyByID[sc.ID] = sc
+	}
 	sharedCoins := make(map[string]bool)
 	for coin, ids := range coinStrategies {
 		if len(ids) > 1 {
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

richkuo:
@claude re-review

PR #615, opened 2026-05-06
Generalize Hyperliquid protection orders to N tiers

richkuo:
## Summary
- Replaced fixed TP1/TP2 protection state with a variable-length `TPOIDs` slice on positions.
- Added SQLite persistence for tiered TP OIDs while keeping legacy TP1/TP2 columns as compatibility inputs.
- Updated Hyperliquid protection sync, cancellation, and execute flows to handle any number of reduce-only TP tiers.
- Extended `check_hyperliquid.py` to place, verify, and detect filled-external TP orders for N tiers.
- Added unit coverage for 3-tier planning, placement sizing, cancellation, and external-fill detection.

## Testing
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test .`
- `uv run --extra test pytest shared_scripts/test_check_hyperliquid.py`

Closes #612

LLM: GPT-5.5 | Effort: high

Issue #592, opened 2026-05-05
tiered TP leaves dust on HL when position size doesn't split evenly into 3 decimals


richkuo:
## Problem

Live Hyperliquid perps strategies that use tiered TP (close 50% at TP1, 50% at TP2) leave behind dust on every trade when the position size doesn't evenly split into two halves that fit HL's sz_decimals (3 for ETH).

Both TP1 and TP2 can round incorrectly, compounding the error.

## Evidence

`hl-tema-eth-live` opened 0.4210 ETH. Tiered TP closed:
- TP1: intended 0.2105, HL filled **0.2100** (rounded to 3 decimals)
- TP2: intended 0.2105, HL filled **0.2100** (rounded to 3 decimals)
- Total closed: 0.4200 ETH
- **Leftover dust: 0.0010 ETH** ($0.10 margin, requires manual cleanup)

The DB recorded TP2 as 0.2105 (forcing total to match the open) but the actual HL fill was 0.2100. A phantom 0.0005 partial was also inserted to reconcile the gap.

This affects any position size where `qty / 2` has more than 3 decimal places (e.g., 0.421, 0.423, 0.425, etc.).

## Root Cause

The tiered TP close logic in `check_hyperliquid.py` calculates `qty * 0.5` for each leg, but:
1. HL enforces `sz_decimals` per asset (3 for ETH, 5 for BTC, etc.) — orders with excess decimal places get rounded by the adapter
2. The adapter already fetches `sz_decimals` from the HL `meta` API and stores it in `self._info.asset_to_sz_decimals` (see `platforms/hyperliquid/adapter.py:230-232`)
3. The tiered TP close in `check_hyperliquid.py` doesn't apply `sz_decimals` rounding before calculating partial quantities
4. Both TP1 and TP2 get rounded down independently by the adapter, and neither accounts for the other's rounding
5. The second leg doesn't use the actual fill size from the first leg — it independently calculates `qty * 0.5` again

## Proposed Fix

The tiered TP close in `check_hyperliquid.py` should:

1. Get `sz_decimals` from the adapter (already available via `self._info.asset_to_sz_decimals`)
2. For TP1 (partial close): round the close quantity **down** to `sz_decimals` before submitting the order
3. After TP1 fills, track the **actual fill size** returned by HL (not the requested amount) to compute remaining position
4. For TP2 (final close): close the full remaining position with no rounding — this ensures zero dust regardless of TP1 rounding

Example:
[...]

PR #620, opened 2026-05-06
fix: manual close isolation — owner guard, sole-peer close, TP OID cancel

### LINKED ISSUE #619: Manual strategy: fix close isolation so manual positions don't flatten peer perps positions on same coin
## Problem

The current config validation blocks `type=manual` strategies on the same coin as perps strategies because of a real race condition — when a manual position is closed, it calls `adapter.market_close(sz=None)` which closes the **entire** wallet position, not just the manual strategy's portion.

This means manual positions and automated perps strategies cannot share a coin, even though they should be able to coexist independently.

## Desired behavior

1. Manual positions should be isolated — closing a manual position only closes that manual position, not any peer automated positions on the same coin.
2. Manual positions should support the same closing strategies as automated strategies (TP1, TP2, SL).
3. No sub-accounts needed — everything on one account, just logically isolated.

## Technical requirements

From the code comments, the specific issue is in `config.go` around the validation that rejects manual+perps on same coin (`Fix #6`). The actual fix needs to:

- Make the manual close path respect position ownership (don't close if `owner_strategy_id` doesn't match)
- Or tag manual positions with their strategy ID so the close is scoped
- The perps path's `shouldCloseFullPosition` sole-peer guard doesn't run for manual closes — this needs to be fixed

## Notes

- Not using sub-account isolation
- Want to be able to manually open/manage positions alongside automated strategies on the same coin

richkuo:
Removes the blanket ban on `type=manual` / `type=perps` coin sharing (#595/#599) and replaces it with the same isolation guarantees the perps close path already provides.

## Changes

1. **OwnerStrategyID guard** — `runManualClose`, `applyManualAction`, and the main-loop manual case all reject a position whose `OwnerStrategyID` does not match the acting strategy before touching on-chain state. A mismatched owner logs an error and aborts; position and cash are untouched.

2. **`shouldCloseFullPosition` sole-peer guard** — manual full-close paths now call `shouldCloseFullPosition` (same logic as perps final-tier TPs). When another live HL strategy shares the coin, it falls back to a sized market open in the opposite direction instead of `market_close(sz=None)`, preventing accidental flattening of a peer's on-chain exposure (#592/#619).

3. **TP OID cancellation on manual full close** — `extraCancelOIDs` is populated from `pos.TPOIDs` before the close execute, mirroring the perps path added in #615.

4. **On-chain protection for `type=manual`** — `buildHyperliquidProtectionPlan` and `hyperliquidPlacesOnChainTPs` are extended to accept `type=manual`. The main loop now runs a protection-sync block for live manual strategies each cycle (same TP/SL OID management as perps).

5. **`hyperliquidConfiguredCoin` helper** — unifies coin resolution: `manual` reads `sc.Symbol`, `perps` reads `hyperliquidSymbol(sc.Args)`. Used in `hlLiveStrategiesForCoin` and `hyperliquidPeerStrategyErrors`.

6. **Peer validation extended** — `hyperliquidPeerStrategyErrors` now includes `type=manual` strategies in the leverage / margin-mode mismatch check (#491 scope extended to #619). Manual strategies still cannot be trailing-stop owners (existing guard unchanged).

7. **Config validation** — the old `conflicts with` error for manual/perps same-coin is removed; leverage-mismatch peer errors still fire.

## Tests

- `TestValidateConfigManualSymbolSharingAllowed` — manual+manual and manual+perps same-coin configs now pass validation.
- `TestValidateConfigManualPerpsPeerLeverageMismatchRejected` — mismatched leverage across manual/perps peers is still rejected.
- `TestBuildHyperliquidProtectionPlanManualStrategy` — protection plan builds correctly for `type=manual`.
- `TestFilterCloseStrategiesForHLOnChainProtection` — manual with `tiered_tp_atr_live` is filtered when on-chain TPs are active.
- `TestShouldCloseFullPosition` — manual+perps same-coin → sized close; solo manual → full close.
- `TestApplyManualActionCloseRejectsOwnerMismatch` — `applyManualAction` returns error and leaves position untouched when owner doesn't match.

Closes #619

---
LLM: gpt-5.5 | high | Harness: codex


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

richkuo:
@claude

richkuo:
## Review

**Verdict: LGTM with a few notes.** The change correctly replaces the blanket #595/#599 ban with the same isolation guarantees the perps path provides. Logic is consistent, well-factored, and adequately tested.

### What's good

- **Owner guard is layered correctly** — checked at `runManualClose` (CLI), `applyManualAction` (queue drain), and twice in the main-loop manual case (top + before exec). The double-check defends against a TOCTOU between the first RLock snapshot and the close exec. `manualPositionOwnedByStrategy` correctly treats empty `OwnerStrategyID` as "owned" so legacy positions don't break.
- **Sole-peer guard symmetry** — `shouldCloseFullPosition` is now consistently used by perps (`runHyperliquidExecuteOrder` via `hlReconcileAll`), main-loop manual, and `runManualClose` (via `hyperliquidCloseScopeStrategies`). All three paths now flatten only when actually solo on the coin.
- **`extraCancelOIDs` semantics are right** — gated on `intentFullClose`, not `closeFullPosition`. Operator-intent full-close still cancels this strategy's per-strategy reduce-only TP OIDs (which are isolated from a peer's TP OIDs), even when the wallet-level close is sized. Mirrors the perps #615 behavior.
- **Protection sync wired into main loop for manual** — `buildHyperliquidProtectionPlan` + `applyHyperliquidProtectionSync` now run for live `type=manual` with a position, matching perps. `hyperliquidPlacesOnChainTPs` extension means `filterCloseStrategiesForHLOnChainProtection` will suppress overlapping close evaluators, avoiding the #604 race.
- **`hyperliquidConfiguredCoin` helper** is the right consolidation; previously `hyperliquidSymbol(sc.Args)` was used everywhere even though manual sets `Symbol` directly.
- **Tests** cover the new positive cases (sharing allowed, manual protection plan, manual sole-peer), the negative case (leverage mismatch still rejected), and the owner-mismatch close rejection.

### Issues / suggestions

1. **CLAUDE.md is now stale.** The repo doc still asserts: *"Validation rejects `type=manual` sharing a symbol with a live HL perps strategy (or another live `type=manual`, #599)."* This PR removes that constraint — the doc should be updated in the same commit (or a doc-sync follow-up like #618) so future agents don't re-add the ban.

1. **Behavioral change in `hlLiveStrategiesForCoin`**: switched coin comparison from `==` to `strings.EqualFold` (`hyperliquid_balance.go:923`). Not called out in the PR description. Likely benign (HL coins are uppercase by convention), but it does mean a config typo like `"eth"` vs `"ETH"` will now resolve as a peer instead of silently isolating. Consider either reverting to case-sensitive (matches the rest of the codebase like `forceCloseHyperliquidLive`) or normalizing in `hyperliquidConfiguredCoin` — pick one and apply consistently.

1. **`hyperliquidPeerStrategyErrors` margin-mode mismatch lacks a manual-mixed test.** New test covers leverage mismatch but not `MarginMode` mismatch across manual+perps peers. One-line addition.

1. **Trailing-stop-owner guard scope.** PR description says "Manual strategies still cannot be trailing-stop owners (existing guard unchanged)" — verified that `hasHyperliquidTrailingStopOwnership` is called inside `hyperliquidPeerStrategyErrors` regardless of type, so a manual strategy with `trailing_stop_pct`/`trailing_stop_atr_mult` would still be flagged in the multi-trailing check. Consider asserting this with a test since the type-set was just widened.

1. **Minor: redundant nil-check.** In the main-loop manual case, `pos != nil && !manualPositionOwnedByStrategy(pos, sc.ID)` — `manualPositionOwnedByStrategy` already nil-guards (returns `true` when nil), so the outer `pos != nil` is just there for clarity. Fine as-is, just noting.

1. **Cancel-fail visibility on full-close-with-peer.** When `intentFullClose=true` and `closeFullPosition=false`, you cancel `pos.StopLossOID` and `pos.TPOIDs`. If the cancel call fails (HL has known cancel-failure modes per `hyperliquid_trailing_stop.go`), the manual strategy is virtually flat but on-chain TP/SL OIDs remain. The perps path treats this as non-fatal; matching that here is fine, but a one-line `logger.Warn` on cancel-failure surface area would be useful.

### Risk assessment

- **Reversibility:** high — no schema changes, no migration; revert is a clean diff revert.
- **Blast radius:** medium — affects all HL manual strategies and the runHyperliquidExecuteOrder peer set for perps. The `hlLiveAll` → `hlReconcileAll` swap at `main.go:1485` is the highest-impact line: perps closes will now treat manual peers as blockers for `market_close(sz=None)`. That's the intended behavior, and `TestShouldCloseFullPosition` asserts it.
- **Operator UX:** users who relied on the validation error to catch accidental same-coin manual+perps configs will now get silent coexistence — worth a release note.

No blocking issues. Recommend merging once CLAUDE.md is synced and the EqualFold case-sensitivity decision is made.

richkuo:
@claude

richkuo:
# Code Review

## Overview

Replaces the blanket ban (#595/#599) on manual/perps coin sharing with the same isolation primitives perps already use:
- `OwnerStrategyID` guard in three close paths (`runManualClose` CLI, `applyManualAction` drain, main-loop manual case).
- `shouldCloseFullPosition` sole-peer guard wired into manual full-close (so a manual full close on a shared coin falls back to a sized close instead of `market_close(sz=None)`).
- `extraCancelOIDs` populated from `pos.TPOIDs` on full close (mirrors #615 perps path).
- On-chain reduce-only TP/SL protection (`buildHyperliquidProtectionPlan` + main-loop sync block) extended to `type=manual`.
- Peer validation (`hyperliquidPeerStrategyErrors`) widened to manual; CB drain widened to use `hlCircuitPeerAll` so a perps CB never silently flattens a manual peer.
- New `hyperliquidConfiguredCoin(sc)` helper unifies coin resolution (`sc.Symbol` for manual, `hyperliquidSymbol(sc.Args)` for perps).

`go build` + `go test ./...` clean (21s).

## Strengths

- **Correct symmetry with perps.** `shouldCloseFullPosition` is reused without duplication; the same pre-existing review #1 doc-comment from #592 now correctly covers manual too.
- **TOCTOU windows handled.** Owner check, plan build, subprocess, and apply each take their own RLock/Lock — matches the established 6-phase pattern. The post-subprocess apply re-validates `Quantity > 0 && Side == plan.Side` (`scheduler/main.go:1578`) to avoid clobbering after an external close.
- **Conservative CB drain.** Switching the shared-coin guard from `hlLiveAll` (perps-only) to `hlCircuitPeerAll` (perps+manual) is the right blast-radius choice — the regression test (`TestRunPendingHyperliquidCircuitCloses_ManualPeerClearsPendingWithoutClose`) exercises exactly this.
- **Operator-facing logging.** `CancelStopLossError` is now surfaced (warn) so orphaned reduce-only OIDs after a manual close are visible.

## Issues / suggestions

1. **`manualPositionOwnedByStrategy` treats empty `OwnerStrategyID` as owned (`scheduler/manual.go:621`).**
   \`\`\`go
   return pos == nil || pos.OwnerStrategyID == \"\" || pos.OwnerStrategyID == strategyID
   \`\`\`
   Presumably for backward-compat with positions opened before #569 stamped owners. Fine, but worth a one-line comment so a future reader doesn't tighten it. Otherwise: any reconciler-discovered position with empty owner would be eligible for close by *any* manual strategy on that coin — which is the precise scenario this PR is hardening against.

1. **No CLI-path owner-mismatch test.** `TestApplyManualActionCloseRejectsOwnerMismatch` covers the queue path; `runManualClose` (`scheduler/manual.go:300`) has the same guard but isn't exercised. Low risk (the guard is a one-liner mirror), but a parallel test would be cheap.

1. **No integration test for the main-loop manual protection sync block** (`scheduler/main.go:1567–1585`). `TestBuildHyperliquidProtectionPlanManualStrategy` only covers the planner. The sync block is the most novel piece — extracting a `runManualProtectionSync(sc, stratState, mu, ...)` helper would let you test \"owner mismatch skips both protection AND close\" and \"protection sync with `Quantity==0` post-subprocess is a no-op.\"

1. **Case-sensitivity caveat.** `hyperliquidConfiguredCoin` returns `sc.Symbol` verbatim for manual and `hyperliquidSymbol(sc.Args)` for perps — neither normalizes case. If an operator writes `\"eth\"` in `symbol:` for a manual peer and `\"ETH\"` in `args:` for a perps peer, peer detection fails silently. The PR description references the second commit \"case-sensitive coin match\" but the diff doesn't show normalization. Consider `strings.ToUpper(strings.TrimSpace(...))` or, at minimum, document the requirement in the field comment.

1. **Protection sync runs every cycle for every live manual strategy with a position** (`scheduler/main.go:1567`). Same pattern as perps, so consistent — but worth confirming the HL `userFills` round-trip is acceptable at manual cadence (which is set per-strategy in `LoadConfig`).

1. **Minor dead-code smell.** `closeIntentFraction` (`scheduler/main.go:1607–1610`) is computed only for the `shouldCloseFullPosition` call and is mathematically identical to `1.0` whenever it's used (since `intentFullClose` gates it). You can pass `1.0` directly and drop the local; the extra clarity isn't pulling weight.

## Risk assessment

**Low.** The changes are additive — closing the gap that the prior blanket ban papered over. The owner guard fails closed (skip + log error), the sole-peer guard fails to a more conservative sized close, and the CB scope widening fails to \"leave on-chain alone.\" Worst-case regression is a manual close that *should* have closed silently being skipped because of an empty `OwnerStrategyID` — which the operator will see immediately.

---

richkuo:
@claude sonnet re-review

richkuo:
## Code review (PR #620)

### Verdict
**LGTM** — `go test -C scheduler ./...` passes on this branch. The change set is coherent: config no longer rejects same-coin manual/perps pairs; runtime isolation matches what the perps path already did (`shouldCloseFullPosition`, TP/SL OID hygiene, protection sync), and circuit-breaker / `computeHyperliquidCircuitCloseQty` paths now treat **manual + perps** as one peer set so a perps CB cannot size a full-wallet close while a manual peer shares the coin.

### What looks strong
- **`manualPositionOwnedByStrategy`** on CLI close, `applyManualAction`, and the main-loop manual branch closes the cross-strategy mutation hole; tests cover mismatch and legacy empty `OwnerStrategyID`.
- **`hyperliquidConfiguredCoin`** + upper/trim normalization fixes fragile peer matching between `symbol` and args coin.
- **`runHyperliquidProtectionSync`** extracts the RLock → subprocess → Lock re-validate pattern cleanly; stub in tests avoids Python.
- **Passing `hlReconcileAll` into `runHyperliquidExecuteOrder`** aligns final-tier `market_close(sz=None)` with the same peer notion as manual — correct fix for shared-coin manual + automated.
- **CB / stuck recovery** using `hlCircuitPeerAll` (`isHLLiveReconcilable`) is the right scope extension (#620).

### Optional / small follow-ups (non-blocking)
1. **Log copy** — In `runHyperliquidExecuteOrder`, when `close_fraction == 1.0` but `shouldCloseFullPosition` is false, the log still says **"HL perps peers"** even though peers can now include `type=manual`. Manual full-close already logs **"HL peers"**. Consider renaming that line to **"HL peers"** for consistency.
2. **Docs** — Only `CLAUDE.md` is updated; if `AGENTS.md` is meant to stay in lockstep with the manual/perps sharing paragraph, mirror the same sentence there.
3. **`syncHyperliquidProtection` package var** — Fine for tests; if anything later uses `t.Parallel()` in this package alongside the stub, watch for races (unlikely today).

### Intentional tradeoff (well documented in code)
- Treating **empty `OwnerStrategyID` as owned** preserves legacy/reconciler positions; operators should know two manuals on the same coin without stamped owners are not owner-gated — sizing / CB guards still limit catastrophe.

Nice work on the tests (`TestValidateConfigManualPerpsPeerMarginModeMismatchRejected`, trailing-stop across manual+perps, CB clear without close, protection sync races).

PR #622, opened 2026-05-06
fix: cap HL SL size at on-chain qty and use actual fill qty in reconciler

### LINKED ISSUE #621: Service restart re-arms SL at wrong quantity after manual TP fill — Detector 3 blind spot for unknown TP OIDs
## Problem

When a user manually places a TP order on Hyperliquid UI (no OID in bot state), the partial fill creates a gap between virtual qty and on-chain qty that the reconciler cannot detect. This causes two failure modes when the oversized SL order is subsequently re-armed at the virtual quantity:

1. **Unprotected position** — HL rejects the oversized reduce-only SL order at placement (sz > actual position), leaving the position with no stop-loss.
2. **State accounting error** — if the SL placement is accepted and later triggers, the bot books the close against `pos.Quantity` (0.422 virtual) rather than the actual fill size (0.211), corrupting cash/PnL.

Note: "double-close" (closing more than the on-chain position) cannot occur because `place_stop_loss` uses `reduce_only=True` — HL caps the fill at the actual position size.

## Reproduction

1. Bot opens long 0.422 ETH, arms SL for 0.422 (`pos.StopLossOID` = some OID, `pos.Quantity` = 0.422)
2. User manually places TP1 limit sell for 0.211 ETH on Hyperliquid UI — bot has no knowledge of this order
3. TP1 fills — on-chain position reduced to 0.211 ETH, but `pos.Quantity` still 0.422 (no TP OID in state to clear)
4. Service restarts (or next cycle for trailing-stop strategies)
5. Reconciler runs: sees delta=0.211, fails all three detectors (Detector 3 can't fire without a cleared bot-placed TP OID), logs gap — virtual state unchanged
6. Per-strategy loop re-arms SL using `hlPosQty = 0.422` (from virtual state) → `place_stop_loss(sz=0.422, reduce_only=True)` on a 0.211 position
7. HL rejects the oversized reduce-only order → **position is unprotected**; or if accepted and triggered, bot records a 0.422 close → **cash/PnL accounting error**

**Which step 6 applies depends on strategy type:**
- **Trailing stop** (`trailing_stop_pct` / `trailing_stop_atr_mult`): fires every cycle at `main.go:1393` — does not require a restart to hit this.
- **Fixed ATR SL** (`stop_loss_atr_mult`): re-arm at `main.go:1443` is gated on `hlStopLossOID == 0`; if the original OID is still in persisted state after restart, re-arming does NOT occur (original oversized order stays resting instead).

## Root Cause

**Detector 3** (`hyperliquid_balance.go:640-680`) detects TP partial fills by calling `hyperliquidHasClearedTPTier`, which requires `pos.TPOIDs` to contain at least one zero (cleared bot-placed OID). When the TP was placed manually, `pos.TPOIDs` is empty and `hyperliquidHasClearedTPTier` returns false immediately (`hyperliquid_balance.go:745`). Detector 3 is therefore blind to any partial fill not originating from a bot-placed TP order.

## Fix

Detector 3 needs to handle the case where a TP order is filled **without** the bot knowing the OID. This requires either:

1. **Periodic TP OID discovery**: On reconciliation, fetch open orders from HL and match them against known TP tiers by price to backfill unknown OIDs — so subsequent partial fills are attributed correctly.

2. **SL quantity guard**: Before arming (or re-arming) a stop-loss, cap the size at `min(virtualQty, onChainQty)`. The on-chain position is already fetched by the reconciler; the per-strategy loop at `main.go:1443` would need that value passed through. This prevents placement of an oversized reduce-only order regardless of whether the restart triggered the issue.

Option 2 is simpler and safer as a defense-in-depth. Option 1 is more complete but requires additional HL API calls and only addresses future fills (not the already-drifted state).

Both options should be paired with correct bookkeeping: when the SL fires and the reconciler closes the position, use the actual fill quantity (from `userFills`) rather than `pos.Quantity`.

## Notes

- This is distinct from issue #619 (manual/automated position isolation) — this is specifically about **on-chain state drift** when TP orders are placed manually outside the bot.
- Not using sub-account isolation.

richkuo:
## Summary

- **Part A — SL quantity guard**: adds `hlSLEffectiveQty()` helper that caps stop-loss placement size at `min(virtualQty, onChainQty)`. A coin→qty map is built from the pre-fetched `hlPositions` before the per-strategy loop and applied at both SL arm sites (trailing stop `main.go:1404`, fixed ATR SL `main.go:1461`). Prevents HL from rejecting an oversized reduce-only order — which leaves the position unprotected — when a manual TP reduced the on-chain position without the bot's knowledge.
- **Part B — bookkeeping fix**: adds `FilledQty float64` to `HLFillLookup`, accumulated from `f.Sz` in both OID and coin+size lookup paths. When the reconciler closes a position via SL and `useFillFee` is true, overrides `pos.Quantity` with `lookup.FilledQty` (when smaller) so PnL and cash credit the actual fill size rather than the stale virtual quantity. Applied at all three reconciler SL-close sites (single-strategy path, Detector 1, Detector 2).
- **13 new tests**: `hlSLEffectiveQty` (4), `hyperliquidHasClearedTPTier` (4, previously zero coverage), `FilledQty` accumulation in fills lookup (3), SL-close bookkeeping with real fill qty vs. fallback (2).

## Test plan

- [x] `go -C scheduler build .` — clean
- [x] `go -C scheduler test ./...` — all pass (21s)
- [x] `gofmt` — no diff

Closes #621

---

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -302,6 +302,13 @@ func reconcileHyperliquidPositionsWithResolver(stratState *StrategyState, sym st
 			lookup, useFillFee := resolveFee(sym, statePos.StopLossOID, statePos.Quantity)
 			oidStr := strconv.FormatInt(statePos.StopLossOID, 10)
 			logHyperliquidReconcileFillLookup(logger, sym, statePos.StopLossOID, statePos.Quantity, lookup, useFillFee)
+			// #621: When userFills returned a real fill qty smaller than the virtual
+			// position (e.g. SL was placed at the on-chain size after a manual TP
+			// reduced the position), use the actual fill qty so PnL/cash are correct.
+			if useFillFee && lookup.FilledQty > 1e-9 && lookup.FilledQty < statePos.Quantity-1e-9 {
+				logger.Info("hl-sync: %s SL close qty adjusted %.6f → %.6f (actual fill from userFills)", sym, statePos.Quantity, lookup.FilledQty)
+				statePos.Quantity = lookup.FilledQty
+			}
 			if recordPerpsStopLossCloseWithFillFee(stratState, sym, statePos.StopLossTriggerPx, lookup.Fee, useFillFee, oidStr, "stop_loss", logger) {
 				return true
 			}
@@ -556,6 +563,12 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 					lookup, useFillFee := resolveFee(coin, pos.StopLossOID, pos.Quantity)
 					oidStr := strconv.FormatInt(pos.StopLossOID, 10)
 					logHyperliquidReconcileFillLookup(logger, coin, pos.StopLossOID, pos.Quantity, lookup, useFillFee)
+					if useFillFee && lookup.FilledQty > 1e-9 && lookup.FilledQty < pos.Quantity-1e-9 {
+						if logger != nil {
+							logger.Info("hl-sync: %s SL close qty adjusted %.6f → %.6f (actual fill from userFills)", coin, pos.Quantity, lookup.FilledQty)
+						}
+						pos.Quantity = lookup.FilledQty
+					}
 					if recordPerpsStopLossCloseWithFillFee(ss, coin, pos.StopLossTriggerPx, lookup.Fee, useFillFee, oidStr, "hl_sync_stop_loss", logger) {
 						changed = true
 					}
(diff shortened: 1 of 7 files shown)


richkuo:
@claude

richkuo:
No blocking findings on PR 622. The change is tightly scoped, and the new `FilledQty` path feeds the existing close-booking code correctly before the position is deleted.

Verified locally on the PR branch:

`env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test ./...` passed  
`env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler build .` passed  
`gofmt -l` returned no files  
`git diff --check` was clean

LGTM from my pre-review.

richkuo:
@claude sonnet re-review

PR #624, opened 2026-05-06
Backfill open trade SL price after fixed-ATR arming

### LINKED ISSUE #623: Trade open Discord alert missing SL price — stampOpenTradeFromPosition not called after live HL SL arm
## Problem

When a live Hyperliquid position opens with a fixed-ATR stop-loss (`stop_loss_atr_mult`), the open trade's `Trade.StopLossTriggerPx` stays `0` in SQLite, so the persisted record (TradingView export, leaderboard reads, future queries) never reflects the SL price even after it's armed on-chain.

The Discord open DM is also missing the SL line, but the cause is two-fold (see "Validation notes" below) — fixing the persisted record alone won't repair the DM.

## Root Cause

Fixed-ATR SL is a **two-cycle** flow:
- **Cycle N (open):** `ExecutePerpsSignalWithLeverage` records the open trade. `pos.StopLossTriggerPx == 0` because EntryATR isn't stamped until after open. `stampOpenTradeFromPosition` is called (main.go:2591, main.go:2615) but no-ops on a zero trigger. Open DM is sent at main.go:1639 with `trade.StopLossTriggerPx == 0`.
- **Cycle N+1 (arm):** With `result.Signal == 0` and `hlPosQty > 0`, the live arm block (main.go:1443-1464) calls `hyperliquidArmFixedATRStopLossLive`, sets `pos.StopLossOID` and `pos.StopLossTriggerPx` (main.go:1456-1457), but never calls `stampOpenTradeFromPosition` — so the existing open `Trade` row never gets backfilled. `trades` is not incremented, so `sendTradeAlerts` early-returns on `trades <= 0` (main.go:2231) and no follow-up DM is sent.

The same gap exists in the **paper** arm at main.go:1426-1442: line 1436 sets `pos.StopLossTriggerPx = newTrigger` with no `stampOpenTradeFromPosition` call.

For comparison, the standard execute path (when the SL OID comes back inside the open fill) does stamp correctly at main.go:2613-2615.

## Fix

Add `stampOpenTradeFromPosition(stratState, db, result.Symbol, pos)` in **both** arm branches, immediately after the `pos.StopLossTriggerPx = ...` assignment:

1. Live: after main.go:1457 (inside the `slResult.StopLossOID > 0` else-if branch).
2. Paper: after main.go:1436 (inside the `newTrigger > 0 && pos.StopLossTriggerPx == 0` else-if branch).

The helper is idempotent (only updates fields currently zero) and short-circuits on a close trade ahead of the open in `s.TradeHistory`, so it's safe to call from the locked block.

## Scope clarification

This fix repairs the **persisted SQLite row** only. The open Discord DM was already sent on cycle N before the SL was armed and won't be retroactively updated. If we also want the DM to show the SL price, a separate change is required — e.g. delay the open DM until the first cycle where SL is either armed or skipped, or emit a follow-up "SL armed" alert on cycle N+1. Out of scope for this issue unless explicitly added.

## Validation notes (from /validate-issue)

- ✅ Live fixed-ATR arm omits `stampOpenTradeFromPosition` (main.go:1455-1459).
- ✅ Paper fixed-ATR arm omits the same call (main.go:1426-1442).
- ✅ Standard execute path stamps correctly (main.go:2613-2615).
- ⚠️ Original framing said the "paper path calls it on immediate fills" — that's not what the paper code does; the paper "immediate" branch invokes `recordPerpsStopLossClose` (a close path, not an open backfill).

richkuo:
## Summary
- Backfill `StopLossTriggerPx` onto the most recent open trade when a fixed-ATR Hyperliquid stop-loss is armed in the paper and live paths.
- Keep the persisted SQLite trade row in sync so TradingView export, leaderboard reads, and later queries see the armed SL price.
- Add regression coverage for the helper’s SQLite update behavior.

## Root Cause
Fixed-ATR stop-loss arming happens one cycle after the open trade is recorded, so the open `trades` row was inserted before the position had a known stop-loss trigger price.

## Testing
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'TestStampOpenTradeFromPosition|TestRunHyperliquidFixedATRStopLossPaper|TestHyperliquidArmFixedATRStopLossLive'`

Closes #623

LLM: GPT-5 | Effort: medium | Harness: Codex desktop app

### DIFF
--- a/scheduler/discord_test.go
+++ b/scheduler/discord_test.go
@@ -2029,4 +2029,30 @@ func TestStampOpenTradeFromPosition(t *testing.T) {
 	if s3.TradeHistory[0].EntryATR != 0 {
 		t.Error("nil pos should be a no-op")
 	}
+
+	// (e) updates the persisted SQLite row for the already-inserted open trade.
+	db, err := OpenStateDB(":memory:")
+	if err != nil {
+		t.Fatalf("OpenStateDB: %v", err)
+	}
+	defer db.Close()
+	ts := time.Date(2026, 1, 2, 3, 4, 5, 0, time.UTC)
+	s4 := &StrategyState{ID: "s4", TradeHistory: []Trade{
+		{Symbol: "ETH", IsClose: false, EntryATR: 0, StopLossTriggerPx: 0, Timestamp: ts},
+	}}
+	if err := db.InsertTrade(s4.ID, s4.TradeHistory[0]); err != nil {
+		t.Fatalf("InsertTrade: %v", err)
+	}
+	stampOpenTradeFromPosition(s4, db, "ETH", &Position{EntryATR: 250.0, StopLossTriggerPx: 2950.0})
+
+	var entryATR, stopLossTriggerPx float64
+	if err := db.db.QueryRow(
+		`SELECT entry_atr, stop_loss_trigger_px FROM trades WHERE strategy_id = ? AND timestamp = ?`,
+		s4.ID, formatTime(ts),
+	).Scan(&entryATR, &stopLossTriggerPx); err != nil {
+		t.Fatalf("query stamped trade: %v", err)
+	}
+	if entryATR != 250.0 || stopLossTriggerPx != 2950.0 {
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #629, opened 2026-05-06
fix: final TP tier absorbs per-tier floor rounding to eliminate position residual

### LINKED ISSUE #628: Protection sync fails to place TP orders when position size falls below minimum order size
## Summary

When a Hyperliquid perps position shrinks to a size below the minimum order quantity (driven by `sz_decimals` precision), the protection sync's `tier_size` calculation results in a quantity that floors to zero, and no TP order is placed. Over repeated partial fills, this leaves the strategy with progressively weaker protection until eventually no TP orders are placed at all.

Additionally, in multi-strategy (same-coin peer) configurations, the protection sync runs per-strategy against each strategy's own virtual quantity. When one strategy's position shrinks below the minimum, it places no TP orders for its remaining virtual qty, while other peer strategies still have their TP orders resting on the shared on-chain position — creating incomplete and asymmetric protection across peers.

## Reproduction

1. Open a HL perps position with tiered TP (TP1 at 50%, TP2 at 50%)
2. TP1 partially fills, reducing virtual qty
3. On next protection sync, `tier_size = size * 0.5` yields a value below the HL minimum order size for that asset
4. `place_take_profit_limit` raises `ValueError: Size floored to zero for ETH (sz_decimals=3)` and the TP is not placed
5. Trade alert shows `tp_oids=[0 0]` — no take-profit protection

**Multi-strategy variant:**
1. Strategy A (0.21 ETH) + Strategy B (0.10 ETH) both on ETH
2. Strategy A's TP1 fills first, shrinking Strategy A's virtual qty toward zero
3. Strategy A's protection sync now places no TP orders (qty too small)
4. Strategy B still has its TP orders open on the shared on-chain position
5. Eventually Strategy A's virtual qty is un-hedged while Strategy B's TPs control the shared position

### Issue 1: Size floored to zero in tier calculation
In `shared_scripts/check_hyperliquid.py`, `syncProtection` computes tier sizes as:
```python
tier_size = size * max(cumulative_fraction - prev_fraction, 0.0)
```
When `size` is small and `cumulative_fraction - prev_fraction` (e.g. 0.5 for TP1) yields a product below the HL minimum for that asset's `sz_decimals`, the size is passed to `adapter.place_take_profit_limit` which then raises `ValueError` and the TP order is skipped.

The error IS captured (goes into `tp_errors[idx]` → `TPErrors` → logged as "protection partially failed"), but the TP is not placed and the virtual quantity remains un-protected.

### Issue 2: Same-coin peers with asymmetric protection state
In same-coin peer configurations (multiple strategies on the same HL coin sharing a single on-chain position), each strategy's protection sync runs against its own virtual quantity. When one strategy's position shrinks below the minimum, it contributes nothing to the shared on-chain protection while its peers still have their orders in the book. The result is incomplete protection for that strategy's virtual qty while the shared position is partially covered by peers.

This is especially problematic because the protection sync has no awareness of the shared on-chain position or peer strategies' orders.

### For Issue 1 (size floored to zero):
- Before placing a TP order, check if `tier_size` would round to zero given `sz_decimals`. If so, skip the order and log a clear warning that the position is below minimum order size.
- Optionally: if remaining size is below minimum but not zero, consider combining remaining tiers into a single order at the final TP price (close-to-market approach).
- Alternatively: surface this as a distinct error state (e.g. `tp_skipped_reason: below_minimum_order_size`) so operators know the position is un-protected.

### For Issue 2 (same-coin peer asymmetry):
- Track shared on-chain position state and distribute protection across peers more intelligently.
- Or: when a strategy's remaining virtual qty is below minimum, defer to peer strategies' protection rather than trying to place orders that will fail.

### General:
- Fix `sz_decimals not found for ETH` warning — adapter should fetch/cache asset decimal info from HL so it doesn't default to 3.

## Files Affected

- `shared_scripts/check_hyperliquid.py` — tier size calculation and placement logic
- `platforms/hyperliquid/adapter.py` — `sz_decimals` lookup and `_floor_size`
- Possibly `scheduler/hyperliquid_protection.go` — multi-strategy coordination

richkuo:
## Summary

Closes #628

### Issue 1 — per-tier floor rounding leaves a permanent residual

Each tier independently applied `_floor_size` to its fraction of the virtual position size. When `size × fraction` wasn't an exact lot multiple, the truncation losses accumulated and the last tier left a permanent gap that no TP ever covered.

**Before / after — 0.003 ETH, default 50/50 split, `sz_decimals=3`:**

| | TP1 | TP2 | Total | Gap |
|---|---|---|---|---|
| Before | 0.001 | 0.001 | 0.002 | **0.001 ETH stranded** |
| After  | 0.001 | 0.002 | 0.003 | 0 |

Fix: non-final tiers accumulate their floored size via the new `adapter.floor_size(symbol, sz)`; the final tier uses `size - placed_size_total` (exact remainder).

### Issue 2 — float64 drift pushes virtual qty below lot boundary

Go's `pos.Quantity -= closeQty` uses float64 arithmetic. Subtracting two valid lot-boundary quantities can drift just below the boundary — e.g. `0.011 - 0.010 = 0.0009999999999999992`. `_floor_size` then truncates this to zero, causing all tiers to fail with "Size floored to zero" even though the virtual qty represents exactly one lot.

Fix: `adapter.round_size(symbol, sz)` (using `round()`, consistent with how `place_stop_loss` already sizes SL orders) normalizes `size` at the top of the TP tier block before any tier calculation runs.

## Test plan

1. Verified zero gap across representative ETH position sizes (0.001–0.015 ETH) via inline simulation.
2. Verified `0.011 - 0.010` drift case normalizes to `0.001` and covers position fully.
3. `python3 -m py_compile` passes on both modified files.

---

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -426,6 +426,11 @@ class HyperliquidExchangeAdapter:
             symbol, is_buy, sz, limit_px, order_type, reduce_only=True
         )
 
+    def floor_size(self, symbol: str, sz: float) -> float:
+        """Return sz floored to the asset's lot precision — mirrors place_take_profit_limit."""
+        sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3) if self._info else 3
+        return _floor_size(sz, sz_decimals)
+
     def open_order_oids(self, symbol: str | None = None) -> set[int]:
         """Return currently open order OIDs, optionally filtered by coin (#601).
 
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -431,6 +431,17 @@ class HyperliquidExchangeAdapter:
         sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3) if self._info else 3
         return _floor_size(sz, sz_decimals)
 
+    def round_size(self, symbol: str, sz: float) -> float:
+        """Round sz to the asset's lot precision (nearest, not floor).
+
+        Use this to normalize a virtual qty that may have drifted just below a
+        lot boundary due to float64 subtraction in Go (e.g. 0.011 - 0.010 =
+        0.0009999...).  place_stop_loss already uses round(); this makes TP
+        tier sizing consistent.
+        """
+        sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3) if self._info else 3
+        return round(sz, sz_decimals)
+
     def open_order_oids(self, symbol: str | None = None) -> set[int]:
         """Return currently open order OIDs, optionally filtered by coin (#601).
 
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -427,7 +427,9 @@ class HyperliquidExchangeAdapter:
         )
 
     def floor_size(self, symbol: str, sz: float) -> float:
-        """Return sz floored to the asset's lot precision — mirrors place_take_profit_limit."""
+        """Exposes the same lot-precision flooring `place_take_profit_limit`
+        applies internally, so callers can pre-compute the on-chain size each
+        tier will occupy and absorb the remainder into a final tier."""
         sz_decimals = self._info.asset_to_sz_decimals.get(symbol, 3) if self._info else 3
         return _floor_size(sz, sz_decimals)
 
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude fix suggested issues

richkuo:
@claude re-review

Issue #561, opened 2026-05-03
feat: position line quantity precision + ATR in position summary and trade DMs


richkuo:
## Problem

**1. Quantity precision**
Position lines display raw float precision:
```
hl-temac-eth-30 LONG ETH x0.4275141768103672 @ $2,339.10 (-$1.18) ...
```
The `%g` verb in `collectPositions` (discord.go:1031) passes the full float through. Should be capped at 3 decimal places: `x0.428`.

**2. ATR missing from position line**
`pos.EntryATR` is available on the `Position` struct and already used to compute TP1/TP2 hints, but the ATR value itself is never shown. Operators can't easily verify whether the ATR-based TP targets are reasonable without knowing the underlying ATR.

**3. ATR + TP1/TP2 + SL missing from trade DMs; layout improvements**
`FormatTradeDM` / `FormatTradeDMPlain` (discord.go:1050) currently emit:
```
🟢 TRADE EXECUTED
Strategy: hl-tema-eth-live (Hyperliquid perps)
ETH — LONG 0.428 @ $2,336
Value: $999 | Mode: live
```
Issues:
- ATR, TP1, TP2, SL are never shown even when the strategy has them.
- `Mode:` is buried on the value line — it belongs on the header so the live/paper distinction is immediately visible.
- `Value:` is on its own line below symbol/price; it reads more naturally on the same line as the trade.

### discord.go — `collectPositions`

1. Format quantity with `%.3f` instead of `%g` (line 1031).
2. When `pos.EntryATR > 0`, append `| ATR: $<value>` before the TP1/TP2 block.

Example after:
```
hl-temac-eth-30 LONG ETH x0.428 @ $2,339.10 (-$1.18) | ATR: $9.20 | TP1: $2,348.30 (+0.4%) | TP2: $2,357.50 (+0.8%) | SL: $2,295.50 (-1.87%) [May 03 22:05]
```

### Trade struct + FormatTradeDM

`Trade` has no `EntryATR` or `StopLossTriggerPx` fields today. Two options were considered:

[...]

PR #631, opened 2026-05-06
fix: persist SL trigger px on protection-sync stamp (#625)

### LINKED ISSUE #625: Trade alert missing SL price — protection sync doesn't re-stamp TradeHistory
## Summary

When the protection-sync step places a stop-loss on exchange post-open, `pos.StopLossTriggerPx` is correctly updated — but `stampOpenTradeFromPosition` is never called to propagate that value back into `TradeHistory` (in-memory) or the SQLite `trades` row. The trade alert reads `t.StopLossTriggerPx`, which remains 0, so it shows no SL price; analytics/TradingView export reading from SQLite are likewise missing the value.

## Root Cause

After `applyHyperliquidProtectionSync` updates `pos.StopLossTriggerPx` (`hyperliquid_protection.go:334`), there is no matching call to `stampOpenTradeFromPosition` to backfill that value into the strategy's trade history.

The stamp IS called after `RecordTrade` (`main.go:2643`), but that path only runs when `trades > 0` on the execute step. When the SL is placed by the protection sync instead (ATR-based or tiered TP), the execute path leaves `t.StopLossTriggerPx = 0` and the protection sync never corrects it.

## Affected Flow

1. `runHyperliquidExecute` → opens position, no SL placed (`sl_pct=0.00` at execute time)
2. `stampEntryATRIfOpened` → stamps position ATR (but SL was never placed, so nothing to stamp for SL)
3. `runHyperliquidProtectionSync` → places SL + TP1 + TP2 on exchange, updates `pos.StopLossTriggerPx`
4. **Missing**: `stampOpenTradeFromPosition(stratState, stateDB, symbol, pos)` — neither in-memory TradeHistory nor SQLite `trades.stop_loss_trigger_px` is updated

## Fix

Partially addressed by 4ddbd5f, which added the in-memory stamp with `db=nil`. That fixes the trade alert but leaves `trades.stop_loss_trigger_px` at 0 in SQLite, inconsistent with #561/#624 which populate that column for analytics and TradingView export.

Remaining work:

1. Add a `*StateDB` parameter to `runHyperliquidProtectionSync`.
2. Update the three call sites in `main.go` (around lines 1490, 1520, 1560) to pass `stateDB`.
3. Replace the `nil` in the `stampOpenTradeFromPosition` call with the threaded `db`.

`stampOpenTradeFromPosition` already invokes `UpdateTradeStampedFields` when `db != nil` and is idempotent (only writes when target field is 0), so no helper changes are needed.

## Files Affected

- `scheduler/hyperliquid_protection.go` — `runHyperliquidProtectionSync` signature; pass `db` into `stampOpenTradeFromPosition`
- `scheduler/main.go` — three call sites that invoke `runHyperliquidProtectionSync`

## Reproduction

1. Open a HL perps position with no `stop_loss_pct` configured (so execute skips SL placement) but with `stop_loss_atr_mult` or tiered-TP enabled.
2. Protection sync places the SL using ATR-based protection.
3. Trade alert shows: `LONG N @ <price> | ATR: <atr> | TP1: <tp1> | TP2: <tp2>` — no SL price (fixed by 4ddbd5f).
4. Inspect `state.db`: `SELECT stop_loss_trigger_px FROM trades WHERE strategy_id=... ORDER BY ts DESC LIMIT 1;` — still 0 (this is what remains).
5. Exchange shows the SL correctly placed; `pos.StopLossTriggerPx` is correct in memory.

richkuo:
## Summary

Threads `*StateDB` into `runHyperliquidProtectionSync` so the `stampOpenTradeFromPosition` call after `applyHyperliquidProtectionSync` updates `trades.stop_loss_trigger_px` in SQLite, not just in-memory `TradeHistory`.

4ddbd5f added the in-memory stamp with `db=nil` — that fixed the trade alert but left the SQLite row at 0, inconsistent with #561/#624 which populate that column for analytics and TradingView export.

## Changes

- `scheduler/hyperliquid_protection.go`: add `db *StateDB` parameter; pass to `stampOpenTradeFromPosition`.
- `scheduler/main.go`: pass `stateDB` at the three call sites (open-no-trade, post-trade, manual).
- `scheduler/hyperliquid_protection_test.go`: update three existing call sites; add regression test `TestRunHyperliquidProtectionSyncStampsTradeInDB` that asserts SQLite `stop_loss_trigger_px` is backfilled.

## Test plan
- [x] `go build` and full `go test ./...` pass in `scheduler/`.
- [x] New regression test asserts SQLite row updated, not just `TradeHistory`.

Closes #625

---

### DIFF
--- a/scheduler/hyperliquid_protection.go
+++ b/scheduler/hyperliquid_protection.go
@@ -302,6 +302,7 @@ func applyHyperliquidProtectionSync(pos *Position, result *HyperliquidProtection
 func runHyperliquidProtectionSync(
 	sc StrategyConfig,
 	stratState *StrategyState,
+	db *StateDB,
 	symbol string,
 	mu *sync.RWMutex,
 	notifier *MultiNotifier,
@@ -336,7 +337,7 @@ func runHyperliquidProtectionSync(
 	// by the protection sync (#625). Without this, execute-path SL=0 leaves the
 	// trade's StopLossTriggerPx unset even though the sync correctly populated
 	// pos.StopLossTriggerPx — the alert then shows no SL price.
-	stampOpenTradeFromPosition(stratState, nil, symbol, pos)
+	stampOpenTradeFromPosition(stratState, db, symbol, pos)
 	if logger != nil {
 		logger.Info("%s (sl_oid=%d tp_oids=%v)", logTag, pos.StopLossOID, pos.TPOIDs)
 	}
--- a/scheduler/hyperliquid_protection_test.go
+++ b/scheduler/hyperliquid_protection_test.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"sync"
 	"testing"
+	"time"
 )
 
 func TestBuildHyperliquidProtectionPlanUsesDefaultTieredATR(t *testing.T) {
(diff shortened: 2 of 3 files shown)


richkuo:
@claude

PR #633, opened 2026-05-06
feat: place SL and TP[n] orders inline on manual-open

### LINKED ISSUE #632: manual-open: place SL and TP[n] orders inline at open time
## Problem

When `go-trader manual-open` is invoked, the SL trigger is placed immediately after the fill (`RunHyperliquidUpdateStopLoss`, manual.go:214), but TP[n] reduce-only orders are deferred until the next scheduler cycle when `runHyperliquidProtectionSync` runs against the materialized `Position`.

This leaves a window — potentially minutes depending on `interval_seconds` — where the position is live with SL but no TP orders on-chain. If `--atr` is omitted entirely and the strategy uses ATR-based protection (the default for `type=manual`), both SL and TPs are skipped, leaving the position fully naked.

## Expected behavior

`manual-open` should place SL **and** all configured TP[n] orders inline immediately after the fill, before returning.

If `--atr` is not provided, a fallback ATR should be computed from the fill so protection still posts, with a warning sent via the notifier.

## Fallback ATR when `--atr` is omitted

```
fallbackATR = 0.1 × fillPrice / leverage
```

This is leverage-aware: it represents a price move that risks 10% of deployed margin at SL (1× ATR) and yields 20% of margin if TP2 (2× ATR) closes the full remaining position.

**Example — $100 margin, 20× leverage, ETH @ $2,500, qty = 0.8 ETH:**

| Order | Price | P&L |
|-------|-------|-----|
| SL (1× ATR) | $2,487.50 | −$10 = −10% of margin |
| TP1 (1× ATR, 50%) | $2,512.50 | +$5 = +5% of margin |
| TP2 (2× ATR, remaining 50%) | $2,525.00 | +$10 = +10% of margin |

Full close at both TPs nets +$15 = +15% of margin. SL closes at −$10 = −10% of margin (isolated margin assumed).

When the fallback is used, a warning must fire through the notifier (Discord/Telegram), not just stderr.

## Implementation sketch

After the fill and SL arm in `runManualOpen` (manual.go):

1. If `entryATR == 0` and `effectiveATRMult > 0`, compute `fallbackATR = 0.1 * resolvedFillPrice / sc.Leverage`, set `entryATR = fallbackATR`, and fire a notifier warning.
2. Call `RunHyperliquidSyncProtection` inline with fill price, qty, ATR, tier config, and the just-placed SL OID. Requires `strategyUsesTieredTPATRClose(sc)` to be true.
3. Plumb the returned TP OIDs through `PendingManualAction` (new `TPOIDs []int64` field, parallel to the existing `StopLossOID`) so `applyManualAction` sets `pos.TPOIDs` correctly on drain.

Guards:
- Skip inline TP placement if `strategyUsesTieredTPATRClose(sc)` returns false (no tiers configured).
- `--record-only` skips both SL and TP placement (operator is responsible for on-chain orders).
- The subsequent scheduler-cycle protection sync is idempotent — it will no-op if OIDs are already present and valid.

## Notes

- `buildHyperliquidProtectionPlan` currently requires a `*Position`, but the inline call can bypass it and call `RunHyperliquidSyncProtection` directly with the known values.
- Isolated margin is assumed for the fallback ATR math; cross margin users should always pass `--atr` explicitly.

---

richkuo:
## Summary

- `manual-open` now places SL **and** TP[n] reduce-only orders inline at fill time, before returning. Previously TPs were deferred to the next scheduler cycle.
- When `--atr` is omitted on a strategy that uses ATR-based protection, a leverage-aware fallback `0.1 * fillPrice / leverage` is used. This risks ~10% of margin at SL (1× ATR). A warning is sent via notifier (owner DM + all channels) and stderr.
- `buildNotifierFromConfig` extracted from `main()` into `notifier_build.go`; `main()` now calls it directly — single source of truth for backend assembly.
- `PendingManualAction.TPOIDs []int64` added with `tp_oids_json` DB column + idempotent migration; `applyManualAction` stamps `pos.TPOIDs` so the scheduler-cycle protection sync is a no-op on already-placed OIDs.

**Worked example** — $100 margin, 20× leverage, ETH @ $2500, qty = 0.8 ETH, fallback ATR = $12.50 (default `type=manual` tiers: TP1 @ 2× ATR, TP2 @ 3× ATR):

| Order | Price | P&L |
|-------|-------|-----|
| SL (1× ATR) | $2,487.50 | −$10 = −10% of margin |
| TP1 (2× ATR, 50%) | $2,525.00 | +$10 = +10% of margin |
| TP2 (3× ATR, remaining 50%) | $2,537.50 | +$15 = +15% of margin |

Full close at both TPs nets +$25 = +25% of margin.

## Test plan

- [ ] `go test ./...` passes (6 new tests: fallback ATR table, no-tiers skip, TP error surfacing, nil notifier guard, DB round-trip, drain stamps pos.TPOIDs)
- [ ] `manual-open` with explicit `--atr` → SL + TP OIDs printed immediately
- [ ] `manual-open` without `--atr`, leverage=20, ETH @ $2500 → fallback ATR=12.5 warning fires via Discord; SL at $2487.50 + TPs at $2525.00/$2537.50 posted
- [ ] `manual-open --record-only` → no SL or TP placed (existing behavior preserved)
- [ ] After scheduler cycle drains: `pos.TPOIDs` matches printed OIDs; `runHyperliquidProtectionSync` no-ops

Closes #632

---
LLM: Claude Opus 4.7 (1M) | high | Harness: Claude Code

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -210,6 +210,7 @@ CREATE TABLE IF NOT EXISTS pending_manual_actions (
     entry_atr REAL NOT NULL DEFAULT 0,
     realized_pnl REAL NOT NULL DEFAULT 0,
     is_full_close INTEGER NOT NULL DEFAULT 0,
+    tp_oids_json TEXT NOT NULL DEFAULT '',
     created_at TEXT NOT NULL
 );
 `
@@ -302,6 +303,8 @@ func (sdb *StateDB) migrateSchema() error {
 		"ALTER TABLE positions ADD COLUMN tp2_oid INTEGER NOT NULL DEFAULT 0",
 		// Variable-length per-strategy HL reduce-only take-profit OIDs (#612).
 		"ALTER TABLE positions ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
+		// Inline TP OIDs for manual-open actions so the scheduler drain sets pos.TPOIDs (#632).
+		"ALTER TABLE pending_manual_actions ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
 	}
 	for _, ddl := range migrations {
 		if _, err := sdb.db.Exec(ddl); err != nil {
@@ -1521,7 +1524,8 @@ type PendingManualAction struct {
 	StopLossTriggerPx float64
 	EntryATR          float64
 	RealizedPnL       float64
-	IsFullClose       bool // close-only: operator/scheduler intent flag (avoids tolerance heuristics on the drain side)
+	IsFullClose       bool    // close-only: operator/scheduler intent flag (avoids tolerance heuristics on the drain side)
+	TPOIDs            []int64 // open-only: TP OIDs placed inline at manual-open time (#632)
 	CreatedAt         time.Time
 }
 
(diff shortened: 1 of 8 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #635, opened 2026-05-06
fix: cleanup orphaned manual-open orders if queue insert fails

### LINKED ISSUE #634: manual-open: orphaned on-chain orders if DB insert fails after fill
## Problem

In `runManualOpen` (manual.go), on-chain orders are placed in this sequence:

1. Fill (market order)
2. SL trigger order
3. TP[n] reduce-only orders
4. `InsertPendingManualAction` → SQLite

If step 4 fails (disk full, DB locked), real HL orders exist but the scheduler has no virtual position record. The next cycle's `reconcileHyperliquidAccountPositions` sees an unowned on-chain position and may not handle it correctly.

This was a pre-existing gap for the SL alone (#569); PR #633 amplified it by adding N TP orders to the inline placement path.

## Fix sketch

Move `InsertPendingManualAction` to before the on-chain placements, or on DB insert failure attempt to cancel the placed SL/TP OIDs as a cleanup step. The former is simpler but introduces its own ordering issue if the fill itself fails.

---

richkuo:
Closes #634

When the manual-open CLI placed the on-chain fill, SL, and TP[n] orders successfully but `InsertPendingManualAction` then failed (disk full / DB locked), real on-chain orders existed with no virtual position record. The next reconcile cycle would see an unowned on-chain position with orphaned reduce-only triggers.

On insert failure the CLI now submits a reduce-only market close sized to `fillQty` that simultaneously cancels the SL + TP OIDs in one shot. Sized to `fillQty` so peer positions on the same coin are preserved. Skipped in `--record-only` mode (operator owns the pre-existing fill). If cleanup itself fails, a MANUAL INTERVENTION REQUIRED notification fires with enough detail to flatten via the HL UI.

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 59 in / 21253 out | Cache: 2904856 read / 202276 written | Cost: $3.25

### DIFF
--- a/scheduler/manual.go
+++ b/scheduler/manual.go
@@ -283,7 +283,26 @@ func runManualOpen(args []string) int {
 		CreatedAt:         time.Now().UTC(),
 	}
 	if err := stateDB.InsertPendingManualAction(action); err != nil {
-		fmt.Fprintf(os.Stderr, "error queuing action: %v\n", err)
+		// On-chain fill (and SL/TPs) succeeded but the queue insert failed —
+		// the scheduler will never adopt this position, so reconcile would see
+		// an unowned on-chain position with orphaned reduce-only triggers.
+		// Skip cleanup in --record-only because the operator's pre-existing
+		// fill is theirs to manage; we never placed those on-chain orders.
+		if *recordOnly {
+			fmt.Fprintf(os.Stderr, "error queuing action: %v\n", err)
+			return 1
+		}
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet fix the pr findings

richkuo:
@claude sonnet re-review

Issue #641, opened 2026-05-07
refactor(backtest): align Backtester API with co-located strategy refs


richkuo:
## Summary

Follow-up to #640. Once the live config model moves to co-located strategy refs (`{name, params}` per strategy), the Backtester API should match so configs are portable between live and backtest without reshaping.

## Current Backtester API

`backtest/backtester.py` takes parallel arguments:

```python
Backtester(
    close_strategies=["tiered_tp_atr"],
    close_params={"tiered_tp_atr": {"tiers": [...]}},
)
```

And `run_backtest.py` exposes them as separate CLI flags:

- `--close-strategy NAME` (repeatable)
- `--close-params '{"NAME": {...}}'`

This is the map-keyed shape: name list in one field, params map keyed by name in another. It works, but diverges from the live config model after #640 lands.

There is no `open_strategy`/open params concept yet on the Backtester — the strategy name is the positional `--strategy` arg and the registry default params are used unmodified.

## Target shape

After #640, live config blocks look like:

```json
{
  "open_strategy": {"name": "tema_cross_bd", "params": {"short_period": 5}},
  "close_strategies": [
    {"name": "tiered_tp_atr", "params": {"tiers": [...]}}
  ]
}
```

Backtester should accept the same shape:

```python
Backtester(
    open_strategy={"name": "tema_cross_bd", "params": {"short_period": 5}},
    close_strategies=[
        {"name": "tiered_tp_atr", "params": {"tiers": [...]}},
    ],
)
```

CLI:

- `--open-strategy '{"name":"...", "params":{...}}'` (or `--open-strategy NAME --open-params JSON` as a thinner alternative)
- `--close-strategy '{"name":"...", "params":{...}}'` (repeatable, replaces the parallel `--close-strategy` + `--close-params` pair)

## Implementation areas

- `backtest/backtester.py` — `Backtester.__init__` accepts co-located refs; merge with registry defaults at run time
- `backtest/run_backtest.py` — collapse `--close-strategy` + `--close-params` into a single repeatable JSON flag; add `--open-strategy` (or keep positional + new `--open-params`)
- `backtest/optimizer.py` — same shape for any optimizer that wraps Backtester
- `backtest/run_backtest.py` (or a config-loader helper) — accept a JSON file matching the live `StrategyConfig` shape so an operator can backtest the exact live config without translation
- tests covering:
  - co-located close refs round-trip through Backtester
  - registry-default merge for both open and close
[...]

PR #642, opened 2026-05-07
refactor: co-located strategy refs (name + params) for open/close split

### LINKED ISSUE #640: refactor: co-located strategy refs (name + params) for open/close split configs
## Summary

Open/close split Hyperliquid perps strategies currently have no way to provide **close-strategy-specific params**. As a result, when an operator tries to customize `tiered_tp_atr` TP levels via `params.tiers`, that close-only config is passed into the **open** strategy too and crashes signal evaluation for strategies that do not accept `**kwargs`.

This showed up with:

- `open_strategy: tema_cross_bd`
- `close_strategies: ["tiered_tp_atr"]`
- `params.tiers = [{atr_multiple: 2, close_fraction: 0.5}, {atr_multiple: 3, close_fraction: 1.0}]`

The next signal-check cycle fails with:

```text
TypeError: tema_cross_bd_strategy() got an unexpected keyword argument 'tiers'
```

## Why this matters

The current behavior does not just cause a crash — it also exposes a missing configuration capability.

Operators need to be able to say:

- open strategy params → go only to the open strategy
- close strategy params → go only to the close strategy / protection logic

Right now `StrategyConfig.Params` is effectively treated as open-strategy params, while close strategies in an open/close split mostly run with defaults unless code is added for a separate param path.

So this is not just a `tiers` filtering bug. It is a config-model gap for open/close split strategies.

## Root Cause

In `shared_scripts/check_hyperliquid.py`, `strategy_params_override` is merged into `strategy_params` before signal evaluation:

```python
if strategy_params_override:
    merged = {**strategy_params_override, **strategy_params}
    strategy_params = merged
```

That merged dict is then passed into `evaluate_open_close(...)`, and the open strategy is run with those params:

```python
open_result = run(open_name, params)
```

So for a config like:

```json
"params": {
  "tiers": [
    {"atr_multiple": 2.0, "close_fraction": 0.5},
    {"atr_multiple": 3.0, "close_fraction": 1.0}
  ]
}
```

`tema_cross_bd_strategy(df, tiers=[...])` is invoked and crashes.

Separately, `shared_tools/strategy_composition.py` currently documents the same limitation:

```python
# Until per-close params blocks exist, only the implicit-self close
# shares the open strategy's params. Distinct close strategies run with
# their own defaults so open-only params do not break another function.
```

That means the system already implicitly knows per-close params are missing; `tiers` just makes the gap user-visible.

## Reproduce

Configure a `type=perps` Hyperliquid strategy like:

```json
{
  "id": "hl-temacb-btc",
  "type": "perps",
  "platform": "hyperliquid",
  "args": ["tema_cross_bd", "BTC", "1h", "--mode=paper"],
  "open_strategy": "tema_cross_bd",
  "close_strategies": ["tiered_tp_atr"],
  "params": {
    "tiers": [
      {"atr_multiple": 2, "close_fraction": 0.5},
      {"atr_multiple": 3, "close_fraction": 1.0}
    ]
  }
}
```

On the next cycle, the scheduler invokes the Python check script and the open strategy crashes with:

```text
TypeError: tema_cross_bd_strategy() got an unexpected keyword argument 'tiers'
```

## Fix direction

**Co-locate strategy name and params** in a single block per strategy. `open_strategy` and `close_strategies` change shape — name and params live together, no separate map lookup:

```json
{
  "open_strategy": {
    "name": "tema_cross_bd",
    "params": {
      "short_period": 5,
      "mid_period": 13,
      "long_period": 34
    }
  },
  "close_strategies": [
    {
      "name": "tiered_tp_atr",
      "params": {
        "tiers": [
          {"atr_multiple": 2.0, "close_fraction": 0.5},
          {"atr_multiple": 3.0, "close_fraction": 1.0}
        ]
      }
    }
  ]
}
```

Go shape:

```go
type StrategyRef struct {
    Name   string                 `json:"name"`
    Params map[string]interface{} `json:"params,omitempty"`
}

type StrategyConfig struct {
    // ...
    OpenStrategy     StrategyRef   `json:"open_strategy"`
    CloseStrategies  []StrategyRef `json:"close_strategies,omitempty"`
    // Params field removed.
}
```

### Why this shape

- Everything for one strategy lives in one place — no two-step lookup
- Naturally namespaced: open and close registries can never collide
- Self-describing: a config block tells you exactly what runs and how
- An empty `params` (or no `params` key) means "use registry defaults" — no ambiguity

### Trade-offs

- Bigger schema change than a map-keyed approach (replaces three fields instead of one)
- Backtester API today uses map-keyed `close_params={...}` — follow-up tracked in #641

## Migration plan

Defaults exist at the registry level on both sides, so a clean structural migration is possible — no operator intervention, no config-DM prompt. **Legacy `params` and string-typed `open_strategy` / `close_strategies` are removed at the version bump; no backwards-compat fallback.**

**Open params** — each strategy carries its own defaults in `shared_strategies/open/registry.py` (e.g., `tema_cross_bd` → `{"short_period": 5, "mid_period": 13, "long_period": 34}`). Merged in `apply_strategy`: `{**strat["default_params"], **(params or {})}`.

**Close params** — each close strategy carries its own defaults in `shared_strategies/close/registry.py`:

```python
register("tiered_tp_atr",      ..., {"tiers": list(DEFAULT_ATR_TIERS)})
register("tiered_tp_atr_live", ..., {"tiers": list(DEFAULT_ATR_TIERS), "atr_source": "live"})
register("tiered_tp_pct",      ..., {"tiers": list(DEFAULT_PCT_TIERS)})
```

`DEFAULT_ATR_TIERS = ({1×, 0.5}, {2×, 1.0})` — same canonical default the Go-side `hyperliquidProtectionTiers` falls back to. The migration uses the close registry's `default_params` as the source of truth for which legacy `Params` keys belong to which close strategy — no hardcoded key list.

### Steps

1. Add `StrategyRef` struct. Change `OpenStrategy` from `string` to `StrategyRef`. Change `CloseStrategies` from `[]string` to `[]StrategyRef`. Drop `Params`.
2. On `LoadConfig`, when the persisted config is at the previous version:
   - Resolve the open strategy name: prefer legacy `OpenStrategy` (string) if non-empty, else `args[0]`. Build `OpenStrategy = StrategyRef{Name: <name>}`.
   - Build a `StrategyRef` for each name in legacy `CloseStrategies []string`.
   - Walk legacy `Params`: for each key, if any close strategy's registry `default_params` contains that key, move it into that close ref's `Params`. Otherwise leave it for the open ref.
   - Assign remaining legacy `Params` to `OpenStrategy.Params`.
   - Delete legacy `Params`, legacy string-typed `OpenStrategy` / `CloseStrategies`.
3. Bump `CurrentConfigVersion` and persist the migrated config back to disk.
4. `hyperliquidProtectionTiers` reads from the matching `CloseStrategies[i].Params["tiers"]`, falling back to the canonical default `[{1×, 0.5}, {2×, 1.0}]`.
5. Thread per-strategy params through each `run*Check` site; Python check scripts forward to `evaluate_open_close`.
6. Extend `evaluate_open_close` to accept the open + close refs (name + params each) directly, instead of separate name list and params dict.

The migration is unambiguous because each legacy `Params` key has at most one owner — either no close strategy claims it (→ open) or exactly one close strategy's registry default mentions it (→ that close).

## Expected behavior after fix

For open/close split strategies:

- open strategy receives only its own params (`open_strategy.params`)
- each close evaluator / Hyperliquid protection tier builder receives only its own params (`close_strategies[i].params`)
- custom ATR TP tiers work without crashing signal evaluation
- legacy `params` field and string-typed strategy refs are gone — configs validate cleanly under the new schema after migration

## Suggested implementation areas

- `scheduler/config.go` — add `StrategyRef`, change `OpenStrategy` and `CloseStrategies` types, drop `Params`, add migration logic, version bump
- `scheduler/config_migration.go` — version bump entry (no DM prompt needed since migration is structural)
- `scheduler/hyperliquid_protection.go` — read tier config from `CloseStrategies[i].Params`
- `scheduler/main.go` + `executor.go` — thread per-strategy params to each check script
- `scheduler/config_reload.go` — SIGHUP allowlist for the new shape
- `shared_scripts/check_hyperliquid.py` (and other check scripts using open/close split) — parse and forward per-strategy params
- `shared_tools/strategy_composition.py` — extend `evaluate_open_close` to accept refs instead of name list + flat params
- `backtest/` — align Backtester API with co-located refs (follow-up, tracked in #641)
- tests covering:
  - migration of legacy flat `params` (incl. close-only `tiers`) → co-located refs
  - migration handles `open_strategy` empty / falls back to `args[0]`
  - open/close split strategy with close-only `tiers`
  - no `tiers` leakage into open strategy
  - custom TP tiers reaching `tiered_tp_atr` / protection logic

## Follow-ups

- #641 — align the Backtester API with the co-located ref shape so live configs are portable into backtests without translation.

## Notes

- This issue was triggered while trying to change TP tiers from the defaults (`1x`, `2x`) to custom values (`2x`, `3x`) on `tema_cross_bd` + `tiered_tp_atr` Hyperliquid paper strategies.
- `type=manual` strategies are a separate case; they already have special handling and do not represent the general open/close split config model.

richkuo:
Closes #640.

## Summary

- Replace the flat `StrategyConfig.Params` + string-typed `open_strategy` / `[]string close_strategies` with co-located `StrategyRef` (`{name, params}`) on each role
- Add v12 → v13 migration that splits a legacy flat `params` map between the open ref and each close ref using each close strategy's registry `default_params` (no hardcoded `tiers` list — driven by `closeStrategyOwnedKeys`)
- New `--strategy-refs` JSON CLI flag in every check script supersedes the old `--params` / `--open-strategy` / `--close-strategies` trio; `evaluate_open_close` gains `close_params_by_name` so each close evaluator receives only its own params

This fixes the bug where setting `params.tiers` on a perps strategy with `tiered_tp_atr` close caused `tema_cross_bd_strategy(df, tiers=...)` → `TypeError`. After this change, `tiers` lives on the matching close ref and never reaches the open evaluator.

## Migration

Synchronous inside `LoadConfig` — runs before `json.Unmarshal` when it sees `config_version < 13`, then re-reads the rewritten file. No DM prompt; the structural rewrite is unambiguous because each legacy `params` key has at most one owner (no close strategy claims it → open; one close strategy's `default_params` mentions it → that close).

When the legacy `open_strategy` field is empty, the migration falls back to `args[0]` so configs that relied on the positional script arg keep working.

## Test plan

- [x] `go test ./...` — 21s, all pass
- [x] `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 858 pass
- [x] New `TestMigrateV13StrategyShape`: legacy flat config (incl. close-only `tiers`) → co-located refs with `tiers` correctly routed to the close ref
- [x] New `TestBuildStrategyRefsArg`: builds correct `--strategy-refs` JSON shape
- [x] `TestAppendOpenCloseArgs*`: position-context flags unchanged; legacy `--open-strategy`/`--close-strategies` no longer emitted
- [x] `TestCloseStrategiesSuppressedMatchesTieredTPATRClose`: HL on-chain protection filter works on the new ref shape
- [ ] Smoke run on a v12 config: confirm migration writes correct v13 shape and scheduler runs cleanly

## Follow-ups

- #641 — align `Backtester` API with co-located refs so live configs are portable into backtests without translation

---


richkuo:
@claude

richkuo:
@claude re-review

richkuo:
@claude sonnet re-review

PR #648, opened 2026-05-08
Atomic update script as single source of truth (#647)

### LINKED ISSUE #647: External/asymmetric deploy can break Python-backed check scripts after #642
## Problem

A runtime version mismatch can break Python-backed strategy checks when a deployment path updates the `go-trader` binary without updating the corresponding Python runtime from the same version.

During the observed failure window, the running binary passed `--strategy-refs` to local Python check scripts that did not support it, producing errors like:

```text
check_hyperliquid.py: error: unrecognized arguments: --strategy-refs {...}
```

This stopped strategy signal processing for the affected runtime until the binary and scripts were brought back into sync.

## Confirmed facts

- The failure mode was real: the running binary passed `--strategy-refs`, while local scripts on disk did not accept it.
- Repo HEAD already supports `--strategy-refs` end-to-end. This is **not** a current HEAD bug in `check_hyperliquid.py`.
- The exposure is **not Hyperliquid-only**. `--strategy-refs` is wired into multiple Python-backed check scripts, including:
  - `shared_scripts/check_hyperliquid.py`
  - `shared_scripts/check_okx.py`
  - `shared_scripts/check_robinhood.py`
  - `shared_scripts/check_topstep.py`
  - `shared_scripts/check_strategy.py`
- The in-tree updater is **not** the direct culprit here. `scheduler/updater.go` aborts on `git pull --ff-only` failure before rebuild, so that path should not create this mismatch by itself.
- One important hardening step is **already shipped**: the startup compatibility probe from #645/#646 (`b100f36`) invokes each unique configured check script with `--probe-only` and the runtime argv shape (including `--strategy-refs`) and exits on mismatch. That prevents this class of desync from silently running going forward.

## Remaining gap

The remaining exposure is specifically **external or asymmetric deploy paths** — for example, any path that copies/rebuilds the binary independently of a matching `git pull` / runtime update for `shared_scripts/` and related Python files.

## Impact

- Affected Python-backed checks crash on every tick until runtime versions are re-aligned.
- No new entries/exits are processed for the affected strategies during the mismatch window.
- This can silently miss market moves if the failure is only noticed from logs or summary behavior.

## What should still be fixed

The deployment/update path should guarantee that the `go-trader` binary and Python check scripts are deployed **atomically from the same version** outside the in-tree updater as well.

Open hardening items:
- standardize external rebuild/update commands so full runtime sync is part of the flow
- ensure external deploy paths treat binary + Python runtime as one deploy unit
- document that copying/rebuilding the binary alone is unsafe after #642

richkuo:
## Summary
- New `scripts/update.sh`: `git pull --ff-only` → `uv sync` → `go build` (with version stamp) under `set -euo pipefail`. Optional restart via `--restart` or `RESTART=1`.
- `applyUpgrade` (auto-update DM path) calls the script without `--restart`, retaining its own state-save + `restartSelf()` ordering. `formatUpdateMessage` operator DM now points at `bash scripts/update.sh --restart` instead of a multi-line snippet that omitted `uv sync`.
- Closes the prevention half of #647: the startup probe (#646) detects argv drift, this script makes drift hard to produce by deploying binary + Python from the same SHA atomically.
- Drops now-unused `goChanged`/`gitDiffHasFiles`/`resolveBuildVersion` (script handles version stamping itself).

## Test plan
- [x] `go -C scheduler test ./...` — full suite passes
- [x] `go -C scheduler build` — binary builds cleanly
- [x] `bash scripts/update.sh --help` and `--bogus` arg handling
- [x] Updated `updater_test.go` for new `formatUpdateMessage` signature
- [ ] Smoke-test on prod: trigger auto-update via Discord DM, confirm script output appears in DM and binary restarts cleanly
- [ ] Confirm `uv sync` runs as expected when `pyproject.toml` / `uv.lock` are dirty

Closes #647

---


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #653, opened 2026-05-08
fix: rebaseline portfolio peak after strategy prune (#650)

### LINKED ISSUE #650: Strategy prune does not re-baseline portfolio peak — immediate kill switch trigger on restart
## Summary

When strategies are pruned from the config and the service restarts, the portfolio peak value is not re-baselined to match the new (lower) portfolio value. This causes an immediate portfolio drawdown kill switch trigger on the very first risk check cycle.

## Root Cause

In `scheduler/main.go`, the strategy prune logic (lines 133-137) deletes removed strategies from `state.Strategies` but does **not** adjust `state.PortfolioRisk.PeakValue`:

```go
// Prune strategies from state that are no longer in config
configIDs := make(map[string]bool)
for _, sc := range cfg.Strategies {
    configIDs[sc.ID] = true
}
for id := range state.Strategies {
    if !configIDs[id] {
        delete(state.Strategies, id)
        fmt.Printf("  Pruned stale strategy: %s\n", id)
    }
}
```

Later, peak is only initialized when it is **zero** (line 150):
```go
if state.PortfolioRisk.PeakValue == 0 {
    total := computeInitialPortfolioPeak(cfg.Strategies, nil)
    state.PortfolioRisk.PeakValue = total
}
```

Since the peak is a non-zero stored value from the prior run, the initialization is skipped. The peak remains at the old high value while the current portfolio value (sum of only the remaining strategies) is significantly lower, creating a large apparent drawdown.

## Reproduction

1. Run with N strategies, portfolio peak builds to $X
2. Remove/prune M strategies (config change)
3. Restart — service loads with (N-M) strategies, current value = Y where Y << X
4. First `CheckPortfolioRisk` call computes drawdown = (X - Y) / X
5. If drawdown exceeds `max_drawdown_pct` (e.g. 25%), kill switch fires immediately

Example from `go-trader-paper-testing`:
- Peak: $15,148.90
- Current value: $9,034.24
- Drawdown: 40.4% (limit: 25%)
- Result: `[CRITICAL] Portfolio kill switch: portfolio drawdown 40.4% exceeds limit 25.0%`

## Fix

After pruning strategies, re-baseline `PeakValue` to the current portfolio value. Something like:

```go
// After the prune loop, if any strategies were removed, re-baseline peak
if wasPruned {
    state.PortfolioRisk.PeakValue = currentPortfolioValue
}
```

Alternatively, compute the new peak from the sum of capital for the remaining strategies at startup (similar to `computeInitialPortfolioPeak` but called even when `PeakValue != 0` when strategies were pruned).

Note: A similar mechanism already exists in `ClearLatchedKillSwitchSharedWallet` (risk.go:285-320) which re-baselines peak on startup when a kill switch is latched and shared wallets are involved — but this doesn\'t run when the kill switch hasn\'t latched yet.

## Files Affected

- `scheduler/main.go` — strategy prune logic (lines ~133-137)
- `scheduler/risk.go` — `ClearLatchedKillSwitchSharedWallet` (existing related mechanism)

richkuo:
## Summary

When strategies are pruned from config and the service restarts, `PortfolioRisk.PeakValue` keeps its pre-prune value. The first risk-check cycle compares it against the now-smaller surviving portfolio and immediately latches the kill switch.

The peak-init guard at `scheduler/main.go:150` only fires on `PeakValue == 0`, so the stale value survives. `ClearLatchedKillSwitchSharedWallet` re-baselines for shared wallets only after the latch — too late.

`rebaselinePortfolioPeakAfterPrune` sums surviving strategies' `RiskState.PeakValue` (per-strategy high-water marks already tracked by `CheckRisk`), falls back to configured `Capital` for cold-start strategies, and floors at `computeInitialPortfolioPeak(remaining)` so we never under-baseline. Wired into the prune block in `main.go` when any strategy was removed.

Per-strategy peaks are summed (rather than scaling the old peak by cash ratio or resetting to cap-sum) because they're the most accurate signal already on hand — they preserve accrued gains for surviving strategies without over-attributing them.

## Test plan

- [x] `go test ./...` from `scheduler/` passes (21s)
- [x] New tests in `shared_wallet_test.go`:
  - sums per-strategy peaks of surviving strategies
  - floors at sum-of-capitals when surviving strategies have no recorded peak
  - falls back to capital per-strategy when peak is missing
  - regression: pre-fix kill switch latched, post-fix `CheckPortfolioRisk` allows trading

Closes #650

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -129,10 +129,25 @@ func main() {
 	for _, sc := range cfg.Strategies {
 		configIDs[sc.ID] = true
 	}
+	pruned := false
 	for id := range state.Strategies {
 		if !configIDs[id] {
 			delete(state.Strategies, id)
 			fmt.Printf("  Pruned stale strategy: %s\n", id)
+			pruned = true
+		}
+	}
+
+	// #650: After prune, rebaseline portfolio peak from surviving strategies'
+	// per-strategy peaks. Without this, the stale portfolio peak (sized for the
+	// pre-prune strategy set) can immediately latch the kill switch on the
+	// first risk-check cycle once current value drops to the surviving subset.
+	if pruned && state.PortfolioRisk.PeakValue > 0 {
+		oldPeak := state.PortfolioRisk.PeakValue
+		newPeak := rebaselinePortfolioPeakAfterPrune(state, cfg)
+		if newPeak != oldPeak {
+			state.PortfolioRisk.PeakValue = newPeak
+			fmt.Printf("  Portfolio peak rebaselined after prune: $%.0f -> $%.0f\n", oldPeak, newPeak)
 		}
 	}
 
--- a/scheduler/shared_wallet.go
+++ b/scheduler/shared_wallet.go
@@ -329,3 +329,46 @@ func computeInitialPortfolioPeak(strategies []StrategyConfig, fetcher WalletBala
 	}
 	return total
 }
+
+// rebaselinePortfolioPeakAfterPrune recomputes PortfolioRisk.PeakValue from the
+// per-strategy peaks of remaining strategies after one or more strategies are
+// pruned from config. Without this, the stale portfolio peak (which reflects
+// the pre-prune strategy set) can immediately latch the kill switch on the
+// first risk-check cycle since current portfolio value drops to the sum of
+// only the surviving strategies. See issue #650.
+//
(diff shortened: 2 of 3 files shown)


richkuo:
@claude

PR #660, opened 2026-05-08
fix: drive tiered ATR TP display from config, not hardcoded 1×/2×

richkuo:
Closes #659

## Summary

Discord position extras, Discord trade DM, and Telegram trade DM all hardcoded TP1=1×ATR / TP2=2×ATR — silently correct for default tiers, wrong whenever an operator overrides `tiers` (e.g. 2×/3×). Replace with a shared `tieredTPATRPrices` helper that reads tiers via the same `hyperliquidProtectionTiers(sc)` the on-chain placer uses, and renders TP1…TPn so future N-tier configs no longer truncate to two slots.

## Test plan
- [x] `go test ./...`
- [ ] Live smoke: open a HL perps position under a `tiered_tp_atr_live` strategy with custom `tiers` and verify DM TP prices match on-chain reduce-only orders

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 58 in / 18694 out | Cache: 2826196 read / 195861 written | Cost: $3.1

PR #663, opened 2026-05-08
feat: owner DM on HL TP/SL fill detection

### LINKED ISSUE #661: Feature: DM notification when TP/SL orders fill
## Summary

Request DM notification when a take-profit (TP) or stop-loss (SL) order fills on the exchange.

## Current Behavior

TP/SL fills are detected and booked across two paths, neither of which emits a DM:

1. **Protection sync** (`scheduler/hyperliquid_protection.go`): `applyHyperliquidProtectionSync` consumes `tp_filled_externally` / `stop_loss_filled_externally` from `check_hyperliquid.py` and clears the corresponding OIDs on the in-memory position (`pos.StopLossOID = 0`, `pos.TPOIDs[idx] = 0`). It does NOT adjust quantity or book a trade.

2. **Reconciler** (`scheduler/hyperliquid_balance.go` → `reconcileHyperliquidAccountPositions`): when on-chain qty drift is observed, Detectors 1/2/3 book the close via `recordPerpsStopLossCloseWithFillFee`, `recordPerpsExternalCloseWithFillFee`, or `recordPerpsExternalPartialCloseWithFillFee` (`scheduler/portfolio.go`).

`sendTradeAlerts` is gated on the per-strategy execute loop's local `trades` counter (`main.go:1644-1652`); the reconciler runs in its own phase (`main.go:1064`) and never increments that counter, so reconciler-booked closes bypass DM alerts entirely.

Operators only see these fills indirectly via:
- Portfolio value changes in the next cycle summary
- Log inspection

## Requested Behavior

Send an owner DM when:
1. **Any TP tier fills** (partial or complete)
2. **SL triggers and fills** (full or partial)

DM content should include:
- Strategy ID
- Symbol + side
- Fill type: tier index (e.g. "TP1", "TP2", … sourced from configured tier slot, not hardcoded — see #660) or "SL"
- Fill price
- Remaining position size
- Realized PnL (if available)

Example:
```
🎯 TP1 filled — hl-tema-eth-live
ETH LONG — 0.216 @ $2340.90
Remaining: 0.216 ETH
```

## Motivation

Currently there's no real-time notification when protective orders execute. Operators may not notice a position has partially or fully closed until the next scheduled cycle summary arrives (up to 5 minutes later). For live trading, immediate awareness of SL/TP fills is important.

## Scope

This would be a **new feature** — not a bug fix. Single emission site at the reconciler's three `record*` call sites (avoids double-fire from the protection-sync OID-clear path, which doesn't book trades anyway).

Likely touches:
- `scheduler/hyperliquid_balance.go` — where the three `record*` calls live (Detector 1/2/3); the DM hook fires here
- `scheduler/portfolio.go` — the `recordPerps{StopLoss,External,ExternalPartial}CloseWithFillFee` helpers themselves (alternative hook site if we want one shared emission)
- `scheduler/discord.go` / `scheduler/telegram.go` — DM formatting
- `scheduler/notifier.go` — notification routing (likely a new `MultiNotifier.SendProtectionFillAlert` method)
- Config: optional toggle to enable/disable (`notify_tp_sl_fills: true/false`)

Tier label must read from `hyperliquidProtectionTiers(sc)` (see #660) rather than hardcoded TP1/TP2 — operators can configure N tiers.

## Related

- Issue #659 — Discord summaries show wrong TP1/TP2 values (display bug, not fill detection)
- PR #660 — tiered ATR TP display driven from config; tier label here must follow the same pattern
- `sendTradeAlerts` currently only fires on open/close trades from the per-strategy execute loop, not on reconciler-booked TP/SL fills

richkuo:
## Summary
- Reconciler-detected TP/SL fills now emit an owner DM the same cycle, instead of waiting for the next summary post (up to 5 minutes later). `sendTradeAlerts` is gated on the per-strategy execute loop's `trades` counter, which the reconciler bypasses entirely.
- Hooks at the three booking sites in `reconcileHyperliquidAccountPositions`: Detector 1 SL-owner full close, Detector 2 SL-owner partial, Detector 3 TP partial.
- TP tier label sourced from `hyperliquidProtectionTiers(sc)` via new `hyperliquidClearedTPTier(sc, pos, closeQty) (int, bool)` so operator-configured tier counts are honored (no hardcoded TP1/TP2 — see #660).
- Toggle: `notify_tp_sl_fills` (top-level, `*bool`, default true). Nil/missing keeps the alert on; explicit `false` disables.

## DM format
```
TP1 filled — hl-tema-eth-live (partial)
ETH LONG — 0.216000 @ $2340.9000
Remaining: 0.216000 ETH | PnL=$12.34
```

## Test plan
- [x] `go test ./scheduler/...` — full suite passes (21.5s)
- [x] `gofmt -l scheduler/` — clean
- [x] New tests: `TestFormatProtectionFillAlert_*`, `TestNotifyProtectionFill_*`, `TestTPTierLabel`, `TestLastBookedTradePnL`, `TestNotifyTPSLFillsEnabled_DefaultsToTrue`, `TestHyperliquidClearedTPTier_TierIndex`
- [x] Existing reconciler test call sites updated for the two new args (`notifier`, `notifyTPSLFills`); behavior unchanged when both nil/false

## Notes
- `applyHyperliquidProtectionSync` (the OID-clear path in `hyperliquid_protection.go`) was deliberately NOT chosen as the hook site — it doesn't book trades, so a DM there would either fire before the booking succeeds or double-fire alongside the reconciler. Single emission at the three `record*` call sites avoids both.
- `syncHyperliquidAccountPositions` (used by `--once` and tests) passes `nil, false` for the new args. The notifier-aware path is only the main loop's reconcile call.

Closes #661

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -122,7 +122,18 @@ type Config struct {
 	SummaryFrequency       map[string]string          `json:"summary_frequency,omitempty"`          // #30 — per-channel summary cadence; keys match Discord/Telegram channel keys (e.g. "spot", "options", "hyperliquid"). Values: Go duration ("30m", "2h"), alias ("hourly", "every"/"per_check"/"always"), or empty for legacy default (continuous: every channel run; spot: hourly)
 	RiskFreeRate           *float64                   `json:"risk_free_rate,omitempty"`             // #397 — annualized risk-free rate used in Sharpe-ratio calculations (e.g. 0.02 for 2%). Nil/missing falls back to DefaultAnnualRiskFreeRate; an explicit 0 is respected so backtest comparisons can pin to a 0% benchmark.
 	DefaultStopLossATRMult *float64                   `json:"default_stop_loss_atr_mult,omitempty"` // #605 — top-level default applied to HL perps/manual strategies that omit all stop_loss_* / trailing_stop_* fields. Nil/missing falls back to 1.0; explicit values let operators tune the ATR stop without recompiling.
+	NotifyTPSLFills        *bool                      `json:"notify_tp_sl_fills,omitempty"`         // #661 — owner DM when HL on-chain TP/SL fills are detected by the reconciler. Nil/missing → enabled; explicit false disables.
 	TradingViewExport      TradingViewExportConfig    `json:"tradingview_export,omitempty"`         // #3 — optional symbol overrides for TradingView portfolio CSV exports
+}
+
+// NotifyTPSLFillsEnabled reports whether reconciler-detected TP/SL fills should
+// trigger an owner DM. Nil pointer (missing field) defaults to true so existing
+// configs get the alert without an explicit opt-in.
+func (c *Config) NotifyTPSLFillsEnabled() bool {
+	if c == nil || c.NotifyTPSLFills == nil {
+		return true
+	}
+	return *c.NotifyTPSLFills
 }
 
 // ParseSummaryFrequency converts a summary_frequency value to a duration.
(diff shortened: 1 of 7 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #664, opened 2026-05-08
fix: mark filled TP tiers with ✓ in position summaries

### LINKED ISSUE #662: Position summaries don't reflect TP tier fill state
## Summary

After a take-profit (TP) tier fills on the exchange, the periodic position summary still displays the TP as if it were pending. The summary does not track or reflect which TP tiers have already executed.

## Current Behavior

For a live Hyperliquid perps position with tiered TP ATR protection:

**Example scenario:**
- Position: ETH LONG 0.432 @ $2316.90
- TP1 order at $2340.90 fills, closing 0.216 ETH
- Protection sync zeros out `pos.TPOIDs[0]` and reconciler books a partial close
- Position quantity drops to 0.216 ETH

**What the summary still shows:**
```
hl-tema-eth-live ETH LONG x0.432 @ $2316.90
ATR: $12.01 | SL: $2298.90 | TP1: $2340.91 | TP2: $2352.91
```

The summary continues to display **TP1 as pending** even though it already filled and reduced the position size.

## Root Cause

`scheduler/discord.go` → `collectPositions()` only checks `strategyUsesTieredTPATRClose()` and `pos.EntryATR` to decide whether to show TP hints. It does **not** inspect `pos.TPOIDs` to see which tiers are already filled.

```go
// HARDCODED math was fixed in #659 — this is about FILL STATE, not pricing
if strategyUsesTieredTPATRClose(sc) && pos.EntryATR > 0 {
    // Shows TP1 and TP2 regardless of whether they already filled
    tp1 = ...
    tp2 = ...
}
```

The position state already tracks fill state:
- `pos.TPOIDs` — array of order IDs, zeroed out when a tier fills
- `pos.Quantity` — reduced after reconciler books the partial close
- Trade history — records the `External partial close` from the reconciler

But the summary formatter ignores all of this and always renders both tiers as pending.

## Expected Behavior

When a TP tier has filled, the summary should reflect that. Options:

1. **Omit filled tiers** — only show TP tiers that still have active OIDs
2. **Mark as filled** — e.g. `TP1: $2340.91 ✓` or `TP1 filled` + remaining TP2
3. **Show remaining size** — make it clear the position is now half-size and only TP2 remains

## Related

- #659 — fixed hardcoded TP1/TP2 pricing in trade execution alerts
- This issue covers the **periodic position summary** path (`collectPositions`)

## Implementation Notes

- `pos.TPOIDs` is all-zero before the first protection-sync places tiers, so `OID == 0` alone can't distinguish "not yet placed" from "filled." Gate the filled marker on `pos.Quantity < pos.InitialQuantity` as well.
- Confirm `tieredTPATRPrices` and `TPOIDs` iterate tiers in the same order before indexing one by the other.

## Files Affected

- `scheduler/discord.go` — `collectPositions()` TP display logic (shared formatter; fix flows to both Discord and Telegram summaries)

richkuo:
## Summary

`collectPositions` rendered every tier from `tieredTPATRPrices` regardless of fill state, so periodic summaries kept showing TP1 as pending even after it filled and shrunk the position. Mark a tier with `✓` when its `TPOIDs[i]==0` **and** the position has shrunk vs `InitialQuantity` — `TPOIDs` are also all-zero before the first protection-sync places tiers, so `OID==0` alone is ambiguous.

Both Discord and Telegram summaries share this formatter via `FormatCategorySummary`, so a single fix covers both surfaces.

## Test plan

- [x] `go test ./...` passes
- [x] New regression `TestCollectPositions_TieredTPATR_FilledTierMarked` covers the post-fill case
- [x] New regression `TestCollectPositions_TieredTPATR_NoFillBeforeProtectionSync` covers the pre-placement (all-zero OIDs, full quantity) case so the marker doesn't leak

Closes #662

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/discord.go
+++ b/scheduler/discord.go
@@ -1039,7 +1039,15 @@ func collectPositions(sc StrategyConfig, ss *StrategyState, prices map[string]fl
 			}
 		}
 		if tps := tieredTPATRPrices(sc, pos.Side, pos.AvgCost, pos.EntryATR); len(tps) > 0 {
+			// A zero TPOID alone is ambiguous (tiers also hold zero before the
+			// first protection-sync places them); require an observed shrink
+			// vs. InitialQuantity to mark a tier as filled (#662).
+			partiallyClosed := pos.InitialQuantity > 0 && pos.Quantity+1e-9 < pos.InitialQuantity
 			for i, tp := range tps {
+				if partiallyClosed && i < len(pos.TPOIDs) && pos.TPOIDs[i] == 0 {
+					extras += fmt.Sprintf(" | TP%d: $%s ✓", i+1, fmtComma2(tp))
+					continue
+				}
 				pct := percentFromEntry(pos.Side, pos.AvgCost, tp)
 				extras += fmt.Sprintf(" | TP%d: $%s (%s)", i+1, fmtComma2(tp), fmtPnlPct(pct))
 			}
--- a/scheduler/discord_test.go
+++ b/scheduler/discord_test.go
@@ -1534,6 +1534,67 @@ func TestCollectPositions_TieredTPATR_OmittedWhenEntryATRZero(t *testing.T) {
 	}
 }
 
+// TestCollectPositions_TieredTPATR_FilledTierMarked verifies #662: once a TP
+// tier has filled (TPOID zeroed AND position quantity dropped below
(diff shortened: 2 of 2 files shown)


richkuo:
@claude

Issue #665, opened 2026-05-09
Trade alert DM: add OID, reorder SL, show ATR multipliers


richkuo:
## Feature Request

Trade alert DMs (Discord/Telegram notifications for opens/closes) have three formatting gaps:
1. No exchange order ID (`exchange_order_id`) — hard to cross-reference with exchange order history.
2. SL renders after the TP tiers, burying the most operationally relevant protection level.
3. SL/TP prices show no ATR multiplier, so it's not obvious how each level relates to the entry ATR (`collectPositions` already shows ATR mult on SL — `(1.00x)`).

### Scope

Trade-alert DMs only (`FormatTradeDM` in `scheduler/discord.go`). Three changes:
1. Append `| OID: <exchange_order_id>` to the symbol/side/qty/price/value line.
2. Reorder extras so SL renders immediately after ATR, before the TP tiers.
3. Append `(<n>x)` ATR multiplier to SL and each TP tier.

Position summaries (`collectPositions`) are **out of scope** — a position can span multiple fills (open + adds, partial closes), so there is no single unambiguous OID to render. SL already shows its ATR mult there; TP ATR mult parity in `collectPositions` can be a separate follow-up if desired.

### Proposal

Current:
```
🟢 **TRADE EXECUTED - LIVE**
Strategy: hl-eth-perps (Hyperliquid perps)
ETH — LONG 0.432 @ $2,306 | Value: $996
ATR: $12.34 | TP1: $2,330.68 | TP2: $2,355.36 | SL: $2,298.90 (-0.3%)
```

Proposed:
```
🟢 **TRADE EXECUTED - LIVE**
Strategy: hl-eth-perps (Hyperliquid perps)
ETH — LONG 0.432 @ $2,306 | Value: $996 | OID: 418206313303
ATR: $12.34 | SL: $2,298.90 (-0.3%) (1.0x) | TP1: $2,330.68 (1.0x) | TP2: $2,355.36 (2.0x)
```

### Context
- `exchange_order_id` is already stored on `Trade` (`scheduler/portfolio.go:377`) and populated for live HL/OKX/RH/TopStep fills + manual opens.
- TP ATR multipliers come from `hyperliquidProtectionTiers(sc)` (already used by `tieredTPATRPrices` to compute the prices); render needs to surface the per-tier `atr_mult` alongside the price.
- SL ATR mult is computed as `|AvgCost - StopLossTriggerPx| / EntryATR` (same formula as `collectPositions`).

### Implementation notes
[...]

PR #671, opened 2026-05-09
feat: persist SL ATR mult + TP tier snapshot on trades for analytics

### LINKED ISSUE #669: Persist SL arming method + TP tier snapshot on trades for analytics
## Problem

The displayed SL ATR multiplier is back-computed (`|entry − SL| / EntryATR`), and TP tier multiples are read from current config. Both are fine for display but lose information needed for **historical analytics**:

1. **SL arming method is unrecoverable from trade history.** A back-computed `(0.83x)` could mean ATR-armed at 0.83×, OR pct-armed at 5% on a position where 5%/EntryATR happened to equal 0.83. There's no way to tell after the fact which arming method ran.
2. **Tier config changes erase the historical record.** If you change `sc.CloseStrategies[].Params["tiers"]` from `[1×/2×]` to `[1.5×/3×]` mid-strategy-life, any analysis querying current config will misattribute older trades' TP plan.

The moment tier config changes on a live strategy, you've lost the ability to reconstruct what older trades were placed against.

## Proposed schema

Two additive nullable columns on `trades`:

- `stop_loss_atr_mult REAL` — non-NULL = ATR-armed at fill time, value = `EffectiveStopLossATRMult(sc)` resolved at open. NULL = armed via `StopLossPct` / `StopLossMarginPct` / `TrailingStopPct` (or no SL). Nullness alone is enough to gate the SL display suffix correctly (`(<n>x)` only when non-NULL), eliminating the misleading-mult problem in #665 review finding 2.
- `tp_tiers_json TEXT` — full tier snapshot at fill time: `[{\"atr_multiple\":1.0,\"close_fraction\":0.5},{\"atr_multiple\":2.0,\"close_fraction\":1.0}]`. NULL = strategy doesn't use tiered_tp_atr*. Storing the snapshot (not just multiples) means future tier-shape changes don't lose history.

Both nullable. No backfill — legacy rows stay NULL = unknown.

## Analytics queries unlocked

- \`SELECT AVG(realized_pnl) FROM trades WHERE stop_loss_atr_mult IS NOT NULL GROUP BY ROUND(stop_loss_atr_mult, 1)\` — PnL by SL ATR-mult bucket
- \`SELECT AVG(realized_pnl) FROM trades WHERE stop_loss_atr_mult IS NULL\` vs the above — ATR vs pct/margin arming methods
- A/B comparing tier configs across a strategy's lifetime by parsing \`tp_tiers_json\` snapshots
- Per-tier hit rate: join trades on \`(strategy_id, position_id)\`, count which tier OID fired

## Implementation notes

- Stamp at trade-open time (\`ExecutePerpsSignal\` / \`ExecuteSpotSignal\` paths) so the values reflect the config that armed the order, not whatever's in config when the trade closes later.
- Migration: \`ALTER TABLE trades ADD COLUMN ...\` guarded by \`PRAGMA table_info\` (existing idempotent pattern in \`scheduler/db.go\`).
- Update \`stampOpenTradeFromPosition\` to backfill these on the most-recent open row when SL is armed late (mirrors how \`stop_loss_trigger_px\` and \`entry_atr\` already work).
- After this lands, the SL display suffix in \`tradeAlertExtras\` (#665) and \`collectPositions\` should switch to gating on \`trade.StopLossATRMult IS NOT NULL\` (or the equivalent on Position) rather than back-computing from the trigger price.

## Out of scope

- Backfilling legacy trades (NULL is correct for unknown).
- Changing on-chain order placement — purely a record-keeping change.
- Display fixes for the back-compute issue (covered as a follow-up to #665).

richkuo:
Closes #669

## Summary
- Two additive nullable `trades` columns: `stop_loss_atr_mult REAL` (non-NULL iff ATR-armed at fill time) and `tp_tiers_json TEXT` (full [{atr_multiple, close_fraction}] snapshot).
- Mirrored on `positions` to support same-cycle backfill via `stampOpenTradeFromPosition`, following the same pattern as `entry_atr` / `stop_loss_trigger_px`.
- Stamped at fill time on every trade-open path (perps live/paper, spot, dispatch, manual, protection sync) and propagated onto close legs from `Position` so round-trip analytics queries see both sides.
- Helpers `tradeOpenStopLossATRMult`, `tradeOpenTPTiersJSON`, `stampPositionProtectionSnapshot` resolve values from `sc`.

## Test plan
- [x] `go test ./...` full suite passes
- [x] 7 new tests cover ATR/pct/zero resolution, tier JSON snapshot shape, stamper idempotency, in-memory + SQLite backfill, NULL preservation

## Out of scope
- Display fix for the back-computed `(<n>x)` suffix (#665 follow-up noted in the issue)
- Backfill of legacy trade rows (NULL = unknown by design)

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 129 in / 57447 out | Cache: 11980535 read / 235152 written | Cost: $8.9

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -76,6 +76,8 @@ CREATE TABLE IF NOT EXISTS positions (
     tp1_oid INTEGER NOT NULL DEFAULT 0,
     tp2_oid INTEGER NOT NULL DEFAULT 0,
     tp_oids_json TEXT NOT NULL DEFAULT '',
+    stop_loss_atr_mult REAL,
+    tp_tiers_json TEXT NOT NULL DEFAULT '',
     PRIMARY KEY (strategy_id, symbol)
 );
 
@@ -159,7 +161,9 @@ CREATE TABLE IF NOT EXISTS trades (
     exchange_order_id TEXT NOT NULL DEFAULT '',
     exchange_fee REAL NOT NULL DEFAULT 0,
     is_close INTEGER NOT NULL DEFAULT 0,
-    realized_pnl REAL NOT NULL DEFAULT 0
+    realized_pnl REAL NOT NULL DEFAULT 0,
+    stop_loss_atr_mult REAL,
+    tp_tiers_json TEXT NOT NULL DEFAULT ''
 );
 
 CREATE INDEX IF NOT EXISTS idx_trades_strategy ON trades(strategy_id);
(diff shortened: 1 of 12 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #675, opened 2026-05-09
fix: attribute TP-fired closes to TP fills, not SL trigger price

### LINKED ISSUE #673: HL reconciler misattributes TP-fired close as stop-loss when position fully flattens
## Summary

When both TPs fire on a Hyperliquid perps position and flatten it on-chain, the sole-owner reconciler books the close as if the **stop-loss** fired at the SL trigger price — even though the SL never executed (HL auto-cancels the resting reduce-only SL once the position is flat). This produces a large negative mis-booking on what should be a winning trade.

## Observed

Strategy `hl-tema-eth-live` (single-owner ETH long), Cycle #17059:

```
LONG ETH x0.432 @ $2,315.70 (+$2.18) | ATR: $4.74
SL: $2,308.60 (-0.3%) (1.5x) | TP1: $2,325.19 (+0.4%) | TP2: $2,329.93 (+0.6%)
```

Between Cycle #17059 and #17061, ETH ran to ~$2,330 and **both TPs filled on-chain**. The reconciler then ran and the strategy `Value` column dropped from **185 → 180** — the opposite direction of the actual realized PnL.

Expected realized PnL: `0.216 × ($2,325.19 − $2,315.70) + 0.216 × ($2,329.93 − $2,315.70) − fees ≈ +$4.4`
Booked PnL: `0.432 × ($2,308.60 − $2,315.70) − fee ≈ −$3.4`
Mis-booking ≈ **−$8**, matches the observed value drop.

## Root cause

`scheduler/hyperliquid_balance.go:298-316` (`reconcileHyperliquidPositionsWithResolver`, sole-owner path):

```go
} else if onChainPos == nil && statePos != nil {
    // Position in state but not on-chain — closed externally.
    ...
    if statePos.StopLossOID > 0 && statePos.StopLossTriggerPx > 0 {
        lookup, useFillFee := resolveFee(sym, statePos.StopLossOID, statePos.Quantity)
        ...
        if recordPerpsStopLossCloseWithFillFee(stratState, sym, statePos.StopLossTriggerPx, lookup.Fee, useFillFee, oidStr, "stop_loss", logger) {
            return true
        }
    }
```

The branch assumes "position flat on-chain + `StopLossOID > 0` = SL fired". It calls `recordPerpsStopLossCloseWithFillFee` with `closePx = statePos.StopLossTriggerPx` regardless of whether the SL OID actually has any fills in `userFills`.

When TPs fire and flatten the position, HL auto-cancels the resting reduce-only SL (see comment at `scheduler/portfolio.go:172-180`). `lookupHyperliquidFillByOID` for the SL OID returns `ok=false` (no matching fills), so `useFillFee=false` and the modeled fee is applied — but the close price is still the SL trigger price, producing a fictitious loss.

## Why existing TP-aware logic doesn't catch it

Detector 3 in the shared-coin path (`hyperliquid_balance.go:716-790`, `hyperliquidSharedPartialCloseDrift` + `hyperliquidClearedTPTier`) correctly attributes TP fills to specific tiers, but:

1. It only runs in the **shared-coin** branch (`reconcileHyperliquidAccountPositions` line 471 onward). Sole-owner strategies skip it.
2. It requires `onChainQty > tol` (partial drift). When TPs fully flatten the position, the condition fails — the sole-owner branch (or shared-coin Detector 1, full external close) takes over.

The shared-coin Detector 2 (SL-owner partial close at line 692) has the same shape of bug for the full-close case: it unconditionally calls `recordPerpsStopLossCloseWithFillFee(... slOwnerPos.StopLossTriggerPx ...)` regardless of whether the SL OID returned a fill.

## Suggested fix

Before falling through to the SL-attribution branch, query `userFills` for the SL OID AND each `TPOIDs[i]`:

- If SL OID has fills → existing SL path (book at SL trigger).
- If SL OID has no fills but one or more TP OIDs do → book each TP fill using its actual `Px`/`Fee`/`ClosedPnl` from `HLFillLookup`. `lookupHyperliquidFillByOID` already returns the data needed.
- Neither has fills (rare race — manual UI close before fills indexed) → fall back to current `recordClosedPosition(... 0, 0, "hl_sync_external" ...)` zero-PnL path.

Apply the same TP-OID check in shared-coin Detector 2 (line 692) so an SL-owner that flatens via TPs doesn't get booked at the SL price.

## Repro

Single-owner HL perps strategy with `tiered_tp_atr` / `tiered_tp_atr_live` close evaluator placing on-chain N-tier TPs + SL via `hyperliquid_protection.go`. Run a position into both TPs in a single cycle gap so on-chain reaches `qty=0` before the next reconcile.

richkuo:
## Summary

- When both reduce-only TPs flatten a HL perps position, HL auto-cancels the resting SL on-chain. The reconciler's sole-owner external-close path (`reconcileHyperliquidPositionsWithResolver`) previously assumed `StopLossOID > 0 + on-chain flat = SL fired`, booking the close at the SL trigger price and producing a fictitious loss (observed: strategy value 185 → 180 on a winning trade).
- New `hlAttemptCloseFromTPFills` helper checks whether the SL OID has fills before falling through to the SL path. When SL has no fills but TP OIDs do, books each tier as a partial close at its actual VWAP fill price + fee from `userFills`.
- `HLFillLookup` gains a `Px` field (size-weighted avg fill price) computed in both `lookupHyperliquidFillByOID` and the coin+size fallback. The cached resolver pre-fetches TP OID lookups so the apply phase has them ready under `mu.Lock()`.

## Test plan

- [x] `TestReconcilePosition_TPFillsAttributedNotSL` — reproduces the exact #673 scenario (ETH long, both TPs fill, SL OID has no fills); asserts cash delta is positive and two close trades are booked at TP prices.
- [x] `TestReconcilePosition_SLFillStillTakesSLPath` — SL OID has fills; asserts legacy SL path is unchanged and TP attribution does not interfere.
- [x] `TestReconcilePosition_NoFillsFallsBackToZeroPnL` — neither SL nor TP fills; asserts `hl_sync_external` zero-PnL fallback is preserved.
- [x] Full `go test ./...` passes (245 lines added, 0 existing tests broken).

Closes #673

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -299,6 +299,13 @@ func reconcileHyperliquidPositionsWithResolver(stratState *StrategyState, sym st
 		// Position in state but not on-chain — closed externally.
 		logger.Info("hl-sync: %s position (%.6f %s) no longer on-chain, removing",
 			sym, statePos.Quantity, statePos.Side)
+		// #673: When the position has TP OIDs and the SL OID has no fills,
+		// the position was flattened by TPs (HL auto-cancels the resting
+		// reduce-only SL once flat). Book each TP fill at its actual price
+		// rather than mis-attributing to the SL trigger price.
+		if hlAttemptCloseFromTPFills(stratState, sym, statePos, resolveFee, logger) {
+			return true
+		}
 		if statePos.StopLossOID > 0 && statePos.StopLossTriggerPx > 0 {
 			lookup, useFillFee := resolveFee(sym, statePos.StopLossOID, statePos.Quantity)
 			oidStr := strconv.FormatInt(statePos.StopLossOID, 10)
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #676, opened 2026-05-09
ci: update Claude bot comment when workflow is cancelled or fails

richkuo:
## Summary

`claude-code-action@v1` posts an in-progress comment on the issue/PR when invoked. The existing "Append LLM footer" step (`.github/workflows/claude.yml:196`) only runs on success — `if:` defaults to `success()` — so when a run is cancelled mid-stream or fails, the bot comment stays stuck in its in-progress state with no model/effort metadata and no indication the workflow aborted.

This adds a sibling step gated on `cancelled() || failure()` that:

- Looks up the most recent `claude[bot]` comment (same pattern as the success-path footer step).
- Strips any stale LLM footer + any previously appended status note (idempotent on retries).
- Appends a status note linking to the Actions run log, plus the standard `LLM: <model> | <effort> | Harness: …` footer.
- Skips ccusage to avoid burning a sub-minute on `npx` install when the run is being torn down.

`MODEL_NAME` falls back to `(model not resolved)` for the case where the failure happens before `Resolve model from @claude invocation` runs.

No untrusted input is interpolated into shell — `RUN_URL` is built from `github.{server_url,repository,run_id}` and `IS_CANCELLED` is a function result.

## Test plan

- [ ] Trigger `@claude …` on an issue, cancel the run mid-execution; confirm the bot comment gets a "Workflow cancelled before completion" note + LLM footer.
- [ ] Force a failure (e.g., temporarily break a setup step on a test branch); confirm the bot comment shows "Workflow failed before completion" + LLM footer.
- [ ] Verify a normal `@claude` run is unchanged (the new step has no effect when neither `cancelled()` nor `failure()` is true).

---

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -246,3 +246,54 @@ jobs:
           gh api "repos/$REPO/issues/comments/$COMMENT_ID" \
             --method PATCH \
             --field body="$NEW_BODY"
+
+      - name: Note workflow cancellation/failure on Claude comment
+        if: always() && steps.verify_invocation.outputs.invoked == 'true' && (cancelled() || failure())
+        env:
+          GH_TOKEN: ${{ github.token }}
+          MODEL_ID: ${{ steps.resolve_model.outputs.model_id }}
+          EFFORT: ${{ steps.resolve_model.outputs.effort }}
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

PR #677, opened 2026-05-09
[codex] Refactor open trade snapshot recording

### LINKED ISSUE #674: Refactor: unify open-trade path (single INSERT, shared helper across Execute* and manual)
## Context

Open-leg trades currently land in SQLite via two writes:
1. `RecordTrade` → `StateDB.InsertTrade` (most fields)
2. `stampOpenTradeFromPosition` → `StateDB.UpdateTradeStampedFields` (backfills `EntryATR`, `StopLossTriggerPx`, and as of #671 also `StopLossATRMult` / `TPTiersJSON`)

The UPDATE exists because `Execute*Signal*` records the trade before the caller knows `result.Indicators["atr"]` (Python-derived) or the SL/TP OIDs that the post-fill protection sync places on-chain.

`manual.go` (`applyManualAction`) already shows the cheaper pattern: build the `Position` with all snapshot fields, call `stampPositionProtectionSnapshot`, populate the `Trade` literal directly from `pos` + the queued action, single `RecordTrade`.

For HL perps live opens with a fixed-ATR SL, the picture today is two writes in the typical path: INSERT inside `Execute*Signal*` → UPDATE via `stampOpenTradeWithProtectionSnapshot` after `runHyperliquidProtectionSync` returns (`main.go:1505-1507`, `hyperliquid_protection.go:373`). A third UPDATE happens only on the paper / missed-arm fallback at `main.go:1448` (cycle N+1, gated on `signal==0 && posQty>0 && hlStopLossOID==0`).

## Proposal

Unify the open-trade path so **every** position-based open (manual + auto, paper + live, all `Position`-based platforms) goes through one helper that does a single INSERT once same-cycle protection placement returns.

### Current live HL fixed-ATR flow (corrected)

1. `runHyperliquidExecuteOrder` → Python `--execute` places the open order. `EffectiveStopLossPct` returns 0 for `StopLossATRMult > 0`, so no `--stop-loss-pct` is passed and Python does not place an SL alongside the fill.
2. `executeHyperliquidResult` runs `ExecutePerpsSignal*` → `RecordTrade` INSERTs the trade. Then `stampEntryATRIfOpened` writes `result.Indicators["atr"]` onto `pos.EntryATR`.
3. `runHyperliquidProtectionSync` (same cycle, `main.go:1505-1507`) builds `hlProtectionPlan` from `sc + pos` (now that `EntryATR > 0`), calls `--sync-protection` which places the SL + TP tier orders on-chain, returns OIDs and the SL trigger price.
4. `applyHyperliquidProtectionSync` writes OIDs + trigger px onto `pos`, then `stampOpenTradeWithProtectionSnapshot` does the UPDATE so the trade row reflects the on-chain values.

The cycle-N+1 branch at `main.go:1448` is a **fallback** for paper-mode arming and the rare case where step 3 didn't land an SL (sync error, partial state). It's not the primary live path.

### Shared core helper

```go
type PositionOpenParams struct {
    Symbol, Side, ExchangeOrderID                   string
    Quantity, FillPrice, FillFee, EntryATR          float64
    StopLossOID                                     int64
    StopLossTriggerPx                               float64
    TPOIDs                                          []int64
    Leverage, Multiplier                            float64
    Manual                                          bool
}

// recordPositionOpen builds a Position + Trade from sc + p, applies the
// fill-time protection snapshot, and persists with a single RecordTrade.
func recordPositionOpen(s *StrategyState, sc StrategyConfig, p PositionOpenParams) Trade
```

Used by both `applyManualAction` and every `executeXxxResult` in `main.go`.

### Path-specific upstream (unchanged)

- **Sizing math** — leverage / cash / `margin_per_trade_usd` for Execute paths vs. `--size` / `--notional` / `--margin` for manual.
- **Signal-driven flip semantics** in `ExecutePerpsSignalWithLeverage` (close + open synthesis) — manual has no flip case.
- **Source of `EntryATR` and SL trigger:**
  - manual: CLI `--atr` (or fallback) + inline-placed SL trigger captured into `PendingManualAction`.
  - Execute: `result.Indicators["atr"]` + protection-sync results (`StopLossOID`, `StopLossTriggerPx`, `TPOIDs`).

### Trade-row protection OID snapshot

Today `StopLossOID` and `TPOIDs` are mutable operational fields on `Position`; they are not preserved on the historical `Trade` row. That is fine for live state, but it loses the fill-time on-chain protection snapshot once OIDs are cleared, replaced, or zeroed after fills/cancels.

As part of this refactor, add immutable open-trade snapshot columns:

- `trades.stop_loss_oid`
- `trades.tp_oids_json`

`positions.stop_loss_oid` / `positions.tp_oids_json` remain the mutable operational state. The new `trades.*` columns are historical: they capture the SL/TP OIDs known at open time after same-cycle protection placement returns. `recordPositionOpen` should write OIDs onto `pos` for operations and also copy the open-time OID snapshot into the `Trade` before the single INSERT.

This requires a SQLite migration plus matching `Trade`, `InsertTrade`, `SaveState`, `LoadState`, and trade-history query updates.

### Sequencing the single INSERT

To collapse to one write, `RecordTrade` for an Execute-path open must run **after** the same-cycle protection sync returns, so `StopLossOID` / `StopLossTriggerPx` / `TPOIDs` are concrete by then. Concretely:

1. `Execute*Signal*` no longer calls `RecordTrade`; it returns a partially-populated `Trade` (open metadata, fees, OID, quantity, side, fillPrice — everything that doesn't depend on protection sync).
2. `executeXxxResult` stamps `EntryATR` from `result.Indicators` onto the returned `Trade` and the `Position`.
3. For HL perps live: caller invokes `runHyperliquidProtectionSync` (or its successor) **before** the INSERT. The sync now returns its placed SL/TP details; caller copies OIDs onto `pos`, and copies trade-relevant snapshot fields (`StopLossOID`, `StopLossTriggerPx`, `TPOIDs`, `StopLossATRMult`, `TPTiersJSON`) into the `Trade`.
4. `recordPositionOpen` (or a thin wrapper) does the single `RecordTrade` with everything filled in.

For paper mode and platforms without a post-fill protection sync (OKX, RH, TopStep, spot, futures), the INSERT can happen immediately after `Execute*Signal*` returns — no protection sync to wait on.

### On drift

The earlier draft of this issue claimed a precomputed fixed-ATR trigger price has "no drift" because cycle N+1's ATR equals cycle N's ATR. That claim is correct only when the trigger is computed **post-fill** using actual `AvgCost`. Two failure modes invalidate it:

- Computing pre-fill (e.g. against mid/limit) and passing the trigger into `--execute` would drift vs. the actual fill `avg_px`.
- HL tick-rounds the on-chain trigger; the un-snapped Go-computed value would differ at sub-tick granularity from what the exchange records.

Sequencing as above (read SL trigger back from the protection-sync response, not from a pre-fill prediction) sidesteps both.

### Concrete deletions

- `stampOpenTradeFromPosition` (`portfolio.go:1763`)
- `stampOpenTradeWithProtectionSnapshot` (`trade_protection_snapshot.go:73`) — 8 call sites in `main.go`
- `StateDB.UpdateTradeStampedFields` (`db.go:579`)

The cycle-N+1 fallback branch at `main.go:1448-1474` stays — it covers paper mode and missed-arm cases where the same-cycle sync didn't land an SL. Its `stampOpenTradeWithProtectionSnapshot` call gets replaced with whatever per-cycle backfill mechanism the refactor settles on (e.g. a targeted UPDATE that the same-cycle path no longer needs).

### Net effect

- One INSERT per normal same-cycle open across **every position-based path** (manual, HL perps live/paper, OKX, RH, TopStep, spot, futures). Options (`OptionPosition`) keep their separate path for now. Rare missed-arm/fallback recovery may still use a later targeted UPDATE.
- `manual.go:548-591` and ~8 `executeXxxResult` blocks collapse to one helper invocation each with truly identical bookkeeping.
- Open trade rows preserve the fill-time protection snapshot, including SL/TP OIDs, while `Position` remains the mutable operational source of truth.
- Eliminates the divergence between `manual.go` and `Execute*` that prompted this issue.

## Caveats

- Larger PR than the original issue described — touches `manual.go`, every `executeXxxResult` path in `main.go`, all live-execute branches, the protection-sync return shape, trade persistence/schema code, and a sizable chunk of tests. Reasonable to land in stages: (1) add trade-row OID snapshot columns + helper shape; (2) migrate Execute paths for non-HL platforms (no protection-sync wait); (3) thread protection-sync results through to delay the HL INSERT; (4) migrate `manual.go`.
- Pure cleanup / consistency in steady state. The cycle-N+1 fallback branch stays — collapsing it is out of scope for this refactor since it covers paper mode and missed-arm recovery, not just primary live behavior.
- Options (`OptionPosition` / `ExecuteOptionsSignal`) intentionally out of scope — they go through their own `recordClosedOptionPosition` flow with a different position struct and lifecycle.

Surfaced from PR #671 review (optional issue 1) where reviewer flagged drift between `manual.go` and the Execute* paths.

richkuo:
## Summary

- add immutable trade-row protection snapshots for `stop_loss_oid` and `tp_oids_json`
- add deferred-open execution paths so scheduler callers can record open trades once after ATR/protection snapshots are known
- route manual, spot, futures, Robinhood, OKX, and Hyperliquid open paths through `recordPositionOpen`
- preserve deferred open fills with a bare trade fallback if the position disappears before final snapshot recording
- keep fallback trade-stamping support for later/missed protection arming

Closes #674

## Tests

LLM: GPT-5.5 | Effort: high | Harness: Codex CLI

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -162,10 +162,10 @@ CREATE TABLE IF NOT EXISTS trades (
     exchange_fee REAL NOT NULL DEFAULT 0,
     is_close INTEGER NOT NULL DEFAULT 0,
     realized_pnl REAL NOT NULL DEFAULT 0,
-    stop_loss_oid INTEGER NOT NULL DEFAULT 0,
-    tp_oids_json TEXT NOT NULL DEFAULT '',
     stop_loss_atr_mult REAL,
-    tp_tiers_json TEXT NOT NULL DEFAULT ''
+    tp_tiers_json TEXT NOT NULL DEFAULT '',
+    stop_loss_oid INTEGER NOT NULL DEFAULT 0,
+    tp_oids_json TEXT NOT NULL DEFAULT ''
 );
 
 CREATE INDEX IF NOT EXISTS idx_trades_strategy ON trades(strategy_id);
@@ -311,9 +311,6 @@ func (sdb *StateDB) migrateSchema() error {
 		"ALTER TABLE positions ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
 		// Inline TP OIDs for manual-open actions so the scheduler drain sets pos.TPOIDs (#632).
 		"ALTER TABLE pending_manual_actions ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
-		// Immutable trade-row snapshot of protection OIDs known at open time (#674).
-		"ALTER TABLE trades ADD COLUMN stop_loss_oid INTEGER NOT NULL DEFAULT 0",
-		"ALTER TABLE trades ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
 		// SL arming method + TP tier snapshot at fill time (#669). stop_loss_atr_mult
 		// stays nullable: NULL = legacy/unknown or non-ATR arming (pct/margin/trailing-pct);
 		// non-NULL = ATR-armed at the recorded multiplier. tp_tiers_json carries the
@@ -324,6 +321,9 @@ func (sdb *StateDB) migrateSchema() error {
 		"ALTER TABLE trades ADD COLUMN tp_tiers_json TEXT NOT NULL DEFAULT ''",
 		"ALTER TABLE positions ADD COLUMN stop_loss_atr_mult REAL",
 		"ALTER TABLE positions ADD COLUMN tp_tiers_json TEXT NOT NULL DEFAULT ''",
+		// Immutable trade-row snapshot of protection OIDs known at open time (#674).
+		"ALTER TABLE trades ADD COLUMN stop_loss_oid INTEGER NOT NULL DEFAULT 0",
+		"ALTER TABLE trades ADD COLUMN tp_oids_json TEXT NOT NULL DEFAULT ''",
 	}
 	for _, ddl := range migrations {
 		if _, err := sdb.db.Exec(ddl); err != nil {
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -162,6 +162,8 @@ CREATE TABLE IF NOT EXISTS trades (
     exchange_fee REAL NOT NULL DEFAULT 0,
     is_close INTEGER NOT NULL DEFAULT 0,
     realized_pnl REAL NOT NULL DEFAULT 0,
+    stop_loss_oid INTEGER NOT NULL DEFAULT 0,
+    tp_oids_json TEXT NOT NULL DEFAULT '',
     stop_loss_atr_mult REAL,
     tp_tiers_json TEXT NOT NULL DEFAULT ''
 );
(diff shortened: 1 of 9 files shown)


richkuo:
@claude

richkuo:
@claude

richkuo:
@claude re-review

PR #679, opened 2026-05-10
ci: fix claude.yml parse error from cancelled() in env block

### LINKED ISSUE #678: CI: cancelled() not valid in env context — claude.yml workflow fails
## Bug

`.github/workflows/claude.yml` fails to parse since #676 (127cc16) landed:

```
(Line: 260, Col: 25): Unrecognized function: cancelled. Located at position 1 within expression: cancelled()
```

**Root cause:** `cancelled()` is a [status check function](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) that is only valid inside `if:` conditionals on jobs/steps. It cannot be evaluated in `env:` blocks.

**Location:** `.github/workflows/claude.yml:260`
```yaml
IS_CANCELLED: ${{ cancelled() }}
```

**Evidence:** every `push` event on the workflow file fails in 0s with "likely failed because of a workflow file issue" — e.g. run [25613496315](https://github.com/richkuo/go-trader/actions/runs/25613496315) on `main`, run [25613936143](https://github.com/richkuo/go-trader/actions/runs/25613936143) on `issue-674-open-trade-helper`. The step's own `if: cancelled() || failure()` gate (line 251) is fine — only the `env:` use breaks parsing.

## Fix

Split the single step into two steps gated by separate `if:` conditions, dropping the `IS_CANCELLED` env var entirely:

```yaml
- name: Note workflow cancellation on Claude comment
  if: always() && steps.verify_invocation.outputs.invoked == 'true' && cancelled()
  env: { ... }   # same env minus IS_CANCELLED
  run: |
    STATUS_NOTE="**Workflow cancelled before completion.** See [run log]($RUN_URL)."
    # ...rest of the script

- name: Note workflow failure on Claude comment
  if: always() && steps.verify_invocation.outputs.invoked == 'true' && failure()
  env: { ... }
  run: |
    STATUS_NOTE="**Workflow failed before completion.** See [run log]($RUN_URL)."
    # ...rest of the script
```

Either factor the shared script into a composite action / reusable shell snippet, or accept the duplication — the body is small.

Alternative (single step, less clean): keep one step gated on `cancelled() || failure()`, set `IS_CANCELLED` from a preceding `if: cancelled()` step that writes to `$GITHUB_ENV`.

richkuo:
## Summary

- `cancelled()` is a status check function only valid inside `if:` conditionals; using it in an `env:` block caused workflow validation to fail with `Unrecognized function: cancelled` since #676 landed (e.g. run [25613496315](https://github.com/richkuo/go-trader/actions/runs/25613496315) on main, run [25613936143](https://github.com/richkuo/go-trader/actions/runs/25613936143) on the issue-674 branch — both fail in 0s with "workflow file issue").
- Add a preceding step gated on `if: cancelled()` that writes `IS_CANCELLED=true` to `\$GITHUB_ENV`; the existing run script reads `\$IS_CANCELLED` unchanged. When the workflow merely fails (not cancelled) the variable stays empty and the script falls through to the "Workflow failed" branch.

Closes #678

## Test plan

- [ ] Workflow file parses (push to branch no longer fails in 0s).
- [ ] On a cancelled `@claude` run, the bot comment gets the "Workflow cancelled before completion." note.
- [ ] On a failed `@claude` run, the bot comment gets the "Workflow failed before completion." note.

---
LLM: Claude Opus 4.7 | high | Harness: Claude Code CLI

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -247,6 +247,10 @@ jobs:
             --method PATCH \
             --field body="$NEW_BODY"
 
+      - name: Mark cancellation for downstream status note
+        if: always() && cancelled()
+        run: echo "IS_CANCELLED=true" >> "$GITHUB_ENV"
+
       - name: Note workflow cancellation/failure on Claude comment
         if: always() && steps.verify_invocation.outputs.invoked == 'true' && (cancelled() || failure())
         env:
@@ -257,7 +261,6 @@ jobs:
           ISSUE_NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
           REPO: ${{ github.repository }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          IS_CANCELLED: ${{ cancelled() }}
         run: |
           case "$MODEL_ID" in
             "claude-opus-4-7[1m]")   MODEL_NAME="Claude Opus 4.7 (1M)" ;;


richkuo:
@claude

PR #683, opened 2026-05-10
update.sh: atomic swap, pre-flight probe, post-restart verify, rollback

### LINKED ISSUE #682: Harden scripts/update.sh: atomic binary swap, pre-flight probe, post-restart verification, rollback
## Problem

`scripts/update.sh` reports success on partial or wedged updates, leaving operators (and the auto-update DM flow) with a silent mismatch between what's on disk and what's running.

Concrete failure modes observed today:

1. **Non-atomic binary swap** — `scripts/update.sh:47` builds directly to `../go-trader`, overwriting the in-use binary. A failed/killed `go build` leaves a corrupt or zero-byte file on disk. Linux lets the running process keep executing via the held inode, but the next restart loads garbage.
2. **No restart verification** — `scripts/update.sh:51` runs `sudo systemctl restart go-trader` and exits 0 as soon as systemd dispatches. The script never confirms the new process actually came up, that the old one is gone, or that the running version matches what was just built. If the service is wedged in `deactivating (stop-sigterm)` (see #681) or fails `--probe-only` on start, the operator finds out by accident.
3. **No pre-flight probe** — version-drift between Go binary and Python (`--strategy-refs`, `--probe-only` argv contract) is only caught *after* downtime begins. The new binary should run `--probe-only` against the new Python *before* the swap.
4. **No rollback** — when a new binary fails to start, there is no automatic restore of the previous binary; the service stays down until an operator intervenes.
5. **No git cleanliness check** — `git pull --ff-only` will refuse on diverged history, but local uncommitted modifications (e.g. an in-progress edit on the box) cause confusing failures partway through.

Symptom in the wild: operators repeatedly need to manually re-run the build or replace the binary because update.sh reported success but the running process is stale or wedged.

## Design

Update flow becomes: **pre-flight → build to staging → probe staging → atomic swap → restart → verify → rollback on failure**.

### 1. Pre-flight checks (before any mutation)

- `command -v uv` (already present at `:30-33`).
- `command -v go` / `/opt/homebrew/bin/go` resolves to a real binary.
- `git diff --quiet` and `git diff --cached --quiet` — refuse to update with uncommitted changes.
- `git status --porcelain` shows no untracked files in `scheduler/` / `shared_*/` / `platforms/` / `backtest/` (warn-only — operator may have intentional state).
- Capture the *currently running* version via `/health` or by reading the symlink target (see step 4) for later rollback comparison.

### 2. Build to staging

- `go build -o ../go-trader.new` instead of overwriting `../go-trader`. A killed/failed build leaves the live binary intact.
- Verify the produced file is non-zero and executable.

### 3. Pre-flight probe of the new binary

- `./go-trader.new --probe-only` (or `./go-trader.new probe`) — already exists in `version_probe.go` for in-process startup; expose it as a CLI flag if not already separately invocable so the script can dry-run the binary against the new Python *before* the swap.
- Non-zero exit aborts the update; live binary unchanged.

### 4. Atomic swap + previous-version retention

- `mv -f go-trader go-trader.prev && mv -f go-trader.new go-trader` — POSIX rename is atomic on the same filesystem.
- Optional: instead of file moves, use a `go-trader -> go-trader-<sha>` symlink and atomic `ln -sfn`. Cleaner version history but bigger refactor; file-move is fine for v1.
- Cap retention at one previous binary (`.prev`) to bound disk use; older `.prev` is removed before the new swap.

### 5. Restart with confirmation

- `sudo systemctl restart go-trader`.
- Poll `systemctl is-active go-trader` every 1s up to ~30s. Must reach `active`.
- Once active, poll `/health` (existing endpoint via `cfg.StatusPort`, default 8099) up to ~10s. Must return 200 *and* report a version field matching `\$ver`.
- If `/health` doesn't currently return version, add it (small additive change in `scheduler/server.go`). Already-acceptable status fields exist; just add `version`.

### 6. Rollback on verification failure

- If restart times out, fails, or `/health` reports the wrong version: `mv -f go-trader.prev go-trader && sudo systemctl restart go-trader && wait-for-active`.
- Exit non-zero with a loud message identifying which check failed (probe / restart timeout / health timeout / version mismatch).
- Operator (and the `applyUpgrade` DM flow at `scheduler/updater.go:97-103`) sees the failure immediately instead of discovering it later.

### 7. Logging

- Each phase prefixed `[update] phase: <name>` so journal/DM output is greppable.
- Emit elapsed time per phase — operators noticing slow updates can localize the cause.

## Interaction With Other Code Paths

- **`applyUpgrade`** (`scheduler/updater.go:97`) calls `bash scripts/update.sh` without `--restart` then handles restart itself via `restartSelf`. The new pre-flight + atomic swap + probe steps still run; the restart-verification phase is a no-op when `--restart` is not passed (in-process path verifies its own restart implicitly by surviving). No change required to `updater.go`.
- **#681** (graceful shutdown): independent. #681 makes `systemctl restart` *return cleanly and quickly*; this issue makes update.sh *know whether restart worked*. Both can ship in either order.
- **Auto-update DM flow** (`updater.go:103`): when verification fails, the script's stderr goes back via `tailForDM`. The richer error messages from this issue improve operator visibility.

## Out of Scope

- Migration of `go-trader` to a versioned-symlink layout (mentioned as alternative under step 4) — file-move atomic swap covers the immediate need.
- Cross-host deployment (Ansible/image bake) — those callers wrap `update.sh`; this issue keeps `update.sh` itself the single source of truth.
- Config schema migration — already handled by the in-process probe and `runConfigMigrationDM`.

## Files Affected

- `scripts/update.sh` — bulk of the work; flow restructured per the seven steps above.
- `scheduler/server.go` — `/health` returns build version (additive field).
- `scheduler/version_probe.go` *(only if needed)* — expose `--probe-only` as a CLI mode invocable on a built binary without entering the trading loop. Likely already accessible; verify before adding.

## Acceptance

- `update.sh` exits non-zero on every observable failure mode: bad git state, build failure, probe failure, restart timeout, health timeout, version mismatch.
- A killed `go build` (e.g. operator Ctrl-C, OOM) leaves the running binary unchanged and the service running.
- A new binary that fails `--probe-only` never reaches the live path.
- A new binary that fails post-restart `/health` triggers automatic rollback to `.prev`; service comes back up on the previous binary; script exits non-zero with the failure cause.
- Successful update logs phase markers and elapsed times to stdout/journal.
- `applyUpgrade`'s in-process restart path still works; no change to `updater.go` required.

richkuo:
Closes #682.

Rewrites `scripts/update.sh` into a seven-phase atomic flow (preflight → pull → sync → build → probe → swap → restart+verify) with auto-rollback to `go-trader.prev` on verify failure. Adds `go-trader probe` + `go-trader version` subcommands and a `version` field on `/health`.

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 87 in / 24558 out | Cache: 5604700 read / 190729 written | Cost: $4.61


richkuo:
@claude

richkuo:
@claude fix review issues

richkuo:
@claude re-review

PR #684, opened 2026-05-10
graceful shutdown: two-phase drain on SIGTERM

### LINKED ISSUE #681: Graceful shutdown: two-phase drain on SIGTERM
## Problem

`systemctl restart go-trader` (operator-driven via `scripts/update.sh --restart` or in-process via `applyUpgrade` → `restartSelf` at `scheduler/updater.go:116-147`) often shows `deactivating (stop-sigterm)` for tens of seconds before systemd SIGKILLs.

The SIGTERM handler at `scheduler/main.go:200-211` saves state and closes `stopCh`, but the main loop only checks `stopCh` between cycles (`scheduler/main.go:466`, `:1908`). A cycle in flight keeps running every queued Python subprocess through the 4-slot `pythonSemaphore` until each hits its 30s `scriptTimeout` (`scheduler/executor.go:17`). Worst case shutdown ≈ `(N strategies / 4) × 30s` plus circuit-breaker drain work.

A naive fix — cancel every in-flight subprocess on SIGTERM — would also kill `--execute` scripts mid-call, leaving on-chain orders with no local Trade record (live/chain divergence). The fix has to distinguish **cancellable read-only work** from **side-effecting work that must finish**.

The four `context.Background()` first-args at `main.go:983,1000,1020,1041` are passed to circuit-breaker drain helpers (`runPendingHyperliquidCircuitCloses` etc.), not directly to `exec.CommandContext` — but those drains internally invoke close scripts via `RunPythonScript`, so the same plumbing reaches them.

## Design — Two-Phase Drain

Modeled on industry-standard graceful shutdown (k8s pod termination, nginx graceful reload, gRPC `GracefulStop`).

### Phase 1 — Drain (signal received → stop accepting new work)

- **Signal handler does one thing.** Do not save state or take locks in the signal goroutine. It sets `atomic.Bool draining=true`, closes `drainCh`, returns. Today's handler at `main.go:200-211` saves state under `mu.Lock()` from the signal goroutine — a panic there leaves the lock held and shutdown wedged. Shutdown sequencing moves to the main goroutine.
- **Main loop refuses new work** as soon as it sees `draining`. Check at cycle top and at each per-strategy dispatch decision. New cycles never start; in-flight cycles stop dispatching additional strategies.
- **Cancel `readOnlyCtx`.** All read-only subprocesses (`check_*.py`, `check_balance.py`, `fetch_*_marks.py`, `fetch_futures_marks.py`, `check_price.py`, etc.) die immediately via existing `exec.CommandContext` + process-group SIGKILL path. They have no side effects; cancelling is free.

### Phase 2 — Quiesce (let side-effecting work finish, bounded)

- **Side-effecting calls run to completion** under their own per-call timeout (already 30s via `scriptTimeout`):
  - `--execute` calls (HL/OKX/RH/TopStep order placement)
  - CB drain closers (`runPending{Hyperliquid,OKX,Robinhood,TopStep}CircuitCloses`)
  - Manual-action drain (`drainPendingManualActions`)
  - Protection-sync calls (`runHyperliquidProtectionSync`, `RunHyperliquidSyncProtection`)
  - Reconciler write paths (e.g. `hlAttemptCloseFromTPFills`, `recordPerpsExternalClose`)
  - Operator-required CB drains (`drainOperatorRequiredPendingCloses`)
- **`sync.WaitGroup` tracks them.** Main goroutine `wg.Wait()` with a hard cap of 15s.
- **Cap fires → force cancel.** `sideEffectCancel()` triggers process-group SIGKILL; this is the last-resort safety valve, not the happy path.

### Phase 3 — Persist (close-out)

1. `SaveStateWithDB(state, cfg, stateDB)` (idempotent — `applyUpgrade` already calls it pre-restart at `updater.go:107`).
2. Flush notifier — `notifier.Close()` waits for buffered Discord/Telegram HTTP. Killing mid-flight drops the shutdown DM.
3. Close `stateDB`.
4. `os.Exit(0)`.

## Caller Classification

Every `RunPythonScript` / `RunPythonScriptWithStdin` call site declares intent. Two new wrappers in `executor.go`:

```go
// Cancellable on shutdown. Use for any script with no on-chain or local-state side effects.
func RunPythonScriptReadOnly(ctx context.Context, script string, args []string) ([]byte, []byte, error)

// Registers with shutdown WaitGroup; runs under sideEffectCtx (cancelled only after the 15s cap).
// Use for any script that places live orders, mutates DB, or sends external messages.
func RunPythonScriptSideEffect(ctx context.Context, script string, args []string) ([]byte, []byte, error)
```

Both `ctx` parameters are derived from `readOnlyCtx` / `sideEffectCtx` at the dispatch site. Existing per-call `context.WithTimeout(parent, scriptTimeout)` stays.

**Audit pass:** every existing `RunPythonScript`/`RunPythonScriptWithStdin` caller is reclassified. Side-effect bucket includes (non-exhaustive): all `--execute` paths, `close_*.py` invocations, `RunHyperliquidExecute`, `RunHyperliquidClose`, manual-action apply, protection-sync, fee/balance backfill writes. Read-only bucket includes: all `check_*.py` strategy evaluation, `fetch_*_marks.py`, `check_price.py`, `check_balance.py`, position/balance fetchers used purely for snapshots.

The four CB drain calls at `main.go:983,1000,1020,1041` switch from `context.Background()` to `sideEffectCtx`.

## Operator-Visible Surface

- **systemd unit:** `TimeoutStopSec=20` in `systemd/go-trader@.service` (15s drain cap + 5s buffer for state save / notifier flush). systemd's SIGKILL becomes a true last-resort backstop, not first response.
- **Status server:** `/health` returns 503 once `draining=true`. Pure additive — no current consumer relies on the existing `/health` body — but future probe-based deployments (k8s, ELB) get correct semantics.
- **Logs:** drain phases written to journal so `journalctl -u go-trader` shows `[shutdown] draining`, `[shutdown] N side-effecting subprocess(es) in flight`, `[shutdown] complete in Xs`.

## Out of Scope

- `version_probe.go` runs pre-loop and exits before the trading loop starts; not wired through.
- One-off CLI commands (`manual-open`, `manual-close`, `backfill hl-fees`, `export tradingview`, `init`) keep `context.Background()` — they are not the long-running daemon.
- `restartSelf`'s `syscall.Exec` fallback at `updater.go:147` bypasses SIGTERM by design (no systemd available); not graceful and not in scope. Production is systemd; the fallback is a dev-box convenience.

## Files Affected

- `scheduler/main.go` — signal handler simplified (no state save, no lock); main goroutine drives Phase 1/2/3; `draining` checks at cycle top and per-strategy dispatch; `sync.WaitGroup` for side-effecting calls; four CB drain call sites switch to `sideEffectCtx`.
- `scheduler/executor.go` — two new wrappers (`RunPythonScriptReadOnly`, `RunPythonScriptSideEffect`); existing `RunPythonScript`/`RunPythonScriptWithStdin` retained as private impl.
- All call sites — mechanical reclassification pass.
- `scheduler/server.go` — `/health` returns 503 when `draining`.
- `systemd/go-trader@.service` — `TimeoutStopSec=20`.

## Acceptance

- `systemctl restart go-trader` returns within ~2s under normal load (no in-flight side-effecting work), within ~15s worst case.
- No live order placement is interrupted mid-call by SIGTERM (verified via integration test that fires SIGTERM during a stubbed `--execute` and asserts the Trade record is written).
- State saved exactly once on shutdown; idempotent if `applyUpgrade` already saved.
- Notifier flush completes before exit (no dropped shutdown DM).
- `/health` returns 503 once draining.
- journal shows the three phase markers on every shutdown.
- `TimeoutStopSec` SIGKILL never fires under normal conditions.

richkuo:
## Summary

- Two-phase drain on SIGTERM/SIGINT — `systemctl restart` no longer hangs in `deactivating (stop-sigterm)`.
- Read-only Python subprocesses (`check_*.py`, fetchers) cancel immediately; side-effecting subprocesses (`--execute`, `close_*.py`, `--sync-protection`) finish under their existing 30s `scriptTimeout` so on-chain orders aren't killed mid-call.
- 15s hard cap with SIGKILL backstop; `TimeoutStopSec=20` in the unit so systemd's SIGKILL is the true last resort.
- `/health` returns 503 once draining (k8s/ELB-friendly).

## Design

Industry-standard graceful-shutdown shape (k8s pod termination, nginx graceful reload, gRPC `GracefulStop`):

**Phase 1** (signal goroutine): set `draining=true`, cancel `shutdownReadOnlyCtx`, close `stopCh`. No state save, no locks, no I/O — a panic here would wedge the daemon.

**Phase 2** (deferred on main goroutine, `runDrain`): wait up to `shutdownDrainCap=15s` on `sideEffectWG`. In-flight side-effecting calls run to completion under their 30s `scriptTimeout`. Cap fires → `shutdownSideEffectCancel` triggers process-group SIGKILL.

**Phase 3** (deferred): `SaveStateWithDB` → `cleanupNotifier` flushes Discord/Telegram → `stateDB.Close` → `logMgr.Close`. Phase-2/3 defer is registered AFTER `cleanupNotifier` so LIFO puts state save BEFORE notifier close, preserving shutdown DMs.

Classification lives at the `executor.go` wrappers (not per-caller):
- `runPythonReadOnly` — check_*.py, fetch_*_marks, price/balance fetchers, `--list-json`. Cancelled at SIGTERM.
- `runPythonSideEffect` — all `Run*Execute` / `Run*Close` / `Run*UpdateStopLoss` / `Run*SyncProtection`. Registers with `sideEffectWG`.

The four `context.Background()` arguments at the CB-drain call sites in `main.go` (HL/OKX/TopStep/RH) switch to `shutdownSideEffectCtx` so the drain cap propagates.

## Test plan

- [x] `go test ./...` — full suite passes (existing + 5 new tests in `shutdown_test.go`).
- [x] `gofmt -l` clean on touched files.
- [x] `go build -ldflags ...` produces a versioned binary.
- [ ] Manual: `./go-trader` then `kill -TERM <pid>` — observe `[shutdown] draining`, `[shutdown] State saved`, `[shutdown] Complete` in stderr; process exits within ~1s when no side-effecting work is in flight.
- [ ] Manual: trigger an HL `--execute` (paper or testnet), send SIGTERM mid-call, verify the Trade record persists and the on-chain fill matches.
- [ ] Operator: `systemctl restart go-trader` returns within ~2s under normal load (vs. multi-minute hangs today).
- [ ] Operator: `curl localhost:8099/health` during drain returns 503 with body `{"status":"draining"}`.

## Out of scope

- `version_probe.go` runs pre-loop and exits the process; not wired into the drain.
- One-off CLI commands (`manual-open`, `manual-close`, `backfill hl-fees`, `export`, `init`) keep their default `context.Background()` — they are not the long-running daemon.
- `restartSelf`'s `syscall.Exec` fallback (`scheduler/updater.go:147`) bypasses SIGTERM by design (no systemd available); production uses systemctl, so this gap only affects dev boxes.
- update.sh hardening (atomic binary swap, post-restart verification, rollback) — tracked separately in #682.

Closes #681

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -119,17 +119,25 @@ type HyperliquidProtectionSyncResult struct {
 	TP2FilledExternally      bool      `json:"tp2_filled_externally,omitempty"`
 }
 
-// RunPythonScript executes a Python script and returns stdout/stderr.
-func RunPythonScript(script string, args []string) ([]byte, []byte, error) {
+// runPython is the shared subprocess spawner. parentCtx is one of
+// shutdownReadOnlyCtx (read-only scripts, cancelled at SIGTERM) or
+// shutdownSideEffectCtx (live order placement / state mutation, allowed to
+// finish under the drain cap). Both default to context.Background() when the
+// daemon entry point hasn't called initShutdownContexts (one-off CLI commands
+// and tests).
+func runPython(parentCtx context.Context, script string, args []string, stdinData []byte) ([]byte, []byte, error) {
 	pythonSemaphore <- struct{}{}
 	defer func() { <-pythonSemaphore }()
 
-	ctx, cancel := context.WithTimeout(context.Background(), scriptTimeout)
+	ctx, cancel := context.WithTimeout(parentCtx, scriptTimeout)
 	defer cancel()
 
 	cmdArgs := append([]string{script}, args...)
 	cmd := exec.CommandContext(ctx, ".venv/bin/python3", cmdArgs...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	if stdinData != nil {
+		cmd.Stdin = bytes.NewReader(stdinData)
+	}
 
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
@@ -143,6 +151,40 @@ func RunPythonScript(script string, args []string) ([]byte, []byte, error) {
 		return stdout.Bytes(), stderr.Bytes(), fmt.Errorf("script timed out after %s", scriptTimeout)
 	}
 	return stdout.Bytes(), stderr.Bytes(), err
+}
+
+// runPythonReadOnly is for scripts with no on-chain or local-state side
+// effects (check_*.py signal evaluation, fetch_*_marks.py, fetch_*_positions
+// for snapshot reads, --list-json, check_balance.py, check_price.py).
+// Cancelled immediately on SIGTERM so the daemon can shut down without
+// waiting on idle work.
+func runPythonReadOnly(script string, args []string) ([]byte, []byte, error) {
+	return runPython(shutdownReadOnlyCtx, script, args, nil)
+}
+
+// runPythonReadOnlyWithStdin mirrors runPythonReadOnly for scripts that
+// receive their input over stdin (currently RunOptionsCheckWithStdin).
+func runPythonReadOnlyWithStdin(script string, args []string, stdinData []byte) ([]byte, []byte, error) {
+	return runPython(shutdownReadOnlyCtx, script, args, stdinData)
+}
+
+// runPythonSideEffect is for scripts that place live orders, mutate local
+// state via on-chain operations, or send external messages (--execute,
+// close_*.py, --sync-protection, trigger updates). Registers with
+// sideEffectWG so SIGTERM waits up to shutdownDrainCap before forcing a
+// kill, preserving local/on-chain consistency.
+func runPythonSideEffect(script string, args []string) ([]byte, []byte, error) {
+	sideEffectWG.Add(1)
+	defer sideEffectWG.Done()
+	return runPython(shutdownSideEffectCtx, script, args, nil)
+}
+
+// RunPythonScript is the public entry for callers outside executor.go
+// (balance.go, init.go); both call sites are read-only. New side-effecting
+// callers should use runPythonSideEffect via a typed wrapper instead of
+// reaching for this function.
+func RunPythonScript(script string, args []string) ([]byte, []byte, error) {
+	return runPythonReadOnly(script, args)
 }
 
 // RunSpotCheck runs check_strategy.py and parses the result.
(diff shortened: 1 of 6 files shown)


richkuo:
@claude

PR #686, opened 2026-05-10
fix: gate SL fallback on confirmed userFills hit (#685)

### LINKED ISSUE #685: TP fill misattributed as SL close when TPOIDs already zeroed
## Bug

A TP2 fill on `hl-tema-eth-live` was recorded as `hl_sync_stop_loss` instead of `hl_sync_tp2_fill`. TP2 firing confirmed via HL UI / userFills.

**Observed:**
```
SL filled — hl-tema-eth-live
ETH LONG — 0.215000 @ $2323.3000
Remaining: 0.000000 ETH | PnL=-$1.47
```

Closed position record:
```
id=45 | ETH long | avg_cost=2329.8 | close_price=2323.3 | qty=0.215 | pnl=-1.47 | close_reason=hl_sync_stop_loss
```

`close_price=2323.3` is the SL trigger px (`AvgCost - 1.5×ATR`), not the TP2 fill price — confirming the reconciler routed through the SL fallback after `hlAttemptCloseFromTPFills` returned false.

## Config
```
close_strategies: tiered_tp_atr with tiers [{fraction: 0.5, multiple: 2}, {fraction: 1, multiple: 3}]
stop_loss_atr_mult: 1.5
```

## Root Cause

Follow-on gap from PR #675 (#673): `hlAttemptCloseFromTPFills` was added precisely to attribute TP-fired closes to TP fills, but it only handles the case where at least one `pos.TPOIDs[i]` is still positive at reconcile time. When all entries have already been zeroed by `applyHyperliquidProtectionSync` (`scheduler/hyperliquid_protection.go:304`), the loop at `scheduler/hyperliquid_balance.go:1163` skips every OID, `len(fills)==0` at `:1173` returns false, and the caller falls through to the SL-trigger-price path (`:527-540`).

The SL fallback does NOT verify the SL OID actually filled — it just checks `statePos.StopLossOID > 0` and books at `StopLossTriggerPx`. `pos.StopLossOID` is never zeroed when HL auto-cancels the resting reduce-only SL on TP flatten, so this path fires whenever both TPs cleared in a sequence that left all `TPOIDs[i] == 0`.

The original draft cited `len(pos.TPOIDs) == 0`; that guard at `:1146` never trips because the slice keeps its length — only the elements get zeroed.

## Suggested Fix

Gate the SL fallback at `hyperliquid_balance.go:527-540` on a verified SL fill. When `useFillFee && lookup.OID == statePos.StopLossOID && lookup.FilledQty > 0`, book as `hl_sync_stop_loss` (current behavior, possibly with the existing #621 qty adjustment). Otherwise fall through to `hl_sync_external` at zero PnL — same disposition as a fully-unknown off-scheduler close.

```go
if statePos.StopLossOID > 0 && statePos.StopLossTriggerPx > 0 {
    lookup, useFillFee := resolveFee(sym, statePos.StopLossOID, statePos.Quantity)
    if useFillFee && lookup.OID == statePos.StopLossOID && lookup.FilledQty > 1e-9 {
        // existing #621 qty adjust + recordPerpsStopLossCloseWithFillFee
    }
    // else: fall through to hl_sync_external
}
```

Rejected alternative: preserving `OriginalTPOIDs` across protection-sync zeroing. That adds parallel state and risks double-booking when reconcile and protection-sync both attribute the same fill across cycles. The SL-fill gate is local to the misattribution site and doesn't grow the position model.

## Regression Test

`hyperliquid_balance_test.go`: SL OID positive, all `TPOIDs` zero, resolver returns no SL fill — expect `hl_sync_external` at zero PnL, not `hl_sync_stop_loss` at trigger px.

## Files Affected
- `scheduler/hyperliquid_balance.go` — SL fallback in the off-chain reconcile path

richkuo:
## Summary

- Off-chain reconcile path mis-attributed TP-fired closes (with all `TPOIDs` zeroed by prior protection-sync cycles) as `hl_sync_stop_loss` at the cancelled SL's trigger price, producing a fictitious loss.
- `hlAttemptCloseFromTPFills` (PR #675) only handles the case where ≥1 `TPOIDs[i]` is still positive; once `applyHyperliquidProtectionSync` zeros them all, the helper returns false and the legacy SL fallback fired without verifying SL actually filled.
- Gate the SL fallback at `scheduler/hyperliquid_balance.go:527` on a userFills-confirmed SL fill (`useFillFee && lookup.OID == StopLossOID && FilledQty > 0`). Otherwise fall through to `hl_sync_external` at zero PnL.
- Existing SL-booking tests updated to supply confirming resolvers/stubs (production always has an account address + real indexer). New regression `TestReconcilePosition_AllTPOIDsZeroedSLNotFilled` locks in the #685 scenario.

## Test plan

- [x] `go test ./...` — all green (24.6s)
- [x] `go build -ldflags ...` clean
- [x] `go vet ./...` clean
- [x] Regression test asserts `hl_sync_external` (not `hl_sync_stop_loss`) when `TPOIDs=[0,0]` and SL has no userFills entry

Closes #685

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -547,6 +547,11 @@ func reconcileHyperliquidPositionsWithResolver(stratState *StrategyState, sym st
 				if recordPerpsStopLossCloseWithFillFee(stratState, sym, statePos.StopLossTriggerPx, lookup.Fee, useFillFee, oidStr, "stop_loss", logger) {
 					return true
 				}
+			} else if useFillFee {
+				// #685 log clarity: the lookup hit (logged above) matched
+				// something other than this SL OID (e.g. a coin+size fallback
+				// onto a TP fill), so this close is NOT being booked as SL.
+				logger.Info("hl-sync: %s SL OID %s unfilled — routing to hl_sync_external (matched oid=%d qty=%.6f)", sym, oidStr, lookup.OID, lookup.FilledQty)
 			}
 		}
 		// Close price is unknown — the fill happened off-scheduler between
@@ -1159,8 +1164,11 @@ func hlAttemptCloseFromTPFills(s *StrategyState, sym string, pos *Position, reso
 	// If the SL OID has fills, leave attribution to the existing SL path —
 	// it knows how to handle the #621 "SL fired on the post-TP residual"
 	// case and we don't want to double-book by also crediting TPs here.
+	// #685: require OID equality so a coin+size fallback hit on a TP fill of
+	// the same size (lookup.OID != StopLossOID) doesn't masquerade as an SL
+	// fill and starve TP attribution.
 	if pos.StopLossOID > 0 {
-		if _, slFilled := resolveFee(sym, pos.StopLossOID, pos.Quantity); slFilled {
+		if lookup, slFilled := resolveFee(sym, pos.StopLossOID, pos.Quantity); slFilled && lookup.OID == pos.StopLossOID && lookup.FilledQty > 1e-9 {
 			return false
 		}
 	}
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #695, opened 2026-05-10
SKILL.md: add critical rule to always use scripts/update.sh

richkuo:
Adds a bolded core rule in SKILL.md requiring [update] phase: preflight
[update] previous binary version: v0.57.0
[update] phase: preflight done (0s)
[update] phase: pull for all go-trader updates instead of manual git pull + go build.\n\nManual steps cause asymmetric deploys (#642). This makes the rule unmissable at the top of the skill file.\n\nReference only: LLM model used: z.ai GLM-5.1


richkuo:
@claude sonnet

PR #701, opened 2026-05-11
fix: reject misplaced subcommands instead of silently starting daemon (#700)

### LINKED ISSUE #700: Bug: misplaced -config before manual-open starts full daemon instead of subcommand
## Problem

Running a CLI subcommand with `-config` **before** the subcommand name can silently start the full scheduler daemon instead of executing the subcommand.

Observed command:

```bash
./go-trader -config scheduler/config.json manual-open manual-eth --side long --margin 50
```

Expected:
- run `manual-open`
- place/queue the manual action
- exit

Actual:
- a second full bot process started
- it connected Discord and posted Hyperliquid summaries to the live channel
- this created confusing duplicate summary PIDs / duplicate live-bot behavior

## Root cause

Subcommand dispatch only checks `os.Args[1]`.

In `scheduler/main.go`, `manual-open` is only dispatched when it is the **first** argument after the binary:

```go
if len(os.Args) > 1 {
    switch os.Args[1] {
    case "manual-open":
        os.Exit(runManualOpen(os.Args[2:]))
    }
}
```

When the command is invoked as:

```bash
./go-trader -config scheduler/config.json manual-open ...
```

then `os.Args[1]` is `-config`, so the subcommand is not recognized. The process falls through to normal flag parsing and starts the main scheduler/Discord daemon.

## Impact

- accidental duplicate live bot instance
- live Discord summaries posted from the wrong PID/process
- operator confusion when debugging summary PIDs
- risk of duplicate notifications or worse if another subcommand is invoked this way

## Expected behavior

One of these should happen:

1. global flags are parsed first, then subcommands are dispatched robustly, **or**
2. if a known subcommand appears after global flags in an invalid position, the binary should **exit with an error**

It should never silently fall through into daemon mode.

## Suggested fix

- make subcommand parsing robust to global flags appearing before the subcommand, or
- explicitly detect misplaced known subcommands (`manual-open`, `manual-close`, `backfill`, `probe`, etc.) after flag parsing and fail hard with a usage error

## Repro

```bash
./go-trader -config scheduler/config.json manual-open manual-eth --side long --margin 50
```

## Notes

This was observed in production and explains why the Hyperliquid live channel showed summaries from two different PIDs: the real `go-trader-live.service` daemon and an accidentally spawned second daemon from malformed CLI invocation.

richkuo:
## Summary
- Subcommand dispatch in `scheduler/main.go` only checks `os.Args[1]`, so `./go-trader -config foo manual-open ...` slipped past the switch, `flag.Parse()` consumed `-config foo`, and the remaining args were silently dropped — booting a second scheduler/Discord daemon instead of running the requested subcommand. Caused duplicate live summaries (#700).
- Add `validateDaemonInvocation(flag.Args())` after `flag.Parse()`. Any leftover positional arg now exits 2 with a clear error; when the leftover matches a known subcommand (`init`, `export`, `manual-open`, `manual-close`, `backfill`, `probe`, `version`) the message names it and reminds the operator that subcommands must precede global flags.
- Pure helper + table-driven tests; a parity test guards `knownSubcommands` against drift from the dispatch switch.

## Test plan
- [x] `go test ./...` from `scheduler/` — all pass
- [x] Repro: `./go-trader -config scheduler/config.json manual-open manual-eth --side long --margin 50` now exits 2 with `subcommand "manual-open" must appear before any global flags …` instead of starting the daemon
- [x] `./go-trader version` and `./go-trader manual-open --help` still work

Closes #700

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -12,6 +12,40 @@ import (
 	"time"
 )
 
+// knownSubcommands lists the leading-position subcommand tokens dispatched in
+// main() before flag.Parse(). Kept in sync with that switch so
+// validateDaemonInvocation can detect a misplaced one and produce a useful
+// error instead of silently starting the daemon (#700).
+var knownSubcommands = []string{
+	"init",
+	"export",
+	"manual-open",
+	"manual-close",
+	"backfill",
+	"probe",
+	"version",
+}
+
+// validateDaemonInvocation rejects stray positional args that survived
+// flag.Parse() on the daemon path. The most common cause is a misplaced
+// subcommand, e.g. `./go-trader -config foo manual-open ...`, where
+// `-config foo` is consumed as a global flag and the remaining args are
+// silently dropped — without this guard the daemon would boot a second
+// scheduler instead of running the requested subcommand (#700).
+func validateDaemonInvocation(extra []string) error {
+	if len(extra) == 0 {
+		return nil
+	}
+	for _, a := range extra {
+		for _, sub := range knownSubcommands {
+			if a == sub {
+				return fmt.Errorf("subcommand %q must appear before any global flags (got: %v)", sub, extra)
+			}
+		}
+	}
+	return fmt.Errorf("unexpected positional arguments after flags: %v", extra)
+}
+
 func main() {
 	if len(os.Args) > 1 {
 		switch os.Args[1] {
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #703, opened 2026-05-11
fix(update): refuse early when scheduler/config.json missing (#702)

### LINKED ISSUE #702: update.sh probe fails when run from source clone without scheduler/config.json
## Problem

`scripts/update.sh` fails with a probe error when run from a bare source clone that doesn't contain a `scheduler/config.json`:

```
[update] phase: probe
probe: failed to load config scheduler/config.json: read config: open scheduler/config.json: no such file or directory
[update] FAIL phase=probe: go-trader.new probe rejected the freshly synced Python — refusing to swap
```

This blocks the standard update workflow when syncing from a source repo to deployment instances.

## Root cause

`runProbe` (`scheduler/probe_cmd.go:16,21-25`) defaults `--config` to `scheduler/config.json` and calls `LoadConfig`, which returns `read config: open ...: no such file or directory` (`scheduler/config.go:561`) when the file is absent. `probeCheckScripts(cfg)` then iterates `uniqueCheckScripts(cfg)` to invoke each configured check script with `--probe-only` (`scheduler/version_probe.go:58-72`) — so it genuinely needs a config to know which scripts to probe.

`scheduler/config.json` is gitignored, so a fresh source clone never has one. `update.sh:218` also reads it directly for `status_port`. The script was designed to run inside a deployment worktree, not a bare source clone.

## Impact

When operators run `update.sh` from the source repo instead of the deployment directory:
- The probe fails and the binary swap is refused
- Operators may improvise alternative sync methods (e.g. rsync) which can delete deployment-specific files like `scheduler/config.json`

## Suggested fix

**Preferred: preflight guard with guidance.** Detect missing `scheduler/config.json` in the preflight phase and fail fast with a clear "run from deployment dir" message before any git/uv mutation. Optionally accept `--config scheduler/config.example.json` as a fallback so source-clone builds can still run a degraded probe against the example config.

**Rejected: skip config-dependent probe when no config.** This makes the source-clone probe a no-op for argv-drift mismatches (e.g. `--strategy-refs`, `--probe-only`) — which is the most common failure mode the probe was added to catch (#683). The probe would only catch binary-level breakage from the source clone, leaving real argv/Python mismatches to surface per-deployment.

## Workaround

Run `update.sh` directly from the deployment directory:
```bash
cd /path/to/deployment && scripts/update.sh --restart
```

Or for multi-instance sync: build in the source repo, then run `update.sh` from each deployment instance.

richkuo:
## Summary

- Add a preflight guard to `scripts/update.sh` that detects a missing `scheduler/config.json` *before* any git pull / uv sync / build mutation, with operator-facing guidance pointing at the deployment directory and `scheduler/config.example.json`.

## Why

Running `update.sh` from a bare source clone used to fail in the **probe** phase with:

```
probe: failed to load config scheduler/config.json: read config: open scheduler/config.json: no such file or directory
[update] FAIL phase=probe: go-trader.new probe rejected the freshly synced Python — refusing to swap
```

By that point `git pull` + `uv sync` + `go build` had already run. Operators sometimes recovered by improvising rsync-based syncs that clobbered deployment-specific files like `scheduler/config.json`.

Failing in `preflight` keeps the tree untouched and tells the operator exactly what to do.

The skip-config-dependent-probe alternative was rejected: it makes the source-clone probe a no-op for `--strategy-refs` / `--probe-only` argv-drift mismatches, which is the failure mode the probe was added to catch (#683).

## Test plan

- [x] `bash -n scripts/update.sh` — syntax OK
- [x] Simulated missing-config run in a scratch repo: exits 1 in `preflight`, prints the guidance block, never reaches `pull` / `sync` / `build`
- [x] Simulated present-config run: guard passes, preflight proceeds to subsequent checks

Closes #702

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -76,12 +76,30 @@ fail() {
 # Phase 1: pre-flight — every check that should abort BEFORE any mutation.
 begin_phase preflight
 
+if ! command -v uv >/dev/null 2>&1; then
+    fail "uv not on PATH — install uv first (see CLAUDE.md → Setup)"
+fi
+
+go_bin=""
+if command -v go >/dev/null 2>&1; then
+    go_bin=$(command -v go)
+elif [[ -x /opt/homebrew/bin/go ]]; then
+    go_bin=/opt/homebrew/bin/go
+else
+    fail "go not on PATH and /opt/homebrew/bin/go missing"
+fi
+
+repo_root=$(git rev-parse --show-toplevel)
+cd "$repo_root"
+
 # scheduler/config.json is gitignored, so a bare source clone never has one.
 # Without it, `go-trader.new probe` later loads no strategies and fails with
 # "read config: open scheduler/config.json: no such file or directory" — but
 # only after git pull + uv sync + build have already mutated the tree (#702).
 # Refuse early so the operator runs update.sh from the deployment directory
 # instead of improvising rsync-based syncs that can clobber deployment state.
+# Check runs AFTER `cd "$repo_root"` so a subdirectory invocation (e.g. from
+# `scheduler/`) doesn't report a misleading "missing config" message.
 if [[ ! -f scheduler/config.json ]]; then
     cat >&2 <<EOF
 [update] scheduler/config.json not found in $(pwd).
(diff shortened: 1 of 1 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #707, opened 2026-05-11
feat(config): effective-config introspection (#704)

### LINKED ISSUE #704: Add clearer effective strategy config introspection for close/protection settings
## Summary

While debugging `hl-tema-eth-live` / `hl-rmc-eth-live`, it was easy to conclude that no TP close strategy was configured because plausible-sounding fields looked empty or missing:

- `close_strategy` (legacy pre-v13; migrated to `close_strategies` in `scheduler/config_migration.go:597`)
- `take_profit_atr_mult` (does not exist in codebase)
- `take_profit_pct` (does not exist in codebase)
- `tp_tiers` (exists only under top-level `manual_defaults` — `scheduler/config.go:141`; per-strategy tiers live in `close_strategies[i].params.tiers`)

But the actual loaded config was using:

```json
"close_strategies": [{"name": "tiered_tp_atr"}]
```

and the strategies also had on-chain protection active (`Position.StopLossOID`, `Position.TPOIDs`).

That led to a wrong diagnosis while investigating why an `SL filled` DM appeared for `hl-rmc-eth-live`.

## Problem

There is no single obvious / low-friction way to inspect a strategy's **effective** loaded close/protection config, and `LoadConfig` silently accepts unknown per-strategy keys — so typos and invented field names (e.g. `take_profit_atr_mult`) produce no warning.

`/status` (`scheduler/server.go:305`) exposes positions/risk but not `OpenStrategy`/`CloseStrategies`/SL source. No `inspect` subcommand exists (`scheduler/main.go:52-64`: `init|export|manual-open|manual-close|backfill|probe|version`).

In practice this means it is too easy for an operator (or an agent) to inspect the wrong field name or the raw JSON shape and come away with the wrong answer about:

- which close strategies are active
- whether TP logic is active
- where SL/TP orders came from
- which defaults were inherited vs explicitly set

## Expected

It should be hard to misread this. A quick inspection path should show the effective loaded values clearly, and config validation should flag field names that don't exist.

## Suggestions

Any one of these would help a lot:

1. Add an `inspect <strategy-id>` subcommand that prints the **effective** strategy config after load/migration/defaulting
   - open strategy (name + params)
   - close strategies (name + params per ref)
   - stop loss source/value (which of the five mutually-exclusive HL stop fields won, default vs explicit)
   - TP source/value (which close ref owns TP, resolved tier list)
   - inherited defaults vs explicit overrides

2. Startup logging should print the resolved strategy summary more explicitly
   - e.g. `close_strategies=[tiered_tp_atr] stop_loss_atr_mult=1.0 (default) tp_source=close_strategy`

3. Reject unknown per-strategy keys in `LoadConfig` (or warn) so typos / invented fields like `take_profit_atr_mult` fail loudly instead of being silently ignored. Also clarify in docs the difference between legacy `close_strategy` and current `close_strategies`.

4. Trade/DM alerts could optionally mention the protection source
   - e.g. `exchange-side SL (oid=...)` vs `close strategy exit`
   - Note: `tradeAlertExtras` already appends `| OID: <id>` on live fills — scope to avoid duplication.

## Why this matters

This wasn't just a docs nit: it directly caused a wrong live-trade explanation during incident/debug flow. The root cause was two-fold — hidden effective state AND no validation on invented field names.

## Repro context

- repo: `richkuo/go-trader`
- strategies: `hl-tema-eth-live`, `hl-rmc-eth-live`
- actual resolved config included `close_strategies: [{name: "tiered_tp_atr"}]`
- initial inspection of fields that either don't exist (`take_profit_atr_mult`, `take_profit_pct`) or live elsewhere (`tp_tiers` under `manual_defaults`) made it look like there was no TP close strategy configured

richkuo:
## Summary

Closes #704. Four observability levers so an operator (or agent) can never again misdiagnose a live trade by inspecting the wrong field — the failure mode that drove the incident on \`hl-rmc-eth-live\`.

1. **LoadConfig rejects unknown per-strategy keys.** Reflection-derived known-keys set from \`StrategyConfig\` json tags; surgical re-parse of \`strategies[]\` only (top-level forward-compat preserved). Targeted hints for the most common misses:
   - \`take_profit_*\` → \"TP lives under \`close_strategies\`\"
   - \`stop_loss_*\` typos → list of the five mutually-exclusive valid fields
   - legacy \`close_strategy\` / \`params\` → pointers to the co-located \`StrategyRef\` shape
2. **\`./go-trader inspect <strategy-id> [--all] [--json]\`** — prints the post-migration, post-default effective shape: open/close refs (+ params), SL field that won \`EffectiveStopLossPct\`'s priority cascade with value + explicit/default provenance, TP close ref + resolved tier list, direction provenance, per-field explicit-vs-default markers derived from re-reading the raw JSON.
3. **Startup logging** — one-line \`[config] <id>: type=X open=Y close=[..] sl=... tp=...\` summary per strategy after \`LoadConfig\`, sharing the same resolver as \`inspect\` so the two can never diverge.
4. **Trade-alert DM \`Source:\` line on close legs** — classifies the trigger as \`exchange SL\`, \`exchange TP1\`, \`close-strategy exit\`, \`external (peer / manual UI)\`, or \`circuit breaker\` based on \`Trade.Details\` prefix. Open legs skip the line.

## Sample output

\`\`\`
\$ ./go-trader inspect hl-rmc-eth-live
strategy hl-rmc-eth-live
  type:                perps
  platform:            hyperliquid
  open_strategy:       range_mean_revert
  close_strategies:    [tiered_tp_atr]
  direction:           long (default long)
  leverage:            5
  sizing_leverage:     5 (inherited from leverage)
  margin_mode:         isolated
  stop_loss:
    source:            stop_loss_atr_mult (default)
    value:             1× ATR (fixed, deferred until EntryATR stamped)
  take_profit:
    source:            close_strategies[0] tiered_tp_atr
    tiers:             [1× ATR @ 50%, 2× ATR @ 100%] — default (canonical [1×@50%, 2×@100%])
  max_drawdown_pct:    50
  interval_seconds:    300
\`\`\`

\`\`\`
\$ ./go-trader inspect --config /tmp/bad-config.json hl-rmc-eth-live
inspect: failed to load config /tmp/bad-config.json: config validation errors:
  strategy[hl-rmc-eth-live]: unknown field \"take_profit_atr_mult\" — TP logic lives under close_strategies (e.g. [{\"name\":\"tiered_tp_atr\",\"params\":{\"tiers\":[...]}}]); manual_defaults.tp_tiers only seeds defaults for type=manual
\`\`\`

## Test plan
- [x] \`go test ./...\` from \`scheduler/\` — all green (24.6s)
- [x] \`go build\` clean
- [x] \`./go-trader inspect\` smoke-tested against \`config.example.json\` — renders defaulted SL + 'TP none', flags hl-momentum-btc's missing TP path
- [x] Unknown-key smoke: \`take_profit_atr_mult\` on a real-shaped config returns the operator-friendly hint
- [x] Startup summary verified locally
- [ ] Manual smoke on a live config in staging after merge

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -579,6 +579,12 @@ func LoadConfig(path string) (*Config, error) {
 	if err := json.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("parse config: %w", err)
 	}
+	// #704: flag unknown per-strategy fields (typos like `take_profit_atr_mult`)
+	// before applying defaults; json.Unmarshal silently drops them and would
+	// otherwise produce a struct indistinguishable from "no protection configured".
+	if unknownErrs := validateStrategyJSONKeys(data); len(unknownErrs) > 0 {
+		return nil, fmt.Errorf("config validation errors:\n  %s", strings.Join(unknownErrs, "\n  "))
+	}
 	if cfg.IntervalSeconds <= 0 {
 		cfg.IntervalSeconds = 600
 	}
--- a/scheduler/config_unknown_keys.go
+++ b/scheduler/config_unknown_keys.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+)
+
+// knownStrategyConfigKeys returns the JSON tag names declared on
+// StrategyConfig. Used by validateStrategyJSONKeys to flag operator typos
+// (e.g. `take_profit_atr_mult`) instead of silently dropping invented fields
+// that would otherwise produce a config that looks correct but is missing the
+// requested protection (#704).
+func knownStrategyConfigKeys() map[string]bool {
+	known := make(map[string]bool)
+	t := reflect.TypeOf(StrategyConfig{})
+	for i := 0; i < t.NumField(); i++ {
+		tag := t.Field(i).Tag.Get("json")
+		if tag == "" || tag == "-" {
+			continue
+		}
+		name := strings.SplitN(tag, ",", 2)[0]
+		if name != "" {
+			known[name] = true
+		}
+	}
+	return known
+}
+
+// unknownKeyHint returns a one-line suggestion for an unknown strategy field,
+// chosen by substring match against the most commonly mistyped categories. The
+// goal is to make the most frequent miss — TP/SL fields that look plausible
+// but never existed — fail loudly with actionable guidance (#704).
+func unknownKeyHint(key string) string {
+	lk := strings.ToLower(key)
+	switch {
+	case strings.Contains(lk, "take_profit") || strings.Contains(lk, "tp_tier") || lk == "tp_tiers":
+		return "TP logic lives under close_strategies (e.g. [{\"name\":\"tiered_tp_atr\",\"params\":{\"tiers\":[...]}}]); manual_defaults.tp_tiers only seeds defaults for type=manual"
+	case strings.HasPrefix(lk, "stop_loss") || strings.Contains(lk, "stoploss"):
+		return "valid SL fields: stop_loss_atr_mult, stop_loss_pct, stop_loss_margin_pct, trailing_stop_pct, trailing_stop_atr_mult (mutually exclusive)"
+	case lk == "close_strategy":
+		return "legacy pre-v13 field; use close_strategies (array of refs)"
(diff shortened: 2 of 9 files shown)


richkuo:
@claude

PR #713, opened 2026-05-12
fix(manual-open): accept interspersed flags and resolve --margin/--notional via mark fetch

### LINKED ISSUE #711: manual-open CLI breaks with documented arg order and --margin/--notional sizing resolves to zero
## Summary

`manual-open` currently has two linked issues in the same command path:

1. The documented usage shows `go-trader manual-open <strategy-id> [flags]`, but Go's standard `flag` parsing stops at the first positional arg, so commands like `go-trader manual-open manual-eth --side long --margin 50` fail with usage output instead of parsing the flags.
2. Even when using the working flag order workaround (`go-trader manual-open --side long --margin 50 manual-eth`), margin/notional sizing can still fail because `resolveManualSize()` is called with `price=0` before the execute path, which resolves to `size=0`, and Hyperliquid rejects the order with `--size must be > 0 unless --close-full-position is set`.

### Bug 1: documented arg order fails

```bash
./go-trader manual-open manual-eth --side long --margin 50
```

Actual result:

```
Usage: go-trader manual-open <strategy-id> [--side long|short] [--size N | --notional N | --margin N] [flags]
```

### Bug 2: margin sizing still fails even with workaround

```bash
./go-trader manual-open --side long --margin 50 manual-eth
```

Actual result:

```
error from HL: --size must be > 0 unless --close-full-position is set
```

## Root cause

- `runManualOpen()` uses Go's stdlib `flag.FlagSet`, which does not support interspersed flags after the first positional arg.
- In `manual.go`, the non-`recordOnly` path calls:

```go
resolveManualSize(*size, *notional, *margin, 0, sc.Leverage)
```

Since `price == 0`, the helper returns 0 before margin-based sizing can be computed.

## Expected behavior

- Either the CLI should truly support the documented `manual-open <strategy-id> [flags]` form, or the usage/help text should be changed to require flags before the strategy id.
- `--margin` and `--notional` should resolve to a non-zero size by fetching/using the current price before calling the HL execute path.

## Suggested fix

- Fix argument parsing so the documented invocation works.
- In the open path, compute sizing from the current mark/fill price before calling the execute script, instead of passing `price=0` into `resolveManualSize()`.
- Add tests covering:
  - `manual-open manual-eth --side long --margin 50`
  - `manual-open --side long --margin 50 manual-eth`
  - `--notional` and `--margin` both producing non-zero size

## Impact

This makes `manual-open` unreliable for normal operator use and forced a manual workaround using precomputed `--size` to open a live position.

richkuo:
Closes #711.

## Summary

- Reorder `manual-open` / `manual-close` args before `fs.Parse` so the documented `manual-open <strategy-id> --flag value` form works.
- Fetch the HL mid before `RunHyperliquidExecute` so `--margin` / `--notional` resolve to a non-zero coin qty instead of 0.
- Tests cover both fixes.

## Test plan
- [x] `go test ./.` passes
- [ ] `./go-trader manual-open manual-eth --side long --margin 50 --dry-run` renders non-zero qty
- [ ] `./go-trader manual-open --side long --margin 50 manual-eth` opens a live position

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 71 in / 27907 out | Cache: 3933836 read / 157987 written | Cost: $3.65

### DIFF
--- a/scheduler/manual.go
+++ b/scheduler/manual.go
@@ -35,6 +35,12 @@ func runManualOpen(args []string) int {
 	recordOnly := fs.Bool("record-only", false, "Register an existing fill without placing a new on-chain order")
 	dryRun := fs.Bool("dry-run", false, "Print planned action without placing order or mutating state")
 
+	// #711: stdlib flag.Parse stops at the first positional arg, so the
+	// documented `manual-open <strategy-id> --flag value` form fails to parse
+	// the trailing flags. Reorder to put the positional last so both
+	// orderings work.
+	args = reorderArgsForPositional(args, manualOpenBoolFlags)
+
 	if err := fs.Parse(args); err != nil {
 		return 2
 	}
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #717, opened 2026-05-12
fix(sl-after): cap SL replace qty at on-chain qty in runPostTPStopLossAdjustment

### LINKED ISSUE #714: [Critical] runPostTPStopLossAdjustment passes virtual qty to SL replace without on-chain cap
## Summary

After PR #710 shipped the live `sl_after` post-TP SL adjustment, `runPostTPStopLossAdjustment` at `scheduler/post_tp_sl.go:444` passes the virtual `pos.Quantity` straight into `runHyperliquidUpdateStopLossFunc` at `scheduler/post_tp_sl.go:530` without applying the `hlSLEffectiveQty` on-chain cap that every other SL placement site already uses.

## Why this is critical

The trailing-stop walker at `scheduler/main.go:1491` and the fixed-ATR SL arm at `scheduler/main.go:1549` both cap the SL size:

```go
slEffectiveQty, capped := hlSLEffectiveQty(result.Symbol, hlPosQty, hlOnChainAbsQty)
```

The `sl_after` cancel+replace path is the only outlier. Per CLAUDE.md:

> **SL size cap:** `hlSLEffectiveQty(coin,virtualQty,hlPositions)` caps placement at `min(virtualQty, onChainQty)` — prevents HL rejection after a manual TP shrinks on-chain qty.

This is the exact scenario `sl_after` targets — a tier just filled, on-chain qty is now below virtual qty, then we try to place a reduce-only trigger sized to virtual qty and HL rejects it.

## Reproduction

1. Open an HL perps position with `tiered_tp_atr` close ref + `sl_after` rules.
2. TP1 fills on-chain, shrinking on-chain qty.
3. Next cycle: `runPostTPStopLossAdjustment` fires, builds a replace order at `qty = pos.Quantity` (still the pre-TP virtual qty until reconciler runs), HL rejects.

## Fix

Thread `hlOnChainAbsQty map[string]float64` into `runPostTPStopLossAdjustment` and apply the same pattern as `main.go:1491-1495` before calling `runHyperliquidUpdateStopLossFunc`. Alternatively, fetch on-chain qty inline at the top of the function under the same RLock used to read `pos`.

## Acceptance

- Test: unit test that supplies `pos.Quantity=1.0` and on-chain qty `0.7`, asserts the subprocess receives `0.7`.
- Manual: trip a TP1 fill on a position with `sl_after`, verify the replace order succeeds.

Caught in retrospective review of #710.

richkuo:
## Summary

- Thread `hlOnChainAbsQty map[string]float64` into `runPostTPStopLossAdjustment` and apply `hlSLEffectiveQty` before the cancel+replace subprocess call
- Mirrors the identical cap already applied at the trailing-stop walker (`main.go:1491`) and fixed-ATR arm (`main.go:1549`)
- All three call sites in `main.go` (perps idle cycle, perps post-trade, manual) updated; `hlOnChainAbsQty` is already in scope at all three (built at `main.go:1168`)

## Why

After a TP tier fills, on-chain qty drops below virtual qty until the reconciler/protection-sync books the fill. `runPostTPStopLossAdjustment` was passing the stale virtual qty as the size for a reduce-only trigger order, which HL rejects as oversized. This is the exact scenario `sl_after` targets.

## Test plan

- `TestRunPostTPStopLossAdjustment_CapsAtOnChainQty` — asserts subprocess receives `0.7` when virtual=`1.0` / on-chain=`0.7`
- `TestRunPostTPStopLossAdjustment_NoCapWhenOnChainGEVirtual` — asserts no cap when on-chain matches virtual
- All existing `TestRunPostTPStopLossAdjustment_*` tests pass with `nil` map (no-op path)
- Full `go test .` suite green (24.8s)

Closes #714

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -1572,7 +1572,7 @@ func main() {
 							}
 							if hyperliquidIsLive(sc.Args) && result.Signal == 0 && hlPosQty > 0 {
 								runHyperliquidProtectionSync(sc, stratState, stateDB, result.Symbol, &mu, notifier, logger, "HL protection synced")
-								runPostTPStopLossAdjustment(sc, stratState, result.Symbol, price, cfg, &mu, notifier, logger)
+								runPostTPStopLossAdjustment(sc, stratState, result.Symbol, price, cfg, &mu, notifier, logger, hlOnChainAbsQty)
 							}
 							if hyperliquidIsLive(sc.Args) && result.Signal != 0 {
 								er, ok2 := runHyperliquidExecuteOrder(sc, result, price, hlCash, hlPosQty, hlPosSide, hlAvgCost, hlStopLossOID, hlTPOIDs, hlReconcileAll, notifier, logger)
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #718, opened 2026-05-12
fix(backtest): suppress same-bar SL hit after sl_after trigger bump (#715)

### LINKED ISSUE #715: [Critical] Backtester sl_after: SL hit check fires on same bar as trigger bump
## Summary

In the `sl_after` parity backtester shipped in PR #712, the end-of-bar SL hit check at `backtest/backtester.py:635-653` can fire on the same bar that `_maybe_apply_sl_after` just bumped `sl_trigger_px`. Live does not do this — the cancel+replace subprocess runs after the partial close is recorded and the new OID only starts protecting from the next cycle.

## Why this is critical

This produces backtest results that diverge from live behavior in exactly the scenario `sl_after` exists to model:

1. Bar N: TP tier fills, `_maybe_apply_sl_after` (called at `backtester.py:564`) bumps `sl_trigger_px` to (say) breakeven.
2. Bar N (same bar): `_sl_hit` at line 650 sees `sl_trigger_px > 0` against the bar's close and fires the SL — using the **bumped** trigger that wouldn't have been live yet.

The whole point of parity is that backtest PnL distributions for strategies using `sl_after` reflect production. Same-bar SL fires after a TP-driven bump are an artifact of the simulation, not the strategy.

## Fix

Set a `sl_after_just_applied` flag inside the branch at `backtester.py:564` where `_maybe_apply_sl_after` returned a trigger change. Guard the SL hit check at line 650 with `not sl_after_just_applied`, and clear the flag at the top of the next bar.

## Acceptance

- Regression test: position opens, TP tier fills on bar N, mark on bar N is below the new (breakeven) trigger — assert position is still open at end of bar N, fills SL at bar N+1 open.

Caught in retrospective review of #712.

richkuo:
## Summary

- Adds a per-bar `sl_after_just_applied` flag to the backtester loop, set when `_maybe_apply_sl_after` actually moves the SL trigger or the trail mult on a partial-close bar.
- Gates the end-of-bar block — **both the trail-walker HWM update and the SL hit check** at `backtest/backtester.py:660-685` — on the flag so a bump and an SL fire can't collapse into the same bar. Skipping the walker on the bump bar is also correct: live wouldn't walk HWM until the next cycle, and `_maybe_apply_sl_after`'s `trail_from_here` path already seeds the HWM at the partial-close fill price.
- Watermark-only advances (empty per-tier rule path) leave the trigger and trail mult untouched and therefore don't suppress, so unrelated SL hits on tier-clear bars still fire normally.
- Resets the flag at the top of every bar and defensively inside the full-close branch alongside the other post-TP SL state resets.

## Test plan

- [x] `test_backtester_sl_after_no_same_bar_fire_after_bump_long` — long-side TP1 → breakeven scenario closes one bar later than pre-fix behavior. Verified the test fails on a reverted gate (exit 2024-01-05) and passes with the fix (exit 2024-01-06).
- [x] `test_backtester_sl_after_no_same_bar_fire_after_bump_short` — short-side mirror of the same scenario; locks in symmetry against future refactors. Same fail-on-reverted-gate verification.
- [x] `test_backtester_sl_after_flag_clears_for_next_bar_long` — confirms the flag clears after the bump bar: bar N+1 SL hit fires normally when close drops below the new trigger.
- [x] Full pytest suite: `958 passed` (`.venv/bin/python3 -m pytest shared_strategies/ shared_tools/ platforms/ backtest/`).
- [x] Go suite: `ok trading-scheduler` (`go test ./...`).

Closes #715

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -497,13 +497,17 @@ class Backtester:
             signal = row["signal"]
 
             # Per-bar reset: when _maybe_apply_sl_after bumps the SL trigger on
-            # this bar, the end-of-bar SL hit check must skip — live places the
-            # new SL OID mid-cycle after the TP fill, and the backtester's
-            # bar-level granularity collapses that delay to zero. Without the
-            # gate, a bar that bumps SL to (say) breakeven and then closes
-            # below breakeven would fire SL on the same bar; live would not,
-            # because the bump and the close happen at different intra-bar
-            # moments. See #715.
+            # this bar, the end-of-bar block (both the trail-walker HWM update
+            # and the SL hit check) must skip — live places the new SL OID
+            # mid-cycle after the TP fill, and the backtester's bar-level
+            # granularity collapses that delay to zero. Without the gate, a
+            # bar that bumps SL to (say) breakeven and then closes below
+            # breakeven would fire SL on the same bar; live would not, because
+            # the bump and the close happen at different intra-bar moments.
+            # Skipping the walker on the bump bar is also correct: live
+            # wouldn't walk the HWM until the next cycle either, and the
+            # trail_from_here path inside _maybe_apply_sl_after already seeds
+            # the HWM at the partial-close fill price. See #715.
             sl_after_just_applied = False
 
             equity = cash + position * mark_price
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -496,6 +496,16 @@ class Backtester:
             mark_price = row["close"]
             signal = row["signal"]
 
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

richkuo:
@claude sonnet fix merge conflict and implement cosmetic note

richkuo:
@claude sonnet re-review

PR #712, opened 2026-05-12
feat(backtest): post-TP stop-loss adjustment parity (sl_after rules)

richkuo:
Mirrors the live #708 `sl_after` machinery in the Python backtester so backtests reflect the same SL bumps a tiered TP fires.

## Summary
- New `shared_strategies/close/post_tp_sl.py` mirrors the Go helpers (`parse_sl_after_rule`, `validate_sl_after_rule`, `parse_strategy_tp_sl_after_rules`, `compute_post_tp_stop_loss_trigger`, `validate_post_tp_stop_loss_rules`).
- `Backtester` accepts `stop_loss_atr_mult`/`stop_loss_pct`/`trailing_stop_*`/`strategy_type` for parity validation, seeds an SL trigger at open from the fixed-SL config, detects tier fills via `closed_qty/initial_qty` against parsed cumulative thresholds, applies the highest-cleared tier's rule, and walks an HWM each bar for `trail_from_here`.
- SL hit check is end-of-bar close-based → fills next bar's open (intra-bar trigger races not simulated, same as the rest of the bar-level engine). When `sl_after` is absent, behavior is unchanged.
- Validation rejects the same bad combos as the live validator (no fixed SL, combined with strategy-level trailing, `trail_from_here` on manual, `sl_after` keys placed under non-tiered_tp_atr* refs).

## Test plan
- [x] `.venv/bin/python3 -m pytest backtest/ shared_strategies/ shared_tools/` — 723 passed
- [x] `go -C scheduler test ./...` — passed
- [x] New `backtest/tests/test_post_tp_sl.py` covers pure helpers + long/short × breakeven/atr_offset/trail_from_here + validation + no-sl_after regression (59 tests).

Closes #709

🤖 

---
LLM: Claude Opus 4.7 (1M) | high | Harness: anthropics/claude-code-action@v1 | Tokens: 97 in / 63924 out | Cache: 9587084 read / 224275 written | Cost: $7.79

PR #710, opened 2026-05-11
feat(hl-protection): post-TP stop-loss adjustment (sl_after rules) (#708)

richkuo:
## Summary

Closes #708.

Adds an `sl_after` rule on tiered TP close strategies (`tiered_tp_atr` / `tiered_tp_atr_live`) that adjusts the remaining position's stop-loss when a TP tier fills. Strategy-level default with optional per-tier override.

## Config

```json
{
  "open_strategy": {"name": "triple_ema_bidir"},
  "close_strategies": [{
    "name": "tiered_tp_atr_live",
    "params": {
      "sl_after": "breakeven",
      "tiers": [
        {"atr_multiple": 2, "close_fraction": 0.5},
        {"atr_multiple": 3, "close_fraction": 1.0, "sl_after": {"atr_mult": 0.25}}
      ]
    }
  }]
}
```

## Modes

| Mode | Shape | Behavior |
|---|---|---|
| breakeven | `"breakeven"` | SL = `AvgCost` |
| signed ATR offset | `{"atr_mult": N}` | SL = `AvgCost ± N·EntryATR`; sign by side, +N = toward profit |
| trail_from_here | `{"trail_from_here": {"atr_mult": M}}` | Cancel fixed SL; trail at `M·EntryATR` behind high-water from this point |

Scope: HL perps + manual. `trail_from_here` is rejected on manual in v1 (trailing walker is perps-only today).

## Validation

`validatePostTPStopLossRules` rejects at load time:
- `sl_after` combined with `trailing_stop_atr_mult` or `trailing_stop_pct` (trailing already walks the SL — would race)
- `sl_after` without any fixed SL configured (nothing to adjust)
- malformed shapes (unknown kind, missing `atr_mult`, etc.)
- `trail_from_here` on `type=manual`

## Execution

After each `runHyperliquidProtectionSync` (both perps and manual call sites), the new `runPostTPStopLossAdjustment`:

1. **RLock** — find the highest cleared TP tier above `pos.SLAdjustedTiersProcessed` (gated on partial close having occurred, so unarmed tiers don't trigger).
2. Apply the **highest cleared tier's** rule (last-wins for multi-tier same-cycle fills).
3. **No-lock subprocess** — cancel+replace the SL OID via `RunHyperliquidUpdateStopLoss` (reuses the trailing-stop primitive).
[...]

PR #719, opened 2026-05-12
sl_after / inspect follow-ups from #707, #710, #712 review

### LINKED ISSUE #716: [Important] sl_after / inspect follow-ups from #707, #710, #712 review
## Summary

Four important-but-not-critical findings from the retrospective review of PRs #707, #710, and #712. Bundled into one issue per repo workflow preference; each numbered item can be its own commit in the PR.

---

### 1. `validateHotReloadStateCompatible` does not guard `sl_after` shape changes (#710)

`scheduler/config_reload.go:303` blocks `trailing_stop_atr_mult` / `stop_loss_atr_mult` nil↔positive while a position is open, but adding or removing `sl_after` rules — or transitioning into a `trail_from_here` rule — mid-position is not gated. A SIGHUP that introduces a `trail_from_here` rule on an open fixed-SL position would stamp `pos.PostTPTrailingATRMult` on the next cleared tier and engage the trailing walker without the usual config-level mode validation the open respected.

**Fix:** Extend `validateHotReloadStateCompatible` to reject `sl_after` rule additions/removals/mode changes when any position is open under that strategy.

---

### 2. `findHighestClearedTier` cannot distinguish filled-tier from never-armed-tier (#710)

`scheduler/post_tp_sl.go:416-430` returns any tier index where `tpOIDs[i] == 0`. A tier that was never successfully armed by `runHyperliquidProtectionSync` (transient subprocess failure on first cycle) has `tpOIDs[i] == 0` indistinguishably from a tier that filled. After any partial close (e.g. from a close-evaluator, not a TP), the top-level partial gate at line 479 passes and the function returns tier 0 — triggering an SL adjustment for a TP that never fired.

**Fix:** Track tier-armed state separately (e.g. add a "was ever non-zero" bit per tier, or only consider tier `i` cleared if a prior snapshot saw `TPOIDs[i] != 0`). Alternatively, gate on observed tier-fill detection (reconciler's `tp_filled_externally[]`) instead of current zero-OID state.

---

### 3. Backtester SL hit check missing `_tp_tier_thresholds` guard (#712)

`backtest/backtester.py:650` runs `_sl_hit` gated only on `sl_after_active`, but `_maybe_apply_sl_after` at line 555 is gated on `sl_after_active and self._tp_tier_thresholds`. If `sl_after` is configured but `parse_tp_tier_close_fractions` returns `[]` (no tiers, or all have zero `close_fraction`), `_maybe_apply_sl_after` never runs but the initial seeded `sl_trigger_px` can still fire spuriously.

**Fix:** Add `self._tp_tier_thresholds` to the line-650 guard, or move the SL hit check inside the `elif sl_after_active and self._tp_tier_thresholds:` block.

---

### 4. Paper-mode trailing SL misclassified as `"exchange SL"` (#707)

`scheduler/discord.go:1191` `tradeAlertCloseSource` matches `"stop loss close"` first and returns `"exchange SL"`, but `recordPerpsStopLossClose(..., "trailing_stop_loss_paper", ...)` at `scheduler/main.go:1474` produces Details starting with `"Stop loss close ETH, PnL: ..."` — which means operators see a paper trailing-SL labeled as an exchange-triggered SL.

**Fix:** Either add a `strings.Contains(d, "trailing")` branch ahead of the `"stop loss close"` match, or change the paper-trailing details prefix to include "Trailing" so it sorts into its own bucket.

---

Caught in retrospective review of PRs #707, #710, #712.

richkuo:
Closes #716.

Four findings from the retrospective review of PRs #707, #710, #712, each as its own commit.

## 1. SIGHUP hot-reload guard for `sl_after` shape changes

`validateHotReloadStateCompatible` (`scheduler/config_reload.go`) now rejects an in-flight reload that adds, removes, or changes the kind/mult of any `sl_after` rule while a position is open under that strategy. Without this guard, the new rule would engage on the next cleared TP tier — and for `trail_from_here`, would hand control to the trailing walker — without the validation the open respected. Numeric changes are still allowed when the strategy is flat.

`tierSLAfterRules.EqualForReload` compares the parsed strategy-level default plus per-tier rules, ignoring trailing empty `PerTier` entries so equivalent shapes compare equal.

## 2. Distinguish never-armed from filled TP tiers

`findHighestClearedTier` previously returned any tier where `pos.TPOIDs[i] == 0`. A tier whose first protection-sync placement failed transiently (still 0, never armed) was indistinguishable from a tier that fired and was zeroed by Python's `tp_filled_externally` signal. A subsequent non-TP partial close (close-evaluator) would then falsely advance the `sl_after` watermark and trigger an SL adjustment for a TP that never fired.

`Position.TPArmedTiers []bool` now records which tiers have been observed with a positive OID at least once. `applyHyperliquidProtectionSync` stamps `armed[i] = true` whenever Python echoes a positive OID, and treats `TPFilledExternally` as proof-of-arming. `findHighestClearedTier` requires `armed[i] == true` before considering a zero OID as cleared.

A new `tp_armed_tiers_json` SQLite column persists the slice; legacy rows get a conservative backfill (`armed[i] = oid[i] > 0`) so an in-flight pre-upgrade position never sees a tier go from OID=0 → cleared without first arming in a later cycle.

## 3. Backtester SL seed gate

In `Backtester.run`, the entry-time SL seed was gated on `sl_after_active` alone, while the tier-fill adjuster `_maybe_apply_sl_after` was gated on `sl_after_active AND self._tp_tier_thresholds`. When `parse_tp_tier_close_fractions` returned `[]` (degenerate tier config), the seed installed a phantom fixed-SL trigger that the post-TP machinery could never adjust — diverging silently from live, where the fixed SL is placed by `runHyperliquidProtectionSync` independently of `sl_after`.

The fix gates the seed on `self._tp_tier_thresholds` (chosen over the issue's proposed gate-the-hit fix because it removes the phantom state at the source rather than letting it linger and be filtered later). The hit check already short-circuits on `sl_trigger_px <= 0`, so no further changes are needed.

## 4. Distinct labels for paper / trailing SL closes

Paper-mode trailing SL closes wrote `"Stop loss close ..."` to `Trade.Details`, which `tradeAlertCloseSource` matched as `"exchange SL"` — misleading operators into thinking the exchange had fired a resting trigger when no exchange was involved at all. The same misclassification affected paper ATR SL closes and live trailing-SL immediate fills (broader scope than the issue called out; fixed in one shot since the prefix string is the load-bearing signal).

`stopLossCloseDetailsPrefix(reason)` now picks the prefix based on the internal reason code:

| Reason | Details prefix | Alert label |
|---|---|---|
| `trailing_stop_loss_paper` | `Paper trailing SL close` | `paper trailing SL` |
| `trailing_stop_loss_immediate` | `Trailing SL close` | `trailing SL` |
| `stop_loss_atr_paper` | `Paper SL close` | `paper SL` |
| everything else | `Stop loss close` | `exchange SL` |

## Test plan

- [x] `go -C scheduler test ./...` — full suite green (24.6s)
- [x] `.venv/bin/python3 -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 956 passed
- [x] Each commit verified to build independently (`go -C scheduler build ./...`)
- [x] Regression test per item:
  - Item 1: `TestApplyHotReloadConfigRejectsSLAfterAddWithOpenPosition`, `TestApplyHotReloadConfigAllowsSLAfterAddWhenFlat`, `TestApplyHotReloadConfigRejectsSLAfterModeChangeWithOpenPosition`
  - Item 2: `TestFindHighestClearedTier_NeverArmedSkipped`, `TestApplyHyperliquidProtectionSyncStampsTPArmedTiers`, `TestRunPostTPStopLossAdjustment_SkipsNeverArmedTier`
  - Item 3: `test_backtester_sl_after_does_not_seed_when_no_tier_thresholds`
  - Item 4: `TestStopLossCloseDetailsPrefix`, expanded `TestTradeAlertCloseSourceClassification`

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -588,7 +588,16 @@ class Backtester:
                     avg_cost = effective_price
                     initial_quantity = shares
                     entry_atr_value = self._stamp_entry_atr(atr_series, idx, effective_price)
-                    if sl_after_active:
+                    # #716 item 3: seed the SL trigger only when sl_after has
+                    # usable tier thresholds. Without thresholds, the post-TP
+                    # adjustment machinery never fires (`_maybe_apply_sl_after`
+                    # is gated on `self._tp_tier_thresholds`), so a seeded-
+                    # then-never-adjusted trigger would represent a phantom
+                    # fixed SL the rest of the engine doesn't actually
+                    # simulate. Mirrors the live shape, where the fixed SL is
+                    # placed by `runHyperliquidProtectionSync` independently
+                    # of `sl_after` configuration.
+                    if sl_after_active and self._tp_tier_thresholds:
                         sl_trigger_px = self._initial_sl_trigger(
                             "long", avg_cost, entry_atr_value,
                         )
@@ -608,7 +617,7 @@ class Backtester:
                     avg_cost = effective_price
                     initial_quantity = shares
                     entry_atr_value = self._stamp_entry_atr(atr_series, idx, effective_price)
-                    if sl_after_active:
+                    if sl_after_active and self._tp_tier_thresholds:
                         sl_trigger_px = self._initial_sl_trigger(
                             "short", avg_cost, entry_atr_value,
                         )
--- a/backtest/tests/test_post_tp_sl.py
+++ b/backtest/tests/test_post_tp_sl.py
@@ -801,3 +801,47 @@ def test_backtester_sl_after_defers_when_sl_unarmed():
     assert processed == 0
     assert trail is None
     assert hwm == 0.0
+
+
+def test_backtester_sl_after_does_not_seed_when_no_tier_thresholds():
+    """#716 item 3 regression: a degenerate sl_after config where
+    `parse_tp_tier_close_fractions` returns [] (e.g., all tiers have
+    close_fraction=0, which the parser drops) must NOT seed a phantom
+    fixed-SL trigger at open. Without tier thresholds the post-TP
+    adjustment machinery never fires; a seeded-then-never-adjusted SL
+    would represent a fixed-SL behavior the rest of the engine doesn't
+    actually simulate.
+
+    Construct a strategy with tiers whose close_fractions are all zero
+    so the parser drops them, then verify the SL never installs.
+    """
+    df = _df_open_then_hold(
+        opens=[100, 100, 100, 80, 80],
+        closes=[100, 100, 80, 80, 80],
+        atrs=[10, 10, 10, 10, 10],
+    )
+    bt = Backtester(
+        initial_capital=1000, commission_pct=0, slippage_pct=0,
+        platform="hyperliquid", strategy_type="perps",
+        stop_loss_atr_mult=1.0,
+        close_strategies=[{
+            "name": "tiered_tp_atr",
+            "params": {
+                "sl_after": "breakeven",
+                # All zero close_fractions are dropped by the parser →
+                # _tp_tier_thresholds == [].
+                "tiers": [
+                    {"atr_multiple": 1.0, "close_fraction": 0.0},
+                    {"atr_multiple": 2.0, "close_fraction": 0.0},
+                ],
+            },
+        }],
+    )
+    # Pre-condition: parser returned no usable thresholds.
+    assert bt._tp_tier_thresholds == []
+    result = bt.run(df, save=False)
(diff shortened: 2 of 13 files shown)


richkuo:
@claude

PR #727, opened 2026-05-12
Support custom systemd units in the update restart path

### LINKED ISSUE #726: Enhancement: make update.sh restart path accept explicit service name
## Summary
`scripts/update.sh --restart` assumes the default service name, which is unsafe for multi-instance deployments.

We have live/paper instances using custom systemd unit names, for example:

- `go-trader-live.service`
- `go-trader-paper-testing.service`
- `go-trader-paper-hl-btc.service`
- `go-trader-paper-hl-eth.service`
- `go-trader-paper-hl-bnb.service`
- `go-trader-paper-hl-sol.service`

The update script should support an explicit service name for restart/verify/rollback instead of assuming the canonical `go-trader` unit.

## Problem
The current safe update path for custom service names is awkward:

```bash
bash scripts/update.sh
systemctl restart go-trader-live.service
curl -s localhost:8099/health
```

Using `scripts/update.sh --restart` can restart or verify the wrong service when the deployment is not named `go-trader.service`.

## Proposal
Add a service-name override, for example:

```bash
bash scripts/update.sh --restart --service go-trader-live.service
```

and/or env fallback:

```bash
GO_TRADER_SERVICE=go-trader-live.service bash scripts/update.sh --restart
```

Backward-compatible default should remain the current/default service name.

## Required behavior
The selected service name should be used consistently for:

- `systemctl restart`
- `systemctl is-active`
- `systemctl show -p MainPID`
- post-restart health/version verification
- rollback restart/verification path

## Additional scope (from validation)

1. **`scheduler/updater.go:137`** — `restartSelf()` hardcodes `systemctl restart go-trader` the same way. The in-process auto-update path (`applyUpgrade` → `restartSelf`) will restart the wrong unit on multi-instance hosts. Fix in the same PR.
2. **Shared env var across both call sites** — `applyUpgrade` invokes `scripts/update.sh` without `--restart` (`scheduler/updater.go:97`), so the proposed `--service` flag only helps operator invocations. Reading `GO_TRADER_SERVICE` from both `scripts/update.sh` and `restartSelf` keeps the two paths consistent from a single source of truth.
3. **Naming nit** — prefer `--unit` over `--service` to match `systemctl`'s vocabulary (`.service` is one of several unit types). Non-blocking.

## Why this matters
Multi-instance deployments are now common. A hardcoded/default restart target is a footgun: the binary can be built correctly but the wrong unit may be restarted or verified.

richkuo:
## Summary
- Added `GO_TRADER_SERVICE` and `--unit`/`--service` overrides to `scripts/update.sh` so restart, active checks, PID verification, and rollback all target the selected systemd unit.
- Updated the in-process auto-update restart path in `scheduler/updater.go` to use the same environment-driven unit selection.
- Added tests for the unit resolver and expanded update-message coverage to mention the custom-unit override.

Closes #726

## Testing
- `bash -n scripts/update.sh`
- `bash scripts/update.sh --help`
- `bash scripts/update.sh --unit --restart` (verifies invalid usage is rejected)
- `bash scripts/update.sh --unit '   ' --restart` (verifies whitespace-only units are rejected)
- `bash scripts/update.sh --service=--restart` (verifies flag-looking unit names are rejected)

### DIFF
--- a/scheduler/updater.go
+++ b/scheduler/updater.go
@@ -11,6 +11,8 @@ import (
 	"time"
 )
 
+const defaultGoTraderSystemdUnit = "go-trader"
+
 // checkForUpdates uses git fetch to check for upstream changes. If new commits are found
 // (and the remote hash differs from lastNotifiedHash), it sends a Discord channel notification
 // and, if OwnerID is configured, a DM offering to auto-upgrade.
@@ -134,7 +136,7 @@ func restartSelf() error {
 	// Try systemctl first (Linux/systemd deployments).
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	if err := exec.CommandContext(ctx, "systemctl", "restart", "go-trader").Run(); err == nil {
+	if err := exec.CommandContext(ctx, "systemctl", "restart", updateSystemdUnitName()).Run(); err == nil {
 		// systemd will SIGTERM this process; give it time to do so.
 		time.Sleep(30 * time.Second)
 	}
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #748, opened 2026-05-13
Run Python subprocesses through uv

### LINKED ISSUE #739: Run Python subprocesses via `uv run` instead of `.venv/bin/python3`
## Motivation

Today every Python subprocess the scheduler spawns goes through `.venv/bin/python3` (see `scheduler/executor.go:136` and the `runPythonReadOnly` / `runPythonSideEffect` wrappers). `uv` is only used at setup/CI time via `uv sync`. Standardizing on `uv run` as the runtime entry point would:

- Make `uv.lock` / `pyproject.toml` the single source of truth for what actually runs, instead of relying on `.venv` having been kept in sync.
- Remove the worktree friction documented in `CLAUDE.md` ("`.venv` not copied — use `<main-repo>/.venv/bin/python3`"). Worktrees would just work.
- Match the package-manager direction already in `scripts/update.sh` (which calls `uv sync` on every deploy).

## Scope

Replace `.venv/bin/python3 <script> <args...>` with `uv run --no-sync python <script> <args...>` (or equivalent) across the runtime path:

- `scheduler/executor.go` — primary call site at `executor.go:136`; covers every check / execute / close / sync-protection script.
- `scheduler/version_probe.go` — startup probe must use the same invocation so `--probe-only` matches runtime behavior.
- `scheduler/backfill_hl_fees.go` — `go-trader backfill hl-fees` path.
- `scripts/update.sh` — preflight / probe / verify steps that shell Python directly (keep `uv sync` where it is).
- `CLAUDE.md`, `SKILL.md`, `README.md` — docs that reference `.venv/bin/python3` or the worktree workaround.
- Test stubs in `scheduler/executor_test.go`, `scheduler/probe_cmd_test.go`, `scheduler/version_probe_test.go`, `scheduler/hyperliquid_stop_loss_test.go` — likely need `writeProbeStub` to emit a `uv`-shaped shim, or a test seam that mocks the executor command.

## Risks / open questions

1. **Startup latency per call.** `uv run` resolves the env on every invocation. With `pythonSemaphore=4` and per-strategy intervals as low as seconds in DD-warn band, the overhead needs to be measured. `--no-sync` should make this near-zero after the first call, but confirm.
2. **CWD drift.** `CLAUDE.md` already warns "`uv run pytest` may fail on CWD drift." Need to confirm `uv run` from the scheduler's working directory resolves the project deterministically (it should — repo root has `pyproject.toml` / `uv.lock`).
3. **Drain semantics.** `runPythonSideEffect` registers on `sideEffectWG` and relies on SIGTERM/SIGKILL reaching the Python process. `uv run` wraps the interpreter in a parent process — verify signals propagate (uv documents passing signals through, but worth a test against `--execute` scripts during a `systemctl restart`).
4. **Probe shape.** `probeOneCheckScript` builds the exact argv that will run in production. The probe must invoke `uv run` too, otherwise startup compatibility validation drifts from runtime.
5. **Systemd unit / PATH.** Need to confirm `uv` is on `$PATH` for the service user, or pin an absolute path the same way `.venv/bin/python3` is hard-coded today.

## Acceptance criteria

- [ ] No runtime call site spawns `.venv/bin/python3` directly. (`rg "\.venv/bin/python3" -t go` returns only deleted/historical refs.)
- [ ] `go test ./...` passes.
- [ ] `scripts/update.sh` still preflights and probes successfully on a clean checkout.
- [ ] One full live restart on a deployed host shows drain completes within `shutdownDrainCap=15s` (no SIGKILL escalation).
- [ ] Worktrees can run `./go-trader --once` without a symlinked `.venv`.
- [ ] Docs (`CLAUDE.md`, `SKILL.md`, `README.md`) updated; the worktree `.venv` caveat removed.

## Out of scope

- Removing `uv sync` from `scripts/update.sh` — keep it; it primes the environment cache.
- Changing pytest invocation (kept as `.venv/bin/python3 -m pytest` per current docs unless `uv run pytest` CWD issue is also fixed here).

richkuo:
## Summary
- Route scheduler Python subprocesses and startup probes through `uv run --no-sync python`.
- Update update-script port parsing, probe test shims, and agent docs for the uv runtime path.

## Test plan
- `go -C scheduler test ./...`
- `rg '\.venv/bin/python3' --glob '*.go'`
- `git diff --check`

Closes #739

Model: Composer 2
Effort: high
Harness: Cursor

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,6 +4,10 @@
 - Go 1.26.2 (`brew install go@1.26` on macOS); use full binary path if not on PATH (e.g. `/opt/homebrew/bin/go`).
 - Python runs through `uv run --no-sync python`; deps via `uv` (`pyproject.toml` / `uv.lock`).
 - Git worktrees work without a copied virtualenv; run `uv sync` once per checkout and ensure `uv` is on PATH.
+- Optional: set `GO_TRADER_UV` to an absolute path when `uv` is not discoverable via `PATH` (e.g. custom install); read in `scheduler/python_cmd.go`.
+- **systemd:** `go-trader.service` and `systemd/go-trader@.service` set `PATH` (includes `%h/.local/bin`) and `UV_CACHE_DIR` under `scheduler/` so `uv run` works under `ProtectSystem=strict`. `/opt/go-trader/.env` (or per-instance `.env`) may override `PATH`, `UV_CACHE_DIR`, or set `GO_TRADER_UV`.
+- Optional: set `GO_TRADER_UV` to an absolute path when `uv` is not discoverable via `PATH` (e.g. custom install); read in `scheduler/python_cmd.go`.
+- **systemd:** `go-trader.service` and `systemd/go-trader@.service` set `PATH` (includes `%h/.local/bin`) and `UV_CACHE_DIR` under `scheduler/` so `uv run` works under `ProtectSystem=strict`. `/opt/go-trader/.env` (or per-instance `.env`) may override `PATH`, `UV_CACHE_DIR`, or set `GO_TRADER_UV`.
 
 ## Setup
 - `uv sync` — install Python deps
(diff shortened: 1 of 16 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #751, opened 2026-05-13
fix(hl): do not re-place consumed TP tiers after TP1 fill (#749)

### LINKED ISSUE #749: Hyperliquid protection sync re-adds TP after TP1 partial fill
## Summary

After TP1 partially fills on Hyperliquid, go-trader reconciles the partial close correctly, but a later protection sync appears to re-add a new TP order into the first TP slot for the remaining half-position.

## What happened

Strategy: `hl-rmc-eth-live`
Symbol: `ETH`

Observed sequence:

- `02:32:07` protection sync logged:
  - `TP1 OID=422857262895 already filled on-chain; not re-placing — reconciler will book the close`
  - `HL protection synced (sl_oid=422857235647 tp_oids=[0 422857285910])`
- `02:32:14` reconciler logged:
  - `External partial close reconciled 0.220000 @ $2294.6500, remaining 0.220000`
- `02:37:16` later protection sync logged:
  - `HL protection synced (sl_oid=422857235647 tp_oids=[423412460553 422857285910])`

So the original TP1 fill was recognized, and the remaining position size was reduced correctly, but a subsequent sync inserted a new TP order `423412460553` into the TP1 slot.

## Why this looks wrong

Expected behavior after TP1 fill on a 2-tier TP setup:

- the filled TP1 should stay retired
- the remaining position should keep only the still-valid remaining TP tier(s)
- protection sync should not recreate a replacement TP1 unless that is explicitly intended by design

Current behavior suggests the sync/rebuild logic may be rehydrating tiers from config without respecting that tier 1 has already been consumed.

## Impact

This can make it look like TP1 is still active after it already filled, or reintroduce an unintended TP level for the remaining position.

## Relevant evidence

Trade row recorded correctly:

- `2026-05-13T02:32:14Z` sell `0.22` @ `2294.65`
- details: `External partial close @ mark 0.220000, PnL: $4.65 (fee $0.07)`
- exchange_order_id: `422857262895`

Current in-memory position after sync showed:

- remaining quantity: `0.22`
- `tp_oids: [423412460553, 422857285910]`

## Suggested fix area

Likely in Hyperliquid protection sync / tier reconciliation logic:

- when a TP tier is confirmed filled and reconciled, mark that tier consumed in a way later syncs respect
- do not recreate earlier TP tiers for the remaining position unless explicitly intended
- if recomputing protection from remaining qty, ensure consumed tiers stay consumed

Reference only: LLM model used: openai-codex/gpt-5.5

richkuo:
## Summary
Fixes #749: after a partial TP1 fill, a later Hyperliquid protection sync was re-creating a TP in the first slot because OID `0` was interpreted as "never placed" while tier sizes were recomputed from the reduced position.

## Approach
- Extend the protection plan / subprocess contract with `TPArmedTiers` (`--tp-armed-tiers-json`), padded to tier count (mirrors `TPOIDs`).
- In `run_sync_protection`, when `prev_oid <= 0` and the tier is armed, skip placement (consumed tier per #716).

## Testing
- `go test -count=1 ./... -short` (scheduler)
- `pytest shared_scripts/test_check_hyperliquid.py::TestSyncProtection`

LLM: Composer | high | Harness: Cursor agent

### DIFF
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strconv"
 	"syscall"
 	"time"
@@ -370,7 +371,9 @@ func RunHyperliquidUpdateStopLoss(script, symbol, side string, size, triggerPx f
 	return parseHyperliquidUpdateStopLossOutput(stdout, string(stderr), err)
 }
 
-func RunHyperliquidSyncProtection(script, symbol, side string, size, avgCost, entryATR, stopLossATRMult float64, tiers []hlProtectionTier, stopLossOID int64, tpOIDs []int64, tpArmedTiers []bool) (*HyperliquidProtectionSyncResult, string, error) {
+// buildHyperliquidSyncProtectionArgv builds argv for check_hyperliquid.py
+// --sync-protection (used by RunHyperliquidSyncProtection and tests).
+func buildHyperliquidSyncProtectionArgv(symbol, side string, size, avgCost, entryATR, stopLossATRMult float64, tiers []hlProtectionTier, stopLossOID int64, tpOIDs []int64, tpArmedTiers []bool) []string {
 	args := []string{
 		"--sync-protection",
 		fmt.Sprintf("--symbol=%s", symbol),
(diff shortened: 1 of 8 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #755, opened 2026-05-13
fix(hl): protection-fill DM from hlAttemptCloseFromTPFills (#754)

### LINKED ISSUE #754: TP2 fill: no DM alert, wrong tier attribution, SL cancel missed
## Validation (code review, current `main`)

Cross-checked against `scheduler/hyperliquid_balance.go` and related paths.

### Still accurate — narrow bug

**Missing owner DM when the flat is booked in `hlAttemptCloseFromTPFills`.** `reconcileHyperliquidPositionsForStrategy` always calls `tryBookSoleOwnerTPFill` first; when it **returns `false`**, reconcile continues into `reconcileHyperliquidPositionsWithResolver`, which may call `hlAttemptCloseFromTPFills` (~517–526). That function has **no** `pendingAlerts` / `ProtectionFillAlert` path (~1168–1223), unlike `tryBookSoleOwnerTPFill` (~441–457) and shared-coin Detector 3 (~1017+).

Typical window: on-chain already flat while state still has a **positive TP OID** (e.g. final tier not yet zeroed by `applyHyperliquidProtectionSync`), so the full-close branch in `tryBookSoleOwnerTPFill` bails (~352–355) but `hlAttemptCloseFromTPFills` can still attribute from `userFills`. Other early `false` returns (e.g. missing `EntryATR` / `AvgCost` ~325–329) hit the same resolver path.

**Imprecise phrasing to avoid:** “`tryBookSoleOwnerTPFill` did not run first” — it **runs first**; say it **returned `false`**.

**Fix direction:** thread `*[]ProtectionFillAlert` into the path that reaches `hlAttemptCloseFromTPFills` (likely `reconcileHyperliquidPositionsWithResolver` today has no such parameter ~466 — not only the inner function).

### Claim 2 — not a current bug as framed

SL trigger-price mis-attribution for TP-flattened closes is addressed by `hlAttemptCloseFromTPFills` (#673 / #675) and the SL fallback’s **`slConfirmed`** gate: `lookup.OID == statePos.StopLossOID` with real fill qty (~528–556, #685). The issue’s “open SL misfire / book at SL trigger” risk does not match current code for that scenario.

### Claim 3 — wrong for TP2 full-close + all TP OIDs zero

`hyperliquidClearedTPTier` returns **`len(tpOIDs)-1` (last tier)** when **all** TP OIDs are zero **and** `closeQty` matches `pos.Quantity` (~1137–1142). Tests: `scheduler/protection_fill_alert_test.go` `TestHyperliquidClearedTPTier_TierIndex` (~232–235). The docstring caveat about “FIRST cleared index” applies when **`hasActive`** (at least one tier still has `oid > 0`) — e.g. TP1 cleared, TP2 still armed (~1134–1135), not the all-zeros full-close case.

### Claim 4 — unverified

No separate code defect traced; needs reproduction if summary still looks wrong after correcting claims 2–3.

---

### Bug Summary (partially superseded above)

When TP2 (final tier of a tiered TP strategy) fills on Hyperliquid, the reporter saw:

1. **No DM** for the final-tier fill — **still valid** when booking goes through `hlAttemptCloseFromTPFills` (no alert emission there).
2. **SL / trigger-price attribution** — **superseded**: current gates prevent the old #670-style SL booking when TPs flattened the position (see Validation).
3. **Wrong TP tier label when both TP OIDs are zero** — **superseded** for full-close + qty match: code attributes to **last** tier (see Validation).
4. **Win not in strategy summary** — **unverified**; may have been confusion from (2)/(3) or needs repro.

### Steps to Reproduce

1. Open a long on hl-rmc-eth-live (or any HL perps sole-owner strategy with `tiered_tp_atr*`)
2. Wait for TP2 to fill (final tier)
3. Observe: no DM for the TP2 fill (especially in the cycle-ordering window before protection-sync zeros the final `TPOIDs` entry)
4. Logs: if mis-attribution appears, capture `userFills` / reconcile branch — SL-at-trigger should not occur under current `slConfirmed` + `hlAttemptCloseFromTPFills` behavior

### Expected Behavior

- Owner receives a TP fill DM when a TP tier flatten is detected and booked (including the `hlAttemptCloseFromTPFills` path).
- Close at TP VWAP / correct tier label; SL path only when SL OID actually filled.
- Wins reflected in summaries consistent with `RecordTrade` / lifetime stats.

### Investigation Notes (historical)

- `tryBookSoleOwnerTPFill` (partial + full-close when applicable) appends `ProtectionFillAlert` when `pendingAlerts != nil`.
- `hlAttemptCloseFromTPFills` had no alert emission.
- Shared-coin Detector 3 path (~960–1018) emits alerts correctly.

richkuo:
Closes #754.

When `tryBookSoleOwnerTPFill` returns false (e.g. flat on-chain but a TP OID is still positive until protection-sync runs), the close is booked via `hlAttemptCloseFromTPFills`, which previously never appended `ProtectionFillAlert`. Plumb `pendingAlerts` through `reconcileHyperliquidPositionsWithResolver` and enqueue alerts after each successful TP fill booking.

Adds `TestSoleOwnerTPFullClose_TPOIDsLag_EmitsProtectionDM` regression coverage.

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -237,7 +237,7 @@ func fetchHyperliquidState(accountAddress string) (float64, []HLPosition, error)
 // cached resolver outside the lock and call
 // reconcileHyperliquidPositionsWithResolver.
 func reconcileHyperliquidPositions(stratState *StrategyState, sym string, positions []HLPosition, accountAddress string, logger *StrategyLogger) bool {
-	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger)
+	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger, nil)
 }
 
 // reconcileHyperliquidPositionsForStrategy is the production entry point with
@@ -276,11 +276,11 @@ func reconcileHyperliquidPositionsForStrategy(
 		// by the booker; the legacy reconciler's qty/side/avgCost resync will
 		// no-op because virtual now matches on-chain. Continue through it for
 		// idempotent housekeeping (multiplier migration, leverage seed).
-		reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger)
+		reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts)
 		return true
 	}
 
-	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger)
+	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts)
 }
 
 // tryBookSoleOwnerTPFill is the sole-owner TP attribution helper. Returns true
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

PR #757, opened 2026-05-14
fix(hl): confirm userFills SL OID on shared-coin reconcile (#756)

### LINKED ISSUE #756: Hyperliquid live mode can create phantom local positions and false SL alerts
### 1. Live execute trusts the SDK `filled` payload without hard confirmation

`shared_scripts/check_hyperliquid.py` parses `statuses[0].filled` from the place-order SDK response into `fill` (avg_px, total_sz, oid). It optionally calls `lookup_fill_fee_by_oid` to enrich fee data and **warns** when userFills returns nothing — but it does **not** fail the execute or clear the fill when userFills is empty; execution JSON is still emitted with whatever OID/size the SDK reported.

The Go path then applies that payload under lock: `executeHyperliquidResultDeferredOpen` in `scheduler/main.go` threads `execResult.Execution.Fill` into `ExecutePerpsSignalWithLeverageDeferredOpen` and stamps `pos.StopLossOID` / `pos.StopLossTriggerPx` from the same JSON when `stop_loss_oid` is present. There is **no** follow-up gate on `clearinghouseState` / position size before mutating durable state.

**Conclusion:** If the SDK returns a plausible `filled` object for an order that never actually landed (or OIDs that HL read APIs later treat as non-existent), local state can still record an open position and resting SL/TP OIDs. This matches the reported "phantom position" class of failure.

### 2. Sole-owner reconcile: SL close is gated on userFills OID match (#685)

In `scheduler/hyperliquid_balance.go`, `reconcileHyperliquidPositionsWithResolver` (non–shared-coin path) only books an SL close when:

`slConfirmed := useFillFee && lookup.OID == statePos.StopLossOID && lookup.FilledQty > 1e-9`

So for **sole-owner** strategies, an empty userFills window should **not** attribute a vanishing position to the SL OID; it should fall through toward external-close handling.

### 3. Shared-coin reconcile: gap — Detectors 1 and 2 still book SL + owner DM without the #685 gate

The same file's **shared-coin** "Detector 1" block (on-chain flat while virtual still shows positions) still calls `recordPerpsStopLossCloseWithFillFee` and appends a `ProtectionFillAlert` with `FillType: "SL"` for every peer that has `StopLossOID > 0`, **without** requiring `lookup.OID == pos.StopLossOID` the way the sole-owner vanish path does. Crucially, the `recordPerpsStopLossCloseWithFillFee` call and the subsequent `ProtectionFillAlert` append are not gated on `useFillFee` either — the SL DM fires even when userFills returned nothing at all (`hyperliquid_balance.go:836,841`).

**Detector 2** (SL-owner partial close, `hyperliquid_balance.go:934,941`) has the identical missing `lookup.OID == pos.StopLossOID` gate for the same reasons.

So a **phantom** local position that carried a bogus `StopLossOID` can still produce an "SL filled (oid=…)" owner DM on the next reconcile when:

- The coin is **shared** by 2+ HL strategies on the same wallet (so this code path runs), and  
- On-chain is flat (Detector 1 condition) — which is exactly true if the position never existed.

If the operator only runs one HL strategy on that coin, the false SL DM is less likely from this specific branch; the phantom **position** issue in §1 still stands.

### 4. Operator-reported HL API state

The report that read APIs show empty `userFills` / `historicalOrders` / positions while related OIDs behave as unknown is consistent with "SDK response and indexer disagree." This repository does not currently implement an `orderStatus` / `unknownOid`-style guard in-tree (no matches for those strings), so tightening confirmation would be new work.

---

## Summary

In live Hyperliquid trading, local state can be updated from the Python `--execute` path using the SDK `filled` structure, with userFills enrichment **best-effort only**. Reconciliation then tries to square local state with on-chain reality. The sole-owner SL path is partially defended (#685), but **shared-coin Detectors 1 and 2** can still book an SL close and emit an SL owner alert without proving the SL OID filled in userFills — including when userFills returned nothing at all.

## Root cause

Durable state and alerts are driven off **optimistic** execute/reconcile interpretations. There is no mandatory, same-session confirmation against independent read APIs (fills, order status, clearinghouse position) before committing an open + protection OIDs, and shared-coin SL attribution is weaker than the sole-owner SL gate.

## Expected behavior

Before recording a live open + SL/TP OIDs (and before treating flat-on-chain + local SL OID as an SL fill), confirm via at least one read path, e.g.:

- `userFills` / `userFillsByTime` contains the entry order OID (and protection OIDs as applicable), and/or  
- `clearinghouseState` / user state shows the expected position size and side, and/or  
- explicit order-status API shows the OID as filled (not unknown / missing).

If unconfirmed after a bounded retry, **do not** mutate local position state or persist protection OIDs; surface a desync / manual-review alert instead.

## Acceptance criteria

1. Unknown or missing entry-fill OIDs cannot create or persist local open positions / trades.  
2. Unknown or missing SL/TP OIDs cannot be stored as active protection on a "live" position without confirmation.  
3. Reconciliation treats flat-on-chain vs local mismatch without a confirmed fill as **desync / manual-review**, not a confirmed SL/TP fill — **including shared-coin Detectors 1 and 2**, aligning with the sole-owner `#685` OID gate.  
4. Tests: unconfirmed SDK `filled` response (e.g. userFills miss / modeled fee path) does not create position rows; shared-coin + bogus SL OID + flat chain does not emit SL protection-fill DM; regression for `orderStatus`-style unknown OID if/when wired.

richkuo:
Closes #756.

Shared-coin Detectors 1 and 2 now require the same userFills confirmation as the sole-owner vanish path (#685): `lookup.OID == StopLossOID` and positive `FilledQty` before booking `hl_sync_stop_loss` or emitting an SL protection-fill alert. Unconfirmed closes route to `hl_sync_external` (mark when available, otherwise the existing zero-price removal).

Tests: stubbed indexer responses where SL attribution is intended; new regression when indexer returns a mismatched OID for the SL lookup query.

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -543,7 +543,7 @@ func reconcileHyperliquidPositionsWithResolver(stratState *StrategyState, sym st
 			// mis-attributed to the cancelled SL at its trigger price. Match must
 			// be by exact OID; the coin+size fallback can spuriously hit a TP
 			// fill of the same size.
-			slConfirmed := useFillFee && lookup.OID == statePos.StopLossOID && lookup.FilledQty > 1e-9
+			slConfirmed := hlReconcileSLFillConfirmed(lookup, useFillFee, statePos.StopLossOID)
 			if slConfirmed {
 				// #621: When userFills returned a real fill qty smaller than the virtual
 				// position (e.g. SL was placed at the on-chain size after a manual TP
@@ -840,14 +840,16 @@ func reconcileHyperliquidAccountPositions(dueStrategies, allStrategies []Strateg
 							}
 							pos.Quantity = lookup.FilledQty
 						}
+						// Snapshot alertSide/alertQty/alertTriggerPx before
+						// recordPerpsStopLossCloseWithFillFee mutates state.
+						// lastBookedTradePnL relies on the just-completed RecordTrade
+						// inside the booker; do not insert another RecordTrade between
+						// here and pendingAlerts append.
 						alertSide := pos.Side
 						alertQty := pos.Quantity
 						alertTriggerPx := pos.StopLossTriggerPx
 						if recordPerpsStopLossCloseWithFillFee(ss, coin, pos.StopLossTriggerPx, lookup.Fee, useFillFee, oidStr, "hl_sync_stop_loss", logger) {
 							changed = true
-							// lastBookedTradePnL relies on the just-completed
-							// RecordTrade inside the booker; do not insert another
-							// RecordTrade between here and the booker call.
 							pendingAlerts = append(pendingAlerts, ProtectionFillAlert{
 								StrategyID:      id,
 								Symbol:          coin,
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

richkuo:
@claude re-review

richkuo:
@claude re-review

PR #760, opened 2026-05-14
fix(hl): prevent sole-owner double TP fill booking (#758)

### LINKED ISSUE #758: Double TP fill booking: same OID recorded twice in trades table
## Summary

When a TP1 order fills, the same fill can be recorded twice in the `trades` table — once via `tryBookSoleOwnerTPFill` and again via `hlAttemptCloseFromTPFills`. Both paths emit a `ProtectionFillAlert` for the same OID, resulting in two identical Discord DMs to the owner.

## Evidence

DB query on live state.db (`scheduler/state.db`):

```sql
SELECT rowid, strategy_id, timestamp, symbol, side, quantity, price, exchange_order_id, is_close, realized_pnl 
FROM trades WHERE strategy_id='hl-rmc-eth-live' AND exchange_order_id='425038458169' ORDER BY timestamp DESC;
```

| rowid | timestamp | symbol | side | qty | price | oid | realized_pnl |
|-------|-----------|--------|------|-----|-------|-----|-------------|
| 144 | 2026-05-14T05:57:26 | ETH | sell | 0.222 | 2267.35 | 425038458169 | 3.14 |
| 145 | 2026-05-14T05:57:34 | ETH | sell | 0.222 | 2267.35 | 425038458169 | 3.14 |

8 seconds apart, identical data. Two `ProtectionFillAlert` DMs sent for the same fill.

## Root Cause

Clarified against `scheduler/hyperliquid_balance.go`:

- `reconcileHyperliquidPositionsForStrategy` (not `reconcileHyperliquidPositionsWithResolver`) calls `tryBookSoleOwnerTPFill` first (`~281`). If it books a partial TP via the **cycle-ordering recovery** path, it **intentionally leaves** the matching `pos.TPOIDs[i]` positive so protection-sync can zero it later (`~411–414`).
- `reconcileHyperliquidPositionsWithResolver` only runs `hlAttemptCloseFromTPFills` on the **vanish** branch: on-chain position missing for that coin but virtual state still open (`~528–536`). That path iterates positive `pos.TPOIDs`, resolves userFills per OID, and books each fill.

Double booking is therefore most plausible as a **two-phase / cross-cycle** interaction: `tryBookSoleOwnerTPFill` already booked the fill and left the TP OID live; a **later** reconcile sees the account flat (or otherwise hits the vanish path) while `pos.TPOIDs` still lists that OID, so `hlAttemptCloseFromTPFills` books the **same** OID again. In a single `reconcileHyperliquidPositionsForStrategy` call where `tryBookSoleOwnerTPFill` succeeds on a **partial** drift path, `hlAttemptCloseFromTPFills` is typically **not** reached in the same pass, because the vanish branch requires `onChainPos == nil` while partial recovery implies `onChainPos != nil`.

Neither path “clears” prior trade rows; both append via `RecordTrade`.

## Expected Behavior

Only one trade record and one DM per fill.

## Fix Direction

After `tryBookSoleOwnerTPFill` successfully books via the **cycle-ordering recovery** path, immediately mirror a **consumed** tier the same way a successful protection-sync would: clear that tier’s `pos.TPOIDs[i]` (OID `0`) and set the matching `pos.TPArmedTiers[i] = true` so `hlAttemptCloseFromTPFills` never treats that OID as an outstanding TP on a later vanish reconcile.

**Must ship with:** a regression test reproducing the double-book path, plus a quick trace that Python protection-sync / plan-build (`--tp-armed-tiers-json`, skip-already-filled tiers) stays correct when Go advances tier state early. This replaces the current “leave OID positive for sync” window for the **already-booked** recovery case only.

Goal: idempotent attribution for a given exchange order id regardless of ordering between protection-sync, sole-owner recovery, and vanish reconcile.

richkuo:
## Summary

After sole-owner **cycle-ordering recovery** books a TP partial in `tryBookSoleOwnerTPFill`, stamp the tier as consumed (`TPOIDs[i]=0`, `TPArmedTiers[i]=true`) so a later vanish reconcile cannot re-book the same OID in `hlAttemptCloseFromTPFills`.

## Test plan

Closes #758

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -290,6 +290,29 @@ func reconcileHyperliquidPositionsForStrategy(
 	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts)
 }
 
+// stampSoleOwnerRecoveryTierConsumed mirrors post-protection-sync state for a
+// tier that tryBookSoleOwnerTPFill attributed via the cycle-ordering recovery
+// path (#758). Without this, pos.TPOIDs[i] stays positive until Python
+// protection-sync runs, and hlAttemptCloseFromTPFills on a later vanish
+// snapshot can book the same TP OID again.
+func stampSoleOwnerRecoveryTierConsumed(s *StrategyState, sym string, tierIdx int) {
+	pos := s.Positions[sym]
+	if pos == nil || tierIdx < 0 {
+		return
+	}
+	n := len(pos.TPOIDs)
+	if tierIdx >= n {
+		return
+	}
+	if len(pos.TPArmedTiers) < n {
+		ext := make([]bool, n)
+		copy(ext, pos.TPArmedTiers)
+		pos.TPArmedTiers = ext
+	}
+	pos.TPOIDs[tierIdx] = 0
+	pos.TPArmedTiers[tierIdx] = true
+}
+
 // tryBookSoleOwnerTPFill is the sole-owner TP attribution helper. Returns true
 // when a TP-tier fill was detected and booked via
 // recordPerpsExternalPartialCloseWithFillFee — covers both the partial-drop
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #761, opened 2026-05-14
feat(hl): reconcile fill hints for sync-protection (#759)

### LINKED ISSUE #759: Redundant HL userFills queries in Python protection sync causing 429 rate limits
## Summary

The Python protection sync (`check_hyperliquid.py --sync-protection`) can re-query HL `userFills` (via `lookup_fill_fee_by_oid`) for SL/TP OIDs that are absent from the current open-orders snapshot. The Go reconcile phase in the same scheduler cycle may already have resolved overlapping fill data via its cached `userFillsByTime` prefetch. That overlap adds redundant HL indexer traffic and can contribute to rate limiting (HTTP 429) under load.

**Correction (prior issue text):** Go does **not** collapse reconcile into a single `userFills` HTTP call. It dedupes `(coin, oid, qty)` candidates, prefetches each into an in-memory resolver before the locked apply phase, but each distinct candidate still triggers `userFillsByTime` through `lookupHyperliquidFillByOID` (with retries). Python similarly calls `user_fills_by_time` inside `lookup_fill_fee_by_oid` (with retries). The redundancy is duplicate **endpoint usage** for the same logical “did this OID fill?” question, not “Go already downloaded everything in one request.”

**Correction:** Startup `--probe-only` for **signal-check** and **`--fetch-atr`** modes exits after `argparse` and does **not** construct `HyperliquidExchangeAdapter` (#645). Live signal checks (`--probe-only` omitted) still instantiate the adapter for OHLCV (`shared_scripts/check_hyperliquid.py`).

## Evidence

Rate limit error logged at 20:03:42 and 20:08:54 tonight on the live instance:

```
[hl-tema-eth-live] [ERROR] Failed to initialize Hyperliquid Exchange client: (429, None, 'null', None, {...})
```

**Note:** That exact message is raised from live `_HLExchange(...)` construction in `HyperliquidExchangeAdapter.__init__`, not from `lookup_fill_fee_by_oid` / `userFills` (`platforms/hyperliquid/adapter.py`). It still indicates HL API pressure on the live path; redundant `userFills` traffic remains a separate contributing factor for protection sync.

Both recovered on the next cycle, but rate limits increase with more strategies running.

## Root Cause

During a protection sync cycle, two HL `userFills`-class query paths can overlap:

**Go reconcile (pre–due-strategy loop):**

- `reconcileHyperliquidAccountPositions` builds `resolveFee` via `buildCachedHyperliquidReconcileFillResolver`: RLock pass collects deduped lookup keys for mismatched on-chain vs virtual positions, then **prefetches** each key with `lookupHyperliquidReconcileFillFee` → `userFillsByTime` per candidate (`scheduler/hyperliquid_fills.go`).
- Cached results are used under lock for fill attribution, partial closes, and `ProtectionFillAlert` (`scheduler/hyperliquid_balance.go`).

**Python protection sync (later in the same loop, per strategy):**

- `run_sync_protection` calls `adapter.open_order_oids(symbol)` then, when a recorded SL/TP OID is **not** in that open set, `_oid_filled_externally` → `lookup_fill_fee_by_oid(oid, since_ms)` with a **7-day** lookback (`shared_scripts/check_hyperliquid.py`).
- That is a second indexer pass for “OID filled vs cancelled?” when Go may have just evaluated related OIDs in reconcile.

Python does **not** call `lookup_fill_fee_by_oid` for every OID every cycle—only when `open_order_oids` succeeds and the OID is missing from the set (otherwise it defers or treats as still open).

## Impact

- Multiple strategies × multiple protection OIDs × `userFills` retries ⇒ redundant API usage vs a single source of truth passed from Go.
- Contributes to 429s on HL during peak traffic alongside other HL calls (adapter init, OHLCV, open orders, etc.).
- Worsens as strategy count grows.

## Fix Direction

1. **Pass fill-resolution data Go → Python:** After reconcile, include resolved fill status (OID → filled / not-filled + fee/px where known) in argv for Python’s protection sync so Python can skip `lookup_fill_fee_by_oid` when Go already has a fresh answer for that cycle.
2. **`--probe-only`:** Signal-check and `--fetch-atr` probes already skip adapter init; any further work here would be other code paths or tightening docs—no change needed for those two probe shapes unless new modes are added without the same early exit.

Secondary: collapsing detection into one owner (Go) may also reduce double-booking risk called out in #758 (separate fix; same area).

Ref: Related to #758 (separate fix, same code path area).

richkuo:
Closes #759.

Pass OID fill snapshots from the Go reconciler prefetch cache to `check_hyperliquid.py --sync-protection` via `--reconcile-fill-hints-json` so Python skips duplicate `userFills` lookups when Go already resolved the OID in the same scheduler cycle.

- `buildCachedHyperliquidReconcileFillResolver` returns hints alongside the cached resolver
- `reconcileHyperliquidAccountPositions` returns `(changed, fillHints)`; main marshals once per cycle
- Python `_oid_filled_externally` prefers hints when present

Tests: `go -C scheduler test ./...`, `pytest shared_scripts/test_check_hyperliquid.py`.

### DIFF
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -372,7 +372,7 @@ func RunHyperliquidUpdateStopLoss(script, symbol, side string, size, triggerPx f
 
 // buildHyperliquidSyncProtectionArgv builds argv for check_hyperliquid.py
 // --sync-protection (used by RunHyperliquidSyncProtection and tests).
-func buildHyperliquidSyncProtectionArgv(symbol, side string, size, avgCost, entryATR, stopLossATRMult float64, tiers []hlProtectionTier, stopLossOID int64, tpOIDs []int64, tpArmedTiers []bool) []string {
+func buildHyperliquidSyncProtectionArgv(symbol, side string, size, avgCost, entryATR, stopLossATRMult float64, tiers []hlProtectionTier, stopLossOID int64, tpOIDs []int64, tpArmedTiers []bool, reconcileFillHintsJSON []byte) []string {
 	args := []string{
 		"--sync-protection",
 		fmt.Sprintf("--symbol=%s", symbol),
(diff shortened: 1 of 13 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #765, opened 2026-05-15
update.sh: Linux go fallback and ExecStart vs swap warning (#764)

### LINKED ISSUE #764: update.sh: Linux go path, non-systemd multi-instance, and systemd ExecStart drift
## Problem

`scripts/update.sh` is built around a **systemd `--restart` + `/health` verify** flow. That fits a single unit updating `./go-trader` at the deploy root. It is **awkward** for **multiple** instances run as **bare processes** (separate git worktrees, `nohup`, no `systemctl`): each tree can still run the script for pull/sync/build/swap, but automated restart/verify expects systemd (`scripts/update.sh` restart path).

## Issue 1: Go binary not found

`update.sh` (preflight) resolves `go` like this:

```bash
go_bin=""
if command -v go >/dev/null 2>&1; then
    go_bin=$(command -v go)
elif [[ -x /opt/homebrew/bin/go ]]; then
    go_bin=/opt/homebrew/bin/go
else
    fail "go not on PATH and /opt/homebrew/bin/go missing"
fi
```

On Linux with Go only under `/usr/local/go` and not on `PATH`, neither branch matches and the script exits:

```
[update] FAIL phase=preflight: go not on PATH and /opt/homebrew/bin/go missing
```

**Fix:** Add `/usr/local/go/bin/go` as a third fallback (standard Linux tarball install path).

## Issue 2: No multi-instance / non-systemd restart

**Example deployment (operator-reported):** several instances as bare processes (no systemd), each in its own workspace, e.g. under `/root/.openclaw/workspace/go-trader-*/`, each with its own `scheduler/config.json` and `scheduler/state.db`.

With `--restart`, the script runs `sudo systemctl restart "$service_unit"` (`scripts/update.sh`); that does not restart `nohup` processes in arbitrary worktrees.

**Directional fix (needs design, not just code):** a "signal" or non-systemd restart path sounds simple, but **killing the old PID is easy — respawning with the same effective behavior is not**:

- **Load-bearing order:** read **`/proc/<pid>/cwd`** and **`/proc/<pid>/cmdline`** (and anything else you need) **before** `kill`, not after — those paths are only meaningful **while the process is alive**. Intended sequence: **capture → kill → wait (process exit) → respawn**, never **kill → capture**.
- **`/proc/<pid>/cmdline`** uses **null bytes** between argv entries; naive shell parsing breaks easily.
- **Working directory** still matters for any relative argv paths; even with a pre-kill snapshot, the contract should be explicit so implementers do not "optimize" into the wrong order.
- **Environment / `.env`:** bare `nohup` children **inherit env from whoever launched them**; that parent shell (and its exports) is often **gone** by update time, so you cannot reliably "reconstruct" the same effective env from memory later. A generic respawn must **not guess** live vs paper secrets.
- **Wrapper script pattern (preferred for bare processes):** a per-instance `run.sh` that `cd`s to the right tree, optionally `set -a; source .env; set +a`, then starts the binary with **fixed, known args**. Then `update.sh` only **kills the old PID and re-invokes that wrapper** detached (e.g. `setsid nohup bash run.sh &` or equivalent) — cleaner than parsing `/proc/.../cmdline` and trying to re-hydrate env by hand.
- **Verify + rollback parity:** signal mode is not done at "respawned a process." It must **reuse the same `/health` poll** (expected `version`, PID freshness vs pre-restart where applicable) and the same **`go-trader.prev` / git+`uv` rollback** path as the systemd flow if verify fails — otherwise it ships **without the safety net** the systemd path already has (`scripts/update.sh` verify + `do_rollback`). Two specific lines in the existing path are systemd-only and need non-systemd equivalents:
  - **PID freshness** (`scripts/update.sh:377-384`): uses `systemctl show -p MainPID`; signal mode needs another source — e.g. a pidfile written by the wrapper on startup, or the pre-kill PID captured from `/proc`.
  - **`do_rollback` restart** (`scripts/update.sh:326`): ends with `sudo systemctl restart`; signal mode rollback needs its own branch — re-invoke the wrapper with `./go-trader.prev` in place.

**PR / scope recommendation (review feedback):**

1. **Ship first (low risk, should not block on signal mode):** Issue 1 Go-path fallback + Issue 3 **`ExecStart` vs swap-target warning** (or equivalent operator-visible check / docs pointer).
2. **Defer to a follow-up issue/PR:** automatic "signal mode" or auto-detecting systemd vs bare process. If signal mode is added later, prefer **explicit opt-in** (e.g. `RESTART_MODE=signal`) rather than silent auto-detect, so production never gets a half-defined respawn heuristic by surprise — and include **verify + rollback** as part of that PR, not a later add-on.

## Issue 3: systemd `ExecStart` path mismatch

On hosts that **do** use systemd, the **effective** unit on disk may point at a different binary path than `./go-trader` at the deploy root (e.g. `ExecStart=.../scheduler/go-trader` while `update.sh` swaps `.../go-trader`). Post-restart verify can fail while the wrong binary still serves `/health`.

**Fix:** Warn or align: read `ExecStart` (e.g. `systemctl cat <unit> | grep ExecStart`) vs swap target, or copy the staged binary to the path the unit runs.

## Build output path (clarification)

An earlier draft claimed `go -C scheduler build -o ../go-trader.new .` writes **above** the repo on Linux. That does **not** match current `go` behavior: the build plan changes into `scheduler/` first, so `../go-trader.new` is the **parent of `scheduler/`** — i.e. **repo root** — same on Linux and macOS. Using `-o "$(pwd)/go-trader.new"` after `cd "$repo_root"` is still reasonable for readability, but it is **not** required to fix a Linux-vs-macOS output-dir bug.

## Suggested implementation (sketch)

```bash
# Go discovery — add Linux fallback (good first-PR candidate)
go_bin=""
if command -v go >/dev/null 2>&1; then
    go_bin=$(command -v go)
elif [[ -x /opt/homebrew/bin/go ]]; then
    go_bin=/opt/homebrew/bin/go
elif [[ -x /usr/local/go/bin/go ]]; then
    go_bin=/usr/local/go/bin/go
else
    fail "go not on PATH and not found in standard locations"
fi

# Optional: absolute output for clarity (repo root = $(pwd) after cd)
# "$go_bin" -C scheduler build -ldflags "-X main.Version=$ver" -o "$(pwd)/go-trader.new" .

# ExecStart vs ./go-trader warning — good first-PR candidate (Issue 3)
# Signal / bare-process auto-restart — separate PR; explicit opt-in if added;
# must include /health verify + rollback parity with systemd path
```

## Workaround (current)

For multi-instance bare-process deployments, run `update.sh` from each workspace **without** `--restart`, then restart each process manually (or align on systemd units per instance).

```bash
for d in /root/.openclaw/workspace/go-trader-*/; do
    (cd "$d" && bash scripts/update.sh)
done
# then kill + respawn each instance per your runbook
```

richkuo:
## Summary
- **Go discovery:** `scripts/update.sh` now falls back to `/usr/local/go/bin/go` after PATH and Homebrew (typical Linux tarball install).
- **ExecStart drift:** before `systemctl restart`, warn when the unit`s `ExecStart` binary (including structured `{ path=... }` payloads) does not resolve to the repo-root `./go-trader` this script just swapped; continues after warning.
- **CI:** `scheduler/update_sh_syntax_test.go` runs `bash -n` on `scripts/update.sh`.

Also includes prior commit `339512c` (docs: LLM footer vs Co-authored-by) that was on local `main` before branching.

Closes #764

LLM: GPT-5.2 | high | Harness: create-pull-request

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,7 +106,7 @@
 - Reference issue with `Closes #<N>` in body. In GitHub comments avoid `#N` for list items (auto-links); use `1.`.
 - Fetch latest bot review: `gh api repos/richkuo/go-trader/issues/<N>/comments --jq '[.[] | select(.user.login=="claude[bot]")] | last | .body'` (top-level summary lives on issues endpoint, not pulls).
 - Before merging long-running PR: `git fetch origin main && git diff origin/main..HEAD -- <paths>` catches silent reverts.
-- Replace default Claude Code footer with `LLM: <model> | <effort> | Harness: <action>`. No `Co-Authored-By`.
+- **Commits and PR bodies:** end with `LLM: <model> | <effort> | Harness: <action>` (same line as the default Claude Code footer slot). Do **not** append `Co-authored-by: Cursor <cursoragent@cursor.com>` or any other `Co-authored-by:` / `Co-Authored-By` trailer.
 
 ### PR review format (`@claude review`)
 Top-level review comment is one of two shapes — no preamble, no closing:
--- a/scheduler/update_sh_syntax_test.go
+++ b/scheduler/update_sh_syntax_test.go
@@ -1,0 +1,23 @@
+package main
+
+import (
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func TestUpdateShellScriptSyntax(t *testing.T) {
+	t.Parallel()
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("runtime.Caller failed")
+	}
+	schedDir := filepath.Dir(thisFile)
+	repoRoot := filepath.Join(schedDir, "..")
+	script := filepath.Join(repoRoot, "scripts", "update.sh")
+	out, err := exec.Command("bash", "-n", script).CombinedOutput()
+	if err != nil {
+		t.Fatalf("bash -n scripts/update.sh: %v\n%s", err, out)
+	}
+}
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Atomic update with pre-flight probe, staging build, atomic binary swap,
-# post-restart verification, and rollback (#682).
+# post-restart verification, and rollback (#682). #764: extra go(1) lookup
+# paths + ExecStart vs swap-target warning before systemd restart.
 #
 # Phases:
 #   preflight  — git/uv/go sanity checks
@@ -70,6 +71,8 @@ while [[ $# -gt 0 ]]; do
             echo "  STATUS_PORT=<n>          override /health port (default: read from config, else 8099)"
             echo "  ACTIVE_TIMEOUT=<sec>     systemctl is-active poll timeout (default: 30)"
             echo "  HEALTH_TIMEOUT=<sec>     /health version-match poll timeout (default: 60)"
+            echo ""
+            echo "Go is resolved from: PATH, then /opt/homebrew/bin/go, then /usr/local/go/bin/go"
             exit 0
             ;;
         *)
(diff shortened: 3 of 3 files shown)


richkuo:
@claude

PR #776, opened 2026-05-23
fix(validator): allow invert_signal with direction="short" (#775)

### LINKED ISSUE #775: Allow invert_signal with direction="short" for inverse short-only strategies
## Problem

`v0.65.0` (PR #774, which introduced `invert_signal`) rejects configs that combine:

```json
"direction": "short",
"invert_signal": true
```

The validator treats this as an invalid double-flip, but the two settings operate at different layers and the combination is semantically distinct from either flag alone.

## Why this matters

`invert_signal` and `direction="short"` are NOT redundant:

- **`direction="short"` alone:** opens short when Python emits raw `SELL` (i.e. on the strategy's long-EXIT trigger). Raw `BUY` closes the short.
- **`direction="short"` + `invert_signal=true`:** opens short when Python emits raw `BUY` (i.e. on the strategy's long-ENTRY trigger, mirrored). This is an "inverse short-only" strategy and fires on different bars than the plain `direction="short"` case.

Intended signal pipeline:

```text
raw strategy signal -> optional invert_signal (Go layer) -> direction execution rules
```

Example:

```text
raw BUY -> invert_signal -> SELL -> direction="short" -> opens short
```

The validator's stated rationale ("double-flips back to long semantics") is incorrect: `direction="short"` does not flip the signal sign — it changes which sign opens vs. closes. So `invert + short` results in opening short on raw-BUY triggers, never long.

## Current behavior

Config validation fails at startup with:

```text
strategy[...]: invert_signal=true with direction="short" double-flips back to long semantics; drop one of the two
```

This forces operators wanting an inverse short-only strategy to either drop `invert_signal` (which silently changes which bars open positions) or drop `direction="short"` (which loses the short-only constraint).

## Expected behavior

Allow `invert_signal=true` with `direction="short"` for Hyperliquid perps/manual strategies.

The validator should still reject `invert_signal` where it is a no-op or unsupported, e.g. non-Hyperliquid platforms or non-perps/non-manual strategy types — that part of the check is correct and should stay.

## Suggested fix

Update validation/docs/tests to make signal ordering explicit:

```text
raw strategy signal -> optional invert_signal -> direction execution rules
```

Tests to add/adjust:

- `direction="short"` without `invert_signal`: raw `SELL` opens short, raw `BUY` closes short.
- `direction="short"` with `invert_signal=true`: raw `BUY` flips to `SELL` and opens short, raw `SELL` flips to `BUY` and closes short.
- non-HL/non-perps use of `invert_signal` remains rejected.
- hot-reload remains blocked while positions are open (existing behavior).

## Operational note

A live update had to remove `invert_signal` from short-only strategies to pass the `v0.65.0` probe. That workaround changes which bars the strategies enter on, so this is not a cosmetic validator nit — it materially altered live trading behavior until the validator is relaxed.

richkuo:
## Summary
- Drop the validator's `invert_signal=true + direction="short"` double-flip rejection — it was based on an incorrect read of the executor (#775).
- `direction="short"` does not flip the signal sign; it changes which sign opens vs closes. Composed with `invert_signal` (Go-layer), the pipeline is `raw -> invert -> direction`, so `invert + direction="short"` opens short on raw-BUY triggers (inverse short-only), which is distinct from plain `direction="short"` opening on raw-SELL.
- Keep the unrelated rejection for `invert_signal` on non-HL or non-perps/non-manual strategies — the Go executor only flips signals on the HL perps/manual path, so the flag would be a silent no-op elsewhere.
- Add `TestValidateConfigInvertSignal` covering the four accept cases (HL perps × long/short/both, HL manual short) and two reject cases (HL spot, non-HL perps).

Closes #775

## Test plan
- [x] `go test ./...` (scheduler) — all pass
- [x] `go build` (scheduler) — clean
- [x] `gofmt -w`

---
LLM: Claude Opus 4.7 | high

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -313,7 +313,7 @@ type StrategyConfig struct {
 	MaxDrawdownPct         float64             `json:"max_drawdown_pct"`
 	IntervalSeconds        int                 `json:"interval_seconds,omitempty"`           // per-strategy override (0 = use global)
 	HTFFilter              bool                `json:"htf_filter,omitempty"`                 // higher-timeframe trend filter
-	InvertSignal           bool                `json:"invert_signal,omitempty"`              // HL perps/manual only: flip BUY<->SELL on a non-zero signal before execution (HOLD/0 is never flipped). Lets inverse variants reuse the same open/close refs. Rejected with direction="short" (double-flip) and outside HL perps/manual.
+	InvertSignal           bool                `json:"invert_signal,omitempty"`              // HL perps/manual only: flip BUY<->SELL on a non-zero signal before execution (HOLD/0 is never flipped). Lets inverse variants reuse the same open/close refs. Composes with Direction — invert runs in the Go layer before direction interprets the resulting sign (e.g. direction="short" + invert_signal=true opens short on raw-BUY triggers, distinct from plain direction="short" which opens on raw-SELL). Rejected outside HL perps/manual.
 	AllowShorts            bool                `json:"allow_shorts,omitempty"`               // DEPRECATED — use Direction. Perps only; legacy boolean retained on the struct so pre-v14 JSON unmarshals cleanly. Read via EffectiveDirection / PerpsAllowsShort / PerpsAllowsLong, never directly. Migrated to Direction in v14 (#656).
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #778, opened 2026-05-23
fix(hl): book filled TP tiers when on-chain dust remains

### LINKED ISSUE #777: Reconciler should book all filled TP tiers when tiny residual dust remains
## Problem

The Hyperliquid reconciler can get stuck — or silently corrupt virtual state — when all TP tiers have filled but a tiny residual position remains on-chain. The visible symptom (persistent gap) occurs in the shared-coin path; the silent variant occurs in the sole-owner path.

Observed live case (shared-coin path):

```text
strategy: hl-vwap-btc-1h
opened: 0.013 BTC short @ 75249
TP1 filled: 0.006 BTC @ 74605, oid=438613562733
TP2 filled: 0.006 BTC @ 74284, oid=438613569010
on-chain residual: -0.001 BTC
local virtual before repair: -0.013 BTC
reconciliation gap: virtual=-0.013000 on-chain=-0.001000 delta=-0.012000
```

The bot logged both TP OIDs as already filled and zeroed the TP OID list:

```text
TP1 OID=438613562733 already filled on-chain; not re-placing — reconciler will book the close
TP2 OID=438613569010 already filled on-chain; not re-placing — reconciler will book the close
HL protection synced (sl_oid=438613556644 tp_oids=[0 0])
```

But later reconcile cycles continued to warn instead of booking the TP closes.

## Current behavior

`hyperliquidClearedTPTier` (`scheduler/hyperliquid_balance.go:1243`) only treats an all-zero TP list as attributable when the observed close quantity fully closes the strategy position:

```go
// scheduler/hyperliquid_balance.go:1274
if math.Abs(pos.Quantity-closeQty) <= 1e-6 {
    return len(tpOIDs) - 1, true
}
return 0, false
```

In the observed case, `pos.Quantity=0.013` and `closeQty=0.012` (from `hyperliquidSharedPartialCloseDrift` at `scheduler/hyperliquid_balance.go:1217`), because a `0.001 BTC` dust residual remained. The reconciler therefore left the gap visible and did not book either TP close.

### Sole-owner path has the same root cause (silent variant)

When only one strategy trades the coin, the flow is `tryBookSoleOwnerTPFill` → `reconcileHyperliquidPositionsWithResolver`:

1. `tryBookSoleOwnerTPFill` (`scheduler/hyperliquid_balance.go:381`) computes `closeQty = statePos.Quantity - onChainAbs = 0.012`, calls `hyperliquidClearedTPTier`, gets `false` (same all-zero + dust failure), then its cycle-ordering recovery path requires `lookup.OID > 0` matching a non-zero `pos.TPOIDs[i]` — impossible when all TPOIDs are zero — so it also returns false.
2. Control falls through to `reconcileHyperliquidPositionsWithResolver` (`scheduler/hyperliquid_balance.go:562-568`), which detects `statePos.Quantity != onChainQty` and silently overwrites `statePos.Quantity = 0.001` with no `RecordTrade` / no PnL / no fee booking. The 0.012 BTC of real TP fills are lost from trade history.

## Expected behavior

When all configured TP tiers are filled/cleared and the on-chain residual is a same-side tiny remainder, both reconciler paths should book the filled TP quantity from userFills (per-tier, at VWAP fill price + real fee) and leave the residual open.

Expected post-reconcile state for the observed case:

```text
local virtual: -0.001 BTC
on-chain:      -0.001 BTC
delta:          0
```

And trade history should include the two TP close fills from userFills:

```text
TP1 close 0.006 BTC @ 74605, fee=0.064458
TP2 close 0.006 BTC @ 74284, fee=0.06418
```

## Scope

Fix covers both paths:

- **Detector 3 (shared-coin)** in `reconcileHyperliquidAccountPositions` — the visibly-stuck symptom from the live case.
- **Sole-owner** via `tryBookSoleOwnerTPFill` / `reconcileHyperliquidPositionsWithResolver` — the silent data-loss variant.

The existing per-tier booking already lives in `hlAttemptCloseFromTPFills` (`scheduler/hyperliquid_balance.go:1305`), which books each filled TP OID at its actual VWAP price + real fee. The fix should reuse that mechanism rather than reinvent attribution.

## Suggested fix

Teach the TP attribution logic to handle:

```text
all TP tiers cleared/armed + same-side on-chain residual dust (closeQty < pos.Quantity)
```

Possible approaches:

- When `TPArmedTiers[i] == true` for every tier (or every TPOID was non-zero before being zeroed — `backfillTPArmedTiers` at `scheduler/db.go:425` already infers this for legacy rows) AND same-side on-chain qty < virtual qty, book the virtual/on-chain delta using userFills, even if `closeQty != pos.Quantity`.
- Prefer userFills OID matches when available so multiple filled tiers are booked accurately at actual fill prices/fees.
- Preserve the residual position instead of requiring full flatten.
- Critical guard: the existing all-zero branch was written to prevent never-placed TPs (TPOIDs=[0,0], TPArmedTiers=[false,false]) from being treated as attributable. The new branch must require `TPArmedTiers` evidence (or a positive userFills OID match against the strategy's known TP OIDs) so that guard isn't lost.

## Regression tests

**Positive case (must book):**

```text
short position qty=0.013
TPOIDs=[0,0]
TPArmedTiers=[true,true]
on-chain size=-0.001
userFills contain TP1 0.006 and TP2 0.006
```

Expected: two close trades recorded (or the total `0.012` close booked accurately), position quantity becomes `0.001`, and `ReconciliationGaps[BTC].DeltaQty == 0`.

**Negative case (must NOT book — preserves never-placed-TP guard):**

```text
short position qty=0.013
TPOIDs=[0,0]
TPArmedTiers=[false,false]
on-chain size=-0.001
no matching userFills entries
```

Expected: no trades recorded, virtual qty unchanged (or whatever the legacy behavior was), gap remains visible. This proves the new branch doesn't regress the existing guard at `scheduler/hyperliquid_balance.go:1274`.

**Sole-owner parity:** same two cases, but with a single strategy on the coin so the path runs through `tryBookSoleOwnerTPFill` → `reconcileHyperliquidPositionsWithResolver`. The silent overwrite at `scheduler/hyperliquid_balance.go:562-568` must be replaced or guarded so the positive case books the TPs and the negative case does not corrupt qty silently.

## Operational note

The live deployment was manually repaired by inserting the missing TP close trades and reducing the local position to `0.001 BTC`. This issue is for the upstream automatic reconciler behavior so future TP+dust cases self-heal.

richkuo:
## Summary
- Adds `hlAttemptCloseFromArmedTPClears` to book each armed/cleared TP tier at userFills VWAP when a same-direction dust residual remains after all TPs fill.
- Wires the path through sole-owner reconcile and shared-coin Detector 3; guards against silent qty resync when all tiers are armed/cleared.
- Prefetches per-tier fill lookups in the reconcile fee cache; regression tests cover positive (armed+dust) and negative (never-placed TP) cases.

Closes #777

## Test plan
- [x] `go test -C scheduler ./... -run 'TPDust|HyperliquidAllTiersArmed|SoleOwnerTP|HyperliquidHasCleared|AttemptCloseFromTP|ReconcileSharedCoin_TP'`

LLM: Composer 2.5 | high | Harness: cursor

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -279,7 +279,7 @@ func fetchHyperliquidState(accountAddress string) (float64, []HLPosition, error)
 // cached resolver outside the lock and call
 // reconcileHyperliquidPositionsWithResolver.
 func reconcileHyperliquidPositions(stratState *StrategyState, sym string, positions []HLPosition, accountAddress string, logger *StrategyLogger) bool {
-	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger, nil)
+	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger, nil, StrategyConfig{})
 }
 
 // reconcileHyperliquidPositionsForStrategy is the production entry point with
@@ -318,11 +318,11 @@ func reconcileHyperliquidPositionsForStrategy(
 		// by the booker; the legacy reconciler's qty/side/avgCost resync will
 		// no-op because virtual now matches on-chain. Continue through it for
 		// idempotent housekeeping (multiplier migration, leverage seed).
-		reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts)
+		reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts, sc)
 		return true
 	}
 
-	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts)
+	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, resolveFee, logger, pendingAlerts, sc)
 }
 
 // stampSoleOwnerRecoveryTierConsumed mirrors post-protection-sync state for a
@@ -440,6 +440,13 @@ func tryBookSoleOwnerTPFill(
 		return false
 	}
 
+	if onChainPos != nil && hyperliquidAllTiersArmedAndCleared(sc, statePos) {
+		onChainAbs := math.Abs(onChainPos.Size)
+		if hlAttemptCloseFromArmedTPClears(stratState, sc, sym, statePos, onChainAbs, resolveFee, logger, pendingAlerts) {
+			return true
+		}
+	}
+
 	lookup, useFillFee := resolveFee(sym, 0, closeQty)
 	logHyperliquidReconcileFillLookup(logger, sym, 0, closeQty, lookup, useFillFee)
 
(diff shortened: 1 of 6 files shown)


richkuo:
@claude

PR #780, opened 2026-05-23
feat(hl): regime-aware directional policy with hold-on-transition (#779)

### LINKED ISSUE #779: Add regime-aware transition from inverse short mode back to long mode
## Problem

The live BTC 1h Hyperliquid strategies can be configured as inverse short-only strategies:

- `direction="short"`
- `invert_signal=true`

That is correct while the broader regime is `trending_down`: a normal long signal is intentionally inverted into a short entry.

The gap is what happens when the regime is no longer `trending_down`. Today this is a manual/operator decision. The scheduler config can say `direction` and `invert_signal` (`scheduler/config.go:316`), and the signal runner can compute a `regime` (`shared_tools/regime.py`), but there is no first-class policy that automatically says:

> If regime leaves `trending_down`, stop using inverse short mode and return to normal long behavior.

The only regime-driven code paths today are `regimeBlocksOpen` (entry gate in `scheduler/regime.go:28`) and the `regime_atr` close evaluator (per-regime SL/TP multipliers). Neither mutates `direction` or `invert_signal`. Reload guards (`scheduler/config_reload.go:334`) actively *block* operator hot-swaps of these fields while a position is open, so any automatic flip must layer above the config-mutation path.

In the current live setup, top-level regime detection is also disabled, so the running bot is not making this decision itself.

## Current live observation

Manual probes on 2026-05-23 returned `regime="trending_down"` for both BTC 1h HAE and VWAP, so no live flip was made.

Current live config remains:

- `hl-hae-btc-1h`: `direction="short"`, `invert_signal=true`
- `hl-vwap-btc-1h`: `direction="short"`, `invert_signal=true`

There is also an existing tiny open short on `hl-vwap-btc-1h` (`0.001 BTC`), so changing direction/invert in-place would need to handle the open-position transition safely.

## Desired behavior

Add an explicit regime-aware policy for this strategy family, for example:

- In `trending_down`:
  - allow inverse short behavior (`direction="short"`, `invert_signal=true`), or an equivalent strategy policy
- Outside `trending_down`:
  - stop opening inverse shorts
  - return to normal long behavior (`direction="long"`, `invert_signal=false`) once safe

The important part is that this should be deterministic and encoded in code/config, not a manual edit each time the regime changes.

## Open design question

Where does the policy live?

1. **Config rewrite** — scheduler mutates `direction`/`invert_signal` in `scheduler/config.json` based on regime. Conflicts with the existing "config is operator-authored, hot-reload validates" model and the `validateHotReloadStateCompatible` open-position guards.
2. **Runtime override layer** — a new per-strategy `regime_directional_policy` block selects effective `direction`/`invert_signal` per regime at signal time, leaving the persisted config untouched and hot-reload guards intact.

Option 2 is likely the better fit; please confirm before implementation.

## Safety requirements

- Do not hot-swap `direction`/`invert_signal` underneath an open position without an explicit transition rule.
- If a short is already open when regime changes away from `trending_down`, define whether the bot should:
  - hold until normal close/SL/TP,
  - flatten immediately,
  - or enter an operator-required transition state.
- Surface the current regime and active directional policy in `/status` so operators can understand why the bot is short-mode or long-mode.
- Add tests for:
  - `trending_down` selects inverse short policy
  - `trending_up`/`ranging` selects normal long policy
  - open short position blocks or safely handles transition back to long mode

## Notes

This is related to but distinct from #775 (fixed by PR #776). PR #776 fixed validation so inverse short-only config is allowed. This issue is about when that inverse short-only config should be active as market regime changes.

richkuo:
## Summary

Adds `regime_directional_policy` per-strategy — a per-regime override for `Direction` + `InvertSignal` so a single HL perps strategy can run inverse short in `trending_down` and plain long otherwise, switching deterministically as regime changes instead of needing operator hot-edits.

Closes #779.

## Design

**Runtime override layer**, not config rewrite. Persisted config is operator-authored; the policy resolves the *effective* `(Direction, InvertSignal)` pair at signal time. This preserves the existing operator-authored config model and the `validateHotReloadStateCompatible` open-position guards (`config_reload.go:334`).

**Shape:**

```json
\"regime_directional_policy\": {
  \"trend_regime\": {
    \"trending_down\": { \"direction\": \"short\", \"invert_signal\": true },
    \"trending_up\":   { \"direction\": \"long\",  \"invert_signal\": false },
    \"ranging\":       { \"direction\": \"long\",  \"invert_signal\": false }
  }
}
```

**Hold semantics (transition rule):** when a position is open, the resolver uses `pos.Regime` (the regime stamped at open) — *not* the current regime. The position runs to its natural exit via existing SL / TP / close-evaluator paths under the policy that opened it. Once flat, the next entry uses the current regime's policy. No forced flatten, no operator-required transition state.

**Wired in `runHyperliquidCheck`** after `result.Regime` is known and **before** `applySignalInversion`, so the invert decision uses the effective `InvertSignal`. The function signature changes from `sc StrategyConfig` to `sc *StrategyConfig` so the mutation propagates to downstream `EffectiveDirection` / `perpsLiveOrderSize` / `PerpsOrderSkipReason` calls in the same cycle. The mutation is scoped to the loop-local `sc` copy; `cfg.Strategies` is never touched.

## Validation + reload

- **HL perps only** (mirrors `invert_signal`'s surface — both override the same Go-layer signal that only HL consumes numerically).
- **Requires top-level `regime.enabled=true`** — without regime detection, `result.Regime` is empty, the resolver always no-ops, and the policy silently defeats. Reject at startup so the asymmetric config never ships.
- **All canonical regime labels required** (`trending_up`, `trending_down`, `ranging`) so there's never an undefined runtime fallback.
- **Unknown keys / unknown labels / invalid direction values rejected** at config-load time.
- **Hot reload:** shape change while open → REJECTED (state-compat parity with `stop_loss_atr_regime` in `regime_atr.go`); change when flat → hot-applies on next cycle.
- **invert_signal added to `strategyRestartShape`** (existing #775 oversight, surfaced here): a pure invert toggle no longer trips the immutable-fields DeepEqual gate.

## /status surfacing

Per-strategy now exposes:
- `direction` / `invert_signal` — base config values
- `effective_direction` / `effective_invert_signal` — resolved for the active regime
- `regime_directional_policy` (bool) — whether the strategy has a policy block
- `effective_policy_regime` — the regime key the resolver used (`pos.Regime` while open, current regime when flat)

So operators can answer \"why is the bot in long/short mode\" without grepping logs.

## Test plan

- [x] `go build` (with `-ldflags Version`)
- [x] `go test ./...` — 24.4s, all pass including 3 new test groups (16 subtests):
  - `TestRegimeDirectionalPolicyResolveRaw` — parsing + validation (canonical labels, valid direction enum, wrapper key, unknown keys, invert default)
  - `TestRegimeDirectionalPolicyEqualForReload` — reload-equality semantics
  - `TestApplyRegimeDirectionalPolicy` — resolver: flat→current, open→pos.Regime hold, post-close fresh, no-policy no-op, unknown-regime no-op, legacy-position fallback
  - `TestValidateConfigRegimeDirectionalPolicy` — HL-perps-only, requires regime.enabled, hot-reload shape + add/remove blocked while open
- [x] `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 1044 passed
- [x] `gofmt -w` on touched files
- [ ] Operator: enable on `hl-hae-btc-1h` / `hl-vwap-btc-1h` (note: existing 0.001 BTC short on `hl-vwap-btc-1h` will run under the policy that opened it until natural exit)

---
LLM: Claude Opus 4.7 (1M) | high | Harness: Claude Code

### DIFF
--- a/backtest/run_backtest.py
+++ b/backtest/run_backtest.py
@@ -109,6 +109,19 @@ def load_strategy_config(config_path: str, strategy_id: str) -> dict:
                 f"{config_path}: strategy {strategy_id!r} has no open_strategy.name; "
                 f"the migrated config should always populate it."
             )
+        # #779: regime_directional_policy is HL-live-only — the backtester
+        # has no per-cycle resolver hook that mirrors runHyperliquidCheck,
+        # so the per-regime Direction / InvertSignal flip wouldn't apply
+        # and results would silently use the static config instead. Reject
+        # at config-load time so the operator sees the gap (parity with
+        # the regime-aware `sl_after` rejection at backtester init).
+        if sc.get("regime_directional_policy"):
+            raise ValueError(
+                f"{config_path}: strategy {strategy_id!r} uses "
+                f"regime_directional_policy, which is HL-live-only in this "
+                f"release (backtester parity deferred — see #779). Use the "
+                f"static `direction` / `invert_signal` fields for backtesting."
+            )
         close_refs = []
         for ref in sc.get("close_strategies", []) or []:
             if isinstance(ref, dict) and ref.get("name"):
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -297,39 +297,40 @@ type StrategyRef struct {
 
 // StrategyConfig describes a single strategy job.
 type StrategyConfig struct {
-	ID                     string              `json:"id"`
-	Type                   string              `json:"type"`                // "spot", "options", "perps", "futures", or "manual"
-	Platform               string              `json:"platform"`            // "deribit", "ibkr", "binanceus", "hyperliquid", "topstep"
-	Symbol                 string              `json:"symbol,omitempty"`    // manual strategies: trading symbol (e.g. "ETH")
-	Timeframe              string              `json:"timeframe,omitempty"` // manual strategies: OHLCV timeframe (e.g. "1h")
-	Script                 string              `json:"script"`
-	Args                   []string            `json:"args"`
-	OpenStrategy           StrategyRef         `json:"open_strategy"`              // entry strategy ref (name + params). Migrated from legacy string-typed open_strategy / args[0] in v13 (#640)
-	CloseStrategies        []StrategyRef       `json:"close_strategies,omitempty"` // exit strategy refs (name + params); max close_fraction wins (#480). Migrated from legacy []string in v13 (#640)
-	AllowedRegimes         []string            `json:"allowed_regimes,omitempty"`  // gate entries: skip signal when current regime not in this list; empty = allow all (#482)
-	Capital                float64             `json:"capital"`
-	CapitalPct             float64             `json:"capital_pct,omitempty"`     // 0-1; dynamic capital = wallet_balance * capital_pct (overrides capital)
-	InitialCapital         float64             `json:"initial_capital,omitempty"` // fixed starting balance for PnL display (never overwritten by capital_pct)
-	MaxDrawdownPct         float64             `json:"max_drawdown_pct"`
-	IntervalSeconds        int                 `json:"interval_seconds,omitempty"`           // per-strategy override (0 = use global)
-	HTFFilter              bool                `json:"htf_filter,omitempty"`                 // higher-timeframe trend filter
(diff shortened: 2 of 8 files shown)


richkuo:
@claude

richkuo:
@claude fix findings

richkuo:
@claude re-review

PR #786, opened 2026-05-26
fix(update.sh): signal fallback when systemd unit missing

### LINKED ISSUE #785: update.sh: systemd restart fallback missing when unit not found
`scripts/update.sh --restart` aborts during the restart phase when the configured systemd unit is not loaded. Under `set -euo pipefail`, a failed `sudo systemctl restart` exits the script immediately with **systemctl's** exit status (typically **5** for "unit not found") — not a script-defined error code.

**What happened:**
```
Failed to restart go-trader.service: Unit go-trader.service not found.
```
Phases through atomic swap already completed (`go-trader.new` → `go-trader`, `.prev` retained), but the old trader PID was never replaced, so the process kept running stale code.

**Expected behavior:**
When default `--restart-mode systemd` cannot restart the unit (unit missing / not loaded), fall back to the existing signal path: SIGTERM/SIGKILL the PID from `GO_TRADER_PIDFILE`, respawn via `GO_TRADER_RUN_SH`, then the same `/health` + PID-freshness verify/rollback used by explicit `--restart-mode signal`.

**Root cause:**
- `RESTART_MODE` defaults to `systemd` (`scripts/update.sh:37`), independent of `GO_TRADER_SERVICE` (unit name only).
- Restart branch calls `sudo systemctl restart "$service_unit"` with no error handling (`scripts/update.sh:586`); `set -e` aborts on failure.
- Signal restart exists (`scripts/update.sh:599-610`) but only when `--restart-mode signal` / `RESTART_MODE=signal` is set explicitly (#766).
- Docs/README describe **two explicit deploy modes** (systemd vs signal), not an automatic fallback after systemd failure — the gap is implementation, not documented behavior today.

**To reproduce:**
1. Linux deployment with `scheduler/config.json`, pidfile + `run.sh` (bare-process), **no** systemd unit installed/enabled.
2. `bash scripts/update.sh --restart` (default systemd mode).

**Environment:**
- `GO_TRADER_SERVICE` unset → unit name `go-trader` (systemd may display as `go-trader.service` in errors).
- `RESTART_MODE` unset → `systemd`.

**Out of scope:** `applyUpgrade` in `scheduler/updater.go` invokes `scripts/update.sh` **without** `--restart` and restarts via `restartSelf()` — this bug affects manual `update.sh --restart` (and `--all --restart` children), not the auto-update DM path.

**Suggested fix:**
On `systemctl restart` failure with exit status 5 (unit not found), retry restart using the signal code path already in the script. Reuse signal-mode preflight where possible: Linux `/proc`, readable pidfile, executable `GO_TRADER_RUN_SH` (defaults `./go-trader.pid`, `./run.sh`; see `scripts/create-run-sh.sh`). If prerequisites are missing, fail with a clear message instead of silent no-op.

Also consider: `do_rollback` in systemd mode calls `systemctl restart` with `|| true` (`scripts/update.sh:330`) — rollback cannot recover a missing unit either; align rollback behavior if signal fallback is added.

richkuo:
## Summary
- On `systemctl restart` exit 5 (unit not found), `update.sh` falls back to the existing signal restart path (SIGTERM pidfile PID, respawn via `run.sh`) instead of exiting after swap.
- Verify and rollback use pidfile semantics when fallback ran; rollback also tries signal when systemd unit is missing.
- README and `--help` document the behavior.

## Test plan
- [x] `bash -n scripts/update.sh`
- [x] `go -C scheduler test -run TestUpdateShellScriptSyntax`
- [ ] On bare-process Linux host without systemd unit: `bash scripts/update.sh --restart` logs fallback and respawns
- [ ] Systemd deploy unchanged: `sudo bash scripts/update.sh --restart`

Closes #785


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #788, opened 2026-05-26
fix(probe): skip live credential checks during update.sh probe

### LINKED ISSUE #787: update.sh: probe fails when HYPERLIQUID_SECRET_KEY not in shell env
update.sh probe phase rejects the binary when `HYPERLIQUID_SECRET_KEY` is not set in the shell environment, even though probe is a dry run that never connects to the exchange.

**What happened:**
```
probe: failed to load config scheduler/config.json: config validation errors:
  strategy[hl-tema-eth-live]: --mode=live requires HYPERLIQUID_SECRET_KEY env var
  strategy[hl-rmc-eth-live]: --mode=live requires HYPERLIQUID_SECRET_KEY env var
[update] FAIL phase=probe: go-trader.new probe rejected the freshly synced Python — refusing to swap
```

The binary was built successfully but the swap was refused because probe couldn't validate live-mode strategies without the secret key.

**Workaround:**
Manually pass the env var: `HYPERLIQUID_SECRET_KEY=*** bash scripts/update.sh`

**Root cause:**
`--mode=live` strategies validate the presence of `HYPERLIQUID_SECRET_KEY` at config load time. The probe phase loads the full config to verify binary compatibility but doesn't actually connect to any exchange — so the secret check is unnecessary during probe.

**Suggested fixes (pick one):**
1. Probe should set a flag that tells config validation to skip secret-key checks (e.g. `--probe` mode)
2. `update.sh` should source env vars from the running process (`/proc/<pid>/environ`) before running probe
3. Config validation should defer secret-key checks until actual exchange connection is attempted, not at load time

**Environment:**
- go-trader v0.66.2 → v0.66.3
- Live instance with `--mode=live` HL strategies
- No `HYPERLIQUID_SECRET_KEY` in shell environment (set elsewhere at runtime)

richkuo:
## Summary

- Adds `LoadConfigForProbe` so `go-trader probe` (and `scripts/update.sh` pre-swap validation) skips live-mode credential env checks (#787).
- Probe only verifies the Python argv contract; it never connects to exchanges. Secrets may exist only in the running process (systemd `EnvironmentFile`, etc.), not in the shell running `update.sh`.
- Normal `LoadConfig` / startup still requires `HYPERLIQUID_SECRET_KEY` and other live credentials.

## Test plan

- [x] `TestValidateConfigHLLiveRequiresSecretKey` — startup path still fails without secrets
- [x] `TestLoadConfigForProbeSkipsLiveCredentialChecks` — probe path accepts live HL config without env
- [x] `TestRunProbeSkipsLiveCredentialChecks` — end-to-end `runProbe` with `--mode=live`

Closes #787

LLM: Composer | default | Harness: agent

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -569,7 +569,21 @@ func EffectiveInitialCapital(sc StrategyConfig, ss *StrategyState) float64 {
 	return sc.Capital
 }
 
+// LoadConfig loads and validates a config file. Live-mode strategies require
+// platform credential env vars to be set in the process environment.
 func LoadConfig(path string) (*Config, error) {
+	return loadConfig(path, false)
+}
+
+// LoadConfigForProbe loads config for `go-trader probe` / update.sh pre-swap
+// validation. Skips live-credential env checks (#787): probe only verifies the
+// Python argv contract and never connects to exchanges; secrets may live only in
+// the running process (systemd EnvironmentFile, etc.).
+func LoadConfigForProbe(path string) (*Config, error) {
+	return loadConfig(path, true)
+}
+
+func loadConfig(path string, skipLiveCredentialChecks bool) (*Config, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, fmt.Errorf("read config: %w", err)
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

PR #791, opened 2026-05-26
update.sh: --rsync-from safe sync and EnvironmentFile warning

### LINKED ISSUE #790: update.sh: add --rsync-from mode with hardcoded rsync exclusions
Two related safeguards for update.sh:

## 1. `--rsync-from` mode with hardcoded rsync exclusions

Add a `--rsync-from <source-dir>` flag (not `--sync` — that name collides with the existing `uv sync` phase in `update.sh`) that rsyncs the full codebase from a source repo into the deployment directory with a hardcoded exclusion list:

- `.env`
- `scheduler/config.json`
- `scheduler/state.db` (default `db_file`; honor configured path when implementing)
- `trading_bot.db*` (legacy `shared_tools` path; low risk but safe to exclude)
- `.venv/`
- `node_modules/` (not used in-repo today; harmless to exclude)
- `__pycache__/`
- `go-trader` / `go-trader.new` / `go-trader.prev` (live + staged + rollback binaries)
- `go-trader.pid`

This prevents operators from accidentally deleting deployment-critical files when manually syncing from a source clone (see SKILL.md #702/#703).

## 2. Preflight warning for missing EnvironmentFile

When `--restart` is set and restart mode is systemd: if the unit’s `EnvironmentFile=` path (parse from `systemctl cat`, not hardcoded — paths differ for `go-trader@.service` instances) does not exist, print a bold warning before the restart phase. Don’t abort — let the operator decide.

Units use `EnvironmentFile=-…` (optional), so a missing file does **not** block `systemctl restart`; the process still starts but secrets (e.g. `DISCORD_BOT_TOKEN`) are absent. The warning catches accidental deletion (e.g. `rsync --delete` without `.env` / `scheduler/config.json` exclusions) before a restart leaves the instance running without credentials.

richkuo:
## Summary
- Add `--rsync-from <source-dir>` to rsync code into a deployment with hardcoded exclusions (replaces `git pull` for that run).
- Warn in bold before systemd restart when the unit's `EnvironmentFile` paths are missing (non-blocking).
- Rollback uses `tree_mutated` so rsync runs revert via `git reset --hard` like pull updates.

## Test plan
- [x] `bash -n scripts/update.sh`
- [x] `go test -C scheduler -run TestUpdateShell -count=1`
- [ ] Manual: `bash scripts/update.sh --rsync-from /path/to/clone` (no `--restart`) on a deployment dir
- [ ] Manual: `bash scripts/update.sh --restart` with a missing `.env` referenced by the unit — confirm bold warning, service still restarts

Closes #790

### DIFF
--- a/scheduler/update_sh_feature_test.go
+++ b/scheduler/update_sh_feature_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+func updateShellScriptPath(t *testing.T) string {
+	t.Helper()
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("runtime.Caller failed")
+	}
+	return filepath.Join(filepath.Dir(thisFile), "..", "scripts", "update.sh")
+}
+
+func TestUpdateShellHelpDocumentsRsyncFrom790(t *testing.T) {
+	t.Parallel()
+	script := updateShellScriptPath(t)
+	out, err := exec.Command("bash", script, "--help").CombinedOutput()
+	if err != nil {
+		t.Fatalf("bash %s --help: %v\n%s", script, err, out)
+	}
+	text := string(out)
+	for _, want := range []string{
+		"--rsync-from",
+		"hardcoded exclusions",
+		".env",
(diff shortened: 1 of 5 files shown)


richkuo:
@claude

richkuo:
@claude re-review

PR #796, opened 2026-05-28
Opt-in 7-state composite regime classifier (per-window selector)

### LINKED ISSUE #795: Opt-in 7-state composite regime classifier (per-window selector)
## Motivation

The current ADX regime detector has a known blind spot: slow directional drifts
that are choppy bar-by-bar get labeled `ranging`. Concrete example: ETH fell
~12% over 30 days on 30m bars; ADX(1500) labeled the window `ranging` because
per-bar +DM and -DM cancelled, even though the move is unambiguous on a chart.

This issue adds a composite 7-state classifier as an **opt-in alternative**
selectable per window, alongside the existing ADX classifier (default unchanged).

## Per-window classifier selector

The classifier becomes a property of each window (#792's unit), not a global
mode. ADX stays the default.

```json
"regime": {
  "enabled": true,
  "windows": {
    "fast":  {"classifier": "adx",       "period": 14},
    "macro": {"classifier": "composite", "period": 720,
              "thresholds": {"return_pct": 0.05, "range_pct": 0.03, "adx": 25}}
  }
}
```

Bare-int JSON (existing format) parses as `{classifier: "adx", period: N}` for
backwards compat.

Per-strategy routing is unchanged:

```json
"regime_gate_window": "fast",
"regime_directional_window": "macro",
"regime_atr_window": "fast"
```

A single strategy can mix coarse (ADX) and fine (composite) windows.

## Composite classifier design

Three measurements per window:

1. **% return over window**, ATR-normalized for cross-asset comparability.
   Catches slow drifts ADX misses.
2. **ADX** — persistence / move quality. Distinguishes clean from choppy.
3. **Range / ATR-normalized width** — distinguishes asleep from awake.

The 3-tuple maps to one of 7 labels:

| Label                  | % return | ADX  | Range  | Description                          |
|------------------------|----------|------|--------|--------------------------------------|
| `trending_up_clean`    | big up   | high | wide   | strong, persistent uptrend           |
| `trending_up_choppy`   | big up   | low  | wide   | real uptrend but noisy               |
| `trending_down_clean`  | big down | high | wide   | strong, persistent downtrend         |
| `trending_down_choppy` | big down | low  | wide   | real downdrift but choppy (ETH case) |
| `ranging_quiet`        | small    | low  | narrow | market asleep                        |
| `ranging_volatile`     | small    | low  | wide   | whipsaw / chop                       |
| `ranging_directional`  | small    | high | wide   | big move that round-tripped          |

## Validation

Each window emits labels from its own classifier's vocabulary. ADX windows emit
the existing 3-state set; composite windows emit the 7-state set. Startup
validation:

1. `allowed_regimes` keys must match the classifier feeding `regime_gate_window`.
2. `regime_directional_policy` keys must match the classifier feeding
   `regime_directional_window`.
3. `*_atr_regime` and `sl_after_regime` keys must match the classifier feeding
   `regime_atr_window`.
4. Mismatch fails startup with a clear strategy + field + expected-vocabulary
   message.

`use_defaults` continues to fill unspecified per-regime entries.

## SIGHUP / state compat

- Changing a window's `classifier` while any open position references that
  window is blocked in `validateHotReloadStateCompatible` (same pattern as
  `*_atr_regime` shape changes today).
- When flat, swap applies on the next cycle.
- `Position.Regime` and `Position.RegimeWindows` stamps stay in the vocabulary
  active at open time for the position's lifetime. `regimeLabelAtOpen` and
  `EffectiveDirectionForPosition` read raw without re-validation.

## Required work

- [ ] `RegimeWindowSpec` struct replacing `map[string]int`; bare-int JSON still
      parses to `{classifier: \"adx\", period: N}` for back-compat
- [ ] `shared_tools/regime.py` gains `classify(method, bars, params) -> label`
      dispatcher with `adx` (existing) and `composite` (new) implementations
- [ ] Composite implementation: ATR-normalized % return, ADX (reuses existing
      `_compute_adx_components`), ATR-normalized range; 3-tuple → 7-state mapper
- [ ] `validRegimeLabels` becomes a per-classifier set: `validRegimeLabelsADX`
      (3) and `validRegimeLabelsComposite` (7)
- [ ] Config validators on `allowed_regimes`, `regime_directional_policy`,
      `*_atr_regime`, `sl_after_regime` resolve the relevant window's classifier
      and validate keys against the right vocabulary
- [ ] Check scripts (5 standard + options): replace `--regime-period` /
      `--regime-adx-threshold` with `--regime-windows-spec-json`
      (per-window classifier + thresholds)
- [ ] `version_probe.go` `probeArgv` + `executeProbeArgv` updated for all 5
      check scripts to include the new flag (covers both classifiers)
- [ ] Backtester parity: `ensure_regime_columns` becomes classifier-aware per
      window; lookahead regression suite (`backtest/tests/test_backtester_lookahead.py`)
      extended to cover composite
- [ ] SIGHUP guard in `validateHotReloadStateCompatible`: block classifier swap
      on a window referenced by any open position
- [ ] `inspect`, `/status`, Discord summary, leaderboard render fine labels
      with the classifier name (e.g. `macro=trending_down_choppy [composite]`)
- [ ] Calibration notebook: BTC/ETH/SOL 30m + 1h + 1d, 1y window, ADX vs
      composite on a forward-return metric. Output sets default composite
      thresholds shipped in code.
- [ ] SKILL.md / CLAUDE.md / AGENTS.md updated for the new selector + 7-state
      label set

## Calibration

Before defaults are committed, the calibration notebook compares ADX-only vs
composite on:

1. State frequency per asset/timeframe (sanity check on rare states like
   `ranging_directional`)
2. Forward 1-day-return distribution per label (does the label predict?)
3. Confusion matrix against ADX baseline (which cases do they disagree on?)

Default thresholds in `RegimeConfig` are set from this analysis. Operators can
override per window.

## Rollout

1. Ship with composite as opt-in, ADX default unchanged. No migration.
2. Enable composite on one paper strategy. Compare alerts and PnL against
   coarse-equivalent on a parallel strategy for 2-4 weeks.
3. If composite clearly wins, promote on one live strategy. Otherwise tune
   thresholds or remove.
4. After 2-3 months of live evidence either way: flip default to composite, or
   delete the classifier and keep the per-window selector framework for future
   experiments (HMM, slope+range, etc.).

## Honest trade-offs

**Pro:** ADX default unchanged, reversible per-strategy, lets calibration
happen in production on one strategy without committing the whole system.
Right path for an unproven idea.

**Con:** two classifiers to maintain forever (or until one wins). Every future
regime feature (next #779-style policy) must support both label spaces. Real
long-term operational cost.

## Out of scope

- Removing ADX classifier (kept as default until composite proves out
  on live PnL)
- HMM-based regime classifier — orthogonal future work; per-window selector
  framework makes that easy to add later as `classifier: \"hmm\"`
- Options regime support (`check_options.py` defers to underlying; unchanged)
- A 4th measurement (volume, funding rate) — orthogonal future work

## Risks

- **Calibration is the long pole.** 7 buckets means rare states will be
  under-sampled per asset; thresholds need real-data tuning before defaults
  are committed.
- **Composite is unvalidated.** No paper, no desk track record. The fact that
  ADX has a known blind spot doesn't automatically mean composite is better
  on out-of-sample.
- **Operator burden.** Strategies using composite must specify behavior across
  7 states per consumer (`allowed_regimes`, directional policy, ATR). In
  practice operators will copy-paste, which dilutes the value of fine labels.

richkuo:
## Summary
- Adds per-window `RegimeWindowSpec` (`adx` default, `composite` opt-in) with bare-int JSON back-compat for `regime.windows`.
- Implements composite classifier in `shared_tools/regime.py` (ATR-normalized return + ADX + range → 7 labels) and threads `--regime-windows-spec-json` through all five check scripts.
- Validates `allowed_regimes`, `regime_directional_policy`, and `*_atr_regime` keys against the classifier on each consumer window; blocks classifier swaps on SIGHUP when open positions reference that window.

## Test plan
- [x] `go -C scheduler test ./...`
- [x] `uv run --no-sync python -m pytest shared_tools/test_regime.py`
- [x] `uv run --no-sync python -m pytest backtest/tests/test_backtester_regime.py`
- [ ] Paper strategy with one composite window vs ADX gate window (rollout step 2 from issue)

Closes #795


richkuo:
@claude

richkuo:
@claude re-review

PR #803, opened 2026-05-29
fix(regime): resolve regime-aware ATR labels from window classifier (#802)

### LINKED ISSUE #802: validateRegimeATRConfig hardcodes ADX 3-state labels, breaking composite 7-state stop_loss_atr_regime
### Summary
When using the composite regime classifier (7-state labels: `trending_up_clean`, `trending_up_choppy`, `trending_down_clean`, `trending_down_choppy`, `ranging_quiet`, `ranging_volatile`, `ranging_directional`), config validation fails if `stop_loss_atr_regime` or `trailing_stop_atr_regime` is configured with explicit per-regime entries.

`validateRegimeATRConfig()` calls `ResolveSurface()` which unconditionally uses `canonicalTrendRegimeLabels` (ADX 3-state: `trending_up`, `trending_down`, `ranging`). It produces spurious "unknown regime label" / "missing required regime labels" errors for the 7-state composite vocabulary, which fail the config load. `validateStrategyRegimeVocabulary()` validates the same fields correctly against the strategy's window classifier, but its correctness can't rescue the load because both passes accumulate into the same error slice (see Root Cause).

### Steps to Reproduce

1. Configure a composite regime window:
```json
"regime": {
  "enabled": true,
  "windows": {
    "daily": {"classifier": "composite", "period": 24, "thresholds": {"return_pct": 0.05, "range_pct": 0.03, "adx": 20}}
  }
}
```

2. Configure a strategy with `stop_loss_atr_regime` using 7-state labels:
```json
{
  "id": "hl-test",
  "regime_atr_window": "daily",
  "stop_loss_atr_regime": {
    "trend_regime": {
      "trending_up_clean": {"atr": 2.0},
      "trending_up_choppy": {"atr": 1.2},
      "trending_down_clean": {"atr": 2.0},
      "trending_down_choppy": {"atr": 1.2},
      "ranging_quiet": {"atr": 1.5},
      "ranging_volatile": {"atr": 1.0},
      "ranging_directional": {"atr": 1.5}
    }
  }
}
```

3. Run `go-trader --config config.json --once`

### Expected Behavior
Config loads successfully because all 7 composite labels are present and the `daily` window uses the composite classifier.

### Actual Behavior
Config validation fails with:
```
strategy[hl-test].stop_loss_atr_regime.trend_regime: unknown regime label "trending_up_clean" (expected one of: trending_up, trending_down, ranging)
strategy[hl-test].stop_loss_atr_regime.trend_regime: missing required regime labels: trending_up, trending_down, ranging (must be exhaustive — no silent fallback)
```

### Root Cause
In `scheduler/regime_atr.go`, `validateRegimeATRConfig()` resolves the SL/trailing surfaces via `ResolveSurface()` (`regime_atr.go:637`, `regime_atr.go:666`):
```go
sub := sc.StopLossATRRegime.ResolveSurface(prefix+".stop_loss_atr_regime", regimeSurfaceStopLoss)
```

`ResolveSurface()` (`regime_atr.go:123-125`) hardcodes `canonicalTrendRegimeLabels`:
```go
func (b *RegimeATRBlock) ResolveSurface(ctxLabel string, surface regimeATRSurface) []string {
    return b.ResolveSurfaceWithLabels(ctxLabel, surface, canonicalTrendRegimeLabels)
}
```

`validateStrategyRegimeVocabulary()` in `scheduler/regime_window_spec.go` resolves labels correctly via `regimeLabelsForStrategyWindow()` (`regime_window_spec.go:298,302,309`). **Note: it runs *before* `validateRegimeATRConfig()`, not after** — `validateStrategyRegimeVocabulary()` is called at `config.go:1714` and `validateRegimeATRConfig()` at `config.go:1763`. Both passes append to the *same* `errs` slice and there is no early return between them; the load only fails at `config.go:1765` (`if len(errs) > 0`). So ordering is not the problem — the problem is that `validateRegimeATRConfig()` uses hardcoded ADX labels and leaves spurious errors in the slice that fail the load regardless of which pass runs first.

### Scope
The same hardcoded-ADX defect affects the regime-aware TP tier close refs. `tiered_tp_atr_regime` / `tiered_tp_atr_live_regime` tiers are parsed by `parseRegimeTPTiers()`, which passes `canonicalTrendRegimeLabels` to `parseRegimeATRBlock()` (`regime_atr.go:564`). These tier refs are **not** validated by `validateStrategyRegimeVocabulary()`, so a fix scoped only to `stop_loss_atr_regime` / `trailing_stop_atr_regime` would still leave composite-classifier TP tiers broken. The fix should cover the tier path as well (or the issue should explicitly declare it a non-goal).

### Proposed Fix
Make `validateRegimeATRConfig()` resolve labels from the strategy's configured `regime_atr_window` classifier, consistent with `validateStrategyRegimeVocabulary()`:

```go
atrLabels := canonicalTrendRegimeLabels
if regimeEnabled {
    atrLabels = regimeLabelsForStrategyWindow(*sc, cfg.Regime, "atr")
}

if sc.StopLossATRRegime != nil {
    sub := sc.StopLossATRRegime.ResolveSurfaceWithLabels(prefix+".stop_loss_atr_regime", regimeSurfaceStopLoss, atrLabels)
    // ...
}
if sc.TrailingStopATRRegime != nil {
    sub := sc.TrailingStopATRRegime.ResolveSurfaceWithLabels(prefix+".trailing_stop_atr_regime", regimeSurfaceTrailing, atrLabels)
    // ...
}
```

Apply the equivalent label-resolution to the TP tier path (`parseRegimeTPTiers` / `regime_atr.go:564`) so `tiered_tp_atr_regime` and `tiered_tp_atr_live_regime` accept the composite vocabulary too. Given the two passes are now redundant for SL/trailing, consider consolidating the validation into a single pass.

### Workaround
Use `"use_defaults": true` instead of explicit `trend_regime` entries. The `use_defaults` path calls `expandRegimeATRDefaultsForLabels()` (`regime_atr.go:300-319,363`) which maps the 3-state baseline to whatever label set is passed in, so it survives both validation passes. However, this loses the ability to customize per-regime multipliers.

### Environment
- go-trader version: v0.67.0 (commit `20deacd`)
- Go version: 1.26.2

richkuo:
## Summary
- Fixes #802: `validateRegimeATRConfig` hardcoded the ADX 3-state vocabulary when resolving `stop_loss_atr_regime` / `trailing_stop_atr_regime` and `tiered_tp_atr_regime` tiers, so explicit per-regime config under a **composite** (7-state) `regime_atr_window` was rejected at config load with spurious "unknown/missing regime label" errors.
- `validateRegimeATRConfig` now resolves the ATR vocabulary from the strategy's `regime_atr_window` classifier (this is the authoritative pass that also populates the typed runtime fields and runs the mutex checks), and threads those labels into the tier validator.
- Runtime tier resolution (`resolveRegimeTPTiers`) now infers the vocabulary from the already-validated raw config, so it stays classifier-agnostic without needing `RegimeConfig` in scope. The `use_defaults` tier path maps composite labels onto their ADX-family baseline via a shared `mapRegimeToBaselineFamily` helper.
- Removes the now-redundant SL/trailing re-resolution from `validateStrategyRegimeVocabulary` (avoids double-reporting label errors).

## Notes
- Also fixes a latent runtime bug: with the old ordering, `use_defaults` SL/trailing blocks were overwritten with a 3-state map by the last-running validation pass, so composite runtime SL/trailing resolution would silently miss. The authoritative pass now populates the composite vocabulary.
- **Out of scope (follow-up):** `sl_after` regime variants (`post_tp_sl.go`) share the same hardcoded-ADX defect via a separate re-parsed-at-runtime path. They continue to **fail closed** at config load for composite windows (safe), and warrant their own focused change.

## Test plan
- [x] `go build -ldflags ...` (release build)
- [x] `go test ./...` (full scheduler suite, green)
- [x] New regression tests in `scheduler/regime_atr_composite_test.go`:
  - composite 7-state `stop_loss_atr_regime` / `trailing_stop_atr_regime` validate and populate the 7-label runtime map
  - composite missing-label rejected with composite vocabulary + classifier context (not ADX)
  - composite `tiered_tp_atr_regime` explicit tiers validate; runtime `resolveRegimeTPTiers` and `use_defaults` resolve composite labels
  - `mapRegimeToBaselineFamily` / `regimeLabelsFromTierRaw` unit coverage
  - end-to-end `LoadConfig` repro (full `ValidateConfig` pipeline)
- [x] `gofmt` clean, no new lints

LLM: Claude Opus 4.8 | high | commit-push-pr


richkuo:
@claude

PR #821, opened 2026-05-29
Stop systemd restart loop on startup probe failure

### LINKED ISSUE #820: Stop systemd restart loop on startup probe failure
## Problem

When the startup compatibility probe fails (e.g. missing `shared_scripts/strategy_tuner_schema.py` after a binary-only deploy), `go-trader` already logs the error, DMs the owner, and exits.

systemd units use `Restart=always` with `RestartSec=10`, so the process is restarted every 10 seconds until the deploy is fixed. Each attempt sends another Discord DM. The failure is easy to misread as a repeated in-app retry.

The generic probe error text also says "binary/Python version mismatch" when Python cannot open the script file (`[Errno 2] No such file or directory`), which sent operators down the wrong debugging path.

## Example

```
Startup probe failed — check-script CLI mismatch:
shared_scripts/strategy_tuner_schema.py rejected --probe-only argv ...
can't open file '.../shared_scripts/strategy_tuner_schema.py': [Errno 2] No such file or directory
```

Root cause was missing Python tree, not argv drift.

## Proposed fix

1. Exit with a dedicated status (**78**, EX_CONFIG) on probe failure so units can opt out of auto-restart.
2. Add `RestartPreventExitStatus=78` to `go-trader.service` and `systemd/go-trader@.service`.
3. Detect "can't open file" / "No such file" in probe stderr and report "script missing from deploy tree" instead of CLI mismatch.
4. One DM per failed start (service stays down until operator syncs scripts and restarts).

## Operator notes

- Deploy fix remains: full `scripts/update.sh` (or pull/rsync `shared_scripts/` + `uv sync` + rebuild), not binary-only.
- Custom supervisors (e.g. OpenClaw) should not restart on exit 78, same as systemd.

## Acceptance

- Probe failure: single stderr line + one owner DM, process exits 78, systemd does not restart until manual `systemctl start`.
- `go-trader probe` returns 78 on failure (update.sh still aborts on any non-zero).
- Unit test for missing-script error wording.

richkuo:
## Summary

- Exit **78** (`ExitProbeFailure`) when the startup compatibility probe fails so operators get one failure, not a `Restart=always` loop every 10s.
- Add `RestartPreventExitStatus=78` to `go-trader.service` and `systemd/go-trader@.service`.
- Report missing probe scripts (`can't open file`) as deploy-tree gaps, not "CLI mismatch".
- `go-trader probe` returns 78 on failure (update.sh still treats any non-zero as abort).

Closes #820

## Test plan

- [x] `go test ./scheduler/ -run 'Probe|FormatProbe'`
- [ ] After deploy: induce probe failure (rename `strategy_tuner_schema.py`); confirm exit 78, single DM, `systemctl status` shows **failed (Result: exit-code)** — not activating in a loop
- [ ] `sudo systemctl daemon-reload` after installing updated unit file
- [ ] Fix deploy with `scripts/update.sh`; confirm probe passes and service starts

### DIFF
--- a/go-trader.service
+++ b/go-trader.service
@@ -12,6 +12,8 @@ WorkingDirectory=/opt/go-trader
 ExecStart=/opt/go-trader/go-trader --config /opt/go-trader/scheduler/config.json
 Restart=always
 RestartSec=10
+# Probe/deploy failures exit 78 — do not restart until scripts are synced.
+RestartPreventExitStatus=78
 StandardOutput=journal
 StandardError=journal
 
--- a/scheduler/exit_codes.go
+++ b/scheduler/exit_codes.go
@@ -1,0 +1,7 @@
+package main
+
+// ExitProbeFailure is used when the startup compatibility probe fails (missing
+// Python scripts, argparse mismatch, probe timeout). systemd units should set
+// RestartPreventExitStatus=78 so the service stays down instead of restarting
+// every RestartSec and spamming Discord. (EX_CONFIG from sysexits.h.)
+const ExitProbeFailure = 78
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -400,17 +400,14 @@ func main() {
 	}
 
 	// #645: Verify each configured check script accepts the binary's argv
-	// shape before entering the trading loop. An asymmetric deploy (binary
-	// from one commit, Python scripts from an older one) caused 18h of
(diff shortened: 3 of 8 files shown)


richkuo:
@claude

richkuo:
@claude sonnet re-review

PR #823, opened 2026-05-30
Perps: auto-resolve orphan positions from regime/direction flips

### LINKED ISSUE #822: Perps: auto-resolve orphan positions from regime/direction flips
## Problem

When a perps position is opened under one regime/direction and the regime later flips (e.g., `trending_down_clean` → `trending_up_clean`), the position becomes an **orphan** — its stamped regime/direction no longer matches the current strategy config. Example warning:

```
perps state-vs-config gap: strategy hl-vwap-btc-60 has long BTC qty=0.0137
(effective_direction="short" from stamped regime="trending_down_clean"; base_direction="short").
Position was likely seeded by migration, paper→live handoff, or a prior conflicting direction.
Close manually before the next signal.
```

The system currently has three **passive** defense layers that detect but don't resolve:

1. **`ValidatePerpsDirectionConfig`** (startup) — warns, does nothing
2. **`PerpsOrderSkipReason` / `executePerpsSignalWithLeverage`** — skips signal, logs "orphan" warning, position stays
3. **`reconcileHyperliquidPositionsWithResolver`** — syncs virtual qty/side to on-chain, ignores direction conflicts

## Impact

- Orphan position sits indefinitely until manually closed
- Executor sizes **fresh opens** assuming flat → stacks a new position on top of the orphan on-chain
- Virtual state desyncs from exchange (cash accounting drifts)
- If multiple strategies share the coin via shared-wallet, the gap compounds

### Option A: Auto-close orphan on regime flip (recommended)
When the reconciler detects a position whose stamped regime conflicts with the **current** regime's effective direction:
- Close the position (on-chain + virtual) at market
- Book the trade with reason `"regime_direction_flip"`
- Emit a DM notification to the operator

This mirrors what a human operator does today but removes the latency window where the orphan can compound.

### Option B: Configurable orphan policy
Add a field to `StrategyConfig`:
```json
"orphan_position_policy": "auto_close" | "warn_only" | "hold"
```
- `auto_close` (default for new configs): auto-resolve as above
- `warn_only`: current behavior
- `hold`: suppress even the warning (for operator-seeded positions that should persist)

### Option C: Direction-aware reconciler
Extend the reconcile phase to detect direction conflicts and auto-close, similar to how it already handles externally-closed positions (on-chain flat, virtual still has position). This keeps the fix centralized in one place.

## Suggested Implementation

The most contained fix is **Option A** in the reconcile phase:

1. In `reconcileHyperliquidPositionsWithResolver`, after the qty/side resync block, check if `perpsPositionConflictsDirection(pos.Side, effectiveDir)` using the **current** regime (not stamped)
2. If conflict: close on-chain via existing `RunHyperliquidClose`, book the close via `bookPerpsCloseWithFillFee`, log + notify
3. The startup validator (`ValidatePerpsDirectionConfig`) remains as a safety net but should no longer fire in normal operation

### Key considerations:
- Must use the **current** regime, not the stamped one — the stamped regime reflects what *was*, the current regime reflects what the strategy expects *now*
- Should not auto-close positions opened via `manual-open` (those have `ManualOpen=true` or similar flag)
- Must respect shared-wallet strategies — only auto-close if the strategy is the sole owner of that coin
- Notification is critical so the operator knows it happened and can audit

richkuo:
## Summary
- During sole-owner HL hl-sync reconcile, detect when an open perps position conflicts with the **current** regime's effective direction (not the stamped open regime).
- Queue a reduce-only market close after `mu` unlock, book the fill with `regime_direction_flip`, and DM the operator.
- Skip shared-wallet coins, paper mode, and non-owner positions; startup `ValidatePerpsDirectionConfig` remains a safety net.

Closes #822

## Test plan
- [ ] Smoke on live HL perps with `regime_directional_policy` after a deliberate regime flip (verify DM + trade reason)

LLM: Composer | default | commit-and-create-pull-request

### DIFF
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -279,7 +279,7 @@ func fetchHyperliquidState(accountAddress string) (float64, []HLPosition, error)
 // cached resolver outside the lock and call
 // reconcileHyperliquidPositionsWithResolver.
 func reconcileHyperliquidPositions(stratState *StrategyState, sym string, positions []HLPosition, accountAddress string, logger *StrategyLogger) bool {
-	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger, nil, StrategyConfig{})
+	return reconcileHyperliquidPositionsWithResolver(stratState, sym, positions, directHyperliquidReconcileFillResolver(accountAddress), logger, nil, nil, StrategyConfig{})
 }
 
 // reconcileHyperliquidPositionsForStrategy is the production entry point with
@@ -308,6 +308,7 @@ func reconcileHyperliquidPositionsForStrategy(
 	resolveFee hlReconcileFillResolver,
 	logger *StrategyLogger,
 	pendingAlerts *[]ProtectionFillAlert,
+	pendingOrphanCloses *[]RegimeDirectionOrphanCloseJob,
 ) bool {
 	if stratState == nil || sym == "" {
 		return false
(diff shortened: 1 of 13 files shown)


richkuo:
@claude

richkuo:
@claude sonnet rereview

richkuo:
@claude sonnet rereview

PR #830, opened 2026-06-01
feat: alert operators when signal scripts fail consecutively

### LINKED ISSUE #829: Alert when strategy signal script fails repeatedly (general)
## Problem

When a signal script (e.g., `check_hyperliquid.py`, `check_okx.py`, etc.) fails repeatedly, the error is logged but no alert is sent. The bot continues cycling and responding to status requests, making it appear "working" even though no actual trading is happening.

Current behavior:
- `logger.Error("Script returned error: %s", result.Error)` logs the failure
- Status port still responds with stale data
- No notification sent to Discord/operators

## Example failure from live run (2026-05-31)

```
[hl-rmc-eth-live] [ERROR] Script returned error: list index out of range
  File ".../hyperliquid/info.py", line 48, in __init__
    base_info = spot_meta["tokens"][base]
IndexError: list index out of range
```

The strategy failed on every cycle from ~16:05 to 16:10 UTC but there was no alert — only the log showed it.

## Root cause (specific case)

The crash in `check_hyperliquid.py` happens at adapter init in `platforms/hyperliquid/adapter.py:_build_info()` when the Hyperliquid API returns a corrupted or stale spot_meta token list. The adapter doesn't handle this gracefully.

## Expected behavior (general)

Track consecutive script errors per strategy across **all** signal scripts. After a threshold (e.g., 3-5 consecutive failures), send an alert via the existing failure alert system (Discord notifier, etc.).

This should apply to any signal script: `check_hyperliquid.py`, `check_okx.py`, `check_binance.py`, `check_robinhood.py`, `check_topstep.py`, etc.

## Proposed solution

Add a consecutive failure counter per strategy in the Go scheduler. After N consecutive failures, trigger an alert through the existing failure alert infrastructure.

**Cover both failure paths, not just `result.Error`.** A script can fail two distinct ways, and "dead in the water" must catch both:
1. **Soft error** — script emits JSON with a non-empty `result.Error`; logged as `logger.Error("Script returned error: %s", result.Error)` (e.g. `main.go:2491` for HL). The example above is this path.
2. **Hard crash** — non-zero exit with no usable JSON (timeout, OOM, import/init crash, missing script); logged as `logger.Error("Script failed: %v", err)` (e.g. `main.go:2481` for HL).

Increment the counter on **both** branches, reset it to 0 on a successful cycle (including a clean HOLD), and apply it at all six `run*Check` sites (spot, options, HL, topstep, robinhood, okx) — each has both a `Script failed` and a `Script returned error` branch. Otherwise a strategy that hard-crashes every cycle would go undetected.

The alert should include:
- Strategy name
- Number of consecutive failures
- Last error message
- Which failure mode (soft error vs. hard crash)

## Additional context

- The Go scheduler process stays alive so status port still responds
- `trade_count` in status reflects the last successful trade, not the current broken state
- This makes the bot look "healthy" when it's actually dead in the water

richkuo:
## Summary

- Adds `ScriptFailureTracker` (`script_failure_alerts.go`) — a per-strategy consecutive signal-script failure counter mirroring `LiveExecFailureThrottle`
- Alerts via `SendToAllChannels` + `SendOwnerDM` after `scriptFailureAlertThreshold=3` consecutive failures, then throttles (every-10th / hourly / on error-signature change); fires a one-shot recovery notice when the script succeeds after an alerted streak
- Instruments **both** failure branches at all six `run*Check` sites and `runManualCloseEval`: `scriptFailureCrash` (hard crash, `Script failed: %v`, no JSON) and `scriptFailureError` (soft `result.Error`)

The hard-crash path was the gap flagged during issue validation — a strategy that timeouts, OOMs, or import-errors every cycle would have gone undetected with a result.Error-only check.

## Test plan

- [ ] `go build` passes
- [ ] `go test ./...` passes (11 new tests: threshold, throttle cadence, hourly re-alert, error-signature change, recovery, per-strategy independence, message formatting)
- [ ] Confirm on live: after 3+ consecutive `check_hyperliquid.py` failures a **SIGNAL SCRIPT FAILING** Discord message fires; after recovery a **SIGNAL SCRIPT RECOVERED** message fires

Closes #829

---


richkuo:
@claude

PR #832, opened 2026-06-01
fix(hl): resolve spotMeta tokens by index, not list position (#831)

### LINKED ISSUE #831: Hyperliquid: IndexError at HLInfo init — spot pair @367 references out-of-bounds token index
## Problem

`check_hyperliquid.py` crashes on every cycle with `IndexError: list index out of range` at `hyperliquid/info.py:48`:

```python
base, quote = spot_info["tokens"]   # line 47
base_info = spot_meta["tokens"][base]  # line 48  ← IndexError here
IndexError: list index out of range
```

The bot is dead in the water: every cycle fails at adapter init in `HLInfo()` while iterating the spot `universe`.

## Root Cause (confirmed via live API fetch)

Not a format change. A **live `spotMeta` fetch** (`POST api.hyperliquid.xyz/info {"type":"spotMeta"}`) shows the response shape is unchanged:

| Data | Live fresh fetch |
|------|------------------|
| `spot_meta['universe']` count | 301 |
| `spot_meta['universe'][0]` | `{"tokens":[1,0],"name":"PURR/USDC","index":0,"isCanonical":true}` |
| universe entries missing `tokens` | 0 (every entry has it) |
| `spot_meta['tokens']` count | 464 (list) |

`spot_info["tokens"]` is a pair of **integer indices** into the `tokens` list. The crash is a single **out-of-bounds token index**:

- Spot pair **`@367`** has `tokens` referencing index **479**, but `spot_meta["tokens"]` only has **464** entries.
- `spot_meta["tokens"][479]` → `IndexError: list index out of range`.

For reference, `@367` is a non-canonical spot market (token #479 vs USDC):

```json
{"tokens": [479, 0], "name": "@367", "index": 367, "isCanonical": false}
```

Token #479 (`WARS`) **is present** — it sits at list **position 459**, not 479. The data is all there; the SDK just indexes it positionally (`tokens[479]`) on a 464-element list → out of range.

This is an upstream change to Hyperliquid's token **indexing** (indices are now sparse), not a missing token or a format change. It crashes the SDK's `Info` constructor for everyone, **regardless of cache** — the cache only "works" because the cached copy predates pair `@367` being added.

A missing `tokens` field (an earlier hypothesis) is ruled out: that would raise `KeyError` at line 47, not `IndexError` at line 48, and the live data shows every universe entry has `tokens`.

### Deeper SDK fragility: token indices are sparse

The SDK does `spot_meta["tokens"][base]`, treating `base` as a **list position**. But the returned token indices are **not contiguous** — the list has 464 entries while indices run up to **513**. So positional indexing is wrong by design: it only happens to work when `base < 464`, and even then the token at that position may not be the token with that `index`. `@367` (base=479) just makes the latent bug fail loudly. The correct lookup is by each token's `index` field, not list position.

## Steps to Reproduce

1. Delete `/tmp/hl_meta.json`
2. Run any hyperliquid strategy — crashes immediately on the fresh fetch.

(Or fetch `spotMeta` directly and check: pair `@367` → token index 479 ≥ len(tokens)=464.)

## Environment

- go-trader at `/root/.openclaw/workspace/go-trader-live`
- SDK: `hyperliquid-python-sdk` 0.23.0 (latest); same code path in 0.22.0 (`info.py:47-48`)
- Cache age: ~40min when crash occurred

## Fix Needed

In the adapter (`platforms/hyperliquid/adapter.py`), **filter out malformed universe entries before constructing `Info`** — in both `_fetch_raw_meta` (fresh) and the cache-read path:

- Drop any `universe` entry whose `tokens` indices aren't resolvable in the tokens array (out-of-bounds, missing, or non-int).
- Optionally log dropped pairs so we notice if the upstream data changes.

Because token indices are sparse (max index 513 vs 464 entries), the robust resolver is to build an **`index -> token` map** from `spot_meta["tokens"]` and look up `base`/`quote` through it, dropping pairs whose indices aren't in the map — rather than relying on the SDK's positional `tokens[base]`. (No need to "map names to indices" or "reconstruct the tokens field" — the data is already present, just sparsely indexed.)

## Additional Context

- Error is consistent across all hyperliquid strategies (the loop iterates the entire spot `universe`, so one bad pair kills every strategy).
- Crash happens at `HLInfo()` init when iterating universe entries.
- The earlier "BTC, szDecimals" sample came from the perp `meta` response, not `spotMeta`.

richkuo:
## Problem

`check_hyperliquid.py` crashes on every cycle with `IndexError: list index out of range` at `hyperliquid/info.py:48` — the SDK's `Info` constructor dies during adapter init, taking down **every** HL strategy.

Root cause (confirmed via a live `spotMeta` fetch): Hyperliquid's spot **token indices are now sparse**. Token index `479` (`WARS`) sits at list **position 459**, but the SDK resolves tokens positionally — `spot_meta["tokens"][479]` on a 464-element list → out of range. The token data is all present; only the SDK's positional assumption is wrong. Spot pair `@367` (`tokens: [479, 0]`) is the trigger.

This is not a response-format change and not a missing token — earlier hypotheses (dict→list, `universe` losing its `tokens` field) are ruled out; those would raise different errors at different lines.

## Fix

`_normalize_spot_meta(spot_meta)` (`platforms/hyperliquid/adapter.py`):
- Rebuilds `tokens` into an **index-aligned (dense)** list so the SDK's positional `tokens[idx]` resolves the correct token for every referenced index (including sparse high indices).
- Drops any spot pair whose base/quote indices are unresolvable (defends against a genuinely truncated payload), logging what was dropped.
- Fast-path identity return when the payload is already dense + aligned, so the common case and cache round-trip are unchanged.

Wired into `_build_info` at the **SDK boundary** on both the cache-hit and fresh-fetch paths, so a poisoned cache is normalized too. The raw upstream payload is still cached verbatim.

Why index-alignment over just dropping `@367`: dropping is whack-a-mole — every new sparse-indexed pair would recur the crash. Resolving by index is correct for any sparse/gapped token list HL ships.

## Verification

- New tests in `platforms/hyperliquid/test_meta_cache.py` re-implement the SDK's exact spot loop to prove the raw payload crashes and the normalized one doesn't, plus drop/identity/malformed cases.
- End-to-end against a live `spotMeta`: raw payload `IndexError`; normalized resolves **all 301 pairs, zero dropped**, `@367`→`WARS` szDecimals correct, `tokens[479].name == "WARS"`.
- `pytest platforms/ shared_tools/ shared_strategies/` → 843 passed.

Closes #831

---

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -179,6 +179,80 @@ def _save_meta_cache(spot_meta, meta, path: str = META_CACHE_PATH) -> None:
                 pass
 
 
+def _normalize_spot_meta(spot_meta):
+    """Make ``spot_meta`` safe for the SDK's positional ``tokens[idx]`` lookup.
+
+    The SDK's ``Info`` constructor resolves each spot pair's base/quote tokens
+    via ``spot_meta["tokens"][idx]`` (info.py:47-49), treating the token's
+    ``index`` as a LIST POSITION. Hyperliquid's token indices used to be dense
+    (position == index) but have become sparse — e.g. token index 479 ("WARS")
+    now lives at list position 459 — so the SDK's ``tokens[479]`` raises
+    ``IndexError`` and kills ``HLInfo`` init for every strategy (#831). The
+    token data is all present; only the positional assumption is wrong.
+
+    We rebuild ``tokens`` into an index-aligned (dense) list so positional
+    lookup resolves the correct token for every referenced index, and drop any
+    spot pair whose token indices can't be resolved at all (defends against a
+    genuinely truncated payload). Returns a shallow copy; the original is left
+    untouched. Non-dict/list inputs pass through unchanged so a malformed
+    payload still reaches the SDK's own validation.
+    """
+    if not isinstance(spot_meta, dict):
+        return spot_meta
+    tokens = spot_meta.get("tokens")
+    universe = spot_meta.get("universe")
+    if not isinstance(tokens, list) or not isinstance(universe, list):
+        return spot_meta
+
+    by_index = {}
+    max_index = -1
+    for tok in tokens:
+        if isinstance(tok, dict) and isinstance(tok.get("index"), int):
+            idx = tok["index"]
+            by_index[idx] = tok
+            if idx > max_index:
+                max_index = idx
+
+    # Drop spot pairs whose base/quote indices aren't resolvable.
+    clean_universe = []
+    dropped = []
+    for entry in universe:
+        pair = entry.get("tokens") if isinstance(entry, dict) else None
+        if (
+            isinstance(pair, list)
+            and len(pair) == 2
+            and all(isinstance(x, int) and x in by_index for x in pair)
+        ):
+            clean_universe.append(entry)
+        else:
+            dropped.append(entry.get("name") if isinstance(entry, dict) else None)
+
+    # Fast path: already dense + index-aligned (position == index) and nothing
+    # to drop → return the original untouched so the cache round-trip is exact.
+    aligned = max_index + 1 == len(tokens) and all(
+        isinstance(t, dict) and t.get("index") == i for i, t in enumerate(tokens)
+    )
+    if aligned and not dropped:
+        return spot_meta
+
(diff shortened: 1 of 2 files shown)


richkuo:
@claude

PR #840, opened 2026-06-01
feat(hl): add per-cycle OHLCV file cache to cut /info 429s

### LINKED ISSUE #839: Consolidate duplicate OHLCV fetches across strategies sharing the same asset
## Problem

Each strategy run independently fetches OHLCV candles from the Hyperliquid `/info` endpoint (`platforms/hyperliquid/adapter.py:401` `get_ohlcv` → `candles_snapshot`; no caching — only `meta` is cached to `/tmp/hl_meta.json`). With ~20 strategies per instance × 7 instances, the same asset+timeframe data is fetched many times per cycle — all from one server IP.

Note: a single strategy can hit `/info` for candles **more than once** per cycle — main signal (`check_hyperliquid.py:151`), HTF filter (`:221`), and ATR recompute (`:1064`) are separate fetches. So per-strategy request count is 1–3×, and "20 identical requests" undercounts the real volume.

This causes HTTP 429 rate limiting from Hyperliquid, which then cascades into script failures across all strategies for that instance. (`get_ohlcv` has no 429 short-circuit — only `lookup_fill_fee_by_oid` does — so a 429 raises uncaught and surfaces as a script-failure alert, #829.)

Example: 20 strategies all fetch `BTC 1h` candles independently within the same 5-minute cycle. That's 20+ identical requests when 1 would suffice.

## Proposal

Consolidate OHLCV fetches so that strategies sharing the same asset+timeframe reuse a single cached response within a cycle.

**Precedent:** #768 already established this pattern for prices — the Go scheduler fetches all mids once per cycle (`fetchHyperliquidMids`, `hyperliquid_marks.go:27`) and forwards `--mark-price` to each strategy subprocess so Python skips its own `/info allMids` call (`main.go:2475`). Option 1 below is the direct extension of that design to candles and should follow it.

Options:
1. **Fetch-once-per-cycle at the scheduler level** — the Go scheduler fetches all unique asset+timeframe combinations once, writes them to a shared temp location or passes them as args, and strategies read from cache instead of fetching themselves. (Mirrors the #768 mids pattern.)
2. **File-based cache with TTL** — `check_hyperliquid.py` checks a shared cache file (e.g. `/tmp/hl_ohlcv_BTC_1h.json`) with a short TTL (e.g. 60s). First strategy to fetch writes it, subsequent strategies read from cache.
3. **Strategy-level dedup within a single instance** — the scheduler groups strategies by asset+timeframe and only invokes the OHLCV fetch once per group.

### Scope caveat — per-instance vs IP-wide

The 429 is triggered by the **aggregate across all 7 instances** ("all from one server IP"), but all three options above only dedup **within a single instance**:

- The 7 instances are separate processes — none of them shares in-memory state or CLI args with the others.
- Both service units set `PrivateTmp=true` (`go-trader.service:28`, `systemd/go-trader@.service:33`), so a `/tmp` cache (option 2) is **not shared across instances** either.

So the realistic outcome is ~7 instances × (unique asset+tf combos per instance), **not** a flat ~10-15. That's still a large reduction (each instance drops from ~20–60 fetches to a handful) and likely enough to clear the 429s, but the "~140 → ~10-15" figure assumes a single IP-wide shared cache that none of options 1–3 provide. If true IP-wide dedup is required, add a 4th option: a shared external cache (e.g. a small local daemon or shared on-disk path mounted outside `PrivateTmp`) that all instances read from.

richkuo:
## Summary

- Each HL strategy subprocess independently fetched candles from `/info`; ~20 strategies sharing a handful of asset+timeframe combos fired identical requests dozens of times per cycle from one IP, tripping HL's 429 rate limit and cascading into script-failure alerts (#829)
- Add `_load_ohlcv_cache` / `_save_ohlcv_cache` / `_ohlcv_cache_path` to `adapter.py`, mirroring the existing `/tmp/hl_meta.json` meta-cache pattern (atomic temp+`os.replace`, log-and-swallow failures)
- `get_ohlcv` now caches transformed candles to `/tmp/hl_ohlcv_<sym>_<interval>_<limit>.json` (TTL=60s) — the first strategy of a cycle fetches, peers read from cache; covers signal (`:151`), HTF-filter (`:221`), and ATR-recompute (`:1064`) paths uniformly since all route through `get_ohlcv`
- Empty/insufficient results are never cached so peers keep retrying live; dedup is per-instance (`/tmp` shared within an instance, isolated across systemd instances by `PrivateTmp=true`); disable via `GO_TRADER_HL_OHLCV_CACHE=0`; no Go/CLI/probe changes

Closes #839

## Test plan

- [ ] 19 new tests in `platforms/hyperliquid/test_ohlcv_cache.py`: enable/disable toggle, path sanitization + uniqueness, load/save round-trip, TTL expiry, empty-not-cached, garbage rejection, atomic-no-leftover-tmp, `get_ohlcv` integration (cache hit skips `candles_snapshot`, disabled always fetches, distinct limits don't share, empty never cached)
- [ ] Two existing `get_ohlcv` tests in `test_adapter.py` updated with `GO_TRADER_HL_OHLCV_CACHE=0` for deterministic fetch-path coverage
- [ ] `uv run --no-sync python -m pytest platforms/hyperliquid/ shared_strategies/ shared_tools/ backtest/ -q` → 1087 + 59 + 56 passed, no regressions
- [ ] Deploy to a live instance and confirm 429 alerts stop after a few cycles

---
LLM: Claude Opus 4.8 (1M) | high


richkuo:
@claude

richkuo:
Addressed the review's two pre-deploy items in 172cbad:

1. **Interval-aware TTL** — `_ohlcv_cache_ttl(interval_ms) = min(OHLCV_CACHE_TTL_S, half-bar)`, so a 1m strategy caps at 30s and can't read a >½-bar-stale snapshot when `effectiveStrategyIntervalSeconds` starts a new cycle inside the window. 3m+ unchanged at 60s.
2. **`interval` sanitized** into the cache path the same as `symbol` (alnum + underscore), closing the slash/dotdot escape gap.

Added 4 tests (interval sanitization + TTL bounds); full suite 1091 passed.

On item 3 (burst capped at `pythonSemaphore`, not 1): agreed — the docs/PR now say "first batch fetches, peers read cache" rather than "first strategy". Left as-is per the review; an `flock` to collapse the first-batch herd to a single fetch is the natural follow-up if 429s persist. Item 4 (lingering `/tmp` files) is bounded and left as noted.

richkuo:
@claude rereview

Issue #841, opened 2026-06-01
Standardize close-strategy JSON configs (canonical keys + unified per-regime block)


richkuo:
## Problem

Close-strategy configs have accumulated significant variance — in **key names**, in the **tier-list key**, in **structure** (regime variants), and in **redundant evaluators**. The regime dimension is smeared across three unrelated places, which makes regime-aware configs hard to author, forces a parser strip-hack, and blocks the atomic per-regime read the dynamic close (#843) needs.

### Variance 1 — trigger/fraction key names

| Surface | Trigger key | Fraction key |
|---|---|---|
| `tiered_tp_atr` / `_live` (py `tiered_tp_atr.py:17`, go `hyperliquid_protection.go:222`) | `atr_multiple` ‖ `multiple` | `close_fraction` ‖ `fraction` |
| `tiered_tp_pct` (`tiered_tp_pct.py:17`) | `profit_pct` ‖ `pct` | `close_fraction` ‖ `fraction` |
| `tp_at_pct` | `pct` (scalar, no tiers) | — (implicit 1.0) |
| **regime blocks** (`regime_atr.go:436,455`) | **`atr`** ← odd one out, no alias | `close_fraction` |

Headline: inside a `trend_regime` block the trigger is `atr`, but everywhere else it's `atr_multiple` — same concept, different name by nesting depth. Plus legacy `multiple` / `fraction` / `pct`-on-a-tier aliases.

### Variance 2 — tier-list key (`tiers` vs `tp_tiers`)

Close-strategy params use `tiers` (`post_tp_sl.go:703`, `regime_atr.go:797`, `hyperliquid_protection.go:104`, `registry.py:96-108`, …), but the manual strategy-level field is already `tp_tiers` (`config.go:142`, `StrategyConfig.TPTiers`). Two names for the same concept depending on where it lives.

### Variance 3 — regime is keyed in THREE places (the structural problem)

For `tiered_tp_atr_regime` / `tiered_tp_atr_live_regime`, the regime dimension is smeared across:
1. the tier's `atr` (regime keyed *inside* each tier),
2. the strategy-level `stop_loss_atr_regime` (SL keyed separately),
3. `sl_after`'s **own** `trend_regime` sub-blocks (`tp_atr_fraction: {trend_regime: {...}}`, backed by `TPATRFractionRegime`/`ATRRegime`/`TrailATRRegime`).

[...]

Issue #844, opened 2026-06-01
feat(close): pure trailing-stop strategy where regime-keyed TP tiers ratchet the trailing ATR multiple instead of closing


richkuo:
## Summary

Add a close strategy built on a **trailing ATR stop**, where each **TP tier both (a) tightens the trailing-stop ATR multiple and (b) optionally closes a fraction of the position at that price**. Each tier carries its own `close_fraction`: a positive value scales out at the tier's price, and **`close_fraction == 0` means no close — that tier only ratchets the trail tighter**. So a tier table can mix "pure trail-tighten" rungs with "scale-out + tighten" rungs. All per-tier values are **regime-keyed**: the tier trigger (`atr_multiple`), the close fraction, and the resulting trailing multiple all differ per regime.

Conceptually:

```
open → trailing stop @ 3.0×ATR
  price reaches TP1 (regime-keyed) → close 0% (no close), tighten trail to 2.0×ATR
  price reaches TP2                → close 30%, tighten trail to 1.0×ATR
  price reverses → trailing stop fires on the remainder → exit
```

A tier with `close_fraction: 0` is a pure trail adjuster; a tier with `close_fraction > 0` scales out *and* tightens. This makes the "pure trailing" strategy just the special case where every tier sets `close_fraction: 0`.

## Motivation

Today, the only way to "tighten the stop after a TP" is to attach an `sl_after` rule (`breakeven` / `atr_offset` / `trail_from_here`) to a tier of a `tiered_tp_atr*` close — but the trail-tighten is a *side effect* of a tier, the tier's `close_fraction` is its primary action, and the on-chain TP ladder coerces the final tier to `close_fraction=1.0` (`scheduler/hyperliquid_protection.go:177-187`; default regime blocks ship the final tier at `1.0` in `scheduler/regime_atr.go:240-242`, documented at `regime_atr.go:219`). There is no single knob per tier that says "tighten the trail to X, and close Y% (possibly 0%) here." This issue requests a close evaluator where **both the close fraction and the new trailing multiple are first-class per-tier outputs**, so `close_fraction: 0` tiers become legal trail-only rungs.

## Relationship to existing work (#843 + the `sl_after` handoff)

> ⚠️ This issue is **assembly of existing runtime**, but two recent landings overlap parts of the goal — read before implementing, and **reuse, don't rebuild**:
>
[...]

PR #846, opened 2026-06-01
feat(close): canonical keys + unified per-regime close block (#841)

richkuo:
Standardizes close-strategy JSON config keys and introduces the **unified per-regime close block**. Part of #841; the on-disk migration and `tp_at_pct` deprecation are split into follow-up #848.

Each commit builds and passes the full Go + Python suites. No production risk to existing configs — legacy keys are still read (with a one-shot `[DEPRECATED]` warning); nothing on disk changes in this PR.

## What's in this PR

**Canonical keys (back-compat reads):**
- Tier-list key `tp_tiers` (legacy `tiers` still read).
- Tier trigger `atr_multiple` everywhere incl. regime blocks (legacy `atr`/`multiple`, `fraction`→`close_fraction` aliases read; both-set rejected).

**Unified per-regime close block** (`tiered_tp_atr_regime` / `tiered_tp_atr_live_regime`) — top-level `trend_regime`, each label owning `{stop_loss_atr?, tp_tiers:[{atr_multiple, close_fraction, sl_after?}]}`:
- **Select-then-scalar** resolution — pick the active regime's block, feed the existing scalar machinery. Tier counts may differ per regime; `sl_after` stays scalar (no regime sub-block).
- Wired through the live HL path: on-chain TP resolver, post-TP `sl_after`, and SL arming (per-regime `stop_loss_atr` → the block **owns the SL**; `EffectiveStopLossPct` defers; strategy-level stop fields rejected as a second owner).
- LoadConfig validates + accepts the shape; Python evaluators resolve it for paper/backtest; hot-reload rejects mid-position block changes.

Works end-to-end for HL live perps. Example + key reference in `SKILL.md` / `CLAUDE.md`.

## Commits
`tp_tiers` key · `atr_multiple` key · unified helpers · on-chain TP resolver · sl_after resolver · SL folding · LoadConfig gate · Python evaluators · sole-owner validation · hot-reload gating · docs.

## Deferred to #848 (separable cleanups)
- Deprecate `tp_at_pct` → single-tier `tiered_tp_pct`
- v15 on-disk migration (rewrite legacy configs to canonical keys + unified shape; then flip defaults + drop alias reads)
- Port legacy regime tests to the unified shape

## Design notes
- Back-compat: legacy `tiers`/`atr`/`fraction`/`multiple` keys read for one window; #848's migration rewrites on disk.
- Registry `default_params` deliberately stay on legacy `tiers` until #848 (they merge under operator params per-key, so a canonical default would shadow an operator's legacy key — caught by the backtester firing-tier tests).

Closes nothing on its own; #841 closes when #848 lands.

---


richkuo:
@claude

richkuo:
@claude

richkuo:
Both blocking issues fixed in 9259ba0 — verified both were real (reachable via a hand-written unified config) before fixing:

1. **Zero-SL regime** — `stop_loss_atr` is now **required on every regime label** in `validateUnifiedRegimeClose` (the block owns the SL, so it must provide one). Closes the gap where a regime omitting it ran with no stop loss.
2. **Single-tier no-TP** — `validateUnifiedTierList` now requires **≥2 tiers per regime**, matching the on-chain resolver and the legacy validator. Single TP-then-trail is #844's archetype, not this one.

Fixed the `unifiedBlock()` fixture (every regime ≥2 tiers + `stop_loss_atr`; `trending_down` now 3 tiers to keep the variable-count coverage) and added error-case tests for both rules. Full Go + Python suites green.

Optional notes: the unified block is **canonical-keys-only by design** (no legacy `atr`/`tiers`/`fraction` aliases inside the new shape) — confirmed intentional. With both issues fixed, the deferred-to-#848 list is unaffected.

richkuo:
@claude sonnet is this good to merge?

Issue #848, opened 2026-06-01
Close-config #841 follow-up: deprecate tp_at_pct + v15 on-disk migration


richkuo:
Follow-up to #841 / PR #846, which standardized the canonical close keys (`tp_tiers` / `atr_multiple` / `close_fraction`) and shipped the unified per-regime close block. #846 left the on-disk migration and the `tp_at_pct` deprecation as separable cleanups (back-compat reads mean old configs keep working in the meantime). This issue covers those.

### A. Deprecate `tp_at_pct` → single-tier `tiered_tp_pct`
`tp_at_pct` (`{pct: X}` → full close at a pct threshold) is exactly a single-tier `tiered_tp_pct` with `close_fraction: 1.0`. Remove the evaluator and migrate:
- `{"name":"tp_at_pct","params":{"pct":X}}` → `{"name":"tiered_tp_pct","params":{"tp_tiers":[{"profit_pct":X,"close_fraction":1.0}]}}`
- one-window read shim that accepts `tp_at_pct` and rewrites it in-memory with a `[DEPRECATED]` warn
- remove `tp_at_pct.py` + its registry entry; `--list-json` snapshot before/after (it leaves the registry list)

### B. v15 on-disk migration
`config_migration.go` v14 → v15 (`migrateV15CloseKeys`): rewrite `config.json` on disk so stored configs become canonical —
- rename `tiers`→`tp_tiers`, `atr`→`atr_multiple` (incl. inside regime blocks), `multiple`→`atr_multiple`, `fraction`→`close_fraction`, `pct`-on-tier→`profit_pct`
- lift legacy tier-keyed `tiered_tp_atr_regime`/`_live_regime` → the unified top-level `trend_regime` shape (fold `stop_loss_atr_regime` → per-label `stop_loss_atr`; inline per-regime `sl_after` scalars)
- fold `tp_at_pct` → single-tier `tiered_tp_pct` (from A)
- after migration, flip registry `default_params` to `tp_tiers` and drop the alias **reads** (the deprecation window closes)

### C. (optional) migrate existing legacy regime tests to the unified shape
The `tiered_tp_atr_regime` / `_live_regime` tests still exercise the legacy tier-keyed shape via back-compat. Once B lands, port them to the unified shape so the legacy path can eventually be removed.

## Constraints
- Live-trading bot: the migration must read the old shape and rewrite atomically; keep `--list-json` byte-identical where required.
[...]

Issue #843, opened 2026-06-01
Dynamic regime-aware close: live SL/TP/sl_after re-resolution (tiered_tp_atr_live_regime_dynamic)


richkuo:
> **Depends on #841** (canonical keys + `tp_tiers` + unified per-regime block) and **#842** (single close strategy). The schema below is the unified shape defined by #841; this issue adds live re-resolution on top. If built greenfield before #841 lands, it must adopt #841's canonical keys exactly so it doesn't introduce a second shape.

## Goal

A new close strategy `tiered_tp_atr_live_regime_dynamic` whose SL, TP ladder, and `sl_after` **all re-resolve live** when the (7-state composite) regime changes mid-position, then sync to the exchange. Today every regime resolver reads `pos.Regime`, which is **stamped once at open and frozen** (`scheduler/regime_multi_window.go:445-456`, `stampPositionRegimeFromPayload:515-539`) — so a position opened in `trending_up` keeps `trending_up` TP/SL/sl_after even after the market shifts.

## Behavior (HL live perps)

1. Each cycle read the **current** ATR-window regime from `stratState.RegimeWindows` (new `strategyCurrentATRRegime` helper; analog of `strategyCurrentDirectionalRegime`).
2. **Confirm-cycle hysteresis:** a new label must persist `regime_confirm_cycles` cycles (default **2**, configurable) before it's applied — filters classifier flicker against the HL 1000-trigger cap.
3. On apply: re-resolve `stop_loss_atr` + **unfilled** TP tiers + future `sl_after` against `pos.EntryATR` (frozen) × the new regime's multiples, then **cancel+replace on-chain**. SL reuses the existing 0.5% min-move guard (`hyperliquid_trailing_stop.go`); TP gets the same guard (already-filled tiers stay filled).
4. Paper reuses the Python evaluator (reads current regime per-tick naturally); **backtester rejects** (no parity — consistent with `regime_directional_policy`).

> Note on existing closes: an already-placed on-chain TP tier is **not** repriced by the current sync — it only places unfilled tiers. So the TP side needs new explicit cancel+replace; SL/sl_after already cancel+replace.

## Schema (unified per-regime block — canonical shape from #841)

[...]

PR #851, opened 2026-06-01
Single closing strategy per trade: collapse close_strategies array (#842)

### LINKED ISSUE #842: Single closing strategy per trade (collapse close_strategies array)
> **Depends on #841** (canonical keys + `tp_tiers` + unified per-regime block). **Blocks #843** (sole-owner enforcement needs the single-close model).

## Problem

A trading strategy's exit is configured as an **ordered array** `close_strategies: []StrategyRef` (`scheduler/config.go:309`), and every entry evaluates each cycle with **max `close_fraction` wins** (`shared_tools/strategy_composition.py:129-137`, loop at `:306-332`). This array-of-closes model has real weaknesses:

1. **Partial exits can't compose.** A returns 0.5 (TP hit), B returns 0.3 (momentum fading) → max picks 0.5, B's intent vanishes. Two partial trimmers silently collapse to the larger.
2. **Cycle-ordering subtlety.** `close_fraction` is "fraction of *current* position" with no cumulative-intent tracking, so "50% of original" vs "50% of remaining" diverges by what fired when. Tiered TP only survives via its own armed-tier state.
3. **Lossy attribution.** Only `best_strategy` is credited; co-firing closes disappear from PnL/leaderboard.
4. **No negotiation.** Closes can only vote "close more"; a profit-taker always beats a let-winners-run rule via max.
5. **The live-perps carve-out already breaks the invariant.** On HL live, `tiered_tp_atr*` is stripped from the max loop and placed on-chain (`filterCloseStrategiesForHLOnChainProtection`, `scheduler/hyperliquid_protection.go:551`), so "max wins" doesn't hold uniformly where real money is.

## Proposal

Collapse `close_strategies` (array) → a **single `close_strategy` ref** per trading strategy. One profit-taking close owns the exit ladder.

Risk-backstops that currently leak into the close array (e.g. `tp_at_pct` used as a hard cap) move to an **explicit separate layer**, consistent with how SL and trailing already live at strategy level (outside the close list). The mental model becomes: *one profit-taking close + risk backstops in a distinct layer* — not N peers fighting through max.

### Why not keep the array

The array's value was "orthogonal composition" (TP ladder + hard cap + time stop). But in practice those backstops belong in the risk layer, not the profit layer; and competing *partial* closes are exactly the lossy case above. Single-owner removes the ambiguity and is what enables embedding SL in the close block cleanly (see #843).

## Scope

- `config_migration.go`: `close_strategies: [x]` → `close_strategy: x`; reject (or migrate to risk layer) multi-entry arrays with operator guidance. Sequenced **after** #841's v15 key/shape migration (one migration pass touches each strategy).
- Decide the risk-backstop layer shape (could be as simple as a strategy-level `hard_tp_pct` / time-exit, or a small dedicated list explicitly named as backstops).
- Update `strategy_composition.py` (drop/simplify `max_close_fraction`), Go `CloseStrategies` handling, backtester `close_strategies=[...]`.
- Hot-reload gating + tests.

## Relationship

- **#841** defines the canonical keys / `tp_tiers` / unified block; migrate the close shape once, then this collapses the array.
- **#843** depends on this — sole-owner validation for the dynamic close is only clean once a strategy has exactly one close.

---

richkuo:
Closes #842.

## What

A strategy's exit was an ordered **array** of close evaluators that all ran each cycle with "max `close_fraction` wins" — a lossy model: co-firing partial closes collapsed to the larger, attribution dropped, and the HL-live on-chain carve-out already broke "max wins" where real money trades. This collapses the array to a **single `close_strategy` ref** per strategy: one profit-taking close owns the exit ladder; risk backstops belong at the strategy level. This is the clean single-owner model #843 (dynamic close) and sole-owner SL validation build on.

**Scope note:** #842 changes the **runtime model only**. No on-disk migration or `config_version` bump — that's deferred to #848 (per the issue's sequencing). Legacy `close_strategies` arrays are read for back-compat.

## Behavior

- `StrategyConfig.CloseStrategies []StrategyRef` → `CloseStrategy *StrategyRef`. `UnmarshalJSON` still reads the legacy `close_strategies` array: a length-1 array is lifted to `close_strategy`; a **length>1 array is rejected at load with the strategy id** (operator collapses to one close + moves backstops to strategy level).
- `close_strategy` is the canonical config / unknown-key / explicit-provenance key; `close_strategies` stays accepted on read and normalizes to `close_strategy` for inspect.
- HL on-chain suppression reworked: the array filter became `closeStrategySuppressedByOnChainProtection(sc) bool`; `strategyConfigWithOnChainProtectionFilter` nils the single close when it overlaps on-chain TPs. **Paper never suppressed.**
- The Go↔Python `--strategy-refs` wire keeps a length-≤1 `"closes"` list, so the Python composition layer is unchanged (max over one element is the element). `run_backtest.load_strategy_config` and `simulate_strategy` read `close_strategy` with legacy-array fallback.

## Tests

- Struct literals collapsed to `&StrategyRef`; multi-close cases became the #842 rejection test (JSON-driven, since len>1 is unrepresentable as a struct literal).
- HL filter test rewritten for single-close drop/keep semantics; inverted legacy-key hint test repurposed to assert both spellings are accepted.
- New Python tests: `load_strategy_config` reads the single key and prefers it over a legacy array.
- Full Go build + vet + test suite green; full Python suite green (1097).

## Smoke (manual)

Verified end-to-end via `inspect`: canonical `close_strategy` shows the single close + TP source; legacy len-1 `close_strategies` reads as the single close; legacy 2-element array is rejected at load with the #842 guidance.

---


richkuo:
@claude

richkuo:
@claude rereview

PR #852, opened 2026-06-01
Add state-DB singleton lock so a duplicate go-trader refuses to start

### LINKED ISSUE #849: Add startup singleton guard so a second go-trader can't silently run against the same state DB
## Problem

There is no guard against running **two `go-trader` instances against the same state DB / exchange account at once**. If a second process starts — e.g. an operator manually launching the binary alongside the systemd-managed instance — both run silently, book trades to the same SQLite DB, and manage the same on-chain positions, leading to double orders and desynced state.

Two reasons a duplicate goes unnoticed today:
- **Status port collision is non-fatal.** `bindWithFallback` (`scheduler/server.go:173`) sweeps to the next port (`DefaultStatusPort=8099` +N attempts) on `EADDRINUSE` instead of refusing, so a second daemon just lands on a fallback port and keeps trading.
- **No state-DB singleton lock.** The DB is opened via `modernc.org/sqlite`, which permits concurrent opens (`scheduler/db.go`), so a second process attaches to the same `state.db` without complaint.

> Note: this is **not** a systemd defect. Under `Type=simple` + `KillMode=control-group` (no `PIDFile=`), systemd correctly tracks and cgroup-kills its managed worker; `PPID=1` is normal. The failure mode is specifically a process systemd never spawned (manual launch) running in parallel — which no restart can reap because it's outside the unit's cgroup.

## Proposal

Add a startup singleton guard so a second instance bound to the same state/account **refuses to start** (or loudly alerts) instead of silently double-trading.

Preferred: acquire an **exclusive advisory lock** keyed on the resolved state-DB path at boot (e.g. `flock` on `<DBFile>.lock` holding the lock for process lifetime). If the lock is already held by a live process, log a CRITICAL, DM the owner, and `os.Exit` with a dedicated non-zero code rather than proceeding.

- Lock the **state DB**, not the status port — the DB is the true shared resource (the port fallback means port-collision is neither sufficient nor reliable as the signal).
- Keep the guard skippable for legitimately-isolated parallelism (sub-accounts / separate DB files already get distinct paths, so per-path locking is naturally safe). Add an escape hatch env var only if a real need surfaces.
- Consider also dropping/relaxing `bindWithFallback`'s silent sweep on the **daemon** path (a daemon that can't bind its configured port is arguably misconfigured), or at least surface the fallback as a louder warning — but the DB lock is the actual fix; the port behavior is secondary.
- **Expose `pid` in `/health`** (currently `{status, version}` only, `scheduler/server.go:240`) so external monitoring can detect a duplicate from the outside — alert when `health.pid` ≠ systemd `MainPID` (or the two ports 8099/8100 both report healthy with different pids). This is a cheap detection aid that complements the lock; it doesn't replace it.

### Relationship to the out-of-cgroup duplicate scenarios

The duplicates seen during signal-mode updates / rollback (a `run.sh`-spawned process landing outside systemd's cgroup while the managed instance is still alive) are the **same shared-state failure this lock prevents**: the live instance holds the DB lock, so the out-of-cgroup process fails to acquire it and exits before booking any trade. The lock is the unifying fix for the *trading-safety* half of those scenarios.

The separate *process-hygiene* half — `update.sh --restart-mode signal` spawning out-of-cgroup processes against a live systemd unit, and rollback not guaranteeing the prior PID is dead — is an `update.sh` concern (e.g. refuse signal mode when `systemctl is-active <unit>`), tracked separately in #850.

### Update/restart safety (critical implementation constraints)

The lock must not interfere with the normal update path. `update.sh` uses `systemctl restart` (sequential: old daemon fully stops/drains → exits → lock released → new daemon starts), so there is **no steady-state overlap window**. But two implementation choices will silently break deploys if gotten wrong:

- **Use a kernel-held `flock`/`fcntl` lock on an open fd — never a pidfile-content check.** The OS auto-releases an fd lock when the process dies (including SIGKILL/crash), so a hard-killed daemon leaves no stale lock. A "write PID to a file, refuse if present" scheme would survive a crash and **permanently block restart**, causing `update.sh` to roll back every deploy. This is the single most important constraint.
- **Scope the guard to the long-running daemon path only.** `update.sh` runs the startup probe (`--probe-only`) **while the old daemon is still alive** (pre-swap), and operators run one-off CLI (`manual-*`, `backfill`, `inspect`, `export`, `version`) against a live daemon. None of these may acquire (or be blocked by) the lock, or the probe fails and aborts the update.
- Ensure `TimeoutStopSec` (currently 20s) stays `≥ shutdownDrainCap` (15s) so the old daemon drains and releases the lock cleanly before the new one starts.
- `--once` runs a trading cycle; decide explicitly whether to guard it. Simplest: leave `--once` unguarded (operator's responsibility — it's also `update.sh`'s post-deploy smoke) and lock only the persistent daemon loop.

### Acceptance
- Starting a second daemon against an in-use `state.db` exits non-zero with a clear "another go-trader is already running (pid N)" message and an owner DM; it does not book any trades.
- A normal single instance, and genuinely separate instances (distinct `DBFile`), start unaffected.
- A `systemctl restart` / `update.sh --restart` cycle succeeds with no false "already running" failure, and a SIGKILLed daemon leaves no stale lock (next start succeeds).
- `--probe-only` and one-off CLI subcommands run successfully while the daemon is live.
- `/health` includes `pid`, and an out-of-cgroup duplicate launched against a live instance fails to acquire the lock and exits without trading.

---

richkuo:
Closes #849.

Acquires an exclusive kernel `flock` on `<DBFile>.lock` for the daemon process lifetime, so a second instance against the same state DB/account refuses to start (exit 79, CRITICAL log + owner DM) instead of silently double-trading. Also exposes `pid` in /health for external duplicate detection, upgrades the port-fallback warning, and registers exit 79 in both service files.

Guard is scoped to the persistent daemon loop only; `--once`, `--probe-only`, and one-off CLI are unaffected. The fd flock auto-releases on crash/SIGKILL, so no stale lock blocks a restart.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 trading_bot.db
 *.db-wal
 *.db-shm
+*.db.lock
 .env
 trading-scheduler
 # Compiled binaries
--- a/go-trader.service
+++ b/go-trader.service
@@ -13,7 +13,10 @@ ExecStart=/opt/go-trader/go-trader --config /opt/go-trader/scheduler/config.json
 Restart=always
 RestartSec=10
 # Probe/deploy failures exit 78 — do not restart until scripts are synced.
-RestartPreventExitStatus=78
+# Singleton-lock refusals exit 79 (#849): another live go-trader already holds
+# the state-DB lock, so staying down (one DM, no restart-loop) is correct — the
+# kernel flock auto-releases on exit, so a legitimate restart is unaffected.
+RestartPreventExitStatus=78 79
 StandardOutput=journal
 StandardError=journal
 
--- a/scheduler/exit_codes.go
+++ b/scheduler/exit_codes.go
@@ -5,3 +5,17 @@ package main
 // RestartPreventExitStatus=78 so the service stays down instead of restarting
 // every RestartSec and spamming Discord. (EX_CONFIG from sysexits.h.)
 const ExitProbeFailure = 78
+
+// ExitSingletonLock is used when the daemon refuses to start because another
+// live go-trader already holds the exclusive lock on the resolved state-DB
+// path (#849). A duplicate against the same state.db/exchange account would
+// double-trade and desync state, so the second process exits here instead.
+//
+// Both service files set RestartPreventExitStatus=79 alongside 78: under a
+// normal `systemctl restart` the managed daemon is the lock holder, so it can
+// only hit this code when an out-of-cgroup duplicate is alive — in that case
+// systemd should stay down (one DM, no restart-loop) rather than respawn every
+// RestartSec into the same refusal. The kernel auto-releases the flock when the
+// holder dies, so this never permanently blocks a legitimate restart. 79 sits
+// just past the sysexits.h range (64–78) to avoid colliding with those codes.
+const ExitSingletonLock = 79
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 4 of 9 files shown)


richkuo:
@claude

richkuo:
@claude fix optional review items

richkuo:
@claude sonnet rereview

PR #855, opened 2026-06-01
fix: prevent signal-mode out-of-cgroup duplicate processes (#850)

### LINKED ISSUE #850: update.sh: signal-mode restart/rollback spawns out-of-cgroup duplicate processes
## Problem

`update.sh` signal-mode restart and rollback can spawn `go-trader` processes **outside systemd's control group**, producing duplicate live processes that systemd can neither see nor reap. This is the process-hygiene half of #849 (the trading-safety half is covered there by the state-DB singleton lock).

### 1. Signal mode against a live systemd unit creates an orphan duplicate

`update.sh` defaults to `restart_mode=systemd` (`scripts/update.sh:44`) and only takes the signal/`run.sh` path when explicitly passed `--restart-mode signal`, or as a fallback when the systemd unit is **missing** (`scripts/update.sh:257-271`). But nothing stops an operator from running `--restart-mode signal` while a systemd unit is actively managing the instance. In that case `signal_launch_wrapper` does `setsid nohup bash run.sh &` (`scripts/update.sh:242-252`), spawning `go-trader` outside the unit's cgroup. systemd still tracks its own (now-killed) `MainPID`, and with `Restart=always` will spawn its own replacement — leaving two processes trading the same state DB.

### 2. Rollback doesn't guarantee the prior PID is dead

On the signal rollback path (`scripts/update.sh:455,476` → `signal_kill_pidfile_process_then_respawn`), rollback restores the old binary and relaunches via `run.sh`, but a failed new process may still be alive on a `bindWithFallback` fallback port (`scheduler/server.go:173`) while the relaunch adds yet another process. Repeated rollback cycles accumulate 3-4 PIDs per instance.

## Proposal

- **Refuse (or auto-redirect) signal mode when a systemd unit manages the instance.** Before taking the signal path, check `systemctl is-active --quiet <unit>`; if active, either hard-fail with a clear message or transparently use `systemctl restart` instead. This prevents the out-of-cgroup duplicate at the source.
- **Make rollback verify the prior process is actually dead** before respawning — reuse the existing `signal_wait_pid_exit` SIGKILL escalation (`scripts/update.sh:226-238`) and additionally sweep for any stray `go-trader` bound to the same state DB / fallback ports before launching.

## Acceptance

- Running `update.sh --restart-mode signal` against a host with an active systemd unit for that instance does not produce a second process (either refused with a clear error or routed through `systemctl restart`).
- A rollback cycle ends with exactly one live `go-trader` per instance; no fallback-port survivors.

## Related

- #849 — the state-DB singleton lock there neutralizes the *trading* risk of these duplicates (the out-of-cgroup process fails to acquire the lock and exits before booking a trade). This issue covers the *process-hygiene* side so duplicates aren't created in the first place.

---

richkuo:
Closes #850

- Redirect explicit `--restart-mode signal` through `systemctl` when an active systemd unit manages this deployment's binary (ExecStart match spares sibling worktrees).
- Sweep stray `go-trader` procs sharing this instance's state DB (by cwd) before signal-mode rollback respawn, so fallback-port survivors can't accumulate.
- New pure helpers + 12 regression assertions in `test_update_helpers.sh`.

---

### DIFF
--- a/scripts/test_update_helpers.sh
+++ b/scripts/test_update_helpers.sh
@@ -41,4 +41,32 @@ if [[ "$warn_out" == *'optional.env'* || "$warn_out" == *'ignore_errors'* ]]; th
     exit 1
 fi
 
+# --- #850: signal-mode redirect decision -------------------------------------
+assert_eq "$(update_signal_redirect_decision active /opt/go-trader/go-trader /opt/go-trader/go-trader)" \
+    "redirect" "active unit running this binary -> redirect"
+assert_eq "$(update_signal_redirect_decision active /opt/other/go-trader /opt/go-trader/go-trader)" \
+    "" "active unit running a different binary -> no redirect (sibling worktree)"
+assert_eq "$(update_signal_redirect_decision inactive /opt/go-trader/go-trader /opt/go-trader/go-trader)" \
+    "" "inactive unit -> no redirect"
+assert_eq "$(update_signal_redirect_decision failed /opt/go-trader/go-trader /opt/go-trader/go-trader)" \
+    "" "failed unit -> no redirect"
+assert_eq "$(update_signal_redirect_decision active '' /opt/go-trader/go-trader)" \
+    "" "unreadable ExecStart -> no redirect"
+assert_eq "$(update_signal_redirect_decision active go-trader /opt/go-trader/go-trader)" \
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

Issue #870, opened 2026-06-03
feat(close): retune default close ladders + per-regime group defaults (depends on #866)


richkuo:
## Summary

Builds on the default-resolution mechanism from #866. That issue ships the three-layer resolver (`system_close_defaults` → `user_close_defaults` → `strategy_close_defaults`), the `user_close_defaults` config plumbing, a conservative working default for every close evaluator (no behavior change for existing strategies), and the backtest `--defaults` selector. **This** issue retunes the actual default **values** and adds per-regime differentiation.

**Depends on #866** — needs its three-layer resolver, per-group expansion machinery, and per-regime opening-trail support to exist first. Land #866, then this.

Scope here is values + the per-regime group model, split out from #866 because it changes live behavior (see the on-chain warning) and deserves independent review + release notes.

## ⚠️ On-chain behavior change (B1)

B1 retunes `defaultHLProtectionTiers()` (`scheduler/hyperliquid_protection.go:163-174`) — the single source of truth that also seeds **HL on-chain TP placement** and `post_tp_sl` `tp_atr_fraction` firing-tier multiples. Changing it moves the actual reduce-only TP orders for **every existing `tiered_tp_atr*` strategy running on defaults**. Must be called out in release notes and validated (backtest + a live smoke) before merge.

## Group model (shared by B2 and C2)

The regime evaluators resolve a *group* ladder per classifier label — three groups:

| Classifier | Label(s) | Group |
|---|---|---|
| ADX | `trending_up`, `trending_down` | choppy |
| ADX | `ranging` | ranging |
| Composite | `trending_up_clean`, `trending_down_clean` | clean |
| Composite | `trending_up_choppy`, `trending_down_choppy` | choppy |
| Composite | `ranging_quiet`, `ranging_volatile`, `ranging_directional` | ranging |

(ADX exposes no clean/choppy quality signal, so its trends map to choppy; the clean group is composite-only.)

## A. Percent take-profit — `tiered_tp_pct` — price-based

Retune `DEFAULT_TIERS` to a 4-tier even-quarter ladder by price move from entry (pure price move, no leverage term — evaluator math unchanged):

```json
"tiered_tp_pct": { "tp_tiers": [
  { "profit_pct": 0.01, "close_fraction": 0.25 },
  { "profit_pct": 0.02, "close_fraction": 0.50 },
  { "profit_pct": 0.03, "close_fraction": 0.75 },
  { "profit_pct": 0.04, "close_fraction": 1.00 }
] }
[...]

PR #871, opened 2026-06-03
feat(close): in-app default resolution mechanism + working ratchet default (#866)

### LINKED ISSUE #866: feat(close): in-app default resolution mechanism + working ratchet default (use_defaults)
## Summary

The `trailing_tp_ratchet` and `trailing_tp_ratchet_regime` close evaluators (#844) currently **require** an explicit `tp_tiers` list/map at config-load. There is no working in-app default: opening a position without supplying the tier table fails validation rather than falling back to a sensible built-in. This issue wires a **layered default** so that:

1. A position opened **without** explicitly providing `tp_tiers` (or with `use_defaults: true`) resolves to a default tier ladder.
2. There is a **hardcoded system default** as the ultimate fallback (single source of truth in-app).
3. An operator can override that default **for all strategies at once** via a top-level `close_defaults` block in `config.json`, without editing source.
4. A per-strategy `tp_tiers` still wins over everything.

## Current behavior (verified against code)

- **Scalar `trailing_tp_ratchet`** advertises a Python-side `default_params` of `{"tp_tiers": [{"atr_multiple": 1.5, "trailing_mult_after": 2.0, "close_fraction": 0.0}]}` (`shared_strategies/close/registry.py:156-159`), but the **Go loader never expands it**:
  - `closeTierListParam` reads **only** `tp_tiers` (`scheduler/post_tp_sl.go:27-35`).
  - `validateTrailingTPRatchetClose` errors `missing tp_tiers` when absent (`scheduler/trailing_tp_ratchet.go:210`).
  - `use_defaults` is accepted as a known key (`scheduler/trailing_tp_ratchet.go:215`) but has **no Go expansion** into a tier list.
- **Regime `trailing_tp_ratchet_regime`** has **empty** `default_params` (`shared_strategies/close/registry.py:161-165`) — no default at all, and requires an exhaustive regime-keyed map.

Net effect: the advertised scalar default is reachable only on the Python paper/backtest evaluator path; on the Go side both names hard-require explicit tiers.

## Reference pattern (already correct for `tiered_tp_atr*`)

`tiered_tp_atr*` has a real in-app hardcoded default with a single source of truth:

- `defaultHLProtectionTiers()` returns `[{1x, 0.5}, {2x, 1.0}]` (`scheduler/hyperliquid_protection.go:163-174`) — "canonical fallback tier ladder used when a tiered_tp_atr* close ref omits explicit tiers".
- `post_tp_sl.go` derives `tp_atr_fraction` firing-tier multiples from it, and `post_tp_sl.py` mirrors it as `_DEFAULT_SCALAR_TP_TIERS`.
- `use_defaults: true` expands into the concrete ladder at load.

The ratchet should follow the same hardcoded-default shape, **plus** an operator-override layer (see below).

## Decided design — three-layer precedence

Resolution order for a close ref that omits `tp_tiers` (or sets `use_defaults: true`), lowest → highest:

1. **Hardcoded system default (Go constant + Python mirror).** Ultimate fallback, always present, cannot be removed.
2. **Operator default — top-level `close_defaults` block in `config.json`.** When present, overrides the hardcoded default for **every** strategy that uses defaults. Lives inside the existing config file, so it rides the existing load / validate / migrate / SIGHUP-hot-reload / unknown-key-guard / probe machinery — **no separate config file**.
3. **Per-strategy `tp_tiers`.** Inline in a strategy's `close_strategy`, highest precedence, overrides both. Already exists today.

A strategy with no `tp_tiers` → use `close_defaults[<name>]` if set, else the hardcoded default.

### Scope

- **All close evaluators that have defaults**, not just the ratchet. `close_defaults` can override `tiered_tp_atr*` defaults too (currently `defaultHLProtectionTiers()`), giving one consistent override mechanism. `trailing_tp_ratchet` / `trailing_tp_ratchet_regime` are the first consumers.

### Default ladder values (hardcoded)

- **Two-rung tighten + scale-out** (ratchets the trail *and* takes profit), e.g.:
  ```
  [
    {atr_multiple: 1.5, trailing_mult_after: 2.0, close_fraction: 0.0},
    {atr_multiple: 3.0, trailing_mult_after: 1.0, close_fraction: 1.0}
  ]
  ```
  (exact values TBD during implementation; must satisfy the invariants below.)
- Note the first rung's `trailing_mult_after: 2.0` implies the strategy-level `trailing_stop_atr_mult >= 2.0` (initial-trail-not-looser validation). Document this coupling for `use_defaults` users, or clamp the first rung to the configured initial trail.

### Regime variant

- `close_defaults` **supports per-regime ladders** for `trailing_tp_ratchet_regime` (a `{label: [tiers]}` map), matching the per-regime override intent. Operators can specify a looser trail in ranging vs trending, etc.
- The **hardcoded** fallback broadcasts its single ladder to **every** classifier label (adx 3-state / composite 7-state) so the exhaustiveness requirement is satisfied automatically when no override is present.

### Validation still applies after resolution

The resolved ladder (from any layer) must satisfy: monotonic tighten (smaller-mult-only), cumulative `close_fraction`, and first-rung-not-looser-than `trailing_stop_atr_mult`.

## Touch points

- `scheduler/trailing_tp_ratchet.go` — `defaultTrailingRatchetTiers()` (hardcoded), expansion in tier resolution + validation (`trailingRatchetTiersForRegime`, `validateTrailingTPRatchetClose`); broadcast hardcoded ladder across all regime labels.
- `scheduler/post_tp_sl.go` — `closeTierListParam` (or a ratchet-specific resolver) to honor `use_defaults` and consult the resolved default.
- **`close_defaults` plumbing** — new top-level `config.json` block: struct field on the config, load/validate, unknown-key-guard entry, migration version bump, SIGHUP hot-reload compatibility, and a default-resolution helper consulted by every close evaluator with a hardcoded default.
- `scheduler/config_migration*.go` — keep `use_defaults` canonicalization intact; add `close_defaults` migration/canonicalization.
- `scheduler/hyperliquid_protection.go` — route `defaultHLProtectionTiers()` through the same `close_defaults` override path (scope: all close evaluators).
- `shared_strategies/close/trailing_tp_ratchet.py` + `registry.py` — mirror the hardcoded default; populate `trailing_tp_ratchet_regime` default_params or document that the Go loader fills it. (Operator `close_defaults` is a Go-side config concern; the Python evaluator continues to receive concrete tiers.)
- Tests: Go `_test.go` for three-layer resolution (hardcoded, `close_defaults` scalar + per-regime, per-strategy override) and validation across both classifiers; Python evaluator parity for the hardcoded default.
- Docs: `CLAUDE.md` close-evaluator notes + `config.example.json`.

---

richkuo:
## Summary

Implements the **mechanism** half of #866 — the three-layer close-default resolution and the fix for the actual reported bug (the `trailing_tp_ratchet*` evaluators had no working in-app default, so `use_defaults` / an omitted `tp_tiers` failed validation instead of falling back).

Three-layer precedence (lowest → highest): **system_close_defaults** → **user_close_defaults** → **strategy_close_defaults** (inline `tp_tiers`).

Preserves all existing default *values* — no behavior change for current strategies. The value retunes + per-regime group differentiation are split into the follow-up #870.

## What's in here

**1. Ratchet system default (the bug fix)**
- `defaultTrailingRatchetTiers()` (Go single source of truth) mirrored as `DEFAULT_RATCHET_TIERS` in `trailing_tp_ratchet.py`; the registry sources the constant (no drift).
- Wired into runtime tier resolution (broadcast across all regime labels for the regime variant) and validation: an omitted `tp_tiers` validates the default instead of erroring; `regime.enabled` and the initial-trail coupling (`validateTrailingRatchetInitialTrail`) are still enforced.

**2. `user_close_defaults` override layer**
- New top-level `user_close_defaults` config block (`CloseDefaultsMap`), resolved by **injection at load**: a user ladder is copied into any close ref that omits `tp_tiers`, after per-strategy normalization and before validation. Explicit per-strategy `tp_tiers` still wins; a ref with no matching entry falls through to the system default.
- Transparent to validation, runtime, and SIGHUP hot-reload (both sides re-load via `LoadConfig`). Additive field → no `config_version` bump, no top-level guard change (none exists).
- Scope: the evaluators that resolve purely via `tp_tiers` — `tiered_tp_pct`, `tiered_tp_atr`/`_live`, `trailing_tp_ratchet`/`_regime`. The `tiered_tp_atr_regime`/`_live_regime` variants are **excluded** (their `use_defaults` `RegimeATRBlock` baseline is per-regime territory handled in #870); excluded names are rejected at validation.

**3. Backtest `--defaults system|user` selector**
- `run_backtest.py --defaults user` applies the config's `user_close_defaults` to refs omitting `tp_tiers` (mirrors the Go injection), so backtests match the live daemon under the operator override; `system` (default) falls through to the Python evaluators' built-in defaults. `user` requires `--config`.

**4. Docs** — `config.example.json` block (probe-validated) + `CLAUDE.md` notes.

## Tests / verification
- New Go tests: ratchet default expansion, initial-trail coupling, regime broadcast, user-layer injection/validation, full `LoadConfig` (migrate → inject → validate) end-to-end.
- New Python tests: ratchet default fallback parity (scalar vs regime broadcast); backtest `--defaults` inject / system / strategy-wins.
- Full Go suite + shared close suite + backtest refs suite green; `config.example.json` probes OK.

## Follow-up
- #870 — retune the default values (A `tiered_tp_pct` 4-tier; B1 `defaultHLProtectionTiers` on-chain change; B2/C2 per-regime group ladders + per-regime opening trail). Depends on this PR's resolver + injection machinery.

Closes #866

---


richkuo:
@claude

richkuo:
@claude rereview

richkuo:
@claude sonnet rereview

PR #876, opened 2026-06-03
Add scale-in support for HL perps and manual

### LINKED ISSUE #873: Add scale-in / pyramiding: increase an open position's size (strategy + manual)
## Summary
There is currently no way to add to (scale into / pyramid) an open position — strategy or manual. A position's size only ever decreases until flat. This issue scopes what a scale-in feature would touch, grounded in the existing code.

## Decision (settled): freeze the original entry
A scale-in **blends only price and size** for PnL purposes — `AvgCost = (oldQty·oldAvg + addQty·addPrice)/(oldQty+addQty)`, `Quantity += addQty` — and **keeps `EntryATR`, the regime label, and the take-profit tier offsets pinned to the first entry**. The stop-loss distance and TP geometry do not move when you add. The only thing re-based is on-chain protection **sizing** (cancel+replace TP/SL so they cover the new total qty at the existing trigger geometry).

Rejected alternative (re-base): recomputing stop/TP/regime around the new blended average on every add. Not chosen — it would shift the operator's original risk plan on each add and fight the `InitialQuantity` and tier-watermark invariants below.

## Current behavior (intentional guards that block it)
Same-direction re-entry is deliberately skipped everywhere (idempotency for strategies that re-emit a buy every cycle while the setup holds):
- `PerpsOrderSkipReason` returns `"already long, skipping buy"` / `"already short, skipping sell"` and gates the live order before the subprocess spawns (`scheduler/portfolio.go:594`).
- `executePerpsSignalWithLeverage` skips same-direction signals (`scheduler/portfolio.go:927`, `:1107`).
- Spot: `"Already long … skipping buy"` (`scheduler/portfolio.go:1318`); `SpotOrderSkipReason` (`:773`), `FuturesOrderSkipReason` (`:800`).
- `manual-open` refuses when a position exists: `"position already open … close it first"` (`scheduler/manual.go:612`).

Partial **closes** exist (decrement `Quantity`, preserve `InitialQuantity`), but there is no partial **open**.

## Touch-points / dependents (resolved by the frozen decision)
1. **Position construction** — open always builds a fresh `Position` with `AvgCost: execPrice`, `InitialQuantity: qty` (`scheduler/portfolio.go:1068`, `:1426`, `:1686`; `scheduler/manual.go:615`). Scale-in needs a blend-into-existing branch instead.
2. **`InitialQuantity` invariant** — adding size makes `Quantity > InitialQuantity`, breaking the "partially closed" test (`Quantity < InitialQuantity`) used for tier-fill detection (`scheduler/discord.go:1135`), partial-close math (`scheduler/post_tp_sl.go:1123`,`:1134`), and fill accounting (`scheduler/hyperliquid_fills.go:472`, `scheduler/hyperliquid_balance.go:1558`). **Resolution:** grow `InitialQuantity` by the added qty so it stays the high-water size and the `Quantity < InitialQuantity` "partially closed" test remains correct.
3. **`EntryATR`** — stamped once (`stampEntryATRIfOpened` rejects if already set). Drives ATR stop distance and tiered-TP offsets. **Resolution: frozen** — keep the first-entry value.
4. **Regime label** — `Position.Regime`/`RegimeWindows` stamped once at open; regime-keyed SL/TP/ratchet resolve from it. **Resolution: frozen** — keep the original label.
5. **On-chain protection (HL live)** — reduce-only TP/SL placed at open sized to position qty. After an add they under-cover the new total; **re-size via cancel+replace to the new total qty at the existing (frozen) trigger prices**. The tier watermark `SLAdjustedTiersProcessed`/`TPArmedTiers` assumes size only decreases (`scheduler/post_tp_sl.go:1138`) — must be preserved across the re-size (an add must not reset cleared tiers).
6. **Trailing stop** — ATR-mult frozen at open (already matches the decision); SL size cap `hlSLEffectiveQty = min(virtualQty, onChainQty)` must track the larger qty after an add.
7. **Live HL order path** — leverage/margin-mode are settable only from flat (`scheduler/executor.go:304`); an add must keep existing leverage and **not** re-send `update_leverage`. `perpsLiveOrderSize` (`scheduler/portfolio.go:640`) only has flat-open / flip / close branches — needs an "add" branch. The skip-reason guards must learn the add case or they suppress the order before spawn (the #298 "fill lands, no Trade recorded" gap).
8. **Trade records / stats** — an add is a second open-side leg on the same position id (`IsClose:false`). `LifetimeTradeStatsAll` counts opens for `#T`; a scale-in would inflate the open count and W/L grouping unless the add leg is distinguished from a fresh open.

## Entry path (opt-in, default behavior unchanged)
Add an explicit entry path rather than overloading the open signal:
- a `manual-add` CLI subcommand, and/or
- an opt-in strategy flag (e.g. `allow_scale_in` with a max-adds / max-notional cap).

The default skip-on-same-direction behavior is preserved for every strategy that does not opt in.

## Acceptance criteria
- A documented, opt-in way to increase an open position's size (manual and/or strategy), gated so existing strategies are unaffected by default.
- `AvgCost` and `Quantity` blend correctly; PnL on the eventual close is computed against the blended average.
- `EntryATR`, regime label, and TP tier offsets remain pinned to the first entry after an add (frozen).
- On-chain TP/SL (HL live) cover the new total size after an add, at the unchanged trigger prices, without resetting the cleared-tier watermark.
- `InitialQuantity` grows with the add so no false "partially closed" state arises.
- Skip-reason guards allow the add path without reopening the #298 live-fill-without-Trade gap.
- Test coverage for the blend math and the protection re-sizing.

---

richkuo:
## Summary
- Add opt-in scale-in handling for Hyperliquid perps and manual positions.
- Preserve first-entry trade geometry while blending average cost and sizing on same-side adds.
- Update config validation, reload behavior, protection logic, trailing-stop handling, and Discord/status surfaces to stay consistent with the new position model.
- Extend tests and probe coverage around config parsing, executor flows, manual actions, and Hyperliquid checks.

Closes #873

## Testing
- Python tests not run: local `uv`/`.venv` environment could not provide `pytest`/runtime deps in this sandbox.

LLM: GPT-5 | medium | Harness: codex


richkuo:
@claude

richkuo:
@claude rereview

richkuo:
Superseded by #882 (merged). Two of this PR's contributions were folded in: (1) the full frozen-entry geometry sweep — re-anchoring every ATR trigger/display site (trailing SL, post-TP SL, ratchet, Discord, reconcile) off the frozen entry rather than the blended avg, which the chosen base had only done for the protection plan; and (2) hot-reload handling (allowed when flat, blocked while open). The synthesis used #877's no-cancel-at-add resize instead of this PR's execute-time cancel/replace to avoid a brief on-chain coverage gap. Thanks.

PR #878, opened 2026-06-03
fix(manual): stamp regime onto manual positions so regime-keyed closes arm (#872)

### LINKED ISSUE #872: Manual positions never record a regime, so regime-keyed closes silently no-op
## Summary
Manually opened positions never record a market regime label, so regime-keyed close strategies silently do nothing on them. Automated strategies stamp the regime onto the position at open (`stampPositionRegimeIfOpened`, wired into all 5 execute dispatches); the `manual` path has no open signal and skips stamping, leaving `Position.Regime` empty for the life of the position.

## Impact
- Regime-keyed Go-side close logic resolves its label from the frozen `Position.Regime` (`positionATRRegimeLabel` → `protectionATRRegimeLabel`). With an empty label, `trailing_tp_ratchet_regime` (and any `*_atr_regime` SL/TP arming) returns no tiers and never arms on manual positions.
- The live per-cycle close evaluator still receives the current regime (`runManualCloseEval` passes `cfg.Regime` to `runHyperliquidCheck`), so the Python-side close sees regime, but the Go-side frozen-label features do not.
- Net: a manual position configured with a regime-aware close behaves as if regime were disabled for the frozen-label paths.

## Proposed fix (Option A — lazy stamp on first close cycle)
`runManualCloseEval` (`scheduler/manual.go:884`) already calls `runHyperliquidCheck` each cycle and gets back a regime payload (`result.Regime`) that it currently discards. Stamp it onto the position once via the existing idempotent helper:

```go
stampPositionRegimeIfOpened(ss, sc.Symbol, regimePayloadValue(result.Regime), sc, cfg.Regime)
```

`stampPositionRegimeFromPayload` only writes when `pos.Regime == ""`, so it stamps exactly once, on the first close-eval cycle after the position opens. Requires threading the payload out of (or stamping inside) `runManualCloseEval`, which today returns only `(closeFraction, closePrice, ok)`.

### Tradeoff
The label is measured one scheduler cycle *after* the fill, not at the exact entry. On manual's multi-minute intervals this is negligible and matches "current regime" semantics closely enough for the regime-keyed closes to arm correctly.

### Alternative (Option B — stamp at true open)
Stamp inside the pending-action apply path where the `Position` is built (`scheduler/manual.go:615`). No regime payload exists there (`manual-open` is a separate CLI invocation that queues an action; the daemon applies it later), so it would require computing the regime at apply time (extra check-script subprocess) or threading an ATR/regime through the queued action. More moving parts; not recommended unless exact-at-fill stamping is required.

## Acceptance criteria
- A manual position opened with a `trailing_tp_ratchet_regime` (or `*_atr_regime`) close has a non-empty `Position.Regime` within one close-eval cycle of opening.
- The regime-keyed ratchet/SL arms on manual positions.
- Stamping is idempotent (does not overwrite a regime already set) and `--record-only` manual positions behave consistently.
- Test coverage for the manual stamp path.

richkuo:
Implements Option A from #872: `runManualCloseEval` now returns the cycle's regime payload, and the manual dispatch stamps it once via the idempotent `stampPositionRegimeIfOpened` under the state lock so frozen-label close features (`trailing_tp_ratchet_regime`, `*_atr_regime` SL/TP) arm on manual positions. Stamp fires on any successful close-eval (not gated on `closeFraction > 0`) so live and --record-only behave identically.

Adds `TestManualStampRegimeOnPosition` and `TestManualCloseEval_FlatReturnsEmptyPayload`.

Closes #872

---

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -1726,17 +1726,21 @@ func main() {
 							break
 						}
 						if pos != nil && hyperliquidIsLive(sc.Args) {
+							// Protection-sync must run before the close evaluator so
+							// the latter sees on-chain-reconciled qty. Post-TP SL
+							// adjustment is deferred to the post-stamp block below so
+							// regime-keyed *_atr_regime SL sees pos.Regime (#878 review).
 							runHyperliquidProtectionSync(sc, stratState, stateDB, sc.Symbol, &mu, notifier, logger, "HL manual protection synced", hlReconcileFillHintsJSON)
-							runPostTPStopLossAdjustment(sc, stratState, sc.Symbol, prices[sc.Symbol], cfg, &mu, notifier, logger, hlOnChainAbsQty)
 						}
 						// #872: run the close evaluator and stamp the regime BEFORE
-						// arming the ratchet/trailing walker below. The regime is the
-						// close-eval's classifier output, and the regime-keyed close
-						// features (trailing_tp_ratchet_regime, *_atr_regime SL/TP)
-						// resolve their tier table / trail distance from pos.Regime.
-						// Arming first (the previous order) left pos.Regime == "" on
-						// the first post-open cycle, so the regime trail no-op'd for
-						// one interval; stamping first arms it correctly on cycle 1.
+						// arming the post-TP SL / ratchet / trailing walker below. The
+						// regime is the close-eval's classifier output, and the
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude rereview

richkuo:
@claude rereview

PR #881, opened 2026-06-03
fix(manual): emit trade alerts for adopted manual fills (#880)

### LINKED ISSUE #880: Manual open/close fills do not emit trade alerts because pending manual actions bypass sendTradeAlerts
## Summary

Manual `manual-open` / `manual-close` exchange fills are adopted into scheduler state via `pending_manual_actions`, but that adoption path does not emit the normal trade alerts / owner DM notifications.

This is why live manual fills appear on exchange and later in dashboard state, but no Discord DM/channel trade alert is sent.

## Observed behavior

In live trading on June 3, 2026:

- `manual-open` placed a real Hyperliquid short fill successfully.
- CLI printed `Queued: hl-manual-btc-live position will appear in the dashboard after the next scheduler cycle.`
- On the next scheduler cycle, the drain path logged:
  - `[manual] applied open: hl-manual-btc-live short ...`
- The position was materialized into AppState and later visible in dashboard/status.
- No trade alert / owner DM was sent for the open.

Same issue appears to apply to manual close actions adopted through the same queue-drain path.

## Root cause

Normal strategy-executed trade alerts are sent only when the main strategy loop sets:

- `trades > 0`
- `detail != ""`

and then calls:

- `sendTradeAlerts(sc, stratState, trades, &mu, notifier)` (`scheduler/main.go:1898`)

Relevant code:

- `scheduler/main.go:1891-1898` (`if trades > 0 && detail != "" { ... sendTradeAlerts(...) }`)

Manual fills do not go through that path.

Instead:

- `manual-open` / `manual-close` immediately place the exchange order
- then enqueue a `PendingManualAction`
- the scheduler later calls `drainPendingManualActions(...)` (`scheduler/manual.go:554`)
- which calls `applyManualAction(...)` (`scheduler/manual.go:591`)
- which records the trade / mutates state (`recordPositionOpen` / `RecordTrade`, appending to `TradeHistory`) / prints `[manual] applied open: ...` (`scheduler/manual.go:652,655,689,699`)
- but does **not** call `sendTradeAlerts(...)` or otherwise emit equivalent notifications

`drainPendingManualActions` does not even receive a `*MultiNotifier` — the notifier built in `manual.go` (lines 252-374) lives in the short-lived **CLI process** and is used only for `warnNotifier` warning paths, never for trade alerts. The scheduler-side drain has no notifier in scope at all.

Relevant code paths:

- `scheduler/manual.go`
  - `drainPendingManualActions(...)` — `:554`
  - `applyManualAction(...)` — `:591`
- `scheduler/main.go`
  - pending manual actions are drained before the dueStrategies loop (`:576`)
  - trade alerts are only emitted from the main strategy execution path (`:1898`)

## Why this matters

Manual fills are real exchange executions and should produce the same operator-facing alerts as strategy fills.

Today this creates a confusing gap:

- exchange order fills
- position later appears in dashboard
- but no notification is sent

That makes it easy to miss that a manual position actually opened/closed.

## Suggested fix

Emit trade alerts for successfully applied pending manual actions.

**Lock-ordering constraint (must observe):** `drainPendingManualActions` is called under `mu.Lock()` (`scheduler/main.go:575-577`), and `sendTradeAlerts` re-acquires `mu.RLock()` (`scheduler/main.go:2500`). `sync.RWMutex` is **not reentrant**, so calling `sendTradeAlerts` directly inside the drain self-deadlocks. The main loop avoids this by alerting outside the write lock (`:1898`). Any fix must collect the applied actions while holding the lock and emit alerts **after** `mu.Unlock()`.

Possible approaches:

1. Thread a `*MultiNotifier` into `drainPendingManualActions(...)`, have it **collect** successfully applied actions per strategy (`sc`, `*StrategyState`, count) under the lock, return them, and call `sendTradeAlerts(...)` at the `main.go:577` call site **after** `mu.Unlock()`. Do **not** call `sendTradeAlerts` from inside the drain (deadlock — see constraint above).
2. Refactor the alert path so both:
   - strategy-executed trades
   - pending manual action trades

   share one common helper, emitted from outside any held write lock.

Note: `sendTradeAlerts(sc, stratState, trades, ...)` alerts the **last `trades`** entries of that strategy's `TradeHistory`, so the drain must track the per-strategy count of trades it appended this cycle and pass it through.

## Acceptance criteria

- Successful `manual-open` adoption sends the same DM/channel trade alert style as a normal live trade.
- Successful `manual-close` adoption does the same.
- No duplicate alerts when the same pending action is drained once and deleted (drain already deletes rows through `maxDrained` after apply — `scheduler/manual.go:583-587`).
- Alert routing should continue respecting configured Discord/Telegram `dm_channels` / `trade_alert_channels`.
- No deadlock: alerts emitted after the drain releases `mu`.

## Notes

This is separate from the reconciler TP/SL fill DM path (`notify_tp_sl_fills`). That path is about reconciled protection fills. This issue is specifically about entry/exit trade alerts for manual actions.

---

richkuo:
## Problem

`manual-open` / `manual-close` exchange fills are adopted into scheduler state via the `pending_manual_actions` drain, but that path never emitted an operator-facing trade alert. The position appeared on-exchange and later in the dashboard with **no Discord/Telegram DM or channel notification**, making it easy to miss that a manual position actually opened or closed.

## Root cause

Normal trade alerts fire from the main strategy loop (`main.go:1898`, gated on `trades > 0 && detail != ""`). Manual fills bypass that loop: they're applied by `drainPendingManualActions` → `applyManualAction`, which records the trade and mutates state but never calls `sendTradeAlerts`. The drain didn't even receive a `*MultiNotifier`.

The naive fix (alert from inside the drain) deadlocks: `drainPendingManualActions` runs under `mu.Lock()` (`main.go:575-577`) and `sendTradeAlerts` re-acquires `mu.RLock()` (`main.go:2500`). `sync.RWMutex` is not reentrant.

## Fix

- `drainPendingManualActions` now **collects** one `manualAlert` per strategy that had ≥1 action successfully applied (with the aggregated trade count) and returns the slice — it no longer alerts inline.
- The call site fires `sendTradeAlerts` for each returned alert **after** `mu.Unlock()`, giving manual fills the same DM/channel routing (`tradeAlertRoutes`, respecting `dm_channels` / `trade_alert_channels`) as normal live trades.
- Emission order is deterministic (id-sorted insertion order), and each action is alerted exactly once since drained rows are deleted after a successful drain.

## Tests

- Extended `TestDrainPendingManualActions` to assert the returned alert (strategy, count, ss pointer).
- New `TestDrainPendingManualActionsAlerts`: per-strategy count aggregation (open + full close = 2 trades), a failed apply contributes no alert, and `trades ≤ len(TradeHistory)` so the alert tail-slice can't underflow.
- Full `go test ./...` passes.

Closes #880

---

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -573,8 +573,15 @@ func main() {
 		// #569: Drain pending manual-open / manual-close actions before the
 		// dueStrategies loop so newly-materialised positions are visible this cycle.
 		mu.Lock()
-		drainPendingManualActions(state, cfg, stateDB)
+		manualAlerts := drainPendingManualActions(state, cfg, stateDB)
 		mu.Unlock()
+		// #880: Emit trade alerts for adopted manual fills AFTER releasing mu.
+		// sendTradeAlerts re-acquires mu.RLock, so alerting inside the drain
+		// (which runs under mu.Lock) would self-deadlock. This gives manual fills
+		// the same DM/channel routing as normal live trades.
+		for _, ma := range manualAlerts {
+			sendTradeAlerts(ma.sc, ma.ss, ma.trades, &mu, notifier)
+		}
 
 		// #87: Resolve capital_pct → capital for strategies with dynamic sizing.
 		// Must run on cfg.Strategies (not dueStrategies) so resolved capital persists
--- a/scheduler/manual.go
+++ b/scheduler/manual.go
@@ -548,20 +548,33 @@ func runManualClose(args []string) int {
 	return 0
 }
 
-// drainPendingManualActions reads all rows from pending_manual_actions and
-// applies them to the in-memory AppState, then deletes the drained rows.
-// Called at the top of each scheduler cycle before dueStrategies is built.
-func drainPendingManualActions(state *AppState, cfg *Config, stateDB *StateDB) {
+// manualAlert captures one strategy's successfully drained manual actions so the
+// caller can emit trade alerts AFTER releasing mu. drainPendingManualActions runs
+// under mu.Lock and sendTradeAlerts re-acquires mu.RLock; since sync.RWMutex is
+// not reentrant, alerting inside the drain would self-deadlock (#880).
+type manualAlert struct {
+	sc     StrategyConfig
+	ss     *StrategyState
+	trades int // count of trades appended this drain for this strategy
+}
+
+// drainPendingManualActions reads all rows from pending_manual_actions, applies
+// them to the in-memory AppState, then deletes the drained rows. It returns one
+// manualAlert per strategy that had >=1 action successfully applied (with the
+// aggregated trade count) so the caller can fire sendTradeAlerts outside the
+// state write lock (#880). Called at the top of each scheduler cycle before
+// dueStrategies is built.
+func drainPendingManualActions(state *AppState, cfg *Config, stateDB *StateDB) []manualAlert {
 	if stateDB == nil {
-		return
+		return nil
 	}
 	actions, err := stateDB.LoadPendingManualActions()
 	if err != nil {
 		fmt.Printf("[manual] failed to load pending actions: %v\n", err)
-		return
+		return nil
 	}
 	if len(actions) == 0 {
-		return
+		return nil
 	}
 
 	scByID := make(map[string]StrategyConfig, len(cfg.Strategies))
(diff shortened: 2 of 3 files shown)


richkuo:
@claude review

PR #886, opened 2026-06-04
feat(manual): resting limit orders for manual-open (#883)

### LINKED ISSUE #883: Support limit orders for manual position open
## Feature Request

Allow placing limit orders, especially for opening manual positions.

**Use case:** `manual-open` should support a target entry price, e.g. "open a long position when ETH hits $1,800". The order sits as a resting limit order on the exchange until filled or cancelled.

**Current behavior:** `manual-open` only supports market orders — it calls `RunHyperliquidExecute` → `check_hyperliquid.py run_execute` → `adapter.market_open(...)` and hard-errors if no immediate fill returns (`manual.go:230`). The position opens immediately at ~mid.

**Proposed:** Add a `--limit-price` flag to `manual-open` that places a resting limit order instead of a market order.

Design notes (this is a new subsystem, not a small flag):

1. **New OID-keyed pending-orders table** (e.g. `pending_limit_orders`), persisted so resting orders survive a scheduler restart. Do **not** reuse `pending_manual_actions` — that queue only holds *already-filled* actions that `manual-open` writes after a synchronous fill for the scheduler to convert to a `Position`. A resting limit order has no fill at placement time.
2. **`manual-open` becomes fire-and-exit:** place the resting limit order, persist its OID, and return — rather than block-until-fill as it does today.
3. **Scheduler-side fill detection:** a per-cycle poll phase reconciles fill status by OID (reuse `hyperliquid_fills.go` / `hl_user_fills.py` `lookup_fill_fee_by_oid`) and, on fill, creates the tracked position.
4. **Defer protection to fill:** stop-loss arming (today synchronous in the CLI, `manual.go:296`) and any `tiered_tp_atr_live` on-chain protection must move to the scheduler-side post-fill path, since there's no fill at placement.
5. **Lifecycle / failure handling:** define behavior for never-fills, partial fills, operator cancellation, and order expiry (GTC sits indefinitely; HL caps 1000 open orders/account).

**Scope:** HL perps / `manual` only.

Note: the HL SDK already supports resting limit orders (`self._exchange.order(..., {"limit": {"tif": "Gtc"}})`, used today for reduce-only TPs), but a non-reduce-only **entry** limit method is net-new.

---
Created with LLM: zai/glm-5.1 | Harness: external

richkuo:
## Summary

Adds `manual-open --limit-price` to place a resting maker limit order instead of a market order, with scheduler-side fill detection and post-fill protection. Closes #883.

The three lifecycle behaviors are the optimal-for-correctness choices:
1. **Post-only by default** (`--tif Alo`) — always rests as a maker; HL rejects an already-crossed price and the CLI surfaces it so you re-price. `--tif Gtc` opts into immediate marketable fills.
1. **Partial fills open immediately** — the scheduler grows the tracked position to the cumulative filled size each cycle and arms SL/TP right after each fill, so no filled coin is ever unprotected; the remainder keeps polling until full or cancelled.
1. **`manual-cancel <id>` + `--expire-after <dur>`** — both route through one cancel→finalize path. The scheduler is the single writer of all fill booking, so the CLI only flags `cancel_requested` and the scheduler cancels on-chain + books any last-moment fill next cycle. No CLI/scheduler race.

## How it's wired

- **Reuses the existing SSoT for protection.** The scheduler already protection-syncs `type=manual` positions every cycle (`main.go` `case "manual"` → `runHyperliquidProtectionSync`, arming SL+TP from a zero OID and re-sizing as the position grows). The limit-fill path reuses that rather than duplicating `manual.go`'s inline arming, and also calls it right after each fill to close the unprotected window regardless of due-ness.
- New `pending_limit_orders` table (resting OID + fill watermarks), distinct from `pending_manual_actions` as the issue specified. Created via the always-run schema DDL — no migration.
- New `scheduler/manual_limit.go`: executor wrappers + parsers, the CLI fire-and-exit path, `manual-cancel`, and `reconcilePendingLimitOrders` (runs at cycle top right after `drainPendingManualActions`).
- Python: `adapter.limit_open` (net-new non-reduce-only entry), `adapter.fills_summary_by_oid` (cumulative size + VWAP — the existing fee-only lookup couldn't provide this), and `check_hyperliquid.py` `--limit-open`/`--limit-status`/`--cancel-order` modes. All three probed at startup so a stale Python deploy fails loudly.

## Scope / notes

- HL perps / `manual` only.
- Protection-override flags (`--stop-loss-atr-mult`/`--stop-loss-pct`) are intentionally ignored on the limit path (SL/TP comes from strategy config via the per-cycle sync); the CLI warns if you pass them.

## Testing

- **Go:** full suite green, incl. 14 new tests — argv/parse contracts, fill math, the create/grow/owner-guard apply path, DB CRUD, and the full reconcile lifecycle (full fill / partial→complete / cancel / expiry / defer-on-unknown-book) with stubbed subprocesses.
- **Python:** 1098 testpaths + 60 check_hyperliquid + 10 new adapter tests passing.
- gofmt/vet clean, release build OK, startup probes exit 0, end-to-end dry-run sizes correctly off the limit price.

---


richkuo:
@claude review

richkuo:
@claude sonnet rereview

richkuo:
@claude sonnet rereview

PR #882, opened 2026-06-03
feat(scale-in): opt-in scale-in / pyramiding — best-of-three synthesis (#873)

richkuo:
## Summary
Synthesizes the three competing scale-in PRs (1. #875, 2. #876, 3. #877) into one implementation that solves #873 and folds in the strongest parts of each. Scope: **HL perps + manual, live + paper**; opt-in, default-off. Implements the settled "freeze the original entry" decision.

A same-direction signal on an open position **blends only price and size** for PnL (`AvgCost`, `Quantity`, `InitialQuantity` grow) and **freezes the risk plan** — `EntryATR`, regime label, and all SL/TP trigger geometry stay pinned to the first entry (`RiskAnchorPrice`); the cleared-tier watermark is never reset. Only on-chain protection **size** is re-based.

## Base + what each PR contributed
Built on #877's architecture (the strongest base) and layered in the best of the others:

**From #877 (base):**
- Single-source `perpsScaleInDecision` computed once in Phase 1, consumed by both the live order and the Trade record — closes the #298 fill-without-Trade gap by construction.
- `RiskAnchorPrice` separates the PnL anchor (blended `AvgCost`) from the frozen risk-geometry anchor.
- `runHyperliquidScaleInOrder` places a same-side market order with **no SL/TP cancel and no `update_leverage`**; existing protection stays resting and the next sync re-sizes it — so there is no cancel-without-replace coverage gap (avoids the window present in #876's execute-time approach) and no Python changes.
- Signed `add_spacing_atr` (`>0` add-to-winners, `<0` average-down), `max_adds`/`max_added_notional_usd` caps, `manual-add` CLI.

**Folded in from #876:**
- **Full frozen-entry geometry sweep.** #877 only re-anchored the protection plan; every other ATR-trigger/display site still used the blended `AvgCost`, which would shift the risk plan after an add. Now `pos.riskAnchorPrice()` drives fixed/trailing SL (`hyperliquid_trailing_stop.go`), post-TP SL adjustment (`post_tp_sl.go`), ratchet profit distance (`trailing_tp_ratchet.go`), Discord SL/TP/ratchet display (`discord.go`), and TP reconcile matching (`hyperliquid_balance.go`). The accessor falls back to `AvgCost` for non-scaled positions, so all existing behavior is byte-identical.
- **Hot-reload handling** (a real gap in #877): `allow_scale_in`/`scale_in` are hot-reloadable when flat and blocked while open; both added to `strategyRestartShape`.

**Folded in from #875:**
[...]

PR #888, opened 2026-06-04
fix(perps): arm ATR-trailing stop-loss inline at open (#885)

### LINKED ISSUE #885: HL perps: ATR/trailing stop-loss not placed at open — full position unprotected for up to one cycle
## Summary
On Hyperliquid **live perps**, a strategy whose stop-loss is **trailing-ATR**-based (`trailing_stop_atr_mult` / `trailing_stop_atr_regime`) opens **without an on-chain stop** and only arms it on the *next* `Signal == 0` cycle. Between the open fill and that cycle, the **entire position has no exchange-side stop** — up to a full strategy interval (e.g. 1h on a 1h strategy). This is the same risk class surfaced in the #882 scale-in review, but for fresh opens rather than adds.

**Scope correction (validated against `main`):** the non-trailing ATR owners — `stop_loss_atr_mult`, `stop_loss_atr_regime`, unified per-regime, and `sl_after` ATR-owned closes — are **NOT** naked. They get their SL placed on-chain in the *same cycle* as the open by the post-trade protection sync (`scheduler/main.go:1752`, the same `--sync-protection` path manual uses). Only the **trailing** owners remain naked, because the trailing walker that arms their SL is gated on `result.Signal == 0` and so first fires the cycle *after* open.

## Current behavior (verified against `main`)
1. At order time the execute call sizes the inline SL from `slPct := EffectiveStopLossPct(sc)` (`scheduler/main.go:2894`), passed to `RunHyperliquidExecute` (`:2931`); Python places it only when `slPct > 0`.
2. `EffectiveStopLossPct` deliberately returns **0** for every ATR/regime/trailing owner because the trigger depends on `Position.EntryATR` (and, for regime, `Position.Regime`), which aren't known until `stampEntryATRIfOpened` / `stampPositionRegimeIfOpened` run after the fill (`scheduler/config.go:622`, `:628`, `:636`, `:642`, `:648`). So at the execute call itself **no** ATR/trailing SL is placed.
3. **Non-trailing ATR owners are armed in the same cycle, not the next one.** After the open trade, `runHyperliquidProtectionSync` runs (`scheduler/main.go:1752`, gated `execResult != nil && trades > 0`) — *after* `executeHyperliquidResultDeferredOpen` stamps `EntryATR` (`:3053`) and `Regime` (`:3054`). `buildHyperliquidProtectionPlan` derives `slMult` from `stop_loss_atr_mult` / `stop_loss_atr_regime` / unified (`scheduler/hyperliquid_protection.go:49-58`) and Python `place_stop_loss` fires for a fresh `stop_loss_oid=0` (`shared_scripts/check_hyperliquid.py:624`, `:629`; `_resolve_missing_oid(0)` → `"place"`). The next-cycle fixed-ATR arm (`scheduler/main.go:1654`, gated `result.Signal == 0 && hlPosQty > 0 && sc.StopLossATRMult != nil && hlStopLossOID == 0`) is then a **no-op fallback** — `hlStopLossOID` is already non-zero.
4. **Trailing owners are the real gap.** `buildHyperliquidProtectionPlan` never reads `trailing_stop_atr_mult` / `trailing_stop_atr_regime`, so the post-trade sync places no trailing SL. The only path that arms it is the trailing walker (`scheduler/main.go:1599`), gated `result.Signal == 0`, which does not run on the open cycle.

Net: a **trailing**-SL perps open is **naked** until the next time the strategy is evaluated. Non-trailing ATR/regime/unified opens get their SL on-chain seconds after the fill, within the same cycle.

## Why it matters
The exposed quantity is the **whole position**, not an increment. On a leveraged perp, a hard adverse move during the gap can run the position to the exchange liquidation price before any stop exists — worst case, total loss of that position's margin. The longer the strategy interval, the larger the window.

## Scope
- **Affected:** HL live perps with `trailing_stop_atr_mult` or `trailing_stop_atr_regime` (and no other inline SL owner).
- **Not affected:** `stop_loss_atr_mult`, `stop_loss_atr_regime`, unified per-regime, and `sl_after` ATR-owned closes — the SL is placed inline in the same cycle by the post-trade protection sync (`scheduler/main.go:1752`); scalar `stop_loss_pct` / `stop_loss_margin_pct` opens (pct known at order time, SL placed inline); paper (no on-chain orders); `type=manual` (places SL+TP inline at open via `placeManualProtectionInline` → `--sync-protection`, #633/#689).

## Proposed fix
Arm the initial trailing SL inline on the open cycle for trailing owners, mirroring how the non-trailing ATR owners are already handled by the post-trade sync. The entry ATR is already available — `stampEntryATRIfOpened` stamps `Position.EntryATR` from the signal-check `result.Indicators` *before* the post-trade sync runs, so no new `--fetch-atr` path is needed. Reuse the existing trailing primitive `runHyperliquidTrailingStopUpdate` (as `scaleInResizeTrailingSLNow` already does for scale-in adds, `scheduler/scale_in.go:64`) immediately after the post-trade protection sync, with the same `hlSLEffectiveQty` on-chain-reconcile cap handling so the SL covers the confirmed fill.

## Acceptance criteria
- An HL live perps position with a **trailing** stop has an on-chain stop placed in the same cycle as the open fill (no naked window).
- The inline trigger matches the value the post-open walker would have produced (no double-arming / no drift); the post-open walker no-ops when the inline SL already exists.
- Non-trailing ATR/regime/unified, scalar-pct, paper, and manual behavior all unchanged.
- Regression test asserting a trailing-SL perps open arms the SL on the open cycle.

---

richkuo:
## Problem
On HL **live perps**, a strategy whose stop-loss is **ATR-based trailing** (`trailing_stop_atr_mult` / `trailing_stop_atr_regime`) opened with **no exchange-side stop** and only armed it on the next `Signal == 0` trailing-walker cycle — the entire position naked for up to a full strategy interval.

Validation narrowed the issue's original scope: the non-trailing ATR/regime/unified owners are **not** affected — the post-trade protection sync (`main.go:1752`) already places their SL the same cycle as the open (`buildHyperliquidProtectionPlan` derives the trigger from the just-stamped `EntryATR`/`Regime`), and scalar-pct owners (including `trailing_stop_pct`) are placed inline at the execute order because `EffectiveStopLossPct` returns a positive pct. The gap is specific to **ATR-trailing** owners, whose `EffectiveStopLossPct` defers to `0` and which the protection plan never reads.

## Fix
`armTrailingStopAtOpenNow` runs right after the post-trade protection sync on the open cycle and arms the initial trailing SL by **reusing the exact walker primitive** (`runHyperliquidTrailingStopUpdate`) and result handler (`applyTrailingStopUpdateResult`) the next-cycle walker uses — one stop-placement path, so the inline trigger uses the **same formula and code path** the deferred walker would have — the only difference is this cycle's mark vs. the next cycle's, and arming sooner is strictly safer. A fresh open carries `currentTrigger==0`, so the primitive computes and rests the initial trigger from a `max(AvgCost, mark)`-seeded high-water (the walker's own ratchet) without `forceResize`.

**Idempotent / tightly scoped** — no-ops unless the position is a live trailing owner with no resting SL (`StopLossOID==0 && StopLossTriggerPx==0`):
- scalar `trailing_stop_pct` → OID already stamped inline at the execute order (`main.go:3082`)
- fixed/regime ATR → OID stamped by the post-trade sync
- partial-close legs → reduce-only SL keeps resting
- the post-open walker no-ops next cycle because the SL now exists

The #621 size cap uses the on-chain qty corrected with the confirmed open fill (the per-cycle reconcile snapshot predates this open); if still capped (shared-coin reconcile lag) it defers to the next walker cycle — never worse than the pre-fix behavior.

This mirrors how `scaleInResizeTrailingSLNow` (#882) already arms/resizes the trailing SL on a non-walker cycle for scale-in adds.

## Scope
- HL **live perps** ATR-trailing owners only. No Python changes (open-side Go decision). Paper, manual, scalar-pct, and fixed/regime ATR behavior all unchanged.

## Tests
- `TestArmTrailingStopAtOpenNow` — asserts inline placement at the correct AvgCost-seeded trigger/qty, plus the no-op guards (existing SL, not-live, non-trailing owner, size-capped deferral).
- Full `scheduler` suite green; `go vet` clean.

Closes #885

---

### DIFF
--- a/scheduler/hyperliquid_open_trailing.go
+++ b/scheduler/hyperliquid_open_trailing.go
@@ -23,10 +23,13 @@ import (
 // Correctness over a single steady-state path: it reuses the exact walker
 // primitive (runHyperliquidTrailingStopUpdate) and result handler
 // (applyTrailingStopUpdateResult) the next-cycle walker uses, so the inline
-// trigger is byte-identical to what the deferred arming would have produced and
-// there is only one stop-placement implementation that can't drift. A fresh
-// position carries currentTrigger==0/currentOID==0, so the primitive computes
-// the initial trigger (AvgCost-seeded high-water) and rests a new SL without
+// trigger comes from the identical formula and code path the deferred arming
+// would have used — the only difference is this cycle's mark vs. the next
+// cycle's, and arming sooner is strictly safer. One stop-placement
+// implementation means the two can't drift. A fresh position carries
+// currentTrigger==0/currentOID==0, so the primitive computes the initial
+// trigger from a max(AvgCost, mark)-seeded high-water (the walker's own ratchet,
+// so positive open slippage is handled the same way) and rests a new SL without
 // needing forceResize.
 //
 // Idempotent and tightly scoped: it no-ops unless the position is a live
--- a/scheduler/hyperliquid_open_trailing.go
+++ b/scheduler/hyperliquid_open_trailing.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"fmt"
+	"sync"
+)
+
+// armTrailingStopAtOpenNow places the initial TRAILING stop-loss on the SAME
+// cycle as a fresh open (or the open side of a flip), closing the #885 naked
+// window. The non-trailing ATR/regime/unified SL owners are already armed this
+// cycle by the post-trade protection sync (buildHyperliquidProtectionPlan
+// derives their trigger from the just-stamped EntryATR/Regime), and the scalar
+// owners (stop_loss_pct, stop_loss_margin_pct, trailing_stop_pct) are placed
+// inline at the execute order because EffectiveStopLossPct returns a positive
+// pct for them. The ATR-trailing owners (trailing_stop_atr_mult /
+// trailing_stop_atr_regime) are the gap: EffectiveStopLossPct defers to 0 (the
+// distance needs per-position EntryATR) and buildHyperliquidProtectionPlan never
+// reads the trailing fields, so the only path that arms them is the Signal==0
+// trailing walker — which first fires the cycle AFTER open. On a long strategy
+// interval that leaves the whole position with no exchange-side stop for up to a
+// full interval.
+//
+// Correctness over a single steady-state path: it reuses the exact walker
+// primitive (runHyperliquidTrailingStopUpdate) and result handler
+// (applyTrailingStopUpdateResult) the next-cycle walker uses, so the inline
+// trigger is byte-identical to what the deferred arming would have produced and
+// there is only one stop-placement implementation that can't drift. A fresh
+// position carries currentTrigger==0/currentOID==0, so the primitive computes
+// the initial trigger (AvgCost-seeded high-water) and rests a new SL without
+// needing forceResize.
+//
+// Idempotent and tightly scoped: it no-ops unless the position is a live
+// trailing owner with NO resting SL (StopLossOID==0 && StopLossTriggerPx==0).
+// That guard is what keeps it from double-placing on scalar trailing_stop_pct
+// (inline OID already stamped at the execute order, main.go), on fixed/regime
+// ATR (OID stamped by the post-trade sync), and on partial-close legs (the
+// reduce-only SL keeps resting). It also means the post-open walker no-ops next
+// cycle because the SL already exists.
+//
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

richkuo:
@claude sonnet rereview

PR #901, opened 2026-06-05
feat(regime): shared regime bundle per (asset, interval) signature (#879)

### LINKED ISSUE #879: feat(regime): shared regime bundle per (asset, interval, period) with 3-state + 7-state views and per-strategy override
## Problem

Regime is currently computed independently inside each strategy's check subprocess, as a byproduct of the open-signal check. (For `manual` it's the same open-signal check — `runManualCloseEval` calls the full HL check `runHyperliquidCheck` and reads `result.Regime` — not a separate close-only eval; and it only computes when a position is open, returning empty while flat.) Two strategies on the same coin + timeframe each recompute the regime math (even when #839 dedupes the HL OHLCV `/info` fetch within an instance), and there's no single source of truth a strategy (or the dashboard) can pull from. Each strategy can also carry its own classifier/period/thresholds, so regime is treated as a per-strategy knob rather than a property of the market.

Today the only live "current regime" storage is `stratState.Regime` (refreshed per cycle, in-memory, not persisted) and the frozen-at-open `pos.Regime` (persisted). Neither is shared across strategies.

## Proposal

A single **global regime calculator** owns all regime computation. Each scheduler cycle it builds a fresh in-memory map: collect the distinct regime signatures of **due** strategies → compute each raw key once → project labels → every strategy and `manual` reads from that map and **does not recompute regime inline**. Custom params are not a private/local path — they are just another signature the global calculator computes and stores, so any peer with identical params shares the same entry.

Conceptually like the layered-default pattern of `user_close_defaults` (#866), but the "override" still lives in the global store rather than being computed privately by the strategy.

**Primary payoff (not mainly fewer network calls):** single-source consistency per signature per cycle, portfolio/dashboard regime view, and bringing `options` + flat `manual` into the same system. #839 already dedupes HL OHLCV I/O across peers within an instance; this work dedupes **regime math** and **label ownership**.

### Where it runs (Go-side calculator)

The calculator and its store live **in the Go scheduler**, not in the per-strategy Python check subprocesses. This is the only place that sees all strategies in a cycle, and the only process that serves the dashboard. Per-strategy subprocesses have private *in-memory* heaps and cannot share a Go map. (Note: subprocesses *can* share derived state on disk — the #839 OHLCV cache is a per-**instance** `/tmp` file cache, `PrivateTmp`-isolated per systemd instance, that dedupes OHLCV I/O *across* subprocesses within an instance; it just stores OHLCV, not regime, so it's a structural-pattern reference only.) A `/tmp`-file regime cache could likewise dedupe regime math across subprocesses, so cross-subprocess sharing is not the blocker; the Go-side store is chosen because the **portfolio/dashboard regime view** lives in the Go process and because it brings **flat `manual` + `options`** into the same regime source independent of the open-signal check — neither of which a leaf-worker `/tmp` cache serves cleanly.

**Cycle placement:** build/clear the map in the main loop **before Phase 3** (subprocess fan-out), after due strategies are known — same orchestration layer as risk assist / price fetch.

**Latency cost (to analyze):** this adds a new serial phase before the check fan-out — one regime subprocess per distinct `(symbol, interval, period)` raw signature, multiplied by multi-window strategies, all gated at `pythonSemaphore=4` with `scriptTimeout=30s`. Quantify the added per-cycle wall-clock before committing to the unconditional split, and confirm the regime subprocess's own OHLCV fetch lands in the #839 `/tmp` cache so the subsequent check reads it instead of re-fetching.

The ADX/efficiency math stays in Python (`shared_tools/regime.py`). The scheduler invokes a **dedicated regime subprocess once per distinct `(symbol, interval, period)` raw signature per cycle** (not once per strategy), gets back raw indicator values + both labels, stores them in the Go-side map, then resolves each strategy's label as a projection. Checks receive the precomputed payload via CLI (see **Injection / deploy** below) instead of calling `prepare_check_regime` on the same OHLCV.

### Why this works structurally

Both classifiers derive from the same underlying math:
- The 3-state ADX vocabulary is `{trending_up, trending_down, ranging}` (`VALID_LABELS_ADX`).
- The 7-state composite is `{trending_up_clean, trending_up_choppy, trending_down_clean, trending_down_choppy, ranging_quiet, ranging_volatile, ranging_directional}` (`VALID_LABELS_COMPOSITE`).
- `latest_regime_composite` already computes `adx` internally (via `compute_regime`) and returns it in `metrics.adx`, alongside `return_eff`, `range_eff`, `efficiency`.
- The 7 composite labels nest under the 3 ADX labels by name prefix, so a 3-state view is derivable from the same computation.

So one computation can emit the raw indicator values, the 3-state label, and the 7-state label.

### Bundle structure (sketch)

Keyed by `(symbol, interval, period)`, computed once per raw key each cycle, reused by peers. In-memory only (recomputed every cycle like `stratState.Regime`); not persisted:

```
RegimeBundle
  key:        { symbol, interval, period }
  bar_time:   closed-bar timestamp        // preserve look-ahead: gate reads N-1
  raw:
    adx                                    // computed at the FULL period (see caveat)
    plus_di, minus_di                      // for exact ADX-classifier direction
    return_eff, range_eff, efficiency, atr_pct
  labels (at DEFAULT thresholds):
    adx3:       trending_up | trending_down | ranging
    composite7: trending_up_clean | ... | ranging_quiet
```

### Signature, storage, per-cycle population (not load-time registry)

Two-layer global store (rebuilt **every cycle** — empty at loop start):
- **Raw layer** keyed by `(symbol, interval, period)` — the ADX/±DI + efficiency components. Shared by everything on that asset/interval/period regardless of classifier or thresholds.
- **Label layer** keyed by the full **regime signature** `(symbol, interval, period, classifier, thresholds)` — a resolved label derived as a cheap projection over the raw layer, computed once per cycle per signature. Threshold-only and classifier-only differences are projections over the same raw entry; only a different `period` forces a new raw computation.

**Per-cycle flow:** clear map → union distinct signatures from due strategies (+ `manual`) → for each missing raw key, one regime subprocess → project label-layer entries → serve reads.

**At load / SIGHUP (validation only):** `validateStrategyRegimeVocabulary` (and existing regime config validation) — **not** pre-populating the ephemeral store. Adding a strategy does not "register" a persistent entry; the next cycle's population step picks up its signature.

### Strategy resolution

- **Any params (default or custom)** → read the label for the strategy's signature from the global store for that cycle.
- Missing signature in the label layer → compute raw (if needed) + project once, then share across peers with the same signature.
- Threshold/classifier differences resolve as projections over the shared raw layer; only a distinct `period` adds a raw-layer entry.

### Failure policy (regime subprocess)

Must be explicit (today inline behavior is inconsistent):
- **Entry gate (`applyRegimeGate`):** empty/missing label → fail-open (do not block opens); matches `regimeAllowsEntry`.
- **Regime subprocess failure** (timeout, non-zero exit, bad JSON): **clear the live store entry to empty (`regime=-` on status) and let every consumer fall back to its existing empty-case behavior** — i.e. option (b). Rejected: (a) reuse-last (silent stale-state correctness risk — a strategy could open against a regime several cycles old during an outage, and the dashboard shows a fake-live label); (c) skip-check (over-couples an advisory overlay to the core signal path, regressing strategies that don't use regime — the check subprocess fetches its own OHLCV, so its failure is already handled by the dispatch `ok` guard independent of this decision). Log + throttled operator alert on sustained failure via the existing `scriptFailureTracker` pattern (`script_failure_alerts.go`).
  - **Why (b) needs no new stale-handling code:** open positions are unaffected because the regime-keyed life-of-position features (`*_atr_regime` SL, dynamic TP/SL re-resolution, directional policy) read the **frozen `pos.Regime`** stamped at open and, for #843, `pos.RegimeAppliedLabel` with confirm-cycle hysteresis — not the live store. So a transient miss yields "reuse-last" exactly where staleness is safe (an already-open position) and "fail-open" everywhere a fresh decision is made. The entry gate's empty → allow (`regimeBlocksOpen`), the directional policy's legacy fail-open, and the `regime=-` status are each already the correct empty-case behavior; no consumer needs a new branch.
  - **One semantic delta to accept (not behavior-neutral):** today regime and the open signal share one subprocess, so a regime-data failure also fails the whole check → the strategy **skips** the cycle. Post-split the check can succeed while regime is empty, so new opens fire **fail-open instead of skip**. This holds "no behavior change" for life-of-position features but is a real change for fresh opens. It is the right tradeoff (an advisory overlay should not block trading), stated here rather than left implicit.
- **Insufficient OHLCV for a raw key:** no bundle for that key; consumers for dependent signatures get empty labels (gate fail-open); do not invent a default `ranging` unless parity tests require it.

Parity goal: pre-migration and post-migration use the **same** policy for equivalent failures, with the single documented exception of the fail-open-vs-skip delta on fresh opens above.

### Injection / deploy cost

Full migration (Scope item 6) **cannot** be Go-only: check scripts still need regime inside Python for open params and live close evaluators (`strategy_params["regime"]`, `market_ctx["regime"]` in `check_hyperliquid.py`, etc.).

Therefore:
- Add CLI injection of the precomputed `RegimePayload` (or per-window labels + metrics) into **each** check script; when injected, skip `prepare_check_regime`.
- Append the new regime script to **`version_probe.go` `probeArgv`** (and keep Go+Python same-SHA deploy per `update.sh`).

**Phased alternative:** Phase 1 — Go consumers only (`applyRegimeGate`, `syncStrategyRegimeState`, `stampPositionRegimeIfOpened`, `strategyCurrentATRRegime`, `applyRegimeDirectionalPolicy`, dashboard); checks still compute until close/open paths accept injection. Phase 2 — remove inline computation everywhere.

### 3-state view: exact vs collapsed

Two ways to get the 3-state label off the bundle:
1. **Run the real ADX classifier on stored raw `adx`/`±DI`** → exact match to a standalone ADX strategy.
2. **Prefix-collapse the composite label** (`trending_up_*`→`trending_up`, etc.) → deterministic but can differ at the trending/ranging boundary.

Prefer (1) so existing ADX strategies don't shift behavior; keep (2) only as a documented fallback.

## Caveats / parity

- **`COMPOSITE_ADX_PERIOD_CAP = 14`**: composite internal ADX uses `min(period, 14)`. For exact 3-state ADX parity when `period > 14`, the bundle must compute ADX at the full period separately (or store both).
- **Look-ahead invariants**: entry gate reads prior-bar regime; closes use closed-bar ATR. Live `latest_regime_composite` and backtest `ensure_regime_columns` → `compute_regime_composite` share `_composite_efficiency_metrics`.
- **Multi-window**: each window resolves to its own signature/entry in the global store.
- **No behavior change for any strategy** — custom-param strategies get the same label they get today, sourced from the global calculator for that signature instead of inline.

## Scope of work

1. Define the global calculator + two-layer store **in the Go scheduler**; per-cycle map (clear → collect signatures of due strategies → compute → project). Dedicated regime subprocess (Python, `shared_tools/regime.py`), once per distinct raw signature per cycle; respect `pythonSemaphore`/`scriptTimeout`; read-only; **probe in `version_probe.go`**.
2. **Load-time validation only** (reuse `validateStrategyRegimeVocabulary`) — do not conflate with per-cycle population (see above).
3. Emit raw values + both labels from one composite pass; full-period ADX for exact 3-state view.
4. Wire strategy resolution from the global store; projection for threshold/classifier; new raw entry only on distinct `period`.
5. Surface the global store to operator display / dashboard as portfolio-level regime.
6. **Migrate every consumer** — no inline `prepare_check_regime` when injection is present:
   - **`syncStrategyRegimeState` dispatches** (six call sites in `main.go`: OKX spot/perps, RH spot, default spot, HL perps, TopStep futures): read global store for signature instead of `result.Regime` from check output.
   - **`options`**: `runOptionsCheck` does not call `applyRegimeGate` or `syncStrategyRegimeState`; regime is computed in `check_options.py` at hardcoded `REGIME_TIMEFRAME = "4h"` via `latest_regime` (3-state ADX). Today it **does** set `stratState.Regime = result.Regime` directly (`main.go:1518`) but without multi-window sync or per-strategy regime config on the wire. Migrate to global store (decide whether 4h-ADX default is the options signature).
   - **`manual`**: HL check only when open (`runManualCloseEval`); open-position `regime=-` fixed (#872). Migration: read bundle while **flat** so status/dashboard show regime without a position. Cycle-1 stamp lag is already fixed (#872/#878) — not in scope.
   - **Downstream readers**: `applyRegimeGate`, `stampPositionRegimeIfOpened`, `syncStrategyRegimeState`, `strategyCurrentATRRegime`, `applyRegimeDirectionalPolicy` — global store by signature.
   - **Check scripts**: accept injected regime; skip inline compute when present (probe argv update).
   - **Backtest**: `ensure_regime_columns` / `compute_regime_composite` (and ADX path) consume the same bundle semantics for parity.
7. Parity tests: `adx3` vs standalone ADX (incl. `period > 14`); collapse fallback documented; live/backtest composite parity; per-consumer label unchanged for default params pre/post migration.

## Coordination with #873 (scale-in / pyramiding)

No design conflict — both rest on the same invariant: **`pos.Regime` is frozen at first entry and never re-stamped.** #873 freezes the regime label on a scale-in add (its point 4); #879 changes only the *source* of the live label (global store vs inline `result.Regime`), not the stamp-once-when-empty semantics of `stampPositionRegimeIfOpened`, and keeps "`pos.Regime` stays the frozen-at-open stamp" as a non-goal. A scaled-in position's frozen TP/SL geometry reads `pos.Regime`, not the live store, so #879's failure policy (clear live store → empty) cannot disturb it.

Interaction is **mechanical only** — both edit the same `main.go` perps + manual dispatch blocks and the `stampPositionRegimeIfOpened` call sites, so whichever merges second rebases:
- If **#879 lands first:** #873's new "add" branch must thread the regime payload from the global store the way the open path does; since an add never re-stamps, it only needs to not regress the existing call.
- If **#873 lands first:** #879's dispatch rewrite must preserve the add branch alongside flat-open / flip / close.

Pre-existing tension neither issue creates: for `tiered_tp_atr_live_regime_dynamic` (#843), "frozen at open" (#873) and live re-resolution (#843) already disagree on whether mid-position regime can move; #879 only relocates the source, so confirming an add interacts cleanly with `RegimeAppliedLabel` hysteresis is #873's concern, not a #879 gap.

## Non-goals

- Removing per-strategy regime params — strategies keep classifier/period/thresholds; the global calculator stores every distinct signature.
- Persisting the store (recomputed each cycle; `pos.Regime` stays the frozen-at-open stamp).

---
Created with LLM: Opus 4.8 (1M) | high | Harness: Claude Code
Updated with LLM: Opus 4.8 (1M) | high | Harness: Claude Code
Updated with LLM: Composer 2.5 | default | Harness: Cursor
Updated with LLM: Opus 4.8 (1M) | xhigh | Harness: Claude Code

richkuo:
## Summary

Moves regime computation out of each strategy's per-cycle Python check subprocess into a single **Go-scheduler-owned global regime store**, computed once per distinct `(symbol, interval, windows-spec)` signature per cycle, injected into every check, and read directly by flat-`manual`, `options`, and the dashboard.

Closes #879.

## Why parity holds (no behavior change)

Regime windows live in **global** `cfg.Regime.Windows` (per-strategy fields are only *selectors*), so the dedup unit is `(symbol, interval)` + a fingerprint of the resolved windows spec. The subprocess computes each window with **its own classifier/period** via the existing `compute_multi_regime`, so labels are byte-identical to the pre-migration inline path for the same candles. An ADX window at `period=20` already uses full-period ADX; `COMPOSITE_ADX_PERIOD_CAP=14` only applies inside composite windows.

I deliberately **rejected the issue's "one composite pass → prefix-collapse for the 3-state view"** sketch — computing each window with its own classifier is exact and avoids the trending/ranging boundary drift that approach risked.

## Mechanism

1. New Go store (`regime_calculator.go`, `regime_run.go`): before Phase 3, collect distinct signatures of due strategies (+ flat `manual`), run one read-only subprocess (`fetch_regime.py`) per signature (concurrent, bounded by the shared `pythonSemaphore`), store the `RegimePayload`.
2. Each check dispatch **injects** the precomputed payload via `--regime-injected --regime-payload-json`; the check skips `prepare_check_regime` (new `resolve_injected_regime` projection helper) and **echoes the payload back** as `result.Regime`. Existing Go consumers (`applyRegimeGate`, `syncStrategyRegimeState`, `stampPositionRegimeIfOpened`, `strategyCurrentATRRegime`, `applyRegimeDirectionalPolicy`) are unchanged — only the *source* of `result.Regime` changes.
3. Flat `manual` reads the store while flat (regime visible without a position). `options` populates the store from its check result (advisory 4h-ADX, embedded in evaluators — store-from-result preserves exact options behavior). Dashboard `/api/regime` serves a per-signature portfolio view.

## Failure policy

A failed regime subprocess clears that signature to **present-but-empty** → every consumer falls back to its existing empty-case (fail-open) behavior. Open positions read the frozen `pos.Regime`/`pos.RegimeAppliedLabel`, never the live store, so a transient outage can't disturb them. Throttled operator alert via a regime-specific `ScriptFailureTracker` (separate key space from signal-script alerts).

**One documented semantic delta:** post-split a check can succeed while regime is empty, so a fresh open fires **fail-open instead of skip** on a regime-data outage (an advisory overlay should not block trading).

## Coordination with #873

No design conflict — `pos.Regime` stays the frozen-at-open stamp. Whichever merges second rebases the shared `main.go` perps/manual dispatch blocks. The manual path already threads the store payload through `runManualCloseEval` without re-stamping.

## Deploy safety

`fetch_regime.py` and the `--regime-injected`/`--regime-payload-json` flags are appended to `version_probe.go` (`probeArgv`, `probeCompositeArgv`, new `fetchRegimeProbeArgv`) so a stale Python fails startup (exit 78) instead of mid-cycle. Go+Python must deploy at the same SHA.

## Latency

The store build adds one serial phase before Phase 3: N distinct `(symbol, interval)` subprocesses, concurrent under `pythonSemaphore=4`. Each fetch lands in the #839 `/tmp` OHLCV cache so the subsequent check reads cache (no double network fetch). Added wall-clock ≈ `ceil(N_distinct/4) × (spawn + ADX compute)`, not `N_strategies`.

## Tests

- Go: store dedup/failure-clear/signature/snapshot, injected-args builder, `/api/regime` endpoint, regime-failure throttle. Full `go test ./...` green.
- Python: `fetch_regime.compute_payload` == `prepare_check_regime` (ADX, multi-window, full-period ADX > 14); `resolve_injected_regime` == `prepare_check_regime` 3-tuple; HL injected-vs-inline wiring; backtest parity (`fetch_regime` vs `ensure_regime_columns`/`compute_regime_composite`). Full pytest suite green (1169 + new).
- Smoke: `--once` with two paper HL strategies + regime enabled — store computed `trending_down` per signature, injected, echoed, `regime=trending_down` on status. No errors.

## Follow-ups (not in this PR)

- Per-script injected-vs-inline parity tests for okx/robinhood/topstep (wiring is byte-identical to HL, covered by shared `resolve_injected_regime` tests + probe).
- Dashboard UI surfacing of `/api/regime` (endpoint added; front-end widget deferred).
- CLAUDE.md/SKILL.md file-ownership entries for the new regime files.

---
Created with LLM: Opus 4.8 (1M) | high | Harness: Claude Code

### DIFF
--- a/backtest/tests/test_regime_bundle_parity.py
+++ b/backtest/tests/test_regime_bundle_parity.py
@@ -1,0 +1,65 @@
+"""#879: parity between the live regime subprocess (fetch_regime.compute_payload)
+and the backtest regime path (ensure_regime_columns / compute_regime_composite)
+for the same candles. Backtest code is unchanged by #879; this guards that the
+relocated live computation stays label-identical to the backtest.
+"""
+import importlib.util
+import pathlib
+import sys
+
+import numpy as np
+import pandas as pd
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(ROOT / "shared_tools"))
+sys.path.insert(0, str(ROOT))
+
+from regime import ensure_regime_columns  # noqa: E402
+
+
+def _load_fetch_regime():
+    spec = importlib.util.spec_from_file_location(
+        "fetch_regime", ROOT / "shared_scripts" / "fetch_regime.py")
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+def _df(n=160, slope=0.7):
+    close = np.linspace(100.0, 100.0 + slope * n, n)
+    return pd.DataFrame({
+        "timestamp": range(n), "open": close - 0.2, "high": close + 0.6,
+        "low": close - 0.6, "close": close, "volume": np.ones(n),
+    })
+
+
+def _last_label(df, **kwargs):
+    out = ensure_regime_columns(df.copy(), **kwargs)
+    return str(out["regime"].iloc[-1])
+
+
+def test_adx_bundle_matches_backtest_period14():
+    fr = _load_fetch_regime()
+    df = _df()
+    spec = {"w": {"classifier": "adx", "period": 14, "adx_threshold": 20}}
+    bundle = fr.compute_payload(df, spec)["w"]["regime"]
+    assert bundle == _last_label(df, period=14, adx_threshold=20)
+
+
+def test_adx_bundle_matches_backtest_full_period_over_14():
+    """period>14 must use the full ADX period in both paths (no COMPOSITE cap)."""
+    fr = _load_fetch_regime()
+    df = _df(200)
+    spec = {"w": {"classifier": "adx", "period": 30, "adx_threshold": 20}}
+    bundle = fr.compute_payload(df, spec)["w"]["regime"]
+    assert bundle == _last_label(df, period=30, adx_threshold=20)
+
+
+def test_composite_bundle_matches_backtest():
+    fr = _load_fetch_regime()
+    df = _df(180)
+    spec = {"w": {"classifier": "composite", "period": 30}}
+    bundle = fr.compute_payload(df, spec)["w"]["regime"]
+    backtest_label = _last_label(
+        df, windows_spec={"w": {"classifier": "composite", "period": 30}}, gate_window="w")
+    assert bundle == backtest_label
--- a/docs/superpowers/plans/2026-06-05-shared-regime-bundle.md
+++ b/docs/superpowers/plans/2026-06-05-shared-regime-bundle.md
@@ -1,0 +1,853 @@
+# Shared Regime Bundle (issue #879) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Move regime computation out of each strategy's per-cycle Python check subprocess into a single Go-scheduler-owned global regime store, computed once per distinct `(symbol, interval, windows-spec)` signature per cycle, injected into every check, and read directly by flat-`manual`, `options`, and the dashboard.
+
+**Architecture:** A per-cycle in-memory two-layer store in the Go scheduler (`regime_calculator.go`). Before Phase 3, the scheduler collects the distinct regime signatures of due strategies (+ flat `manual`/`options`), runs one dedicated read-only regime subprocess (`shared_scripts/fetch_regime.py`) per distinct signature (concurrent, gated by `pythonSemaphore`), and stores the resulting `RegimePayload` keyed by signature. Each check dispatch **injects** the precomputed payload into the check script via `--regime-injected --regime-payload-json <json>`; the check skips `prepare_check_regime` and echoes the payload back as `result.Regime`, so existing Go consumers (`applyRegimeGate`, `syncStrategyRegimeState`, `stampPositionRegimeIfOpened`, `strategyCurrentATRRegime`, `applyRegimeDirectionalPolicy`) are unchanged — the *source* of `result.Regime` is now the store. Flat `manual`, `options`, and the dashboard read the store directly. Failure policy: a failed regime subprocess clears that signature to empty; every consumer falls back to its existing empty-case (fail-open) behavior.
+
+**Why parity holds:** Regime windows live in **global** `cfg.Regime.Windows` (one spec for all strategies; per-strategy fields are only *selectors* — gate/atr/directional window, `allowed_regimes`, ATR-regime blocks). So the dedup unit is `(symbol, interval)` + a fingerprint of the resolved global windows spec. The subprocess computes each configured window with **its own classifier/period** via the existing `compute_multi_regime`/`classify_window` — identical to what each check computes inline today. An ADX window at `period=20` already uses full-period ADX; the `COMPOSITE_ADX_PERIOD_CAP=14` only applies *inside* composite windows. No single-pass prefix-collapse is used, so labels are byte-identical pre/post migration for the same candles.
+
+**Tech Stack:** Go (scheduler, `modernc.org/sqlite`), Python 3 (`shared_tools/regime.py`, platform adapters via `importlib`), subprocess contract (JSON stdout, exit 1 on error), systemd `PrivateTmp` `/tmp` OHLCV cache (#839).
+
+---
+
+## Key invariants (must hold across every task)
+
+1. **No behavior change for any strategy's label** for the same candles — inline-computed `result.Regime` (pre) == store-injected `result.Regime` (post). Enforced by parity tests.
+2. **`pos.Regime` stays the frozen-at-open stamp** (coordination with #873). `stampPositionRegimeIfOpened` semantics unchanged — only the payload source changes.
+3. **Failure = clear-to-empty (policy b).** Regime subprocess failure → store empty for that signature → consumers fall open. Open positions read frozen `pos.Regime`/`pos.RegimeAppliedLabel`, never the live store, so they are unaffected.
+4. **One documented semantic delta:** post-split a check can succeed while regime is empty, so a fresh open fires **fail-open instead of skip** on a regime-data outage. Accepted.
+5. **Look-ahead unchanged:** the subprocess computes on the same closed candles the check would fetch; gate reads the resolved label exactly as today.
+6. **Go+Python same-SHA deploy:** every new required CLI flag is appended to `version_probe.go`.
+
+---
+
+## File Structure
+
+**New files:**
+- `scheduler/regime_calculator.go` — `RegimeSignature`, `RegimeStore`, signature collection, per-cycle build orchestration, store reads.
+- `scheduler/regime_calculator_test.go` — store dedup, failure-clear, signature collection, payload reads.
+- `shared_scripts/fetch_regime.py` — platform-aware OHLCV fetch + `compute_multi_regime` + emit payload JSON. Read-only.
+- `shared_scripts/test_fetch_regime.py` — subprocess output shape + parity vs `prepare_check_regime`.
+
+**Modified files:**
+- `scheduler/main.go` — build store before Phase 3; inject store payload into each `runXxxCheck`; flat-`manual` + `options` read store; dashboard wiring.
+- `scheduler/strategy_composition.go` — `appendInjectedRegimeArgs(args, payload)` builder; keep `appendRegimeArgs` for the windows-spec the subprocess uses.
+- `scheduler/regime_run.go` (new, or in regime_calculator.go) — `runRegimeSubprocess(platform, symbol, interval, specJSON) (RegimePayload, error)`.
+- `scheduler/version_probe.go` — probe `fetch_regime.py`; probe `--regime-injected`/`--regime-payload-json` on each check script.
+- `scheduler/script_failure_alerts.go` — reuse `scriptFailureTracker` for regime subprocess failures (per-signature key).
+- `shared_scripts/check_hyperliquid.py`, `check_strategy.py`, `check_options.py`, `check_okx.py`, `check_robinhood.py`, `check_topstep.py` — accept `--regime-injected`/`--regime-payload-json`; skip `prepare_check_regime` when injected; echo payload back.
+- `shared_tools/regime.py` — `resolve_injected_regime(payload, *, primary_key, atr_window)` helper returning `(stdout_payload, live_atr_label, strategy_params_snapshot)` from an injected multi-window payload (mirrors `prepare_check_regime`'s return contract without recomputing).
+- `scheduler/ui_server.go` / dashboard — portfolio-level regime view from the store.
+- `backtest/` — **no functional change**; add parity test only.
+
+---
+
+## Task sequence (build order)
+
+Phase A (Tasks 1-5): Go store + subprocess + injection helper, fully tested in isolation.
+Phase B (Tasks 6-9): wire HL as the reference platform end-to-end (inject + echo + store reads), parity test.
+Phase C (Tasks 10-13): replicate injection to spot/okx/robinhood/topstep; options; flat-manual.
+Phase D (Tasks 14-16): dashboard view; backtest parity test; probe argv + smoke + full parity matrix.
+
+---
+
+### Task 1: Regime signature + store types (Go)
+
+**Files:**
+- Create: `scheduler/regime_calculator.go`
+- Test: `scheduler/regime_calculator_test.go`
+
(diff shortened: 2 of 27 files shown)


richkuo:
@claude

richkuo:
Thanks — fixed in `01941cf`.

1. **Options overwrite / signature namespacing** — added `Platform` + `Kind` to `RegimeSignature`. Bundle entries are `Kind="bundle"`, the advisory options store-from-result is `Kind="options"`, so an options strategy on underlying BTC at 4h can never clobber a real BTC/4h bundle that a later-dispatched peer reads. Regression test `TestOptionsSignatureCannotClobberBundle`.

1. **Cross-platform collision** — the new `Platform` field (from `regimePlatformForStrategy`) is part of the key, so two exchanges' same-symbol/interval strategies no longer collapse to one signature / wrong-exchange candles. Regression test `TestRegimeSignatureIncludesPlatform`. Also folded `Platform`/`Kind` into `regimeSignatureKey` for the failure-tracker.

1. **`snapshot()` primary window** — `snapshot` now takes `*RegimeConfig` and threads it to `PrimaryLabel`, so the dashboard top-level label honors the configured `medium` instead of the lexicographically-first window. Call site passes `cfg.Regime`. Regression test `TestSnapshotHonorsPrimaryWindow`.

On the two informational items:

- **Non-HL redundant fetch / one-bar skew** — correct, the "lands in the #839 cache, no double fetch" claim holds only for Hyperliquid; OKX/RH/TopStep/binanceus fetch candles twice and can carry up to a one-bar skew between the injected regime and the signal bar. Low impact (regime is advisory and frozen-at-open over a 14+ bar ADX window), and closing it means extending the per-instance `/tmp` OHLCV cache to those adapters — out of scope here. Noting it as a known limitation rather than fixing in this PR.
- **Fail-open on the hard `allowed_regimes` gate** — intended. It's the issue's explicit option (b) and matches the pre-existing `regimeAllowsEntry` empty→allow behavior (a strategy with regime detection disabled already opens in any regime today). So a regime-subprocess outage degrades the hard entry filter to the same posture as "regime unavailable" already had — an advisory overlay shouldn't hard-block trading. Confirmed as the intended risk posture.

Full `go test ./...` + Python regime/parity suites green after the fixes.

richkuo:
@claude rereview

PR #910, opened 2026-06-09
feat(regime): global per-cycle regime store with dedicated subprocess + payload injection

### LINKED ISSUE #879: feat(regime): shared regime bundle per (asset, interval, period) with 3-state + 7-state views and per-strategy override

richkuo:
Closes #879

## What

A single global regime calculator now owns all regime computation. Each cycle the Go scheduler builds a fresh in-memory store — one dedicated regime subprocess per distinct `(data platform, symbol, timeframe, windows-spec)` signature among **due** strategies — and every consumer reads that map. Check scripts no longer compute regime inline: the bundle is injected via a new `--regime-payload-json` flag.

## Design notes vs the issue sketch

- **Signature space collapsed:** classifier/period/thresholds live on the global `regime.windows` config, not per strategy (strategies only select windows), so the two-layer raw/label split degenerates: one bundle per `(platform, symbol, timeframe, spec)` carries every window's snapshot, and per-strategy "projection" is just window selection (`RegimePayload.Label`). The spec string stays in the key so the options ADX/4h signature coexists with the global one.
- **Data-source keying:** the key's platform is the *data source the check script fetches from* (default spot dispatch = BinanceUS regardless of `sc.Platform`), so the bundle is computed on exactly the frame the strategy trades on.
- **3-state view = exact (option 1):** `views.adx3` runs the real ADX classifier at the window's **full** period (exact past `COMPOSITE_ADX_PERIOD_CAP`); `views.composite7` is the complementary view. Consumer labels still come from the configured classifier's snapshot — unchanged by migration (regression-tested).
- **Full migration, not Phase 1:** injection landed in all five check scripts + `check_options.py`, so there is no double computation and `result.Regime` echoes the store payload; the six dispatch gate/sync sites additionally read the store directly and point `result.Regime` at it for the stamp-at-open path.

## Failure policy (option b, as specified)

Failed/missing bundle → no store entry → empty payload everywhere: entry gate fails open, `regime=-` on status, check scripts receive an **empty** `--regime-payload-json` (skip inline compute, resolve empty) — no reuse-last, no inline fallback. Open positions are untouched (`pos.Regime` frozen stamp). Sustained failures alert through the existing `scriptFailureTracker` throttling keyed `regime[<platform>/<symbol>/<tf>]`. **Accepted semantic delta** (stated in the issue): a regime-data failure no longer fails the whole check, so fresh opens fail open instead of skipping the cycle.

## Consumer migration

- Six `syncStrategyRegimeState` dispatches + `applyRegimeGate` + `stampPositionRegimeIfOpened`: store-backed.
- `manual`: flat strategies now show a live regime (pre-#879: `regime=-` without a position); the close-eval no longer returns a regime payload.
- `options`: migrated to the store's `(underlying, 4h, ADX-default)` signature — same hardcoded params as the old inline path, still **not** gated on `regime.enabled`; the injected payload keeps the script's emitted label identical.
- `strategyCurrentATRRegime` / `applyRegimeDirectionalPolicy`: read `stratState` / the echoed-injected payload, both store-sourced now.
- Backtest: no change needed — live and backtest already share `_composite_efficiency_metrics` / `compute_regime`; the bundle uses the same `compute_multi_regime` entry point (parity-tested).

## Latency

The new phase runs before the fan-out, parallel across signatures under `pythonSemaphore=4`, so added wall-clock ≈ one subprocess generation per cycle (peers dedupe to one). For HL, the bundle's OHLCV fetch lands in the #839 `/tmp` cache with the **same limit** the check scripts use (`regimeRequiredOhlcvLimit`), so the subsequent check reads the cache instead of re-hitting `/info` — regime-enabled HL deployments net out to roughly the same `/info` load with strictly less duplicated math.

## Surfaces

- `GET /api/regime`: portfolio-level view — per-signature windows with snapshots + both-vocabulary views, `bar_time`, `built_at`.
- One `Regime: <platform>/<sym>/<tf>=<label>; …` line per cycle in the scheduler log.
- Probe: `check_regime.py` probed at startup; `--regime-payload-json` added to the signal probe argvs (stale-Python asymmetric deploys fail at startup, exit 78).

## Tests

- Go: signature derivation per type/platform (incl. manual, options, luno-on-binanceus-data), peer dedupe, one-compute-per-signature, failure→empty→gate-fails-open, no reuse-last across cycles, injection flag presence semantics, output parsing (raw bytes preserved), probe counts, dashboard projection.
- Python: injected-vs-inline triple parity (multi-window, primary/atr selection), empty-injection fail-open, inline-skip proof (df=None), bundle snapshots == `compute_multi_regime`, full-period `adx3` parity at period>14, end-to-end bundle→injection round-trip.
- Suites: `go vet` + `go test ./...` green; pytest 1172 passed + `shared_scripts/` 148 passed. Live smoke: `check_regime.py` against BinanceUS/HL/OKX; injected + empty payloads through `check_strategy.py`, `check_hyperliquid.py`, `check_okx.py`.

---

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -746,6 +746,13 @@ func main() {
 		if saveFailures >= 3 {
 			fmt.Println("[CRITICAL] State save failed 3x, skipping trades this cycle")
 		} else {
+			// #879: build the per-cycle global regime store BEFORE the check
+			// fan-out — one regime subprocess per distinct (platform, symbol,
+			// timeframe, spec) signature among due strategies. Every regime
+			// consumer this cycle (entry gates, stratState sync, stamp-at-open,
+			// check-script injection, flat manual, options, dashboard) reads
+			// this map; check scripts no longer compute regime inline.
+			populateRegimeStore(globalRegimeStore, dueStrategies, cfg.Regime, notifier)
 			// #42 / #243: Portfolio-level risk check before running any strategy.
 			//
 			// Fetch live Hyperliquid clearinghouseState ONCE per cycle (outside
@@ -1455,12 +1462,17 @@ func main() {
 						if sc.Platform == "okx" {
 							if result, signalStr, price, ok := runOKXCheck(sc, prices, okxPosCtx, cfg.Regime, notifier, logger); ok {
 								prices[result.Symbol] = price
-								if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, okxPosQty); regimeBlocked {
+								// #879: single-source regime — read the global store for this
+								// strategy's signature instead of the check output, and point
+								// result.Regime at it so stamp-at-open inside execute* shares it.
+								storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+								result.Regime = &storeRegime
+								if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, okxPosQty); regimeBlocked {
 									logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 									result.Signal = 0
 								}
 								mu.Lock()
-								syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+								syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 								mu.Unlock()
 								var execResult *OKXExecuteResult
 								liveExecFailed := false
@@ -1480,12 +1492,17 @@ func main() {
 						} else if sc.Platform == "robinhood" {
 							if result, signalStr, price, ok := runRobinhoodCheck(sc, prices, rhPosCtx, cfg.Regime, notifier, logger); ok {
 								prices[result.Symbol] = price
-								if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, rhPosQty); regimeBlocked {
+								// #879: single-source regime — read the global store for this
+								// strategy's signature instead of the check output, and point
+								// result.Regime at it so stamp-at-open inside execute* shares it.
+								storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+								result.Regime = &storeRegime
+								if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, rhPosQty); regimeBlocked {
 									logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 									result.Signal = 0
 								}
 								mu.Lock()
-								syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+								syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 								mu.Unlock()
 								var execResult *RobinhoodExecuteResult
 								liveExecFailed := false
@@ -1503,19 +1520,29 @@ func main() {
 								}
 							}
 						} else if result, signalStr, price, ok := runSpotCheck(sc, prices, spotPosCtx, cfg.Regime, notifier, logger); ok {
-							if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, spotPosCtx.Quantity); regimeBlocked {
+							// #879: single-source regime — read the global store for this
+							// strategy's signature instead of the check output, and point
+							// result.Regime at it so stamp-at-open inside execute* shares it.
+							storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+							result.Regime = &storeRegime
+							if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, spotPosCtx.Quantity); regimeBlocked {
 								logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 								result.Signal = 0
 							}
 							mu.Lock()
-							syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+							syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 							trades, detail = executeSpotResult(sc, stratState, stateDB, result, signalStr, price, cfg.Regime, logger)
 							mu.Unlock()
 						}
 					case "options":
 						if result, signalStr, ok := runOptionsCheck(sc, posJSON, notifier, logger); ok {
+							// #879: options regime now comes from the global store's
+							// (underlying, 4h, ADX-default) bundle instead of the
+							// check script's inline fetch; the injected payload keeps
+							// the script's own emitted label identical.
+							optionsRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
 							mu.Lock()
-							stratState.Regime = result.Regime
+							stratState.Regime = optionsRegime.PrimaryLabel(nil)
 							var harvestDetails []string
 							trades, detail, harvestDetails = executeOptionsResult(sc, stratState, result, signalStr, logger)
 							mu.Unlock()
@@ -1528,12 +1555,17 @@ func main() {
 						if sc.Platform == "okx" {
 							if result, signalStr, price, ok := runOKXCheck(sc, prices, okxPosCtx, cfg.Regime, notifier, logger); ok {
 								prices[result.Symbol] = price
-								if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, okxPosQty); regimeBlocked {
+								// #879: single-source regime — read the global store for this
+								// strategy's signature instead of the check output, and point
+								// result.Regime at it so stamp-at-open inside execute* shares it.
+								storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+								result.Regime = &storeRegime
+								if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, okxPosQty); regimeBlocked {
 									logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 									result.Signal = 0
 								}
 								mu.Lock()
-								syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+								syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 								mu.Unlock()
 								var execResult *OKXExecuteResult
 								liveExecFailed := false
@@ -1552,12 +1584,17 @@ func main() {
 							}
 						} else if result, signalStr, price, ok := runHyperliquidCheck(&sc, prices, hlPosCtx, cfg.Regime, notifier, logger); ok {
 							prices[result.Symbol] = price
-							if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, hlPosQty); regimeBlocked {
+							// #879: single-source regime — read the global store for this
+							// strategy's signature instead of the check output, and point
+							// result.Regime at it so stamp-at-open inside execute* shares it.
+							storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+							result.Regime = &storeRegime
+							if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, hlPosQty); regimeBlocked {
 								logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 								result.Signal = 0
 							}
 							mu.Lock()
-							syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+							syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 							mu.Unlock()
 							var execResult *HyperliquidExecuteResult
 							liveExecFailed := false
@@ -1813,12 +1850,17 @@ func main() {
 					case "futures":
 						if result, signalStr, price, ok := runTopStepCheck(sc, prices, tsPosCtx, cfg.Regime, notifier, logger); ok {
 							prices[result.Symbol] = price
-							if gateRegime, regimeBlocked := applyRegimeGate(sc, regimePayloadValue(result.Regime), cfg.Regime, tsContracts); regimeBlocked {
+							// #879: single-source regime — read the global store for this
+							// strategy's signature instead of the check output, and point
+							// result.Regime at it so stamp-at-open inside execute* shares it.
+							storeRegime := globalRegimeStore.PayloadForStrategy(sc, cfg.Regime)
+							result.Regime = &storeRegime
+							if gateRegime, regimeBlocked := applyRegimeGate(sc, storeRegime, cfg.Regime, tsContracts); regimeBlocked {
 								logger.Info("Regime gate: open signal blocked (regime=%s)", gateRegime)
 								result.Signal = 0
 							}
 							mu.Lock()
-							syncStrategyRegimeState(stratState, regimePayloadValue(result.Regime), cfg.Regime)
+							syncStrategyRegimeState(stratState, storeRegime, cfg.Regime)
 							mu.Unlock()
 							var execResult *TopStepExecuteResult
 							liveExecFailed := false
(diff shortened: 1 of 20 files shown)


richkuo:
@claude

richkuo:
Addressed all four review points in 68e79b8:

1. **Critical path:** `populateRegimeStore` is now `startRegimeStorePopulation` + a wait func. Population runs concurrently with the portfolio risk / kill-switch phase, and the wait sits immediately before the check fan-out (the first store consumer), bounded by `regimeStorePhaseBudget = scriptTimeout + 15s`. On budget exhaustion the store is **sealed** — straggler results are discarded (no mid-fan-out flips) and those signatures fail open for the cycle. A regime hang can no longer delay risk management, and a timeout storm costs at most one budget instead of `ceil(N/4)×30s`.
2. **Manual stamp window:** documented in the manual dispatch block — bundle failure on the open cycle leaves the stamp an empty no-op until a later successful cycle; idempotent on `pos.Regime == ""`, self-heals (fail-open tradeoff).
3. **Notifier concurrency:** failure/recovery alerts now collect per-request errors and fire sequentially after the parallel wave — populate goroutines no longer call `SendToAllChannels`/`SendOwnerDM` concurrently.
4. **Stale dashboard on `saveFailures >= 3`:** that branch now clears the store so `/api/regime` doesn't serve the prior cycle's labels as fake-live while trading is suspended.

New tests: set-after-seal discard, reset-unseals, phase-budget seal with a hung signature (fast peer lands, straggler stays discarded after release); regime tests pass under `-race`, full suites green.

---

richkuo:
@claude rereview

richkuo:
Fixed in a9c47bf — the cross-cycle straggler leak is closed, and I took the optional semaphore-overhang fix too since the same mechanism covers both:

1. **Generation guard:** `resetForCycle` now bumps and returns a generation; population captures it at start and `set()` drops any write whose generation differs from the store's current one. A budget-exceeded cycle N straggler that succeeds after cycle N+1's reset is discarded — `TestRegimeStoreStaleGenerationWriteDropped` is the regression.
2. **Seal-time cancellation (your optional note):** population runs under a context derived from `shutdownReadOnlyCtx` that is cancelled at the budget seal (and on normal completion). In-flight stragglers are SIGKILLed and queued ones fast-fail on the dead context, so prior-cycle regime work no longer holds `pythonSemaphore` slots into the next cycle's fan-out. Budget-cancelled signatures still record toward the 3-strike streak — chronic starvation should surface — with the message naming the phase-budget seal as the cause rather than a generic `context canceled`.

On the alert-ordering note: agreed it's concurrent with the main loop in the budget path and safe per the `snapshotBackends` RWMutex; the commit message wording is now scoped to "off the population goroutine only".

Regime tests pass under `-race`; full Go suite green.

---

richkuo:
@claude rereview

PR #911, opened 2026-06-09
fix(main): reuse shared-wallet-adjusted totalPV in cycle summary log

### LINKED ISSUE #908: Shared-wallet cycle summary logs double-counted total portfolio value instead of using shared-wallet-adjusted total
## Problem

In shared-wallet live setups, the cycle summary `Total value` log can report a materially incorrect number because it sums per-strategy `PortfolioValue(...)` instead of using the already-computed shared-wallet-adjusted total from `computeTotalPortfolioValue(...)`.

## Why this is wrong

For shared-wallet platforms like Hyperliquid, per-strategy cash is virtual by design. Summing per-strategy PV can double-count virtual cash and diverge from the actual exchange wallet value.

During live reconciliation on June 8, 2026, the exchange account was flat with `accountValue = 0.0`, but the cycle summary still logged a much larger total because it used the naive sum path.

## Affected code

- `scheduler/main.go`
- The cycle already computes `totalPV, usedPVFallback := computeTotalPortfolioValue(...)`
- Later, the summary log recomputes `totalValue` by summing `PortfolioValue(s, prices)` across strategies instead of reusing `totalPV`

## Expected behavior

The cycle summary should use the same shared-wallet-adjusted portfolio value that was already computed earlier in the cycle.

## Actual behavior

The summary log reports an inflated or misleading total in shared-wallet mode.

## Suggested fix

- Thread the previously computed `totalPV` through to the summary logger.
- Only fall back to per-strategy summation if the shared-wallet-adjusted value was not computed.

## Notes

This is a logging / operator-visibility bug, but it makes live reconciliation much harder because the bot appears to disagree with the exchange even when positions are already reconciled.

richkuo:
## Summary

- The cycle summary `Total value` log was summing per-strategy `PortfolioValue()` instead of reusing the `totalPV` already computed by `computeTotalPortfolioValue()` earlier in the cycle.
- In shared-wallet setups (e.g. Hyperliquid), per-strategy cash is virtual — naive summation double-counts it and diverges from the real exchange balance.
- Hoist `totalPV` to the outer loop scope so `logMgr.LogSummary` uses the shared-wallet-adjusted value. `saveFailures >= 3` path remains consistent (totalPV stays zero when trades are skipped).

## Test plan

- [ ] `go test ./...` passes (verified locally)
- [ ] On a shared-wallet HL setup, cycle summary `Total value` matches `computeTotalPortfolioValue` output, not the naive per-strategy sum
- [ ] On a non-shared-wallet setup, behavior is unchanged (single strategy, no virtual cash split)

Closes #908

---
Created with LLM: Sonnet 4.6 | high | Harness: Claude Code

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -742,6 +742,12 @@ func main() {
 			fmt.Println()
 		}
 
+		// totalPV holds the shared-wallet-adjusted portfolio value computed during
+		// the portfolio risk check; it is reused in the cycle summary log below
+		// so the summary doesn't double-count virtual cash in shared-wallet
+		// setups (#908). Zero when saveFailures >= 3 (trades skipped).
+		var totalPV float64
+
 		// Process only due strategies
 		if saveFailures >= 3 {
 			fmt.Println("[CRITICAL] State save failed 3x, skipping trades this cycle")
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

PR #912, opened 2026-06-09
[codex] Enrich portfolio warning payload

### LINKED ISSUE #904: Portfolio warning: enrich payload with per-strategy breakdown, time-in-band, trend, and recent activity
## Problem

The portfolio drawdown warning (`CheckPortfolioRisk` in `scheduler/risk.go`, emitted at `main.go:1095`) currently sends a single-line reason that aggregates equity/margin drawdown with the kill-switch limit and a couple of dollar values, then prefixes it with `**PORTFOLIO WARNING**`. Real example from `risk.go:510-514`:

```
portfolio drawdown approaching kill switch limit 25.0% (warn at 15.0%): equity=16.5% (value=$8400.00, peak=$10060.00); perps margin=18.2% (unrealized loss=$250.00, margin=$1500.00)
```

That's enough to know you're in trouble but not enough to act on it. To triage a warning an operator typically needs to answer:

1. **Which strategies are pulling the portfolio down** (per-strategy unrealized P&L and drawdown).
2. **What positions are still open and exposed** (id, side, size, current P&L).
3. **How long the portfolio has been in the warn band** (when did `WarningSent` flip on, or first time the band was entered).
4. **How much room is left before the kill switch** (current DD vs `max_drawdown_pct` — "you're 9.5% away from forced close").
5. **Direction** — is the drawdown still deepening or already recovering? A warning that's getting better shouldn't read the same as one that's getting worse.
6. **What was the last activity** — recent fills, recent manual opens, recent close-strategy triggers, so the operator can correlate the bleed to a specific event.

Without these, every warn-band cycle sends a notification that reads as "something is bad" and forces the operator to SSH in and dig through state to find out what.

## Proposed change

Expand the warning payload (the `reason` returned by `CheckPortfolioRisk` plus the message at `main.go:1095`) to a multi-line block with the following sections. Use the data already available in `state.PortfolioRisk`, per-strategy state, and `PortfolioWarnBandEntered`:

```
**PORTFOLIO WARNING** — BTC-30m trending_down_clean (dd=8.2%) dragged portfolio into warn band
Kill switch: 25.0% drawdown | Warn threshold: 15.0% | In band since: 2026-06-06 05:47 UTC (18m)
Current: equity=16.5% ($8400 / peak $10060) | perps margin=18.2% ($250 loss on $1500 margin)
Distance to kill switch: 8.5% (equity) / 6.8% (margin)
Trend: WORSENING — equity dd +1.2% over last 5m; margin dd +0.8%

Top contributors:
  hl-btc-sma-30   P&L -$420  dd 9.1%  pos: short 0.5 BTC @ $67,800  (-$140 unrealized)
  hl-btc-bb-30    P&L -$185  dd 7.4%  pos: long 0.3 BTC @ $67,950   (-$45 unrealized)
  hl-eth-ema-30   P&L -$95   dd 4.1%  (flat)

Recent activity (last 15m):
  05:51  fill  hl-btc-sma-30  short 0.5 BTC @ $67,800  (signal flip)
  05:48  warn  equity=15.3% (band entered)

Recommended: review open positions above; consider manual close on hl-btc-sma-30 if signal doesn't recover within next cycle.
```

Keep the single-line `reason` for callers that just need a quick summary (e.g. log lines, test assertions), but build the multi-line `warnMsg` from the same fields plus per-strategy and recent-event lookups. A Discord `SendToAllChannels` + `SendOwnerDM` with the full block is fine — Discord handles multi-line just as well as single-line.

## Implementation notes

- **Per-strategy breakdown**: `state.Strategies` already has per-strategy P&L; you need a helper that takes a snapshot at warning time (not at cycle start) so the numbers in the warning match the moment of the alert.
- **Time in band**: persist `warnBandEnteredAt` on `PortfolioRiskState` (set on the transition into the warn band, cleared on recovery, like the existing `WarningSent` flag). Not currently tracked.
- **Distance / trend**: compute from the same snapshot vs the previous warning cycle's snapshot. The kill-switch event log already records each band entry — you can reconstruct the trend from there without new state.
- **Recent activity**: pull the last N rows from `trades` (or whichever table logs per-strategy fills) for the warning window. Cap at ~5 entries.
- **Recommendation**: start with a static rule (e.g. "if any single strategy contributes > 50% of the drawdown, surface it as a candidate for manual close"). Skip the LLM — keep it deterministic.

## Considerations

- **Message size**: Discord's 2000-char limit. With 5 contributors + recent activity, a typical warning should land around 800-1200 chars. Cap per-section to be safe.
- **Not every cycle has all the data**: if no positions are open or no recent fills, omit those sections rather than rendering empty.
- **Owner DM vs channels**: send the same block to both — operators triage from DMs but channel history is the audit trail.
- **Test surface**: existing `risk_test.go` covers `CheckPortfolioRisk` return values. Add a test that asserts the multi-line `warnMsg` builder includes the per-strategy, time-in-band, and trend sections when those fields are populated.
- **Backwards compat**: the existing `reason` string is consumed by callers that just need a one-liner. Keep that signature; the multi-line block is a new field on the notifier payload (or a new helper that the main loop calls when constructing `warnMsg`).
- **Don't notify twice**: the existing dedup-via-`WarningSent` was relaxed in #420 (warn every cycle). The new payload should also fire every cycle, not just on band entry — operators want the trending data, which only updates per cycle.

## Acceptance criteria

- `CheckPortfolioRisk` (or a sibling helper called from `main.go:1095`) builds a multi-line `warnMsg` that includes: warning band threshold + in-band duration, current equity/margin drawdown, distance to kill switch, per-strategy P&L/drawdown for top contributors, recent activity, and a direction (worsening/recovering).
- `warnBandEnteredAt` is added to `PortfolioRiskState`, set on band entry, cleared on recovery, and rendered in the message.
- `notifier.SendToAllChannels(warnMsg)` and `notifier.SendOwnerDM(warnMsg)` at `main.go:1095` use the new block instead of the single-line reason.
- `risk_test.go` has a test asserting the multi-line block contains the expected sections given a populated `state.Strategies` and recent `trades` rows.
- `warnMsg` stays under 2000 chars in realistic cases (test with 5 strategies + 5 recent fills).
- Existing `reason` return value is preserved for callers that depend on the one-liner.

richkuo:
Fixes #904

## Summary
- Add persisted portfolio warning band timing and per-cycle trend deltas
- Send Discord portfolio warnings as a multi-line operator triage block with top contributors, open exposure, distance to kill switch, and recent fills
- Add recent-trade lookup and tests for persistence and warning formatting

## Tests
- env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test ./...

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -192,7 +192,12 @@ CREATE TABLE IF NOT EXISTS portfolio_risk (
     current_margin_drawdown_pct REAL NOT NULL DEFAULT 0,
     kill_switch_active INTEGER NOT NULL DEFAULT 0,
     kill_switch_at TEXT NOT NULL DEFAULT '',
-    warning_sent INTEGER NOT NULL DEFAULT 0
+    warning_sent INTEGER NOT NULL DEFAULT 0,
+    warn_band_entered_at TEXT NOT NULL DEFAULT '',
+    last_warning_equity_dd_pct REAL NOT NULL DEFAULT 0,
+    last_warning_margin_dd_pct REAL NOT NULL DEFAULT 0,
+    warning_equity_delta_pct REAL NOT NULL DEFAULT 0,
+    warning_margin_delta_pct REAL NOT NULL DEFAULT 0
 );
 
 CREATE TABLE IF NOT EXISTS kill_switch_events (
@@ -301,6 +306,12 @@ func (sdb *StateDB) migrateSchema() error {
 		// Portfolio margin drawdown + kill-switch source tracking (#296 review).
 		"ALTER TABLE portfolio_risk ADD COLUMN current_margin_drawdown_pct REAL NOT NULL DEFAULT 0",
 		"ALTER TABLE kill_switch_events ADD COLUMN source TEXT NOT NULL DEFAULT ''",
+		// Portfolio warning diagnostics (#904).
+		"ALTER TABLE portfolio_risk ADD COLUMN warn_band_entered_at TEXT NOT NULL DEFAULT ''",
+		"ALTER TABLE portfolio_risk ADD COLUMN last_warning_equity_dd_pct REAL NOT NULL DEFAULT 0",
+		"ALTER TABLE portfolio_risk ADD COLUMN last_warning_margin_dd_pct REAL NOT NULL DEFAULT 0",
+		"ALTER TABLE portfolio_risk ADD COLUMN warning_equity_delta_pct REAL NOT NULL DEFAULT 0",
+		"ALTER TABLE portfolio_risk ADD COLUMN warning_margin_delta_pct REAL NOT NULL DEFAULT 0",
 		// Per-leaderboard-summary last-post timestamps stored as JSON (#308).
 		"ALTER TABLE app_state ADD COLUMN last_leaderboard_summaries TEXT NOT NULL DEFAULT ''",
 		// Per-channel regular summary last-post timestamps stored as JSON (#474).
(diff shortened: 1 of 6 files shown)


richkuo:
@claude

richkuo:
@claude sonnet rereview

richkuo:
@claude rereview

richkuo:
@claude sonnet review

PR #916, opened 2026-06-09
feat(regime): detect medium vs short composite divergence + trust-short override

### LINKED ISSUE #907: feat(regime): detect medium vs short composite divergence + trust-short override
## Problem

Composite (7-state) regime can be evaluated at multiple windows — typically a "medium" window (e.g. `composite_medium`) used for open/position policy and a "short" window (e.g. `composite_short`) used for tactical/gate logic. **Which window feeds which feature is per-strategy config**, not a fixed system behavior: the `regime_directional_window` / `regime_gate_window` selectors point each feature at a window (`scheduler/regime_multi_window.go:226-246`). The directional policy resolves a **single** window label per cycle (`scheduler/main.go:2872` → `regimeDirectionalLabel`), and **there is no logic that compares two windows** — divergence between short and medium is invisible to the system today.

Concrete recent case (2026-06-08, cycle 80609):
- BTC and ETH both put in a swing low 3 days ago (~Jun 5).
- Both have been trending **up** for the past 3 days.
- ETH at $1691.75 is back near where the downtrend started.
- Yet `hl-rmc-eth-live` and `hl-tema-eth-live` are still resolving `regime=trending_down` on `composite_medium` — and worse, `effective_policy_regime=ranging_volatile`, so the directional policy is keeping `direction=both`, which means the strategies can still short into a tape that's been green for 3 sessions.
- Manual-eth just had a manual short that lived through the entire bounce because the medium composite never flipped.

(The cycle-80609 state above is an operator/runtime anecdote — the config behavior it describes, `ranging_volatile → direction=both`, is a policy-entry choice, not a code default.)

The medium window is doing what it's designed to do (lag = less whipsaw), but **the absence of any "the short window has flipped" awareness is a real blind spot.** Right now we'd have to wait for the medium to catch up, and by then half the move is gone.

## Proposal

Treat the short composite as a **leading indicator** for the medium. When they disagree, surface that disagreement and use it as a transition signal.

Three pieces, in order of value:

### 1. `regime_window_divergence` config (per strategy)

For each strategy that has both windows configured, expose a new field:

```json
"regime_window_divergence": {
  "short_window": "composite_short",
  "medium_window": "composite_medium",
  "on_divergence": "trust_short"   // or "trust_medium" | "alert_only"
}
```

The check script already evaluates **all** configured windows and returns every label in one subprocess call (multi-window `RegimePayload.Windows`); #879's global regime store dedupes that across peers. The divergence logic therefore only needs to **read two existing window labels** and, on each cycle, compare them. A divergence event is:
- `short_label != medium_label`, AND
- the labels are **incompatible** (not just sub-state differences). E.g. `trending_down` vs `trending_up_*` is a hard flip. `trending_down_choppy` vs `ranging_volatile` is a soft divergence (still both bearish-ish). Define a small compatibility matrix.

### 2. Divergence-driven direction override (opt-in, opt-out friendly)

When `on_divergence: "trust_short"`:
- If the medium says `trending_down` but the short says `trending_up_*` (or `ranging_directional` with a positive return_eff), **flip the strategy's effective direction to long** for N cycles OR until the medium catches up, whichever first.
- Conversely: medium says `trending_up`, short says `trending_down` → short-only.
- This extends the existing `regime_directional_policy` machinery (#779/#783) with one more input: a synthesized "leading" pseudo-regime that the policy can read.

**Override-state lifetime (must be specified):** the override counter (`cycles_active`) and the resolved direction are per-strategy mutable state. Define its owner and medium — in-heap per-strategy (parallel to `regimeDirectionalLegacyWarned sync.Map`, lost on restart) vs persisted to state DB — and what happens to an in-flight override across process restart and SIGHUP hot-reload.

The override should be **visible** — a new status field `regime_divergence: { short, medium, resolved_direction, cycles_active }` so operators can see it on `/status` and in trade DMs.

### 3. Trade DM + dashboard surface

Add a line in trade DMs and the position summary when divergence is active:
```
⚠ regime divergence: medium=trending_down short=trending_up (since 3 cycles, trusting short)
```
And a small badge on the dashboard (#809) when the resolved direction is divergence-driven.

## What this is NOT

- Not a replacement for the medium window. The medium is still the default for new opens and life-of-position freeze. This is an **override** layer on top.
- Not related to #879's global regime bundle. That work shares computation across peers; this work adds a *new derived signal* (divergence) that consumers like `regime_directional_policy` can opt into.
- Not retro-rationalizing. Strategies that never configured both windows keep current behavior (no divergence logic, no override). Opt-in only.

## Example walkthrough

**ETH on 2026-06-08 with `on_divergence: "trust_short"`:**
- Cycle N: medium=`trending_down`, short=`ranging_volatile` → soft divergence, no override (both still bearish-ish)
- Cycle N+5: medium=`trending_down`, short=`trending_up_clean` → HARD divergence, override fires
  - `effective_direction: long` (was `both` from `ranging_volatile` policy)
  - Status logs: `regime_divergence: short=trending_up_clean medium=trending_down cycles_active=1`
  - DM on next open: `regime divergence: trusting short (trending_up_clean) over medium (trending_down)`
- Cycle N+20: medium catches up → `regime=trending_up_clean`, divergence flag drops

## Acceptance criteria

1. New `regime_window_divergence` strategy config block (validated, hot-reload-safe).
2. Divergence reads both window labels from the existing multi-window regime payload (already a single subprocess call; #879 dedupes across peers) — **no new bundle return shape required.**
3. A divergence classifier function: `(short, medium) → { kind: hard|soft|none, override_direction: long|short|none }`.
4. `regime_directional_policy` consults the divergence result before resolving effective direction. Add a `divergence_override` slot to the policy table.
5. Status payload exposes `regime_divergence` for any strategy with the block configured.
6. Trade DMs include the divergence line when active.
7. **Scope decision required — backtest parity.** `regime_directional_policy` is HL-live-only and is **explicitly rejected by the backtester** (`backtest/run_backtest.py:160` raises `ValueError`). The divergence override inherits that live-only nature. Either (a) declare the override live-only with no backtest path (matching the base policy), or (b) treat "port `regime_directional_policy` to the backtester" as an explicit prerequisite for `--regime-window-divergence`. Do **not** assume drop-in backtest parity.
8. No behavior change for strategies that don't set the new block. Verified by running the existing test suite + a 1-week backtest on a divergence-disabled strategy.

## Dependencies

- Builds on #792 (multi-window detection) and #779/#783 (regime-aware directional policy; #801 added the composite 7-state vocabulary).
- #879 (global regime bundle) **has merged** (`f9f06cc`) — the regime bundle is available; divergence is a derived consumer of it, not a parallel computation path.
- Independent of #843 (live re-resolution). The divergence is computed on **closed bars** at decision time, not re-resolved mid-position.

## Files likely touched (preliminary)

- `shared_tools/regime.py` — already emits per-window labels; no change needed unless adding divergence metrics
- `shared_strategies/regime_directional_policy*` — read divergence override
- `scheduler/regime_divergence.go` (new) — classifier + override state
- `scheduler/main.go` — call site near `applyRegimeDirectionalPolicy` (`main.go:2872`)
- `scheduler/discord_commands.go` — surface in `/status`
- `scheduler/trade_alert*.go` — divergence line in trade DMs
- `backtest/...` — only if criterion #7 option (b) is chosen

---
Created with LLM: MiniMax-M3 | Harness: OpenClaw

richkuo:
## Summary

Adds a `regime_window_divergence` per-strategy config block (HL perps live only) that reads two named regime windows each cycle, classifies them as none/soft/hard divergence, and on a hard divergence optionally overrides `sc.Direction` so new entries follow the fast (short) window instead of waiting for the slow (medium) window to catch up.

- **No Python changes** — both window labels are already in `RegimePayload.Windows` from the multi-window check script; this is a pure Go-side derivation over existing data.
- **Open positions unaffected** — override applies to `posQty==0` only; hold-on-transition freeze (#779) governs in-flight positions.
- **Live-only** — inherits the same backtester rejection as `regime_directional_policy`; no backtest path (see criterion #7 in the issue).
- **Hard divergence** = one window bullish, other bearish (both non-neutral). **Soft** = one neutral, one directional — detected, logged, no override.
- `ranging_directional` bias resolved from `RegimeSnapshot.Metrics["return_eff"]`.

### Config example

```json
"regime_window_divergence": {
  "short_window": "composite_short",
  "medium_window": "composite_medium",
  "on_divergence": "trust_short"
}
```

`on_divergence`: `trust_short` | `trust_medium` | `alert_only`

## Files changed

| File | Change |
|---|---|
| `scheduler/regime_divergence.go` | New: type, classifier, override, state, DM formatter |
| `scheduler/regime_divergence_test.go` | New: 22 tests |
| `scheduler/config.go` | Add `RegimeWindowDivergence` field + surface/enablement guard |
| `scheduler/regime_window_spec.go` | Wire `ResolveRaw` in `validateStrategyRegimeVocabulary` |
| `scheduler/regime_multi_window.go` | Window-reference validation in `validateRegimeWindowsConfig` |
| `scheduler/config_reload.go` | Hot-reload gate (blocked while open) + flat apply/diff |
| `scheduler/executor.go` | Add `Divergence DivergenceResult` field to `HyperliquidResult` |
| `scheduler/main.go` | Override call in `runHyperliquidCheck`; state update in cycle loop |
| `scheduler/state.go` | `RegimeDivergence *RegimeDivergenceState` on `StrategyState` |
| `scheduler/portfolio.go` | `RegimeDivergenceNote` on `Trade`; stamp at perps open |
| `scheduler/discord.go` | Emit divergence line in `tradeAlertExtras` |
| `scheduler/server.go` | `RegimeDivergence` field on `StratStatus` |
| `scheduler/ui_server.go` | `RegimeDivergence` on `UIStrategyOverview`/`UIStrategyStatus` |
| `scheduler/static/ui/index.html` | Divergence badge element |
| `scheduler/static/ui/app.js` | `updateDivergenceBadge` + wiring in `refreshStatus` |
| `scheduler/static/ui/styles.css` | `.regime-badge--divergence` amber style |

## Test plan

- [x] `go build` clean
- [x] `go test ./...` — full suite passes (28s, all green)
- [x] Classifier: hard/soft/none, all three modes, `ranging_directional` sign
- [x] Config validation: valid, unknown mode, missing keys, same-window, unknown key
- [x] `IsConfigured`/`IsZero`/`EqualForReload` round-trips
- [x] Override mutates `sc.Direction` when flat; no mutation when open (`posQty>0`)
- [x] `alert_only` detects but never mutates
- [x] `CyclesActive` increments; resets on kind/direction change; clears on none

Closes #907

---


richkuo:
@claude

richkuo:
@claude rereview

richkuo:
@claude sonne rereview

PR #917, opened 2026-06-09
Fix #915: deduplicate shared-wallet value in Discord/Telegram and leaderboard TOTAL rows

### LINKED ISSUE #915: Discord/Telegram category summary TOTAL double-counts shared-wallet portfolio value
## Problem

In shared-wallet live setups (e.g. multiple Hyperliquid perps strategies on one account), Discord/Telegram category summary **TOTAL** rows can report a materially inflated portfolio value because they sum per-strategy `PortfolioValue()` instead of using the shared-wallet-adjusted total from `computeTotalPortfolioValue()`.

This is the same class of bug as #908 (stdout cycle summary), which was fixed in #911 — but the notification summaries were not updated.

## Why this is wrong

For shared-wallet platforms like Hyperliquid, per-strategy cash is virtual. Summing per-strategy PV double-counts virtual cash and diverges from the actual exchange wallet value.

Per-strategy rows in the summary table are correct (each bot's virtual book). The **TOTAL** row is where the double-count misleads operators.

## Affected code

- `scheduler/discord.go` — `FormatCategorySummary` accepts a `totalValue` parameter (`discord.go:353`) but **does not use it**; it recomputes `filteredValue` by summing `PortfolioValue(ss, prices)` across channel strategies (`discord.go:508,518`) and passes that to `writeCatTableChunks` for the TOTAL row (`discord.go:572`).
- `scheduler/main.go` — still computes `channelValue` / `assetValue` via naive per-strategy sum (`main.go:2178,2249`) and passes them to `FormatCategorySummary` (`main.go:2237,2254`), but those values are ignored today.
- `scheduler/main.go` — `runSummaryAndExit` (`-summary=` CLI) uses the same naive sum pattern (`main.go:2448,2470`).
- `scheduler/leaderboard.go` — **confirmed** same pattern: per-category TOTAL sums per-strategy `e.Value` (= `PortfolioValue`) at `leaderboard.go:193-197` (rendered `269-280`); `BuildLeaderboardSummary` shares it.

## Expected behavior

Category summary TOTAL rows (and leaderboard aggregates) should use the same shared-wallet-adjusted portfolio value that risk management already computes via `computeTotalPortfolioValue()`. Per-strategy rows should remain virtual-book values.

## Actual behavior

TOTAL rows sum per-strategy `PortfolioValue()`, which double-counts virtual cash in shared-wallet setups. The stdout cycle summary was fixed in #911; Discord/Telegram can still disagree with the exchange balance and with stdout logs.

## Suggested fix

The fix is **per-channel**, not whole-account — note the constraints below before implementing:

1. In `FormatCategorySummary`, render the TOTAL row from the passed-in (shared-wallet-adjusted) value instead of the internally summed `filteredValue`. Do **not** reuse the whole-account `totalPV` directly: it is computed over **all** `cfg.Strategies` (`main.go:940`), so feeding it to a single channel's or asset's TOTAL over-reports when more than one channel/asset exists.
2. In `main.go`, compute the adjusted total **per channel subset** via `computeTotalPortfolioValue(chStrats, …)` and pass it in. This requires hoisting `walletBalances` (declared in the risk block at `main.go:872`) to cycle scope — the summary block at `~main.go:2171` currently only sees function-scope `totalPV`/`sharedWallets`, not `walletBalances`. No re-fetch needed.
3. **Per-asset caveat:** shared wallets are keyed by `{Platform, Account}` (`detectSharedWallets`), i.e. account-wide. A single exchange balance cannot be split across asset-grouped messages, so a naive per-asset subset call would add the full account balance to *each* asset message. Specify intended behavior for per-asset TOTALs on a shared wallet: keep the virtual sum (labeled) or suppress the dedup.
4. **Failure policy:** when the shared balance is missing, `computeTotalPortfolioValue` returns `usedFallback=true` and sums member PVs. The summary should mirror that fallback, not error.
5. Apply the same per-subset pattern to `runSummaryAndExit` and the leaderboard per-category TOTAL.
6. Add a test asserting the TOTAL row matches `computeTotalPortfolioValue` — use a **multi-asset** shared HL wallet fixture (a two-strategy single-channel fixture passes without exposing the per-asset over-count).

## Related

- #908 — stdout cycle summary (fixed)
- #911 — fix for stdout `LogSummary`
- #909 — unrelated (HL external-close fill price booking)

## Notes

Logging / operator-visibility bug only — does not affect trades, risk checks, or kill switch (those already use `computeTotalPortfolioValue`).

---

richkuo:
## Summary

- **Root cause**: Discord/Telegram category-summary TOTAL rows and leaderboard TOTAL rows naive-summed per-strategy `PortfolioValue()`, double-counting virtual cash for strategies that share one on-exchange wallet (e.g. two HL perps strategies on the same account). Per-strategy rows were correct; only the TOTAL was inflated. Risk/kill-switch and the stdout cycle summary (#911) were already correct.

- **New helper `computeSubsetPortfolioValue`**: deduplicates shared wallets that are **fully contained** within a given subset; wallets that straddle a subset boundary (e.g. one member is in a different channel/asset bucket) fall back to virtual sum silently — a single on-exchange balance cannot be split across a partial subset.

- **`computeTotalPortfolioValue`** refactored to delegate to the new helper (whole-account = all wallets fully contained → identical results, single code path).

- **`walletBalances` hoisted** to cycle scope in `main.go` so the summary block and leaderboard auto-post can share the already-fetched balances from the risk loop.

## Files changed

| File | Change |
|------|--------|
| `scheduler/shared_wallet.go` | New `computeSubsetPortfolioValue`; `computeTotalPortfolioValue` delegates to it |
| `scheduler/discord.go` | `FormatCategorySummary` TOTAL row uses caller-supplied adjusted value |
| `scheduler/main.go` | Hoist `walletBalances`; subset-adjusted totals at summary call sites; `runSummaryAndExit` fix; leaderboard auto-post plumbing |
| `scheduler/leaderboard.go` | `formatLeaderboardMessage` + `BuildLeaderboardMessages` + `BuildLeaderboardSummary` adjusted-total threading; `PostLeaderboard` fetches wallet data |
| `*_test.go` | 8 new tests; all existing tests pass |

## Test plan

- [ ] `go test ./...` passes (all 539 net-new lines of test coverage green)
- [ ] Two live HL strategies sharing one account: Discord TOTAL row matches `computeTotalPortfolioValue` output, not the inflated naive sum
- [ ] Single-strategy or non-shared platforms: behavior unchanged (fallback to `filteredValue`/naive sum when `totalValue=0`)
- [ ] `-summary=perps` CLI: TOTAL row uses real wallet balance, not double-counted virtual cash

Closes #915

---

### DIFF
--- a/scheduler/discord.go
+++ b/scheduler/discord.go
@@ -560,7 +560,15 @@ func FormatCategorySummary(
 		})
 	}
 
-	totalPnl := filteredValue - totalInitCap
+	// For the TOTAL row, use the caller-supplied shared-wallet-adjusted value
+	// when one is provided (totalValue > 0). This prevents double-counting
+	// virtual cash in shared-wallet setups (#915). Per-strategy rows are
+	// unaffected — they use bot.value from the loop above.
+	totalRowValue := filteredValue
+	if totalValue > 0 {
+		totalRowValue = totalValue
+	}
+	totalPnl := totalRowValue - totalInitCap
 	totalPnlPct := 0.0
 	if totalInitCap > 0 {
 		totalPnlPct = (totalPnl / totalInitCap) * 100
@@ -569,7 +577,7 @@ func FormatCategorySummary(
 	// Render the strategy table in chunks of catTableMaxRows. The first chunk
 	// is appended to the in-message header; any extra chunks become standalone
 	// continuation messages so the table never overflows the 2000-char limit.
-	tableChunks := writeCatTableChunks(tableBots, filteredValue, totalPnl, totalPnlPct, hasSharedWallet)
+	tableChunks := writeCatTableChunks(tableBots, totalRowValue, totalPnl, totalPnlPct, hasSharedWallet)
 	if len(tableChunks) > 0 {
 		sb.WriteString(tableChunks[0])
 	}
--- a/scheduler/discord.go
+++ b/scheduler/discord.go
@@ -561,11 +561,16 @@ func FormatCategorySummary(
 	}
 
 	// For the TOTAL row, use the caller-supplied shared-wallet-adjusted value
-	// when one is provided (totalValue > 0). This prevents double-counting
-	// virtual cash in shared-wallet setups (#915). Per-strategy rows are
-	// unaffected — they use bot.value from the loop above.
+	// when one is provided. This prevents double-counting virtual cash in
+	// shared-wallet setups (#915). Per-strategy rows are unaffected — they use
+	// bot.value from the loop above.
+	//
+	// Sentinel: a negative totalValue means "no adjustment available" (fall back
+	// to the naive sum). A portfolio value is never negative, so this lets a
+	// legitimately drained shared wallet display $0 instead of being mistaken
+	// for "unset" and falling back to the inflated naive sum (#917 review item 3).
 	totalRowValue := filteredValue
-	if totalValue > 0 {
+	if totalValue >= 0 {
 		totalRowValue = totalValue
 	}
 	totalPnl := totalRowValue - totalInitCap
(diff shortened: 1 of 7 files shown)


richkuo:
@claude

richkuo:
@claude rereview

richkuo:
@claude sonnet rereview

Issue #360, opened 2026-04-20
[phase 2] Per-strategy CB on-chain close: OKX perps (live)


richkuo:
Phase 2 of #357 — OKX perps per-strategy circuit breaker close.

## Problem

Per-strategy \`CheckRisk\` on live OKX perps only mutates virtual state. If the drawdown or 5-loss CB fires, the real on-chain OKX swap position stays open until the portfolio kill switch fires.

## Scope

1. Extend \`OKXLiveCloser\` to accept \`partialSz *float64\` (CCXT \`createOrder(..., {reduceOnly: true, amount: sz})\`).
2. On CB fire, enqueue into \`RiskState.PendingCircuitCloses[\"okx\"]\` (requires phase 1b merged).
3. Drain in main loop with stuck-CB recovery.
4. Unit tests for: sizing math (single owner / shared-account proportional / mixed-units fallback), stuck-CB recovery, latch clear on confirmed flat.
5. OKX balance fetcher for \`defaultSharedWalletFetcher\` (unlocks multi-strategy OKX portfolio value correctness).

## Dependencies

- Depends on #357 phase 1a (shared-wallet registry) — for OKX wallet key recognition.
- Depends on #357 phase 1b (generic pending plumbing) — for \`PendingCircuitCloses\` map.

## Acceptance criteria

- [ ] \`OKXLiveCloser\` supports reduce-only partial close
- [ ] CB fire on live OKX perps enqueues pending close; drain submits it
- [ ] Shared-account proportional sizing parity with HL (\`computeOKXCircuitCloseQty\`)
- [ ] Tests cover sole-owner, shared-account, mixed-units fallback, stuck-CB recovery, clear-on-success
- [ ] No regression to portfolio kill switch path

## References

- #357 — epic
- #356 — HL reference
- scheduler/okx_close.go, scheduler/hyperliquid_balance.go

richkuo:
@claude

PR #920, opened 2026-06-10
feat(#918): exchange-authoritative shared-wallet reconciliation + drift alarm

### LINKED ISSUE #918: Shared-wallet per-strategy virtual values must reconcile exactly to the exchange balance each cycle (eliminate drift)
## Background

Follow-up to #915 (PR #917). For shared-wallet platforms (multiple Hyperliquid/OKX strategies on one on-exchange account), each strategy keeps its **own** virtual cash/equity book. #915 fixed the operator-visible **TOTAL** rows to read the real exchange balance once, but the underlying per-strategy books are still independent forecasts that diverge from reality.

**Invariant we want to hold, exactly, every cycle:**

> The sum of every shared-wallet member's virtual portfolio value equals the real on-exchange account balance, to the cent.

## Why drift happens today

The virtual book is a *model* updated from modeled inputs; the exchange balance is ground truth. They're computed from different numbers, so they diverge:

- **Fees** — the book deducts a *modeled* fee at trade time; the exchange charges the real maker/taker fee on the actual fill.
- **Fill price / slippage** — the book assumes it filled at the signal/mark price; the real order fills at a different VWAP.
- **Funding** — perps accrue funding on the real account every funding interval; the virtual book never credits/debits it.
- **Mark price** — unrealized P&L in the book uses the scheduler's fetched mark at one instant; the exchange uses its own mark at another.
- **Rounding** — size decimals, price ticks, and per-strategy cent rounding accumulate residuals across many strategies.

Even with perfect bookkeeping *logic*, a forecast can't equal ground truth. So the fix is not to model more carefully — it's to **stop modeling and read the real values each cycle**.

## Desired behavior — make the exchange authoritative

Each cycle, for each detected shared wallet, derive per-strategy values from real exchange data instead of the modeled book:

1. **Wallet total = real balance.** Use `marginSummary.accountValue` from the `clearinghouseState` snapshot already fetched for risk/positions (`fetchHyperliquidBalance`, `scheduler/hyperliquid_balance.go`; the cycle-scoped `walletBalances` map, `main.go:755`). This is the authoritative number the rows must sum to.
2. **Per-strategy positions = real position data, with a split for shared coins.** Each strategy owns its positions (`Position.OwnerStrategyID`); the exchange reports each position's size, entry, mark, and unrealized P&L in the same `clearinghouseState` (`assetPositions`). Use those directly instead of `qty × (fetchedMark − avgCost)` **when each member trades a distinct coin**. Note: HL `assetPositions` is **netted per coin** — two members trading the same coin on one account produce ONE on-chain position that can't be cleanly attributed per-strategy. Those shared-coin positions need a by-convention split (e.g. by virtual qty share, the codebase already splits shared-coin fills) — the **same kind of definitional split** as free collateral below, not clean real data. The sum is exact either way.
3. **Fees & funding are already in `accountValue` — do NOT separately fetch them.** Realized fees and funding payments are already reflected in `marginSummary.accountValue` and the on-chain position values read in steps 1–2. Re-fetching real fees (`HLFillLookup.Fee`) or HL `userFunding` to "credit/debit the book" would reintroduce the modeling this issue rejects, and is unnecessary for the invariant. Drop the per-strategy modeled-fee deduction; do not add a funding fetch. (Keeping a corrected book is the *other* design — but then the invariant no longer holds by construction.)
4. **Free collateral = the residual, split by convention.** The exchange reports ONE account balance, not per-strategy balances. Free cash not tied to a position is genuinely shared, so attributing it per-strategy needs a rule (pro-rata by `capital_pct`, or by each strategy's tracked equity share). **This split is a definitional choice, not drift** — the per-strategy numbers are by-convention, but the **sum stays exact to the cent** regardless of how the free cash is divided.

With values *read* rather than *modeled*, the invariant holds **by construction**: `Σ member value == accountValue` every cycle.

## Drift alarm becomes a real bug detector

Once values are exchange-derived, the cent-exact reconciliation check is no longer fighting normal mark/fee noise — it should **never** fire in normal operation. So:

- Each cycle, assert `abs(Σ member value − accountValue) <= $0.01` (compare cents-rounded values).
- If it exceeds that, fire a **throttled operator alert** (Discord/Telegram) naming the wallet `{platform}/{account}`, the virtual sum, the real balance, and the signed difference — this now indicates a genuine accounting/attribution bug, not expected divergence.
- Skip the check for any wallet whose balance fetch failed that cycle (already handled — `computeSubsetPortfolioValue` `usedFallback` freezes the peak; don't false-alarm).
- Throttle like the existing per-key trackers (`LiveExecFailureThrottle` in `failure_alerts.go` / `scriptFailureTracker` in `script_failure_alerts.go`): alert on first detection, backed-off cadence after, recovery notice when back within tolerance.

No extra network I/O: reuse the cycle's `clearinghouseState` for balance and positions, and no funding fetch is needed (it's already in `accountValue`). Run the comparison under the existing state read lock — read-only, no added contention.

## Implementation notes / data availability

- **Already fetched per cycle:** `marginSummary.accountValue` (wallet equity) and raw `assetPositions` (size/entry/leverage) via `fetchHyperliquidState` (`scheduler/hyperliquid_balance.go`), reused through the cycle-scoped `walletBalances` map (`main.go:755`) and `hlPositions`. Shared-wallet membership via `detectSharedWallets`.
- **Go parse must be extended (no new I/O):** the raw `clearinghouseState` response includes `markPx` / `unrealizedPnl`, but `fetchHyperliquidState` currently extracts only `coin/szi/entryPx/leverage/marginMode` — mark and unrealized P&L are dropped. Add them to the Go struct to read position value directly (the HTTP call is unchanged; the Python adapter already parses `unrealizedPnl`).
- **No new fetch needed:** funding/fees are already in `accountValue` (see step 3) — do not add an HL `userFunding` fetch.
- **Attribution rule** is needed for BOTH shared free collateral AND shared-coin (netted) positions — recommend pro-rata by current tracked equity share (or `capital_pct`); document it; the sum is exact either way.
- OKX shares the same shape (one margin account, multiple strategies) — apply the same exchange-authoritative reconciliation via its balance/position fetch.

## Scope

- Exchange-authoritative per-cycle reconciliation (values read, not modeled) so rows sum to the balance exactly: **in scope**.
- Throttled cent-exact drift alarm as a bug detector: **in scope**.
- The attribution convention (free collateral + shared-coin positions): pick one, document it.
- **Decide the book-vs-display boundary:** does the real-derived per-strategy value *replace* the per-strategy book (`s.Cash`, which drives drawdown/peak, kill-switch, and `capital_pct` sizing), or only *shadow* the display/summary rows? Display-only creates a second per-strategy number that itself diverges from the risk-math book — state which one this issue changes.

---

richkuo:
## Summary

Follow-up to #915 / #917. For shared-wallet platforms (multiple Hyperliquid/OKX strategies on one on-exchange account), each strategy's per-strategy display value was a modeled forecast (`s.Cash` + qty × fetched mark) that drifts from the real balance — so the per-strategy operator rows didn't sum to the account balance even after #917 fixed the TOTAL row.

This reads the real values each cycle instead of modeling. For every detected shared wallet, the real account balance is split into per-member display values:

```
value_i = w_i * (accountBalance − U) + ownedUPnL_i
```

- `U` — total exchange-reported unrealized P&L across the wallet's on-chain positions.
- `w_i` — each member's configured-capital weight (the operator-set starting allocation), dividing the genuinely-pooled collateral base.
- `ownedUPnL_i` — each position's real unrealized P&L attributed to its owner; shared-coin positions split by virtual-quantity share (mirrors `hyperliquidKillSwitchFillShare`).

The member values sum to the account balance **exactly to the cent, by construction**.

## Design decisions (confirmed)

1. **Display + alarm only.** Risk math (drawdown/peak, kill-switch, `capital_pct` sizing) keeps using the modeled `PortfolioValue`. The exchange-derived value is stored in-memory on `StrategyState` (never persisted, self-heals each cycle) and consumed only by operator surfaces via `displayStrategyValue`.
1. **Split the shared collateral base by configured capital** (the operator-set starting allocation).
1. **Hyperliquid + OKX** both.

## Drift alarm as a bug detector

Once values are exchange-derived they should reconcile exactly, so a >\$0.01 miss now indicates a genuine attribution bug (an on-chain position no member owns). The reconciler returns the **pre-rounding** drift separately and only absorbs sub-cent rounding into the last member — so an orphan surfaces in a throttled operator alert (first detection, then backed-off cadence, with a recovery notice) instead of being masked into a row. Wallets whose balance fetch failed are skipped (no false alarm).

## No new network I/O

Reuses the cycle's already-fetched `clearinghouseState` / OKX snapshot. The HL parse now keeps `assetPositions.unrealizedPnl`; `fetch_okx_positions.py` emits `unrealized_pnl`. No new CLI flags, so the startup probe contract is unchanged. Bonus: `/status` and `/stats` totals now dedup shared wallets too.

## Changes

- `shared_wallet_reconcile.go` — pure `reconcileSharedWalletMemberValues` (exact split), per-cycle `reconcileSharedWalletDisplayValues` (builds HL/OKX inputs, stores values, resets gates), `displayStrategyValue` helper.
- `shared_wallet_drift_alerts.go` — `SharedWalletDriftTracker` (first-detection alert + throttle + recovery), `reportSharedWalletDrift`.
- `state.go` — in-memory `SharedWalletValue` / `SharedWalletValueSet` on `StrategyState`.
- `hyperliquid_balance.go` / `okx_close.go` / `executor.go` / `fetch_okx_positions.py` — parse exchange unrealized P&L into `HLPosition` / `OKXPosition`.
- `main.go` — reconcile under the risk write lock, alarm after unlock; clear gates on the save-failure fast-exit cycle.
- `discord.go` / `discord_commands.go` / `leaderboard.go` / `server.go` / `ui_server.go` — operator surfaces read `displayStrategyValue` (risk/warning paths unchanged).

## Tests

Pure split math (distinct/shared coins, orphan→drift, cent rounding, capital fallback), cycle gating + fetch-fail fallback, drift throttle/recovery, HL + OKX unrealized-P&L parsing. Full Go suite + `go vet` pass; `fetch_okx_positions.py` compiles.

## Notes

- OKX reconciles to the same balance number its TOTAL row already uses (`get_account_balance`, free+used USDT from #360); its balance semantics are a pre-existing choice, untouched here.
- The full-wallet invariant holds; a partial-subset summary (one channel showing only some members of a wallet) keeps the existing #915 virtual-sum TOTAL behavior.

Closes #918

---
Created with LLM: Opus 4.8 | high | Harness: commit-push-pr

### DIFF
--- a/platforms/okx/adapter.py
+++ b/platforms/okx/adapter.py
@@ -232,10 +232,22 @@ class OKXExchangeAdapter:
         return results[0] if results else {}
 
     def get_account_balance(self) -> float:
-        """Return total USDT-denominated account value for shared-wallet
-        aggregation (#360 phase 2 — unlocks multi-strategy OKX portfolio
-        value correctness). Sums free + used USDT; callers that need to
-        include open-position PnL should rely on ccxt's total field.
+        """Return the total USDT-denominated account VALUE for shared-wallet
+        aggregation (#360 phase 2 — unlocks multi-strategy OKX portfolio value
+        correctness).
+
+        Returns ccxt's ``balance["total"]["USDT"]``, which ccxt's OKX driver
+        maps from the account's equity field (``eq``) — i.e. cash balance PLUS
+        unrealized P&L on open positions for the unified/cross margin account.
+
+        #918 RELIES on this being equity-inclusive: the shared-wallet
+        reconciler computes ``base = accountBalance - Σ unrealizedPnL`` and
+        redistributes the base by capital weight, attributing each position's
+        unrealized P&L to its owning strategy. If this value ever excluded
+        unrealized P&L the per-member equity rows would be skewed by the
+        owners' P&L (the member SUM still reconciles to this number, so the
+        drift alarm stays correct, but the split would misreport per strategy).
+        Verify against a live OKX account if the ccxt mapping changes.
 
         Only available in live mode; raises RuntimeError in paper mode.
         """
--- a/scheduler/discord.go
+++ b/scheduler/discord.go
@@ -505,7 +505,7 @@ func FormatCategorySummary(
 		if ss == nil {
 			continue
 		}
-		pv := PortfolioValue(ss, prices)
+		pv := displayStrategyValue(ss, prices)
 		walletPct := 0.0
 
 		// Shared wallet indicator (no value scaling — cash is already split by capital_pct).
--- a/scheduler/discord_commands.go
+++ b/scheduler/discord_commands.go
@@ -139,7 +139,7 @@ func formatStatusResponse(state *AppState, prices map[string]float64) string {
 	for _, id := range sortedAppStateIDs(state) {
 		s := state.Strategies[id]
 		cash += s.Cash
-		value += PortfolioValue(s, prices)
+		value += displayStrategyValue(s, prices)
 		posCount += len(s.Positions) + len(s.OptionPositions)
 		trades += len(s.TradeHistory)
 		if regime == "" && s.Regime != "" {
@@ -201,7 +201,7 @@ func formatPnLResponse(state *AppState, prices map[string]float64) string {
 	var perStrat []string
 	for _, id := range sortedAppStateIDs(state) {
 		s := state.Strategies[id]
-		pv := PortfolioValue(s, prices)
+		pv := displayStrategyValue(s, prices)
 		cap := s.InitialCapital
 		pnl := pv - cap
 		pnlPct := 0.0
@@ -337,7 +337,7 @@ func formatLeaderboardResponse(cfg *Config, state *AppState, prices map[string]f
 		if ss == nil {
 			continue
 		}
-		pv := PortfolioValue(ss, prices)
+		pv := displayStrategyValue(ss, prices)
 		initCap := EffectiveInitialCapital(sc, ss)
 		pnl := pv - initCap
 		pnlPct := 0.0
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -1094,10 +1094,11 @@ type OKXPositionsResult struct {
 
 // OKXPositionJSON is the per-position payload from fetch_okx_positions.py.
 type OKXPositionJSON struct {
-	Coin       string  `json:"coin"`
-	Size       float64 `json:"size"`
-	EntryPrice float64 `json:"entry_price"`
-	Side       string  `json:"side"`
+	Coin          string  `json:"coin"`
+	Size          float64 `json:"size"`
+	EntryPrice    float64 `json:"entry_price"`
+	Side          string  `json:"side"`
+	UnrealizedPnL float64 `json:"unrealized_pnl"`
 }
 
 // RunOKXFetchPositions runs fetch_okx_positions.py and returns the parsed
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -23,6 +23,12 @@ type HLPosition struct {
 	EntryPrice float64
 	Leverage   float64 // on-chain leverage value (#254)
 	MarginMode string  // "isolated" | "cross" — forwarded to Python so it can skip a duplicate /info call (#768)
+	// UnrealizedPnL is the exchange-reported unrealized P&L for this position
+	// (clearinghouseState assetPositions.position.unrealizedPnl). Used by the
+	// shared-wallet exchange-authoritative reconciliation (#918) to attribute
+	// real position P&L to the owning strategy instead of modeling it from a
+	// fetched mark. Zero when the field is absent or unparseable.
+	UnrealizedPnL float64
 }
 
 // hlExecuteSnapshotForCoin extracts the cycle-local on-chain leverage + margin
@@ -225,6 +231,7 @@ func fetchHyperliquidState(accountAddress string) (float64, []HLPosition, error)
 					Type  string      `json:"type"`
 					Value json.Number `json:"value"`
 				} `json:"leverage"`
+				UnrealizedPnl string `json:"unrealizedPnl"`
 			} `json:"position"`
 		} `json:"assetPositions"`
 	}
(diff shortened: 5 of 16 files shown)


richkuo:
@claude review

richkuo:
@claude rereview

richkuo:
@claude rereview

richkuo:
Addressed the latest review in `6938b28`:

1. **Needs Fixing — subset TOTAL vs rows:** added `computeSubsetDisplayValue` (`scheduler/shared_wallet_reconcile.go`). Gated members (`SharedWalletValueSet`) are summed directly — reconciled values sum to the real balance by construction, so any slice (per-asset, leaderboard top-N) totals exactly to its visible rows; ungated strategies fall back to `computeSubsetPortfolioValue` with the #915 dedup/virtual-sum semantics. Wired into both per-channel/per-asset cycle summary sites, the summary CLI path, and `leaderboardAdjustedTotal`. This also removes the display-side double count for a gated same-account live `manual` strategy. Risk math untouched.
2. **Create Follow-up Issue — risk-path manual double count:** filed as #921.
3. **Recommended Optional — consecutive distinct transients:** implemented via orphan signature. `reconcileSharedWalletMemberValues` now returns `OrphanCoins`; `SharedWalletDriftTracker.Record` keys the confirmation streak on the sorted coin set and restarts it on a signature change, so a limit-fill transient followed by an unrelated manual-open transient no longer sums to a spurious alert + recovery pair. Magnitude is deliberately not part of the streak key (a real orphan's uPnL moves with the mark each cycle, and the same-coin signature persists). Bonus: drift alerts now name the unattributed coins.

Tests: new coverage for partial-slice/mixed/ungated-fallback/manual-no-double-count subset totals and the distinct-transient / same-orphan-changing-magnitude streak cases. Full Go suite + `go vet` pass.

---

richkuo:
@claude rereview

richkuo:
Addressed the round-3 review item in `85cd8c2`:

1. **Per-coin continuity replaces exact-set streak keying.** `SharedWalletDriftTracker` now tracks `coinStreaks` per orphan coin: a coin alerts once it has stayed unowned for the threshold number of consecutive cycles, so a persistent orphan keeps confirming while transients on other coins churn the set (`{A}` → `{A,B}` → `{A,C}` alerts on cycle 2). Per-coin `alertedCoins` marks make a new persistent orphan appearing immediately after a prior alert (no intervening clean cycle) re-confirm deterministically — even when the drift magnitude coincidentally matches the last-notified value. Coinless over-tolerance drift (weighting bug) confirms via a pseudo-coin slot, preserving the bare-counter behavior; the reported count and recovery prior-count are the longest current per-coin streak.

Tests cover persistent-orphan-through-churn, new-orphan-after-alert re-confirmation, and coinless confirmation; the earlier distinct-transient and magnitude-re-alert cases still pass unchanged. Full Go suite + `go vet` pass.

---

richkuo:
@claude rereview

richkuo:
Addressed both round-4 review items in `dfbadb2`:

1. **Re-alert gate anchored to the last notification with a relative threshold.** The "materially changed" check now compares against `lastNotifiedDriftCents` (updated only when a notification actually fires, matching the comment's stated intent) and requires the cumulative move to exceed both 1 cent and `sharedWalletDriftRealertRatio` (10%) of the notified magnitude. Per-cycle mark wiggle on a confirmed orphan's uPnL therefore settles into the every-10th-cycle/hourly back-off, while a genuinely growing (or sign-flipping) drift accumulates against the anchor and re-surfaces. Invariant tested: +4%/+8% wiggle suppressed, +12% cumulative move re-alerts and advances the anchor, then small wiggle vs the new anchor is throttled again.
2. **Wallet-level duration counter.** The tracker keeps `entry.cycles` — consecutive over-tolerance cycles at the wallet level, independent of which orphan coin is responsible — and uses it for the alert/recovery "N cycles" duration and the every-10th-cycle cadence. Per-coin streaks remain the confirmation and `alertedCoins` mechanism. Invariant tested: an episode that churns coins in its final cycle reports the full wallet-level duration, not the final coin's 1-cycle streak.

Distinct-transient and new-orphan-after-alert tests updated to the wallet-level count semantics (their notify assertions are unchanged). Full Go suite + `go vet` pass.

---

richkuo:
@claude sonnet rereview

PR #923, opened 2026-06-10
Require Invariant + Must survive on @claude review findings

### LINKED ISSUE #922: Review findings should state the violated invariant + adversarial cases so fixes cover the failure class in one pass
## Problem

Review fixes are landing example-shaped instead of class-shaped, forcing multiple review passes on one PR. Concrete case from PR 920's drift-alarm work:

1. Round 2 flagged "two distinct one-cycle transients on consecutive cycles produce a spurious alert" — the fix keyed the confirmation streak on the **exact orphan-coin set**, which fixed that example.
1. Round 3 then flagged the inverse failure the same change introduced: a persistent orphan coexisting with churning transients (`{A}` → `{A,B}` → `{A,C}`) never reaches the threshold and never alerts. The correct first-principles design (per-coin continuity) was only reached on the third pass.

The root cause is on both sides: the review finding described one scenario rather than the invariant ("streak continuity must follow the orphan, not the set"), and the fix session implemented against that scenario without enumerating states that break the new design.

## Proposal

**1. Reviewer side — `.github/workflows/claude.yml` (`--append-system-prompt`):** extend the `@claude review` output contract so every `### Needs Fixing` / `### Recommended Optional` item must include, after the description:

- **Invariant:** one sentence stating the general property the code must satisfy (what is violated, independent of the example).
- **Must survive:** 1–3 adversarial cases beyond the example that any fix must handle (compound states, inverse scenario, boundary).

This keeps the existing comment structure (verdict line, H3 sections, numbered bold-title items) and just adds two labeled lines per item.

**2. Fixer side — `CLAUDE.md` (Pull Requests section):** add a rule for sessions addressing review findings: before implementing, restate the finding as an invariant, enumerate states that would break the proposed fix (especially the inverse of the reported scenario and compound/concurrent cases), and add tests for that class — not only the reported example — before pushing.

Both files also need the PR-review-format section of `CLAUDE.md` updated to match item 1 so the documented contract and the enforced prompt stay in sync.

## Acceptance

- `claude.yml` system prompt requires Invariant + Must survive lines per finding.
- `CLAUDE.md` documents both the new finding format and the fixer-side class-coverage rule.
- Next `@claude review` run emits findings in the new shape.

---

richkuo:
## Summary
- Extend `.github/workflows/claude.yml` `--append-system-prompt` so `### Needs Fixing` and `### Recommended Optional` items include **Invariant:** and **Must survive:** lines after the description.
- Document the reviewer contract and a fixer-side class-coverage rule in `CLAUDE.md` (also `AGENTS.md` via symlink).

Closes #922

## Test plan
- [ ] Merge and trigger `@claude review` on a PR with at least one Needs Fixing or Recommended Optional item; confirm findings include Invariant + Must survive.
- [ ] Confirm `### Requires Human Review` / `### Create Follow-up Issue` items are unchanged (no Invariant/Must survive required).

LLM: Composer 2.5 | default | Harness: Cursor


richkuo:
@claude sonnet

PR #938, opened 2026-06-10
fix(hl): add +50 candle fetch margin and trim for sparse data gaps

### LINKED ISSUE #937: [hl] get_ohlcv drops below requested limit when candle data has gaps
## Summary

`HyperliquidExchangeAdapter.get_ohlcv` (`platforms/hyperliquid/adapter.py`) computes the request window as:

```python
end_ms = int(time.time() * 1000)
start_ms = end_ms - interval_ms * limit
```

i.e. exactly `limit` intervals of wall-clock history. Hyperliquid `candleSnapshot` is **sparse**: intervals with no trading activity are omitted rather than emitted as zero-volume bars. When the window contains gaps (exchange downtime, low-liquidity no-trade intervals, etc.), the API returns fewer than `limit` rows.

Callers only hard-fail when `len(candles) < 30` (`shared_scripts/check_hyperliquid.py`); otherwise a shortfall vs `ohlcv_limit` continues silently. Strategies that assume a full bar-count lookback (ATR warm-up via `ensure_atr_indicator`, regime detection, HTF filter refetch, etc.) get fewer bars than requested — distorting rolling indicators and regime labels. The failure mode is **bar-count shortfall**, not calendar staleness of the oldest timestamp (unless gaps cluster at the old end of the window).

## Expected

`get_ohlcv(..., limit=N)` should return **exactly `N` rows** (the most recent `N` candles) even when the raw time window contains gap slots. When gaps prevent reaching `N` bars after reasonable headroom, behavior today (continue if `>= 30`, else error) can remain unless we add an explicit warning.

## Proposed fix

1. **Widen** the request window so sparse gaps do not truncate bar count.
2. **Trim** to the most recent `limit` rows before return **and** before `_save_ohlcv_cache` (Robinhood precedent: `platforms/robinhood/adapter.py` `result[-limit:]`).

```python
GAP_MARGIN = 50  # tunable; consider retry/extend if still short after trim

end_ms = int(time.time() * 1000)
start_ms = end_ms - interval_ms * (limit + GAP_MARGIN)

candles = self._info.candles_snapshot(symbol, interval, start_ms, end_ms)
# ... transform to result list ...

if len(result) > limit:
    result = result[-limit:]
# cache + return result
```

`+50` is a conservative default; illiquid symbols or large `regimeRequiredOhlcvLimit` values may need more headroom or a small extend-until-`>= limit` loop (cap at HL's ~5000-candle API max).

## Acceptance criteria

- [ ] `get_ohlcv(..., limit=N)` returns exactly `N` rows when enough candles exist in the widened window
- [ ] No regression for the no-gap case (still `N` rows after trim, not `N + margin`)
- [ ] Cache key unchanged (`#839`, `(symbol, interval, limit)`); cached payload is the **trimmed** `N`-row list
- [ ] Tests in `platforms/hyperliquid/test_adapter.py` / `test_ohlcv_cache.py` cover gap shortfall + trim + cache write path

## Out of scope

- **Backtest** uses `shared_tools/data_fetcher.py` (ccxt BinanceUS), not `HyperliquidExchangeAdapter` — no backtest change from this fix.

## Environment

- Affects live HL perps/manual paths: `check_hyperliquid.py`, `check_regime.py`, HTF refetch, `run_fetch_atr` (hardcoded `limit=200`)
- Most visible at larger `limit` values and lower-liquidity symbols

## Repro (fill in when observed)

- Symbol / interval / `ohlcv_limit`:
- Observed `len(candles)` vs expected:
- Estimated gap count or context:

---
Updated with LLM: Composer | default | Harness: Cursor

richkuo:
## Summary

- `HyperliquidExchangeAdapter.get_ohlcv` now requests `limit + 50` candles of history (`OHLCV_GAP_MARGIN`) so the response still has at least `limit` rows when the sparse `candleSnapshot` window has gaps, then trims to the most recent `limit` rows before caching/returning.
- Without the margin, strategies that depend on a fixed lookback (ATR warm-up, regime detection) silently lose the oldest data points on gap days — particularly visible at larger `limit` values and lower-liquidity symbols.
- `+50` is headroom, not a duplicate fetch; Hyperliquid returns the candles it has within the wider window.
- **Shortfall warning:** when gaps exceed the margin and the trimmed result is still `0 < len < limit`, emit a `[WARN]` to stderr (adapter convention) instead of letting indicators silently warm up on fewer bars than requested. `+50` is a conservative default; an extend-until-`>= limit` loop remains a possible follow-up for very illiquid symbols.

Closes #937

## Test plan

- [x] `get_ohlcv(..., limit=200)` widens the request window to `limit + 50` intervals
- [x] API returns extra rows → trimmed to exactly `limit`, most-recent kept
- [x] Gaps exceed margin (`len < limit`) → result returned as-is **and** a `shortfall` `[WARN]` is logged to stderr
- [x] Full result (`len == limit`) → no shortfall warning
- [x] Cycle-scoped on-disk cache (`#839`, key = `(symbol, interval, limit)`) stores the **trimmed** `limit`-row list; peers read exactly `limit` rows and skip the network
- [x] Backtester parity: backtest uses ccxt `data_fetcher.py`, not this adapter — no backtest change

Created with LLM: MiniMax-M3 | medium | Harness: OpenClaw


richkuo:
Needs Updates

### Create Follow-up Issue

1. **Flat +50 margin does not guarantee `limit` rows for very gappy / short-interval / illiquid symbols.**
The widen-and-trim approach (`adapter.py:534`, `:558`) plus the new `0 < len < limit` shortfall `[WARN]` (`adapter.py:560`) makes the shortfall visible, but a 1m interval on an illiquid coin can still exceed a 50-candle gap and warm indicators on fewer bars. Track an extend-until-`>= limit` retry loop (capped at HL's ~5000-candle API max) as a follow-up; the issue (#937) already flagged this as a possible enhancement. Not a blocker for this PR.

### Requires Human Review

1. **`OHLCV_GAP_MARGIN = 50` is a fixed absolute constant across all intervals.**
50 candles of headroom is generous for `1h`+ but thin for `1m`/`3m` on low-liquidity symbols. Recommendation: ship the flat constant as-is (simple, covers the common `1h`/`200`-bar case), and let the follow-up extend-loop above handle the long tail rather than per-interval tuning now. Confirm the constant value is acceptable before merge.

PR #948, opened 2026-06-10
fix(hl): extend-until-limit retry when +50 gap margin is insufficient

### LINKED ISSUE #947: [hl] get_ohlcv: extend-until-limit retry when +50 gap margin is insufficient
## Context

Follow-up to #937 (PR #938). That PR widens the `get_ohlcv` request window to `limit + OHLCV_GAP_MARGIN` (50) intervals, trims to the most-recent `limit`, and emits a `[WARN] hl ohlcv shortfall` to stderr when sparse `candleSnapshot` gaps still leave `0 < len < limit`.

The flat `+50` margin is a conservative default. For **short intervals on illiquid symbols** (e.g. `1m`/`3m` on a thin coin), no-trade gaps can exceed 50 candles, so the trimmed result is still short of `limit` and indicators warm on fewer bars than requested. Today that only logs a warning — the caller continues while `len >= 30` (`shared_scripts/check_hyperliquid.py:154`).

## Proposed

When the widened-window fetch still yields `< limit` rows, **extend the window and retry** until `len >= limit` or a cap is hit:

- Bounded retry loop (e.g. double the extra margin each pass, or step back `start_ms` by another `GAP_MARGIN` block).
- **Cap at Hyperliquid's ~5000-candle API max** — never request a window that would exceed it; if still short at the cap, fall through to today's behavior (return what we have + the existing shortfall `[WARN]`).
- Keep the cache contract unchanged: cache key + cached payload stay `limit`-sized after final trim.

## Scope / notes

- HL adapter only (`platforms/hyperliquid/adapter.py` `get_ohlcv`). Other adapters fetch via yfinance/ccxt which emit continuous bars, so they don't share the sparse-gap failure mode.
- Low priority: in practice `regimeRequiredOhlcvLimit` is in the low hundreds, far below the 5000 cap, and `1h`+ intervals rarely gap past 50. Drive prioritization off whether the new shortfall `[WARN]` actually fires in production logs.
- Tests: extend `platforms/hyperliquid/test_adapter.py` to cover the retry-extends-window and hits-cap-still-short paths.

---

richkuo:
## Summary

Follow-up to #937/#938. The flat `+50` gap margin can still come up short for very gappy / short-interval / illiquid symbols, leaving indicators warming on fewer bars than requested (the prior PR only logged a `[WARN]`).

`HyperliquidExchangeAdapter.get_ohlcv` now **extends the request window and refetches** when the widened fetch is still short, terminating on a **returned-count plateau**:

- Start at `limit + OHLCV_GAP_MARGIN` (50) intervals; if `len(result) < limit`, double the window and refetch.
- **Hyperliquid caps the returned *count*** (most-recent ≤ `OHLCV_MAX_CANDLES = 5000` candles in the range), **not the request span**. So we keep widening until: we reach `limit`; the returned count **plateaus** (a wider span adds no candles → the symbol's full history is in-window); we hit the 5000 return cap; the window is empty; or the `OHLCV_MAX_EXTEND_PASSES = 12` backstop trips.
- **Empty windows don't extend** — a halted / untraded symbol won't populate by widening; empty stays the caller's error path (`len < 30` in `check_hyperliquid.py`).
- **No sticky "short symbol" marker** — a brand-new listing accumulates candles over time and must be re-attempted on each cache miss, not pinned short. Per-call `/info` calls are bounded (plateau usually trips in 1–2 widens) and only happen on the cache-miss shortfall path; the cache TTL spares peers.
- Final result is trimmed to the most-recent `limit`; the cache key + cached payload stay `limit`-sized (peers read exactly `limit`, unchanged from #937).
- Persistent shortfall warns to stderr (`[WARN] hl ohlcv shortfall ... after extending the window to the symbol's available history`).

## Review findings (PR #948) — both addressed in-PR

- **Finding 1 (correctness):** the original loop capped the request *span* at 5000 intervals; HL actually caps the returned *count*. Fixed by switching to count-plateau termination so very sparse symbols aren't stopped short by a span ceiling.
- **Finding 2 (efficiency):** plateau termination bounds the cascade for persistently-short symbols to ~1–2 widens (not a full doubling ladder); the shortfall test now asserts `call_count` to pin it.

## Test plan

- [x] Extend reaches `limit`: sparse mock (~60% of span) → two fetches (250→500 spans), trimmed to exactly `limit`
- [x] Plateau stops + warns: 150-candle new listing → stops after 2 fetches at 150 rows, warns `150 of 200`
- [x] Extend passes bounded: pathological +1-candle-per-widen → caps at `OHLCV_MAX_EXTEND_PASSES`
- [x] Shortfall test asserts `call_count == 2` (bounded, not a full ladder)
- [x] Empty window does not extend (single fetch); cache dedup tests supply `>= limit` rows so the loop stays inert; cache key/payload remain `limit`-sized
- [x] Full HL suite: 102 passed

## Out of scope

- HL adapter only. Other adapters fetch via yfinance/ccxt (continuous bars), so they don't share the sparse-gap failure mode.

Closes #947

---


richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

Issue #942, opened 2026-06-10
parity(backtest): --config loader — legacy close keys no-op silently; direction/invert_signal/regime_window_divergence ignored


richkuo:
Found by the #906 audit (D2.8 + D2.1/D2.5). Consolidates issue 943 — both are `run_backtest.py --config` loader fixes enforcing the same invariant, so they ship as one PR.

## Part 1 — v13/v14 version gate lets pre-v15 legacy close keys silently no-op (D2.8)

`run_backtest.py --config` gates on `config_version >= 13` (`backtest/run_backtest.py:138-144`), but the v15 migration is what canonicalizes close-strategy params (`tiers` → `tp_tiers`, `atr`/`multiple`/`fraction` → `atr_multiple`/`close_fraction`) on disk (`scheduler/config_migration_v15.go:270-284, 320-368`). A v13/v14 config passes the backtest gate while still carrying the legacy keys.

The Python close evaluators accept **only** the canonical runtime key: `tier_list_from_params` reads `tp_tiers` exclusively (`shared_strategies/close/_helpers.py:22-26`). `load_strategy_config` passes close params verbatim, so:

- Explicitly-configured tiers in a v13/v14 config are silently dropped; evaluators fall back to the system default ladder.
- With `--defaults user`, `_apply_user_close_defaults` checks only `params.get("tp_tiers")` (`backtest/run_backtest.py:104`) and **injects user defaults over the operator's explicit legacy tiers**.

Live is unaffected — Go canonicalizes on read — so backtest and live silently diverge on the same file.

**Fix:** bump the gate to `config_version >= 15`, reuse the existing "run go-trader once against this file to migrate, then retry" error text. Regression test: v13-shaped config with legacy `tiers` → loader raises.

## Part 2 — direction / invert_signal / regime_window_divergence silently ignored (D2.1/D2.5)

`load_strategy_config` (`backtest/run_backtest.py:119-223`) reads only `open_strategy`/`close_strategy` refs. Three strategy-level fields the live scheduler honors are silently dropped:

1. **`direction`** — live gates which sides may open; backtest opens whatever the strategy emits. A `direction: "long"` config backtests with shorts included.
2. **`invert_signal`** — live flips BUY↔SELL via `applySignalInversion` before `Direction`; backtest ignores it. Worse, the `regime_directional_policy` rejection message (`backtest/run_backtest.py:165`) recommends these exact fields for backtesting.
[...]

PR #950, opened 2026-06-10
audit(backtest): synthesis parity tool — #946 base + #935 grafts (#906)

### LINKED ISSUE #906: audit(backtest): full correctness, parity, and quality audit of the backtesting subsystem
## Summary

Full correctness, parity, and quality audit of the backtesting subsystem. The engine has grown to **4,659 LOC across the 8 core modules** (plus 4 `consolidation_*` research scripts sitting in `backtest/` — see scope note below) with **19 test files**, and 30+ close-strategy / regime / risk features have shipped in the last 90 days — many without explicit backtest parity work. This issue tracks a structured audit so the codebase doesn't accrue silent drift between backtest and live, and so known-bad patterns (#302 look-ahead, #303 parity, #304 reporting, #304 Sharpe annualization) are caught in the new code paths before they ship.

This is intentionally a **scoping/working issue**, not a "found bugs" issue. Sub-tasks should be filed as linked cards (or comments) as the audit surfaces concrete findings; this card tracks the overall audit plan and any *blocking* discoveries.

## Audit scope

**In scope (4 backtest variants + shared infrastructure):**
- `backtest/backtester.py` — main engine (1,706 LOC, 2 classes (`Trade`, `Backtester`) + ~26 helpers)
- `backtest/backtest_options.py` — options-specific
- `backtest/backtest_theta.py` — theta-harvest strategies
- `backtest/backtest_pairs.py` — two-leg pairs strategies
- `backtest/optimizer.py` — walk-forward grid search
- `backtest/run_backtest.py` — CLI surface
- `backtest/reporter.py` — output formatting
- `backtest/registry_loader.py` — strategy registry glue
- `backtest/tests/*.py` — 19 test files (parity, look-ahead, fills, regime, walk-forward, options)

**Out of scope** (tracked elsewhere):
- `backtest/consolidation_*.py` research scripts (one-off research tooling from #896, not part of the engine)
- Live trading engine parity (separate audit if needed)
- Storage / `shared_tools/storage.py` (issue #824)
- Strategy implementations themselves (each strategy's own issue)

## Background — history of backtest-specific bugs

A real audit needs to start from the prior findings to avoid re-litigating them:

| #    | Severity  | Title                                                              | Status   |
|------|-----------|--------------------------------------------------------------------|----------|
| #302 | Critical  | Backtest: critical correctness bugs (look-ahead, IV math)         | Closed   |
| #303 | High      | Backtest: parity gaps with live trading system                    | Closed   |
| #304 | Med/Low   | Backtest: reporting bugs and maintenance cleanup                  | Closed   |
| #534 | Feature   | Wire close-strategy registry into backtester                       | Closed   |
| #543 | Feature   | Backtester regime parity + `--regime-*` CLI flags                  | Closed   |
| #641 | Refactor  | Refactor(backtest): align Backtester API with co-located strategy refs | Closed |
| #709 | High      | Backtester parity for `sl_after` (post-TP SL adjustment)          | Closed   |
| #715 | Critical  | `[Critical]` Backtester `sl_after`: SL hit check fires on same bar | Closed   |
| #730 | Critical  | Harden backtests against look-ahead bias                          | Closed   |
| #737 | Feature   | Backtester parity for #733 regime-aware surfaces                   | Closed   |
| #770 | Feature   | Two-leg pairs backtester                                          | Closed   |
| #824 | Critical  | Startup probe fails under `ProtectSystem=strict` (storage.py)     | Closed   |
| #838 | Cleanup   | Stop offering percentage close strategies that don't support `sl_after` | Closed |

The pattern is clear: **correctness, parity, and reporting bugs have shipped roughly twice a year, with the last critical one (#824) closed ~1 week before this issue was filed**. The audit is overdue, not premature.

### D1 — Correctness: look-ahead contract (#730) integrity

The look-ahead contract is documented in `backtester.py:14-37` and is the engine's central correctness guarantee. Audit must verify it's intact for every column path.

- [ ] D1.1 — Re-verify the `shift(1)` guard is applied to **every** decision input column (signal, open_action, close_fraction, regime). For each: trace from the strategy script emission → DataFrame column → backtester normalization → per-bar read. Look for new columns added post-#730 that bypass the shift.
- [ ] D1.2 — Audit the close-evaluator path (`_evaluate_close_strategies` at `backtester.py:1368`, dispatched from the per-bar loop at `backtester.py:1139`). Close evaluators see bar N's mark + bar N's ATR; their output becomes `pending_close_fraction` applied at bar N+1's open. Verify this is preserved for: ratchet (`trailing_tp_ratchet`/`_regime`), `tiered_tp_atr_*`, `post_tp_sl` SL-after adjustments.
- [ ] D1.3 — Verify bar-level SL/TP intra-bar race resolution (documented under "Backtest" in CLAUDE.md) — the engine resolves at bar close, not intra-bar OHLC walking. Confirm no strategy closes at the bar's *open* when both SL and TP are within the same bar's range.
- [ ] D1.4 — Hunt for upstream peeking. Per #730: "signal shift(1) is the only look-ahead guard and is defeated by upstream peeking." Spot-check each `shared_strategies/open/*.py` and `shared_strategies/close/*.py` for forward-looking patterns (`close.shift(-1)`, `high.shift(-1)`, `low.shift(-1)`, `next-bar` references). Any present = silent look-ahead the backtester can't catch.
- [ ] D1.5 — Verify the `regime` shift path (line 779) survives #879 (shared regime bundle) and #862 (unit-consistent composite). The regime column may be replaced with a multi-column bundle — confirm the backtester uses the column that was shifted, not the raw injected one.

### D2 — Parity: backtester ↔ live scheduler

The parity contract: same `compute_*` function, same window, same params → same decision. Audit must identify the gaps since the last parity issues (#303, #709, #737).

- [ ] D2.1 — **Recent features without explicit backtest parity work.** Audit PRs #842, #857, #866, #870, #872, #880, #882, #883, #885, #887 — those whose PR body says `backtest=N` from the recent merged list. For each, determine if the feature is live-only (and document why) or has a parity gap that needs closing.
- [ ] D2.2 — **Scale-in / pyramiding (#873).** CLAUDE.md explicitly notes "NOT backtested." Decide: is this a permanent constraint (HL perps + manual only) or a feature to add parity for? If permanent, document the constraint in `backtester.py`; if not, scope a parity card.
- [ ] D2.3 — **Resting manual limit orders (#883).** Not in backtester. Scope: backtest can simulate limit fills at bar open (worst case) vs. mid-bar (best case) vs. close (typical). Decide if backtester should model this; if not, document.
- [ ] D2.4 — **`tiered_tp_atr_live_regime_dynamic` (#843).** `run_backtest.py:193-196` rejects this close ref outright. The "dynamic" path has regime hysteresis (`regime_confirm_cycles`) that paper backtests can't reproduce. Verify the rejection message and confirm it's still the right behavior.
- [ ] D2.5 — **Regime directional policy + orphan auto-close (#822 / `regime_directional_policy.go`).** CLAUDE.md says "HL perps only, `regime.enabled=true` required, backtester rejects." Verify the rejection is in place; spot-check that paper/paper-fleet behavior matches the live orphan-close path.
- [ ] D2.6 — **Inline ATR-trailing SL at open (#885, `armTrailingStopAtOpenNow`).** Live arms trailing SL same cycle; backtester may use a different trigger computation. Compare trigger-px formulas.
- [ ] D2.7 — **Default tier retunes (#870, #887).** `defaultHLProtectionTiers` was retuned from `[{1×,0.5},{2×,1.0}]` to `[{1.5×,0.4},{3×,0.8},{5×,1.0}]`. Backtester uses these via the close evaluators — verify the retune flowed through and the parity test (`test_backtester_close_strategies.py`) actually asserts the new tier values, not just the old ones.
- [ ] D2.8 — **v15 close-key migration (#841, #853).** `backtest/optimizer.py:487,494` still emits legacy `tiers` per CLAUDE.md. Confirm: (a) the v15 migration accepts these and rewrites on load, (b) backtest-written configs round-trip cleanly through `LoadConfig`, (c) no third path uses the raw `tiers` key.
- [ ] D2.9 — **Single close strategy ref (#842).** `closeRefs()` adapter handles 0/1-element scans. Verify no backtest code path still iterates `close_strategies` as a length-N array assuming len ≥ 1.
- [ ] D2.10 — **Per-platform fee parity.** `test_platform_fees.py` scrapes `fees.go` to enforce parity per CLAUDE.md. The backtester's `PLATFORM_FEE_PCT` table (`backtester.py:191-198`) covers 6 platforms (`binanceus`, `hyperliquid`, `robinhood`, `luno`, `okx`, `okx-perps`) — `deribit`/`ibkr`/`topstep` have no spot-fee entry (options/futures fees flow through different fee functions). Verify the table matches fees.go and decide whether the missing platforms need entries or a documented exclusion.

### D3 — Engine coverage: which live surfaces have no backtest path

- [ ] D3.1 — List every close strategy name in `close/registry.py` and confirm each has at least one `test_backtester_close_strategies.py` case. Spot-check: `trailing_tp_ratchet_regime` (added #857), `tiered_tp_atr_live_regime_dynamic` (rejected — see D2.4).
- [ ] D3.2 — List every regime label in the composite 7-state classifier (`ranging_quiet`, `ranging_volatile`, `ranging_directional`, `trending_up_clean`, `trending_up_choppy`, `trending_down_clean`, `trending_down_choppy`) and confirm `test_backtester_regime.py` exercises each.
- [ ] D3.3 — Cross-platform coverage in `backtester.py` and the 3 variant backtesters. Confirm spot, perps, options, pairs, theta all have non-trivial coverage in the test fixtures.
- [ ] D3.4 — Multi-timeframe coverage: 1m, 5m, 15m, 30m, 1h, 4h, 1d, 1w. The TIMEFRAME_PERIODS_PER_YEAR table (`backtester.py:167-181`) has 13 entries (adds 2h/6h/8h/12h/1M) — confirm the Sharpe annualization test (added in #304 M3) has cases at ≥ 3 timeframes, not just daily.

### D4 — Reporting / metrics quality

- [ ] D4.1 — Sharpe annualization: `periods_per_year()` (#304 M3) is in place. Verify it's called in **all** Sharpe/Sortino/Calmar paths in the 4 backtesters, not just `backtester.py`.
- [ ] D4.2 — Options annualized return (D4.2 = #304 M5): `len(equity_curve)` was being used as days. Confirm the fix (calendar days between first/last timestamp) is in and is the only `*annualized*` calc.
- [ ] D4.3 — Max-fraction semantics pre-#641: `run_backtest.py:179` reference per CLAUDE.md. Verify the legacy max-fraction path is removed and the new single-ref semantics is the only path.
- [ ] D4.4 — Equity-curve dates: verify `format_results` and the JSON report emit ISO timestamps, not `pd.Timestamp` repr, for downstream consumers (dashboard, optimizer, walk-forward splits).
- [ ] D4.5 — Drawdown calculation: peak-to-trough + recovery duration. Verify recovery time is included as a metric, not just max DD %.
- [ ] D4.6 — Trade-log completeness: per #304 L5, `backtest_theta.py` force-close omits terminal trade-log entries. Verify the fix is in and the parity test (if any) covers it.
- [ ] D4.7 — Net-of-fee vs gross-of-fee: the reporter should emit both. Confirm.
- [ ] D4.8 — Walk-forward output: per-fold Sharpe, per-fold return, and an aggregate "out-of-sample only" report. Verify the report isn't conflating in-sample (warmup + first fold) with out-of-sample.

### D5 — Optimizer quality (`optimizer.py`)

- [ ] D5.1 — `param_ranges` coverage: #304 L3 noted ~15 strategies had no default ranges. Re-audit post-#892 (`momentum_pro`, `mean_reversion_pro` added) and post-#896 (consolidation research). Confirm every strategy in `open/registry.py` has a `default_param_ranges` entry OR a clear "not optimizable" rejection at runtime.
- [ ] D5.2 — Walk-forward splits: the `splits=5` default may not be enough for low-volatility assets (overfits) or enough for high-volatility (underfits). Check if there's a minimum-data heuristic, and if not, scope one.
- [ ] D5.3 — Overfit detection: optimizer has `walk_forward_optimize` but unclear if it has a robustness check (e.g., fold-to-fold Sharpe std, parameter stability). If absent, scope a card.
- [ ] D5.4 — Multi-objective: optimizer likely optimizes a single metric (Sharpe or return). Verify whether `optimize_metric` is configurable and what the alternatives are.
- [ ] D5.5 — Reproducibility: `random_seed` parameter — confirm it's honored end-to-end (not just at one of the stochastic sites).

### D6 — Code health / debt

- [ ] D6.1 — Dead code: per #304 L4, `run_backtest.py:16` imported `fetch_full_history` but only used `load_cached_data`. Re-grep for unused imports across the backtest tree.
- [ ] D6.2 — Duplication: the 4 backtesters (main, options, theta, pairs) likely each implement their own fill / fee / equity-curve logic. Quantify the duplication; if >30%, scope a refactor.
- [ ] D6.3 — Hardcoded constants: per #304 L2, options backtester hardcodes `ccxt.binanceus`. Verify whether this was fixed; if not, scope a `--exchange` flag.
- [ ] D6.4 — HTF filter parity (#304 M2): `--htf-filter` was added to the CLI. Verify it actually plumbs through to the same `htf_trend_filter` the live scheduler uses, with a platform-appropriate `fetch_fn`. (This may be live-only if HTF data isn't in the historical feed.)
- [ ] D6.5 — `close_registry_loader` shim: per `backtester.py:78` comment, the loader exists to avoid the bare `import registry` collision with the open registry. Confirm the collision is still real and the shim is still needed; if not, simplify.
- [ ] D6.6 — Magic numbers: scan for hardcoded SL percentages, fee percentages, slippage assumptions that should be `fees.go`-sourced or test-configurable.

### D7 — Observability / debuggability

- [ ] D7.1 — Verbose mode: `--verbose` (or equivalent) per-strategy trace — confirm what's logged (signal bar, regime bar, close fraction, fill bar) and whether it's enough to debug a parity mismatch without re-running live.
- [ ] D7.2 — Equity-curve export: CSV/Parquet dump of the full equity curve, not just summary stats. Useful for downstream comparison with paper-fleet equity.
- [ ] D7.3 — Trade log export: per-trade CSV with entry/exit/reason/regime_at_entry/regime_at_exit. Useful for backtest-vs-paper reconciliation.
- [ ] D7.4 — Parity-debug helper: a `backtest-vs-paper-diff` tool that takes a strategy ID and a date range, runs both paths, and emits a per-bar diff of (signal, regime, open_action, close_fraction, fill_px, fee). This is the single highest-leverage tool for catching future parity drift.

### D8 — Test coverage gaps

- [ ] D8.1 — Audit `tests/` for which features have *no* test: every close strategy, every regime label, every platform fee, every time-frame's Sharpe annualization, every #304 M-item, every #303 H-item.
- [ ] D8.2 — Property-based tests: strategies are numeric-heavy and benefit from Hypothesis. Check if any test uses `hypothesis`; if not, scope a starter card.
- [ ] D8.3 — Fuzz testing for the JSON config loader: configs can be malformed in subtle ways (legacy `close_strategies` arrays of length >1 post-#842, unknown regime labels, etc.).
- [ ] D8.4 — Regression tests for #302 / #303 / #304 / #715 / #730 / #824: confirm each closed issue has a corresponding test that *would have caught the bug*. If any are missing, add them — this is the audit's primary preventative value.

## Audit process

1. **D1 (correctness) first** — these are silent killers. If a look-ahead exists, every metric downstream is wrong. Don't optimize reporting until the engine is correct.
2. **D2 (parity) second** — these are silent killers in production (paper says one thing, live does another). A new parity-debug helper (D7.4) accelerates every later step.
3. **D3 (coverage) and D4 (reporting) third** — these are quality concerns; backtest results can still be directionally useful even if reporting is rough.
4. **D5 / D6 (optimizer / debt) last** — these don't affect correctness but shape the audit's future velocity.

## Acceptance criteria

- [ ] D1–D4 fully audited. Findings filed as separate issue cards (cross-referenced here).
- [ ] D7.4 parity-debug helper shipped (highest-leverage output of the audit).
- [ ] D8.4 regression tests for every prior backtest bug (closed issues #302, #303, #304, #715, #730, #824) are confirmed present. Gaps are filed as separate cards.
- [ ] A short audit summary doc added to `backtest/` (or `docs/`) capturing the current state of correctness, parity, coverage, and known limitations. Future contributors can read it instead of re-deriving.

## References

- Prior audits: #302 (correctness), #303 (parity), #304 (reporting)
- Look-ahead contract: `backtest/backtester.py:14-37`, test in `backtest/tests/test_backtester_lookahead.py`
- Parity contract: `test_options_adapter_parity.py`, `test_backtester_regime.py:197-220`
- File-ownership map (live side, for parity checks): `CLAUDE.md` under "Repo Structure"
- `fees.go` (live fee source) — parity enforced by `test_platform_fees.py`

---

richkuo:
## Summary

Best-of-N synthesis of the three #906 bake-off PRs. Base is PR 946 (scored 90/100: genuine two-path tool, least code, zero production-code risk, found real bugs #942/#943); grafted on top are the strengths that made PR 935 the runner-up (88/100); nothing was taken from PR 936's single-path tool. Supersedes PRs 946, 935, and 936.

## What's from where

**From #946 (base):** `parity_diff.py` trailing-window replay (default 200 = live `--ohlcv-limit`), stride/expanding-window/CSV/exit-code CLI, `AUDIT.md` (D1–D8 verified findings incl. bugs #942/#943), behavior-asserting tests (clean on window-invariant strategy, catches a full-frame-normalization strategy).

**From #935 (grafts):**
1. **Faithful live path** — the live side now calls the actual check-script helpers (`prepare_check_regime` → `params["regime"]` injection → `evaluate_open_close`/`finalize_decision` with `close_registry_loader.evaluate`) rather than modeling them. This was #935's standout strength and #946's one acknowledged gap.
2. **`--config <live-config> --strategy-id <id>`** — replays the exact live refs via the #641 loader (v13+ gate, #842 single-close collapse, HL-live-only rejections all inherited); symbol/timeframe/registry/regime pulled from the strategy entry.
3. **`--fills`** — full `Backtester` run reporting simulated entry/exit fills (price + modeled fee) to line the decision diff up against booked trades.
4. **`--jsonl`** output + per-bar `backtest_effective_*` (post-`shift(1)`) engine-input columns.
5. **`backtester.py` docstring** — live-parity-limitations decision record (#873/#883/#843/#822/#885).

**New in the synthesis (neither PR had it):**
- **Close-evaluator parity** — registry closes (`tiered_tp_atr` …) run through the same `close_registry_loader.evaluate` on BOTH sides with a shared simulated position context (side/avg_cost/quantities/entry_atr/entry regime). Backtest side feeds engine semantics (bar-N close, full-frame closed-bar ATR/regime); live side feeds check-script semantics (`latest_atr`/`prepare_check_regime` on the trailing window). A diff therefore isolates window-derived evaluator inputs; the shared context can never be the source.
- **Composed-signal comparison** — with close refs, the bt side composes `open_action` + `close_fraction` + position side exactly as `finalize_decision` does, so signal diffs isolate input drift, never the composition.

**Deliberately excluded:** #935's regression-index tests (they assert a test file exists and mentions a bug number, not that behavior is caught); #936's BYO-CSV single-path tool and its silent no-op `--regime-classifier` flag.

## Verification

- 13 `test_parity_diff.py` tests incl. new ones: close-evaluator parity clean **and exercised** (asserts the evaluator actually fired), composed-signal clean, frame-dependent detection survives the close-ref path, config-mode loader fields, fills legs, `backtest_effective_*` = prior-bar inputs.
- Full Python suite: **1190 passed**.
- Real cached data (BTC/USDT 1h): ad-hoc and config modes clean over 521 strided bars with `tiered_tp_atr` + 44 fills, exit 0; a `supertrend --regime` run over 1081 bars surfaced **1 genuine mismatch** — a trailing-window-vs-full-frame ADX boundary flip (`trending_down` vs `ranging`) — exactly the drift class the tool exists to catch.

Closes #906

---

### DIFF
--- a/backtest/AUDIT.md
+++ b/backtest/AUDIT.md
@@ -1,0 +1,138 @@
+# Backtest Subsystem Audit — 2026-06 (#906)
+
+Structured audit of correctness, parity, coverage, reporting, and debt across
+the backtesting subsystem (engine at the time of audit: 4,659 LOC across the
+8 core modules, 20 test files / ~1,180 passing tests). Findings were filed as
+cards #942–#945; this doc captures the verified state so future contributors
+don't re-derive it. Line numbers reference the audit-time tree — anchor on
+symbol names if they drift.
+
+## D1 — Look-ahead contract: INTACT
+
+The contract (`backtester.py` module docstring) holds for every decision
+input. Verified end to end:
+
+- `signal`, `_open_action`, `_close_fraction`, `regime` are all `shift(1)`'d
+  in the normalization blocks (`backtester.py:738, 748, 749, 780`) before the
+  per-bar loop reads them; fills use the current bar's `open` (contractual).
+- Close evaluators run end-of-bar against bar N's close mark and bar N's ATR
+  (`_evaluate_close_strategies`, dispatched at `backtester.py:1139`); their
+  output becomes `pending_close_fraction`, applied at bar N+1's open.
+- `_regime_bar_close` is snapshotted **before** the regime shift
+  (`backtester.py:762`) and is only fed to close evaluators as
+  `market_regime` — bar-N data for an end-of-bar decision, matching live.
+- Same-bar SL re-fire after a post-TP bump (the #715 class) is gated by
+  `sl_after_just_applied` (`backtester.py:1175`).
+- SL/TP intra-bar races resolve at bar close (no OHLC walking), documented in
+  the contract and deterministic: TP evaluators → SL bump → suppressed hit
+  check → next-bar fill.
+- Strategy sweep: no `shift(-n)`, `center=True` rolling, or forward `iloc`
+  in `shared_strategies/open/registry.py` or `shared_strategies/close/*.py`.
+  (Full-frame normalization — e.g. a signal keyed on the whole series' mean —
+  is *not* statically detectable; that class is what `parity_diff.py` exists
+  to catch.)
+- Composite (#861/#862) and shared-regime-bundle (#879) work did not bypass
+  the shift: only the `regime` column is a decision input; `adx`/`regime_score`
+  /±DI columns are display-only.
+
+Regression guards: `test_backtester_lookahead.py`, `test_post_tp_sl.py`
+(same-bar fire), `test_backtester_regime.py` (gate timing + live↔backtest
+label parity).
+
+## D2 — Parity with the live scheduler
+
+| Surface | State |
+|---|---|
+| Fees | ✅ `PLATFORM_FEE_PCT` matches `fees.go`; `test_platform_fees.py` scrapes the Go source so drift fails CI. deribit/ibkr/topstep flow through option/futures fee functions, not the spot table. |
+| Initial/trailing ATR SL (#885) | ✅ Same trigger formula (`entry ± mult×EntryATR`) and same-bar arming on both sides. |
+| Default tier ladders (#870/#887) | ✅ Values synced across Go and all three Python mirrors — but unpinned by tests (#944). |
+| Single close ref (#842) | ✅ `--config` rejects legacy `len>1` arrays with the same semantics live rejects them; the engine's max-wins multi-ref path remains for direct-constructor/test use only. |
+| `tiered_tp_atr_live_regime_dynamic` (#843) | ✅ Loudly rejected at config load (`run_backtest.py:193-196`); no evaluator registered under the name. |
+| `regime_directional_policy` (#822) | ✅ Loudly rejected (`run_backtest.py:160-166`). |
+| Regime-aware `sl_after` (#736) | ✅ Loudly rejected at `Backtester` init; scalar forms backtestable. |
+| `user_close_defaults` (#866) | ✅ `--defaults system\|user` mirrors the live three-layer resolution. |
+| Scale-in (#873), manual limit orders (#883) | Live-only **by design** (HL perps/manual execution mechanics, no signal-path component). Config keys are ignored by the backtest loader; acceptable while the features stay execution-side. |
+| **v13/v14 legacy close keys** | ❌ **#942** — the `--config` gate admits pre-v15 configs whose `tiers`/alias keys silently no-op in the Python evaluators while live canonicalizes them. |
+| **`direction` / `invert_signal` / `regime_window_divergence`** | ❌ **#943** — silently ignored by `--config`; the `regime_directional_policy` rejection message even recommends two of them. |
+
+## D3/D8 — Coverage
+
+- Every registered close strategy has at least one engine-level test except
+  the rejected dynamic variant (correct). `trailing_tp_ratchet_regime` and
+  `tiered_tp_atr_live_regime` are covered.
+- ADX 3-label vocabulary covered; **composite 7-label vocabulary has zero
+  backtest tests** (#944).
+- Sharpe annualization tested at 4 timeframes (1d/4h/1h/1w) — meets the ≥3 bar.
+- Variant depth: main engine ~strong (incl. shorts), pairs and options
+  moderate, **theta thin** (one force-close test, #944).
+- `DEFAULT_PARAM_RANGES` completeness is enforced by
+  `test_registry_loader.py::test_param_ranges_cover_every_registered_strategy`.
+- No property-based (`hypothesis`) tests anywhere; grid search is fully
+  deterministic so seed plumbing is moot.
+- Regression tests exist for every prior backtest bug: #302 (fills + vol
+  math), #303 (regime/BS parity), #304 M3/M5/L5 (annualization, calendar
+  days, theta force-close log), #715 (same-bar SL), #730 (look-ahead),
+  #824 (storage lazy-init).
+
+## D4/D5 — Reporting & optimizer
+
+- `periods_per_year()` drives Sharpe/volatility in the main engine; options
+  uses check-interval-aware sampling; pairs uses `bars_per_year`. Theta
+  hardcodes daily — valid only because theta data is always `1d` (#945).
+- Options annualized return uses elapsed calendar days (the #304 M5 fix is in
+  and commented).
+- Walk-forward separates train/test per fold and aggregates OOS-only stats
+  (`oos_mean_*`, `oos_std_return`). The boundary-bar inclusion at fold start
+  is **intentional** (its raw signal becomes row 1's shifted signal — matches
+  live) and is regression-tested in `test_walk_forward_warmup.py`; don't
+  "fix" it.
+- Optimizer is a deterministic grid; `optimize_metric` accepts any result-dict
+  key. No fold-to-fold robustness gate (param stability is reported via
+  `most_common_best_params`, not enforced).
+- Known gaps, not bugs: drawdown has no recovery-duration metric; the main
+  engine doesn't emit a separate `total_fees`; timestamps serialize as
+  `str(pd.Timestamp)` (space separator), not strict ISO-8601.
+
+## D6 — Debt
+
+- `close_registry_loader` shim is still required (open and close registries
+  both resolve as module name `registry`).
+- `backtest_options.py` exchange is configurable (`--exchange`, #304 L2 fixed).
+- HTF filter plumbs through the same `shared_tools/htf_filter.py` as live,
+  fed by cached candles; missing HTF cache fails open to neutral trend.
+- Residual items in **#945**: unused imports, dead close-name optimizer range
+  entries, theta daily-only comment.
+- Metrics/reporting logic is re-implemented per variant (~150–200 LOC
+  duplication across options/theta/pairs); extraction is optional until one
+  of them next drifts.
+
+## D7 — Observability
+
+- `parity_diff.py` (added by this audit, D7.4) replays the vectorized
+  backtest path and the trailing-window live path over the same candles and
+  emits a per-bar diff of signal / open_action / close_fraction / regime —
+  the tool to reach for first on any backtest-vs-paper mismatch. The live
+  side calls the actual check-script helpers (`prepare_check_regime` →
+  `evaluate_open_close`/`finalize_decision` with
+  `close_registry_loader.evaluate`), not a model of them; registry close
+  evaluators are compared through the same evaluator on both sides with a
+  shared simulated position context, so a diff isolates window-derived
+  inputs (trailing-window ATR/regime vs full-frame). `--config
+  <live-config> --strategy-id <id>` replays the exact live refs via the
+  #641 loader; `--fills` lines the decision diff up against the engine's
+  simulated entry/exit fills; `--csv`/`--jsonl` dump the full frame, which
+  also carries the post-`shift(1)` `backtest_effective_*` engine inputs per
+  bar. See the module docstring for usage.
+- Still absent (scope when needed): per-bar verbose trace inside
+  `Backtester.run`, equity-curve/trade-log CSV export, `exit_reason` and
+  entry/exit regime fields on `Trade`.
+
+## Known live-only surfaces (decision record)
+
+`scale_in` (#873), manual resting limit orders (#883), per-cycle regime
+hysteresis (`tiered_tp_atr_live_regime_dynamic`, #843),
+`regime_directional_policy` (#822), regime-aware `sl_after` (#736), and
+`regime_window_divergence` (#907) are execution/per-cycle mechanics with no
+bar-level equivalent; the first two are silently ignored by design (no signal
+path), the middle three are loudly rejected, and the last should be promoted
+to a loud rejection (#943).
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -42,6 +42,22 @@ SL/TP intra-bar races
 Bar-level granularity — when an SL hit and a TP fill could both occur within
 the same bar, the engine resolves them at bar close, not by intra-bar OHLC
 walking. Documented under ``Backtest`` in CLAUDE.md.
+
+Live parity limitations (#906 audit)
+------------------------------------
+Surfaces intentionally **not** modeled here (use ``backtest/parity_diff.py`` for
+decision-layer parity checks; see ``backtest/AUDIT.md`` for the full matrix):
+
+  • **Scale-in / pyramiding** (#873) — HL perps + manual live-only; same-direction
+    adds are skipped in backtest.
+  • **Resting manual limit orders** (#883) — maker fills / partial OID reconcile
+    have no bar-level simulation.
+  • **``tiered_tp_atr_live_regime_dynamic``** (#843) — rejected at
+    ``run_backtest.load_strategy_config`` (on-chain regime hysteresis).
+  • **``regime_directional_policy``** (#822) — rejected at config load; use static
+    ``direction`` / ``invert_signal`` for backtests.
+  • **Inline trailing SL at open** (#885) — live arms same-cycle; backtest seeds
+    trailing/ratchet triggers on the bar after open (no naked-gap modeling).
 """
 
 import sys
--- a/backtest/parity_diff.py
+++ b/backtest/parity_diff.py
@@ -51,10 +51,15 @@ omission: it needs a second-timeframe fetch and is orthogonal to
 frame-dependence.)
 
 When close refs are configured, both sides share a single simulated
-position context (side / avg_cost / quantities / entry_atr / entry
-regime), seeded from the backtest-effective open/close decisions. The
-context is deliberately identical on both sides so it can never be the
-source of a diff — only the evaluator's window-derived inputs can.
+position lifecycle (side / avg_cost / quantities / entry values),
+seeded from the backtest-effective open/close decisions — shared so the
+position itself can never be the source of a diff. The dict shape each
+evaluator sees mirrors its real caller: the backtest side always passes
+``regime`` (possibly empty) and ``entry_atr``, exactly like
+``Backtester._evaluate_close_strategies`` (#747); the live side omits
+empty keys, exactly like the check scripts. The asymmetry is deliberate
+— a diff at empty-regime bars surfaces that genuine engine-vs-live
+shape divergence instead of masking it.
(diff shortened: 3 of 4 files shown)


richkuo:
@claude

richkuo:
All review findings addressed in a942301:

1. **Platform auto-detect (Needs Fixing 1)** — `--platform` now defaults to `None`; `--config` mode passes `args.platform or ""` so perps/manual auto-detect `hyperliquid`, explicit `--platform` still wins, ad-hoc mode resolves `args.platform or "binanceus"`. Parametrized tests cover all four must-survive cases (perps, manual, futures, explicit override) plus a CLI-plumbing regression that runs `main()` exactly as the shell does and asserts `extract_fills` receives `hyperliquid`.
2. **Zero-bar runs (Needs Fixing 2)** — `main` now exits 2 with a stderr explanation when `bars_compared == 0`; docstring exit-code contract updated ("a vacuous comparison is not agreement"). Regression test feeds 40 candles to a 200-bar window.
3. **Normalizer parity (Recommended Optional)** — the live no-close path now uses the same float-aware `_normalize_signal` as the backtest side; the truncating import was removed. Regression test registers a 0.5/-0.5-emitting strategy and asserts a clean diff.
4. **Engine dict shape (filed as fix, not follow-up)** — `_bt_close_evaluator_fraction` now builds engine-shaped dicts mirroring `Backtester._evaluate_close_strategies` (#747): `regime` always present (possibly empty) in both position and market dicts, `entry_atr` always a float. The live side keeps the check-script omit-when-empty shape — the asymmetry is deliberate and documented so the tool can surface a genuine engine-vs-live divergence at empty-regime bars instead of masking it. Test asserts the dict shape via a stubbed evaluator.

Full suite: 1199 passed. Real-data smoke (BTC/USDT 1h, 1195 bars, --fills): clean, exit 0.

---

richkuo:
@claude review

richkuo:
Both round-2 findings fixed directly in 1e38fb3 (rather than filed as follow-ups):

1. **Decision-bar entry-regime stamp** — `_simulate_position_contexts` now stamps the entry regime from bar i-1's label, mirroring the engine's shifted-regime `_entry_stamp` (backtester.py:796-802) and the label live computes alongside the signal. New test pins the scaffold's stamp to the decision bar, not the fill bar.
2. **Strict signal contract** — `_normalize_signal` now mirrors the engine's `{-1, 0, 1}` rejection on both paths: NaN → 0, any other out-of-domain value (fractional or integer) raises with a message explaining the engine-vs-live divergence. The fractional-signal test now asserts the loud rejection instead of a coerced CLEAN; a contract test covers 0.5/-0.5/2/-3/NaN/float64(1.0).

Full suite: 1201 passed. Real-data smoke (BTC/USDT 1h, --regime): clean.

---

richkuo:
@claude review

richkuo:
Both round-3 findings fixed directly in ab0b55d:

1. **Non-registry close refs** — chose the loud-rejection option, because it is what the engine itself does: `Backtester.__init__` refuses unknown close names (`Unknown close strategy: ...`), so a signal-strategy close ref has no engine path to diff against. `compute_parity_frame` now raises the same error upfront (CLI exits 2), the silent `except ValueError: continue` is gone (remaining evaluator errors propagate like the engine), and the module docstring documents the live-only `legacy_close_fraction_from_signal` surface. The test pins tool and `Backtester` to the same rejection.
2. **Entry-ATR plausibility guard** — the scaffold now mirrors `_stamp_entry_atr` / Go's `stampEntryATRIfOpened`: non-positive or > 50%-of-entry-price ATR stamps 0.0, so the shared context matches both real sides at extreme-volatility bars. Test constructs ATR ≈ 180 on a ~100 price and asserts the context omits `entry_atr`.

Full suite: 1203 passed. Real-data smoke (BTC/USDT 1h, tiered_tp_atr): clean.

---

richkuo:
@claude rereview

richkuo:
Round-4 finding fixed directly in e420cb2 (no follow-up issue):

**Fill-bar open + slippage for the scaffold entry price** — `_simulate_position_contexts` now stamps `avg_cost` from the fill bar open (fallback close) adjusted by the engine slippage — `+slippage` for longs, `-slippage` for shorts — mirroring `effective_price` in `Backtester.run`, and the entry-ATR plausibility guard now thresholds against that same effective price, matching the `_stamp_entry_atr` call site. The slippage constant is read from the `Backtester.__init__` signature default at import, so it cannot drift out of sync. New test pins long and short `avg_cost` to `open * (1 ± 0.0005)` and asserts it is NOT the bar close.

Full suite: 1204 passed. Real-data smoke (BTC/USDT 1h, tiered_tp_atr + regime): clean.

---

richkuo:
@claude rereview

richkuo:
Round-5 finding fixed directly in 0e3b2f9 (no follow-up issue) — took the fold-back option, plus two adjacent fidelity fixes it forced into the open:

1. **Ladder advancement** — the bt-side registry evaluator fraction at bar i now feeds bar i+1 quantity decrement, exactly the engine pending_close_fraction. `_simulate_position_contexts` returns `(contexts, registry_fractions)` and `compute_parity_frame` compares the very fraction the accounting applied, so they can never disagree. The context stays shared (live evaluates with the same decremented quantities), so tier-selection drift still surfaces while the accounting matches what both real paths book: 0.4 once, then 0, then the tier-2 increment.
2. **Current-size booking** — decrements now book against the CURRENT position (engine: `qty_to_close = abs(position) * fraction`); the old initial-size decrement was a latent mismatch on the column path too.
3. **Context timing** — `contexts[i]` is now the position AFTER bar i open fills, the state the engine holds during end-of-bar close evaluation (backtester.py:1150-1156) and what the live bar-i cycle sees; previously it lagged one bar.

New ladder test pins tier 1 firing once, quantity 1.0 → 0.6 → 0.2, and the tier-2 increment expressed against current size (2/3); the three scaffold tests updated for the corrected timing. Full suite: 1205 passed. Real-data smoke (BTC/USDT 1h, tiered_tp_atr + regime + fills): clean.

---

richkuo:
@claude sonnet rereview

Issue #943, opened 2026-06-10
parity(backtest): --config silently ignores direction / invert_signal / regime_window_divergence


richkuo:
Found by the #906 audit (D2.1/D2.5).

## Current behavior

`load_strategy_config` (`backtest/run_backtest.py:119-223`) reads only `open_strategy` / `close_strategy` refs from a live config. Three strategy-level fields the live scheduler honors are silently dropped:

1. **`direction`** — live gates which sides may open; backtest opens whatever the strategy emits. A `direction: "long"` config backtests with shorts included.
2. **`invert_signal`** — live flips BUY↔SELL via `applySignalInversion` before `Direction`; backtest ignores it entirely. Worse, the `regime_directional_policy` rejection message (`backtest/run_backtest.py:165`) explicitly tells the operator to *"Use the static `direction` / `invert_signal` fields for backtesting"* — recommending fields the backtest path does not read.
3. **`regime_window_divergence`** (#907) — HL-live-only, but unlike its siblings (`regime_directional_policy` rejected at `run_backtest.py:160-166`, `tiered_tp_atr_live_regime_dynamic` rejected at `193-196`) it is silently ignored rather than loudly rejected.

## Proposed fix

- Implement `direction` and `invert_signal` in the backtest config path — both are simple frame transforms (mask disallowed-side signals; multiply signal by −1) applied after `apply_strategy`, mirroring live ordering (inversion before direction gating).
- Reject `regime_window_divergence` at config load with the same wording pattern as the `regime_directional_policy` rejection.
- Until/unless the first item lands, correct the `run_backtest.py:165` message so it stops recommending unimplemented fields.

## Invariant

Every strategy-level field that changes live entry/exit decisions is either modeled by the backtester or loudly rejected at config load — never silently dropped.

---
Created with LLM: Fable 5 | high | Harness: Claude Code

richkuo:
Closing as consolidated into issue 942, which now carries this full scope (direction / invert_signal implementation, regime_window_divergence loud rejection, message fix) alongside the v15 gate bump — same file, same invariant, one PR.

Updated with LLM: Fable 5 | high | Harness: Claude Code

PR #951, opened 2026-06-10
parity(backtest): --config v15 gate + direction/invert_signal + regime_window_divergence

### LINKED ISSUE #942: parity(backtest): --config loader — legacy close keys no-op silently; direction/invert_signal/regime_window_divergence ignored

richkuo:
Closes #942 (consolidates #943).

**Part 1 (D2.8)**: bump the --config gate to config_version 15 so pre-v15 legacy close keys (tiers/atr/multiple/fraction) cannot silently no-op vs live.

**Part 2 (D2.1/D2.5)**: apply direction + invert_signal inside Backtester.run (invert before direction gating, path-aware masking); reject short/both with no close evaluator; reject regime_window_divergence (#907) loudly.

**Behavior-change note (spot/futures `--config`)**: `_effective_direction` returns `"long"` for non-perps types (they are long-by-construction in live, matching `EffectiveDirection`). For a spot/futures `--config` backtest that has a close evaluator (the open/close engine path), this now masks `signal<0` short-opens — those `-1`s previously opened erroneous shorts, so existing spot/futures `--config` numbers can shift. The new behavior is the correct one (spot can't short) and is pinned by end-to-end regression tests. The compound case this exposes — a stray `invert_signal` on a non-perps `--config` — is now rejected at load to mirror live, where `LoadConfig` (config.go) refuses such a config at startup (`invert_signal` is HL-perps/manual-only); previously the backtester would have silently applied the inversion.

All Python tests pass (`shared_strategies/ shared_tools/ backtest/`: 953 passed).

---

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -325,7 +325,9 @@ class Backtester:
                  trailing_stop_pct: Optional[float] = None,
                  stop_loss_atr_regime: Optional[dict] = None,
                  trailing_stop_atr_regime: Optional[dict] = None,
-                 strategy_type: str = "perps"):
+                 strategy_type: str = "perps",
+                 direction: Optional[str] = None,
+                 invert_signal: bool = False):
         """
         Args:
             initial_capital: Starting portfolio value.
@@ -397,6 +399,13 @@ class Backtester:
         self.trailing_stop_atr_mult = trailing_stop_atr_mult
         self.trailing_stop_pct = trailing_stop_pct
         self.strategy_type = strategy_type
+        # #942: live strategy-level entry transforms the backtester must mirror
+        # so --config doesn't silently diverge from the daemon. ``invert_signal``
+        # flips BUY<->SELL; ``direction`` gates which side may open. Both are
+        # applied to the raw signal in ``run()`` (see _apply_direction_invert),
+        # mirroring the live order (applySignalInversion before EffectiveDirection).
+        self.direction = (str(direction).strip().lower() if direction else None)
+        self.invert_signal = bool(invert_signal)
         self.stop_loss_atr_regime = (
             dict(stop_loss_atr_regime) if stop_loss_atr_regime else None
         )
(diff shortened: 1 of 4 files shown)


richkuo:
@claude

richkuo:
@claude fix review issues

richkuo:
@claude sonnet rereview

PR #962, opened 2026-06-10
fix(strategies): seed supertrend band recursion from first non-NaN ATR row

### LINKED ISSUE #961: supertrend strategy is dead code: band recursion seeds from NaN, never emits a signal
Found during the #956 audit: `supertrend_strategy` (shared_strategies/open/registry.py) emits **zero signals on any data** — the band recursion seeds from NaN (first ATR rows) and never escapes, so upper/lower bands stay NaN and no cross ever fires. Verified empirically across BTC/ETH/SOL 1h/4h, 12 months: 0 trades everywhere.

## Fix

Seed the recursion from the first non-NaN ATR row (or skip-forward initialization), add a signal-value regression test asserting supertrend produces signals on trending synthetic data, and include it in the smoke test that iterates registered strategies.

---

richkuo:
Closes #961

## Bug
`supertrend_strategy` (shared_strategies/open/registry.py) emitted zero signals on any data. The band recursion started at `i=1`, where `final_upper`/`final_lower` are NaN (rolling ATR warmup, `atr_period-1` bars). NaN comparisons are always False, so both update conditions failed and the "carry previous band" branch propagated NaN forward forever: bands stayed NaN on every bar, `close > NaN` was always False, direction locked at -1, and no flip ever fired.

## Reproduction
On 300 bars of synthetic three-leg trend data (up 100→200, down 200→120, up 120→220, DatetimeIndex): **0 buys, 0 sells, `supertrend` column 100% NaN, direction values `{0, -1}`**.

## Fix
Seed the recursion from the first non-NaN ATR row (skip-forward initialization): find the first valid ATR index, keep the basic-band values as the seed there, and start the carry-forward loop from the next bar. Inputs shorter than the ATR window return cleanly with zero signals. Standard supertrend semantics, function signature, and registry metadata unchanged.

After the fix, the same synthetic data yields buy at bar 14, sell at bar 106, buy at bar 204 — one flip per trend leg — and the `supertrend` line is NaN only during the 9-bar ATR warmup.

## Registry contract
Spot and futures `--list-json` snapshots taken before and after: **byte-identical** (verified with `diff`).

## Tests
New `shared_strategies/open/test_supertrend.py` (4 tests): nonzero buy AND sell signals on trending data, exact signal values/positions and direction per leg, NaN-warmup escape (`isna().sum() == 9`, all valid after), short-input edge case (all-NaN ATR → zero signals). The existing registry parity smoke test iterates every registered strategy (supertrend included) and now exercises the fixed path.

Full suite: `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` → **1283 passed**.

## Sanity backtest
`backtest/run_backtest.py --strategy supertrend --symbol BTC/USDT --timeframe 4h --since 2025-06-10 --mode single` (was 0 trades everywhere pre-fix):

- Total Trades: 30
- Total Return: -26.05% (final capital $739.49 on $1,000)
- Win Rate: 43.3% | Profit Factor: 0.583
- Sharpe: -1.013 | Max Drawdown: -35.42%

The strategy now trades; raw single-mode performance on this window is poor, which is a tuning question, not a correctness one. Note: the backtest run needed `PYTHONPATH=.` because `post_tp_sl.py`'s absolute `shared_strategies.*` import isn't on `sys.path` when invoking `backtest/run_backtest.py` directly — pre-existing on main, untouched here.

---
Created with LLM: Fable 5 | high | Harness: agent

### DIFF
--- a/shared_strategies/open/registry.py
+++ b/shared_strategies/open/registry.py
@@ -474,7 +474,18 @@ def supertrend_strategy(df: pd.DataFrame, atr_period: int = 10, multiplier: floa
     final_lower = basic_lower.copy()
     direction = pd.Series(0, index=result.index, dtype=int)
 
-    for i in range(1, n):
+    # Seed the recursion from the first non-NaN ATR row; the rolling ATR is NaN
+    # for the first atr_period-1 bars and NaN comparisons are always False, so
+    # starting at i=1 would carry NaN bands forward forever and never emit a signal.
+    atr_valid = atr.notna().to_numpy()
+    if not atr_valid.any():
+        result["supertrend"] = np.nan
+        result["st_direction"] = direction
+        result["signal"] = 0
+        return result
+    start = int(atr_valid.argmax())
+
+    for i in range(start + 1, n):
         if basic_upper.iloc[i] < final_upper.iloc[i-1] or result["close"].iloc[i-1] > final_upper.iloc[i-1]:
             final_upper.iloc[i] = basic_upper.iloc[i]
         else:
--- a/shared_strategies/open/test_supertrend.py
+++ b/shared_strategies/open/test_supertrend.py
@@ -1,0 +1,101 @@
+"""Regression tests for supertrend_strategy in registry.py.
+
+Issue #961: the supertrend band recursion seeded from the NaN rows of the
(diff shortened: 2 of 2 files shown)


richkuo:
@claude

Issue #956, opened 2026-06-10
Strategy audit: rank all registered entry strategies, flag deprecation candidates


richkuo:
Part of #955.

## Task

Rank every registered open strategy (`shared_strategies/open/registry.py`, ~35 entries) and flag deprecation candidates.

## Method

- **Live**: leaderboard / lifetime trade stats per strategy where data exists.
- **Backtest**: standardized runs — BTC/ETH/SOL, 1h + 4h, last 12 months, `--mode single`, default close stack via `backtest/run_backtest.py`.
- Rank by out-of-sample Sharpe and max-drawdown-adjusted return; flag negative expectancy or Sharpe < 0.

## Deliverable

- Audit report (table: strategy, platform tags, Sharpe, expectancy, max DD, trade count) posted on this issue.
- Deprecation list with rationale; removals respect the registry `--list-json` byte-identical contract for Go `discoverStrategies` (deprecate via config/docs first, removal is a separate PR).
- Confirms or cuts the five candidate strategies listed in #955 before their issues are opened.

## Follow-up deliverable: render the audit as an HTML report in the dashboard UI

The audit report posted in the comment below must also be surfaced as an easy-to-read HTML page inside the app's UI dashboard, under a new **"Reports"** section.

### Requirements

- **New route** served by `scheduler/server.go` (alongside `/dashboard`, loopback-only, same auth posture as the rest of the UI): a reports index plus the strategy-audit report itself. Suggested: `/reports` (index listing available reports) + `/reports/strategy-audit` (this report).
- **Static, self-contained HTML** styled to match the existing dashboard (`scheduler/static/ui/styles.css`) — readable table of the full ranking (strategy, registry, mean Sharpe, mean return %, vs B&H %, worst max DD %, trades, verdict), the deprecation candidates list, the out-of-sample check table, and the five-candidate verdicts. Sortable columns and color-coded verdicts (keep / watch / deprecate / bug) are nice-to-haves.
- **Navigation**: add a "Reports" link/tab to the dashboard (`scheduler/static/ui/index.html` / `app.js`) so it's reachable from the existing UI, not only by direct URL.
- **Content source**: render from the audit data so the page stays the single source of truth for the report (don't hand-maintain a second copy of the numbers). A small embedded JSON/Go data structure feeding a template is fine; the comment table is the reference content.
- Keep it within the repo's UI conventions (loopback-only bind, no external assets, `version` surface unaffected). No new platform or risk-machinery changes.

### Deliverable

[...]

Issue #955, opened 2026-06-10
Strategy quality initiative: audit existing entries, add higher-edge strategies


richkuo:
## Problem

The open-strategy registry (`shared_strategies/open/registry.py`) holds ~35 strategies, but most are single-indicator classics (SMA/EMA cross, RSI, MACD, Bollinger, PSAR). Live performance is mediocre: edge comes mostly from the close-side machinery (tiered TP, ratchets, regime SL) rather than entry quality.

**This is a tracking umbrella** — work happens in child issues; this closes when all children do.

## Child issues

1. #956 — Audit: rank all existing strategies, flag deprecation candidates. **Done — report posted.** 13 underperformers flagged for reference only — keeping all existing strategies (owner decision, 2026-06-10); the ranking just informs allocation.
2. #957 — Multi-timeframe confluence (HTF trend gate + LTF pullback). **Confirmed by audit.**
3. #958 — Regime-adaptive entry (breakout vs mean-reversion switch from composite metrics). **Confirmed.**
4. #959 — Volatility-targeted momentum (ATR-normalized, efficiency-confirmed, bidirectional). **Confirmed.**
5. #960 — Funding-rate aware perps entry. **Confirmed, blocked on funding data in backtester (in scope of #960).**
6. ~~Volume-profile / liquidity levels~~ — **Cut by audit**: the existing structure/liquidity family ranks mid-to-bottom with no out-of-sample promise.

Related: #961 — supertrend is dead code (found during audit).

## Validation bar (applies to every new strategy)

Registry entry + `PLATFORM_ORDER`; `knownShortNames` + defaults in `init.go`; `DEFAULT_PARAM_RANGES` in `backtest/optimizer.py`; signal-value tests; backtest report with in/out-of-sample split in the PR body; bidirectional ones added to `bidirectionalPerpsStrategies`. Must beat the median incumbent on out-of-sample Sharpe and max-drawdown-adjusted return. No look-ahead (signal at bar N fills at N+1 open).

## Non-goals

No new platforms; no changes to the close-strategy stack or risk machinery. Registry `--list-json` stays byte-identical for existing strategies.

---
Created with LLM: Fable 5 | high | Harness: Claude Code
Updated with LLM: Fable 5 | high | Harness: Claude Code

richkuo:
[...]

PR #963, opened 2026-06-10
feat(strategy): mtf_confluence — HTF trend gate over LTF pullback entries (#957)

### LINKED ISSUE #957: New strategy: multi-timeframe confluence (HTF trend gate + LTF pullback entry)
Part of #955. Confirmed by the audit (#956): the dominant failure mode in the 12-month bear window was high-churn long entries fighting the higher-timeframe downtrend, and 4h consistently beat 1h.

## Strategy

Multi-timeframe confluence as a first-class registry strategy: higher-timeframe trend filter (e.g. 4h EMA slope / structure) gates lower-timeframe pullback entries (1h), instead of the bolt-on `htf_filter`. Long and short variants (perps `variants={"futures":{...}}`).

## Validation bar (from #955)

Registry entry + `PLATFORM_ORDER`; `knownShortNames` + defaults in `init.go`; `DEFAULT_PARAM_RANGES` in `backtest/optimizer.py`; signal-value tests; backtest report with in/out-of-sample split in the PR body; bidirectional → `bidirectionalPerpsStrategies`. Must beat the median incumbent on out-of-sample Sharpe and DD-adjusted return. No look-ahead.

---

richkuo:
Closes #957. Part of #955.

## Design

- **One frame, two timeframes.** The framework passes a single OHLCV frame; the higher-timeframe (HTF) view is derived in-process via an epoch-aligned pandas `resample` to `htf_factor` × the native bar (1h → 4h by default). No extra data fetch. Non-datetime indexes fall back to positional bucketing.
- **HTF trend gate with hysteresis.** Strict *entry* gate: HTF EMA(20) > EMA(40) by ≥ `htf_sep_pct` AND the fast EMA rising bucket-over-bucket (mirror for down). Loose *hold* gate: fast still on the right side of slow. Hysteresis exists because the first iteration (exit on any slope flicker) churned 102 trades and -44% on BTC 1h; with it, trade count drops ~6× and the entry edge survives fees.
- **LTF pullback entry.** Within an HTF uptrend: a prior bar tagged the native EMA(20) within `pullback_window` bars, and the current bar resumes (close reclaims the EMA and breaks the prior bar's high). Short is the exact mirror.
- **Exit** when the HTF EMAs cross against the held side (position-state transition, clamped to {-1,0,1} like `tema_cross_bd`).
- **Long-only base + bidirectional futures variant** via `variants={"futures": {"default_params": {"allow_short": true}}}`.

### Anti-look-ahead mechanism

An HTF bucket labeled `L` spanning `[L, L+H)` becomes readable only at the native bar labeled `L+H-p` — the bar whose own close coincides with the bucket close. The HTF indicator series is re-indexed to those visible-at labels and forward-filled onto the native index, so the in-progress trailing bucket maps **past the end of the frame** and is never read. Buckets are epoch-aligned, so bucketing is independent of where the fetch window starts (live/backtest parity). A signal at bar N may use bar N's own close, matching the engine's fill-at-N+1-open contract (`backtest/tests/test_backtester_lookahead.py`).

Pinned by three regression tests: prefix-consistency (`core(df[:k]) == core(df)[:k]` for every intra-bucket offset), future-row perturbation (crash/moon all rows after a cutoff → signals at/before it unchanged), and a mid-bucket `htf_trend` immutability test.

## Wiring checklist (#955 validation bar)

- [x] `shared_strategies/open/registry.py` — `@register` + `PLATFORM_ORDER` (spot + futures, before `hold`)
- [x] `--list-json` byte-identical for existing strategies in both shims (verified before/after; the only change is the additive `mtf_confluence` entry)
- [x] `scheduler/init.go` — `knownShortNames` (`mtfc`), `defaultSpotStrategies`/`defaultPerpsStrategies`/`defaultFuturesStrategies`, `bidirectionalPerpsStrategies`
- [x] `backtest/optimizer.py` — `DEFAULT_PARAM_RANGES["mtf_confluence"]`
- [x] Signal-value tests (DatetimeIndex, actual values) + look-ahead tests — `shared_strategies/open/test_mtf_confluence.py` (15 tests)

## Test evidence

- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — **1294 passed**
- `go -C scheduler build .` + `go -C scheduler test ./...` — **ok** (gofmt clean)

## Backtest report

Single mode, default params, `binanceus` fee model (audit-identical harness), long leg (spot config — see caveats). DDadj = return / |max DD|. Median incumbent = per-dataset median of the audit's OOS top 8 (`momentum_pro`, `chart_pattern`, `squeeze_momentum`, `mean_reversion_pro`, `donchian_breakout`, `ichimoku_cloud`, `range_scalper`, `sma_crossover`) re-run on the identical harness/window.

**In-sample (since 2025-06-10):**

| Dataset | Return | Sharpe | Max DD | DDadj | Trades | med-inc Sharpe | med-inc DDadj | B&H | Beats median? |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---|
| BTC 1h | -9.5% | -0.31 | -31.0% | -0.31 | 16 | -1.28 | -0.84 | -44.2% | yes |
| BTC 4h | -13.2% | -0.64 | -27.3% | -0.48 | 5 | -0.31 | -0.33 | -43.8% | no |
| ETH 1h | +11.5% | +0.48 | -38.4% | +0.30 | 16 | +0.40 | +0.22 | -39.8% | yes |
| ETH 4h | +16.7% | +0.64 | -27.2% | +0.61 | 4 | +0.16 | +0.01 | -39.5% | yes |
| SOL 1h | -34.9% | -0.80 | -51.0% | -0.68 | 20 | -0.98 | -0.76 | -60.9% | yes |
| SOL 4h | -12.2% | -0.15 | -42.9% | -0.28 | 5 | -0.17 | -0.38 | -60.6% | yes |
| **Mean** | **-6.9%** | **-0.13** | | **-0.14** | 66 | **-0.36** | **-0.35** | -48.1% | **5/6** |

**Out-of-sample (since 2026-01-01):**

| Dataset | Return | Sharpe | Max DD | DDadj | Trades | med-inc Sharpe | med-inc DDadj | B&H | Beats median? |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---|
| BTC 1h | +1.8% | +0.29 | -14.9% | +0.12 | 6 | -1.13 | -0.59 | -30.2% | yes |
| BTC 4h | +4.8% | +0.69 | -9.6% | +0.50 | 1 | -0.30 | -0.25 | -30.2% | yes |
| ETH 1h | -9.1% | -0.50 | -20.3% | -0.45 | 7 | -0.21 | -0.23 | -45.7% | no |
| ETH 4h | -9.3% | -0.80 | -10.9% | -0.86 | 2 | -0.42 | -0.44 | -45.7% | no |
| SOL 1h | -21.9% | -1.36 | -29.5% | -0.74 | 8 | -1.64 | -0.80 | -49.9% | yes |
| SOL 4h | -9.7% | -1.20 | -14.3% | -0.68 | 2 | -0.84 | -0.66 | -50.0% | no |
| **Mean** | **-7.2%** | **-0.48** | | **-0.35** | 26 | **-0.76** | **-0.49** | -42.0% | **3/6** |

### The bar (beat the median incumbent on OOS Sharpe + DD-adjusted return)

**Cleared on aggregate and on the audit's reference dataset; mixed per-dataset.**

- Aggregate OOS: Sharpe **-0.48 vs -0.76** and DDadj **-0.35 vs -0.49** — both clear.
- Audit's published OOS benchmark (#956, BTC 1h): median incumbent Sharpe **-1.155** / return **-15.15%**; mtf_confluence posts **+0.29** / **+1.8%** — clears decisively.
- Per-dataset OOS it wins 3/6 on both metrics (BTC 1h, BTC 4h, SOL 1h) and loses ETH 1h/4h + SOL 4h. ETH OOS is the weak spot: the Jan–Jun 2026 ETH bear had rallies just strong enough to flip the 4h gate up, then failed.
- Tuning hygiene: parameters were swept only on 2025-06-10 → 2025-12-31 (OOS window untouched); the chosen defaults sit in a flat neighborhood of the tuning surface (fast 20–26 / slow 40 / window 4–8 all within ±0.01 Sharpe).

### Caveats / deviations

1. **Long leg only, like the audit.** The plain backtest path is long/flat, so the report uses the long-only configuration (`allow_short=False`, the spot default). The futures variant's short leg — expected to be the stronger leg in this bear window — is unmeasured, the same caveat #956 applies to every `bidirectionalPerpsStrategies` member. Running the futures variant through the long/flat path scores it *worse* (OOS mean Sharpe -0.35 in result-dict scale) purely because `+1` close-short transitions get misread as long entries — an engine-representation artifact shared with `tema_cross_bd`/`triple_ema_bidir`, not a strategy property.
2. **Audit-table Sharpe scale.** `run_backtest.py` CLI prints timeframe-annualized Sharpe; all numbers above use that same scale for both mtf_confluence and the incumbent medians (recomputed on the identical harness rather than trusting cross-run drift).
3. `backtest/run_backtest.py` currently needs `PYTHONPATH=.` from the repo root (pre-existing on main, reproduced there: `post_tp_sl.py` does a `shared_strategies.…` absolute import at `Backtester.__init__`); not fixed here to keep the PR scoped.
4. `DEFAULT_PARAM_RANGES` warmup note: the slow-EMA lookback is in HTF *buckets* (real need ≈ `htf_factor × htf_ema_slow` native bars), which `max_indicator_lookback`'s raw-int heuristic can't see — safe because the strategy holds the HTF trend neutral until primed (documented in the grid entry).

---
Created with LLM: Fable 5 | high | Harness: agent

### DIFF
--- a/backtest/optimizer.py
+++ b/backtest/optimizer.py
@@ -462,6 +462,18 @@ DEFAULT_PARAM_RANGES = {
         "rsi_oversold": [25.0, 30.0, 35.0],
         "rsi_overbought": [65.0, 70.0, 75.0],
     },
+    "mtf_confluence": {
+        "htf_factor": [3, 4, 6],
+        "htf_ema_fast": [10, 20, 26],
+        # NOTE: the real warmup need is ~htf_factor × htf_ema_slow NATIVE bars
+        # (the slow EMA counts HTF buckets), which max_indicator_lookback's
+        # raw-int heuristic can't see. Safe regardless: the strategy holds the
+        # HTF trend neutral until primed, so under-provisioned folds just
+        # trade less at the start — never on look-ahead or junk EMAs.
+        "htf_ema_slow": [40, 50],
+        "ltf_ema": [13, 20],
+        "pullback_window": [4, 6, 8],
+    },
     "consolidation_range": {
         "box_width_pct": [0.03, 0.05, 0.08, 0.10],
         "min_bars": [12, 16, 24],
--- a/scheduler/init.go
+++ b/scheduler/init.go
@@ -75,6 +75,7 @@ var knownShortNames = map[string]string{
 	"momentum_pro":          "mompro",
 	"mean_reversion_pro":    "mrpro",
 	"consolidation_range":   "cr",
+	"mtf_confluence":        "mtfc",
 }
 
 // bidirectionalPerpsStrategies lists strategy IDs that emit signal=-1 as a
@@ -93,6 +94,7 @@ var bidirectionalPerpsStrategies = map[string]bool{
 	"momentum_pro":        true, // emits short on stacked-bearish-EMA trend-pullback breakdowns
 	"mean_reversion_pro":  true, // emits short on overbought reversion in no-trend regimes
 	"consolidation_range": true, // emits short at the top edge of a consolidation box (range-edge mean-reversion)
+	"mtf_confluence":      true, // futures variant (allow_short) shorts LTF pullback rallies in HTF downtrends (#957)
 }
 
 func isBidirectionalPerpsStrategy(id string) bool {
@@ -141,6 +143,7 @@ var defaultSpotStrategies = []stratDef{
 	{ID: "tema_cross", ShortName: "temac"},
 	{ID: "momentum_pro", ShortName: "mompro"},
 	{ID: "mean_reversion_pro", ShortName: "mrpro"},
+	{ID: "mtf_confluence", ShortName: "mtfc"},
 }
 
 var defaultOptionsStrategies = []stratDef{
@@ -164,6 +167,7 @@ var defaultPerpsStrategies = []stratDef{
 	{ID: "session_breakout", ShortName: "sbo"},
 	{ID: "momentum_pro", ShortName: "mompro"},
 	{ID: "mean_reversion_pro", ShortName: "mrpro"},
+	{ID: "mtf_confluence", ShortName: "mtfc"},
 }
 
 var defaultFuturesStrategies = []stratDef{
@@ -190,6 +194,7 @@ var defaultFuturesStrategies = []stratDef{
 	{ID: "tema_cross_bd", ShortName: "temacb"},
 	{ID: "momentum_pro", ShortName: "mompro"},
 	{ID: "mean_reversion_pro", ShortName: "mrpro"},
+	{ID: "mtf_confluence", ShortName: "mtfc"},
 }
 
 // Supported CME futures symbols for the init wizard.
--- a/shared_strategies/open/mtf_confluence.py
+++ b/shared_strategies/open/mtf_confluence.py
@@ -1,0 +1,254 @@
+"""
+MTF Confluence — higher-timeframe trend gate over lower-timeframe pullback
+entries (#957).
+
+The strategy audit (#956) showed the dominant failure mode across the registry
+was high-churn entries fighting the higher-timeframe trend, with 4h runs
+consistently beating their 1h twins. This strategy makes the multi-timeframe
+gate first-class instead of the bolt-on ``htf_filter`` — and it derives the
+higher-timeframe view from the SINGLE frame the framework already passes
+(resampled in-process; no extra data fetch).
+
+Rules (long; short is the mirror, gated by ``allow_short``)
+-----------------------------------------------------------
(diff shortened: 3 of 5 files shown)


richkuo:
@claude

PR #965, opened 2026-06-11
[codex] Allow safe multi-trailing HL peers

### LINKED ISSUE #964: [hl] Allow multiple trailing-stop owners on shared Hyperliquid coins (relax #601 peer validation)
## Problem

`hyperliquidPeerStrategyErrors` rejects any config where two or more strategies on the same Hyperliquid coin both own a trailing stop:

> hyperliquid peers on ETH have multiple trailing-stop owners (strategies A, B, C): trailing_stop_pct/trailing_stop_atr_mult cancel and replace OIDs each cycle, racing on the shared on-chain position; at most one peer may run a trailing stop

Current code confirms this guard at `scheduler/config.go:1214-1224`. Trailing ownership is detected for `trailing_stop_pct`, `trailing_stop_atr_mult`, and `trailing_stop_atr_regime` at `scheduler/config.go:1103-1113`.

This blocks legitimate configs where multiple strategies on the same coin each need per-strategy trailing reduce-only protection, e.g. one RMC strategy, one TEMA strategy, and one manual hold all on ETH perps, all using `trailing_tp_ratchet_regime`.

## Verified current behavior

- `trailing_tp_ratchet_regime` requires `trailing_stop_atr_regime`; scalar `trailing_tp_ratchet` requires `trailing_stop_atr_mult > 0` (`scheduler/trailing_tp_ratchet.go:263-272`). Any strategy using either close therefore necessarily owns trailing.
- Startup validation and hot-reload validation both run `hyperliquidPeerStrategyErrors` (`scheduler/config.go:1845-1852`, `scheduler/config_reload.go:337-340`). Current `origin/main` is not more permissive at startup; a currently running multi-trailing process likely predates this guard or was started from different code/config state.
- Hot reload validates the full next config before applying individual hot-reloadable fields. If the next config still has multiple trailing owners, reload is rejected and the live process keeps the old in-memory config (`scheduler/main.go:512-517`).
- Fixed-distance SL owners are allowed to coexist across same-coin peers because those orders are per-strategy sized and do not continuously cancel/replace (`scheduler/config.go:1208-1213`, `scheduler/hyperliquid_stop_loss_test.go:2075-2095`).
- Live trailing updates are already called with the strategy's virtual position quantity, capped only when the aggregate on-chain position is smaller (`scheduler/main.go:1724-1734`, `scheduler/main.go:2025-2035`, `scheduler/hyperliquid_trailing_stop.go:18-23`). This is necessary but not sufficient: there is no current per-coin arbiter that proves aggregate resting trailing SL quantity is safe.
- The live update path cancels one previous OID and places a replacement through `--update-stop-loss` (`scheduler/hyperliquid_trailing_stop.go:617-650`, `scheduler/executor.go:382-395`, `shared_scripts/check_hyperliquid.py:1051-1121`). Multiple peers can still overlap without scheduler-level coordination.
- The account reconciler has a single-SL-owner partial-close detector, but explicitly treats multiple SL owners as ambiguous and skips that detector (`scheduler/hyperliquid_balance.go:1059-1080`). Multi-owner attribution needs to be updated before the startup blocker can become merely a safety net.

## Why the current validation is too strict

The guard solves the historical cancel/replace race by banning all same-coin multi-trailing setups. That is operationally conservative, but it also makes `trailing_tp_ratchet*` unusable for otherwise valid same-coin peer strategies.

The desired end state is not simply deleting `trailingOwners > 1`. The safer model is to allow multiple trailing owners only once the scheduler can prove and maintain all of these invariants:

- Each trailing SL is sized to that strategy's own virtual leg, never the aggregate coin position.
- The sum of all same-coin trailing reduce-only quantities never exceeds the intended same-side virtual exposure.
- Cancel/replace sequences for same-coin peers are serialized or otherwise coordinated so one peer does not place a new OID while another peer's old OID is still resting in a way that creates unsafe aggregate reduce-only exposure.
- Reconciliation can attribute multi-owner SL fills or failures without falling through the current ambiguous multi-SL-owner path.

## Observed impact

Operators with an already-running multi-trailing live config cannot SIGHUP-reload policy edits while the new config still contains multiple trailing owners. The reload is rejected before otherwise hot-reloadable fields can apply, leaving live in-memory policy drift until the operator flattens and restarts or deploys a fix.

Note: edits such as `invert_signal` are independently blocked while that strategy has an open position (`scheduler/config_reload.go:377-379`), so examples should distinguish peer-validation rejection from normal open-position state guards.

## Proposed direction

Allow N trailing owners on the same Hyperliquid coin only after adding scheduler-owned coordination for trailing SL replacement:

1. Build a per-coin trailing update plan for all same-coin HL perps/manual strategies that currently need a trailing SL update.
2. Serialize cancel/place operations for those peer owners in Go, or provide an equivalent per-coin critical section around `runHyperliquidTrailingStopUpdate`.
3. Track/verify expected resting SL quantities by strategy leg and aggregate same-side exposure before and after replacements.
4. Extend reconciliation so multiple SL-owner fills/failures can be attributed, or fail closed with an explicit operator alert rather than silently skipping the single-owner detector.
5. Replace the current `trailingOwners > 1` startup/hot-reload error with a validation that all same-coin trailing owners use the coordinated per-strategy sizing model.

## Acceptance criteria

- Multi-trailing peer configs load cleanly at startup and hot-reload only when the coordinated sizing/replacement path is active.
- Tests cover same-coin peers with multiple `trailing_stop_atr_regime` / `trailing_tp_ratchet_regime` owners.
- Tests prove no update path can submit aggregate same-coin reduce-only trailing SL quantity above intended same-side virtual exposure, including cancel failure, place failure, missing OID, immediate-fill, and scale-in resize cases.
- Reconciler tests cover multiple SL-owner peers and define exact behavior for ambiguous partial on-chain drops.
- Existing single-owner, fixed ATR, fixed pct, manual ratchet, kill-switch, and circuit-breaker cancellation paths keep working.
- An on-chain stress test or high-fidelity paper/replay harness demonstrates that peer trailing replacements do not create an unsafe reduce-only quantity window.

## Related

- #601 — original same-coin per-strategy protection work
- #604 — review #5 with the OID race discussion
- Follow-up threads on per-strategy sized protection as the actual fix

Validation note: checked against `origin/main` at `7add410`.

richkuo:
## Summary
- allow same-coin Hyperliquid peers to have multiple trailing-stop owners once margin mode/leverage match
- make trailing SL replacement fail closed on open-order lookup/cancel uncertainty and serialize replacements per coin
- reconcile multi-owner SL fills only when exact OID-confirmed fills explain the on-chain residual

## Why
Same-coin HL peers can legitimately need per-strategy trailing protection, especially `trailing_tp_ratchet*`, but the prior validator banned all multi-trailing owners. The unsafe part was not multiple owners by itself; it was placing replacements when old OID state was unknown or overlapping cancel/place could over-size reduce-only protection.

Closes #964

## Validation
- `uv run --extra test pytest shared_scripts/test_check_hyperliquid.py platforms/hyperliquid/test_adapter.py`

LLM: GPT-5 | high | Harness: go test ./... (scheduler); pytest shared_scripts/test_check_hyperliquid.py platforms/hyperliquid/test_adapter.py


richkuo:
@claude

richkuo:
@claude sonne rereview

Issue #698, opened 2026-05-10
Hyperliquid userFills closedPnl appears gross; TP reconciler should net out actual fees


richkuo:
## Problem

Hyperliquid `userFills.closedPnl` appears to be **gross PnL before fees**, while the exchange UI shows **net realized PnL after fees**.

Example from `manual-eth`:
- entry: `0.428 ETH @ 2331.5`
- TP1 fill: `0.214 ETH @ 2354.8`, `fee=0.072565`
- TP2 fill: `0.214 ETH @ 2366.5`, `fee=0.072926`

`userFills` returned:
- TP1: `closedPnl=4.9862`
- TP2: `closedPnl=7.49`

Exchange UI showed:
- TP1: `4.91`
- TP2: `7.42`

UI values match fee-net calculation:
- `0.214 * (2354.8 - 2331.5) - 0.072565 = 4.913635`
- `0.214 * (2366.5 - 2331.5) - 0.072926 = 7.417074`

## Current state (audited)

**No production code path currently mis-uses `closedPnl`.** Audit results:

- `HLFillLookup.ClosedPnL` (`scheduler/hyperliquid_fills.go:28`) is **log-only** (`hyperliquid_fills.go:274`).
- `bookPerpsPartialCloseWithFillFee` (`scheduler/portfolio.go:219-258`) computes PnL locally from `avgCost`, `fillPx`, `qty`, and subtracts the real `fillFee` — never reads `closedPnl`.
- `backfill_hl_fees.go:225` recomputes via `newPnL = t.RealizedPnL + (modeledFee - realFee)` from stored values, not `closed_pnl`.
- Python attaches `closed_pnl` to fill dicts (`shared_scripts/check_hyperliquid.py:353`, `shared_scripts/fetch_hl_user_fills.py:167`), but no Go consumer reads it for booking.

So this is **not** a present-day bug — it's a forward-looking guardrail to prevent future code from trusting `closedPnl` as net.

## Goal

Lock in the local-fee-net PnL contract so any future refactor that wires `closedPnl` into a booking path fails loudly.

## Deliverables

1. **Regression test** against `bookPerpsPartialCloseWithFillFee` using the `manual-eth` numbers above — asserts the booked `Trade.RealizedPnL` equals the local fee-net value (e.g. `4.913635` for TP1, `7.417074` for TP2), not the gross `closedPnl`.
[...]

Issue #455, opened 2026-04-28
Feature: #T and W/L should show lifetime round-trip stats, not current-run counts


richkuo:
## Current Behavior

- `#T` (trade count) counts each open and each close as separate trades — a round-trip = 2 trades
- `#T` and `W/L` reset when the kill switch fires or a circuit breaker resets — they show "current run" stats, not lifetime
- Example: `hl-rmc-eth-live` has 17 trades in the DB but `#T` shows 2 (since last reset)

## Requested Change

1. **1 trade = 1 round-trip** (open + close = 1 trade, not 2). An open position without a close should not count yet.
2. **`#T` should show lifetime completed round-trips**, not current-run count. Count close trades (or pairs) from the trades table.
3. **`W/L` should show lifetime win/loss** based on round-trip PnL, not reset on kill switch / circuit breaker.

## Why

- Current-run stats reset after every kill switch or circuit breaker, making the summary misleading for strategies that have been running for days/weeks
- Round-trip counting is more intuitive — "17 trades" means 17 round-trips, not 17 individual buy/sell actions
- W/L ratio should reflect the strategy's full track record

Reference only: LLM model used: GLM 5.1

richkuo:
@claude

PR #969, opened 2026-06-11
feat(shared-wallet): trade-ledger display PnL + gross convention (#954)

### LINKED ISSUE #954: Shared-wallet display PnL should derive from the local trades ledger, not a balance split
## Summary

For strategies that share a single on-exchange account (one Hyperliquid wallet holding positions for two or more `go-trader` strategies), the **display PnL** is currently derived by reading the live wallet balance and splitting the realized base across members by configured-capital weight, with owned unrealized PnL attributed by position ownership (#918/#920). The **risk PnL** (kill switch, drawdown, peak) is still modeled.

The display path should instead be derived from the local `trades` table — a per-strategy trade-sum (see **PnL convention** below for the exact formula) — the same way it would be for a single-account strategy. To make that trustworthy, three preconditions need to hold first:

1. **Every on-chain fill that affects a tracked position lands in the local `trades` table.** Today the only path that drops a `Trade` row entirely is the no-mark-price reconcile branch (`hyperliquid_balance.go` — `recordClosedPosition` + `delete` with no `RecordTrade`); every other path records a row, but stamps `exchange_fee = 0` whenever the `userFills` lookup misses (PnL is still booked net of the *modeled* fee).
2. **Captured fill data is as accurate as the exchange API allows** (real fill price from `userFills`, real `fee`, real OID) — not a mid/orderbook/stale-mark proxy.
3. **Non-trade cash flows are accounted for.** Funding payments, deposits, and withdrawals move the wallet balance but produce no fill — see **Funding and transfers** below.

When all three hold, the `clearinghouseState` (or equivalent live-balance) becomes a pure drift alarm, no longer a source for the display value itself.

## Current behavior

- #918 (with #920 review follow-ups) introduced exchange-authoritative shared-wallet reconciliation. Each cycle the bot reads `clearinghouseState` (or OKX equivalent) and computes `value_i = w_i × (balance − U) + ownedUPnL_i`: the realized base (balance minus total unrealized PnL) is split by **configured capital weight**, while each position's unrealized PnL is attributed to its owning member(s) by virtual-quantity share (`reconcileSharedWalletMemberValues`, called from `reconcileSharedWalletDisplayValues`). Members with ≤0 configured capital get zero weight; an equal-split fallback applies only when no member has positive capital.
- The split is the only place the on-exchange balance enters the display number; `trades.realized_pnl` plays no role in shared-wallet display.
- The drift alarm ($0.01 threshold, 2-cycle confirmation) measures **orphan unrealized PnL only** — on-chain positions no member owns. The member values sum to the balance minus that orphan amount *by construction*, so the alarm never compares against the trades ledger and is structurally silent for any ledger error.
- On a `clearinghouseState` fetch failure, `SharedWalletValueSet` gating falls members back to the modeled `PortfolioValue` the same cycle — a stale exchange-derived value is never served.
- For non-shared-wallet strategies, display and risk PnL use the modeled virtual book (`StrategyState.Cash` + position marks), which is maintained from trade bookings — equivalent to a trade-sum only when the ledger has no gaps.

## Why the current approach is wrong

- **Realized attribution is non-causal.** The realized base `(balance − U)` is split by capital weight, so two strategies with $500 configured capital each inherit equal shares of every *realized* wallet move (fills, fees, funding), regardless of which one actually traded. (Unrealized PnL is already ownership-attributed and is not the problem.)
- **The drift alarm cannot see ledger errors at all.** Because the split sums to the balance by construction, the alarm only fires on orphan positions. Lost fills, modeled fees, and mark-based close prices in the trades ledger are invisible to it — the ingestion gaps keep recurring with no signal.
- **Display degrades to the modeled book on any fetch failure.** A `clearinghouseState` failure (rate limit, transient error) drops members back to the modeled `PortfolioValue` for that cycle — not stale, but a silent quality downgrade; trade-sum PnL is local and deterministic every cycle.
- **Replays and forensics are hard.** Given just the live wallet balance, you can't reconstruct the historical path. Given a complete `trades` table, you can.

## Desired behavior

**PnL convention (decide first — this gates everything else):**

Today `Trade.RealizedPnL` is stored **net of fees** (`pnl -= fee` before stamping, `portfolio.go`), and `HLFillLookup.ClosedPnLGross` is **gross** — explicitly guarded against being wired into `RealizedPnL` since #698 (`hyperliquid_fills.go`). A naive `Σ(realized_pnl − exchange_fee)` therefore double-counts fees on every existing row, and sourcing `realized_pnl` from `userFills[].closedPnl` flips the convention.

This issue adopts the **gross convention**: `realized_pnl` stores the gross close PnL (`userFills[].closedPnl` for close fills, `(fill_px − entry_px) × qty` for computed closes), `exchange_fee` stores the fee on every row (opens and closes), and the display sum is `Σ(realized_pnl − exchange_fee)` over all rows (open rows contribute `−exchange_fee`). This requires:

- Migrating historical net-convention rows in the backfill (add back the fee that was subtracted at booking), keyed off a schema marker or cutoff so the migration is idempotent.
- Re-scoping the #698 guard: `ClosedPnLGross` becomes the *correct* source under the new convention; the guard moves to "never mix conventions in one sum."

**Funding and transfers:**

The wallet balance includes funding payments, deposits, and withdrawals; the trades ledger does not (acknowledged as "ignored funding" in `shared_wallet_reconcile.go`). Left unaddressed, the trade-sum-vs-balance comparison drifts persistently on any held perps position and the $0.01 alarm fires forever. Therefore:

- Book funding payments as ledger rows (a `funding` trade-type per strategy per coin, sourced from the HL funding history endpoint), attributed to the position owner(s) by the same virtual-quantity share as uPnL.
- Book deposits/withdrawals as per-wallet transfer rows or a per-wallet non-trade-flow accumulator, excluded from per-strategy PnL but included in the drift comparison.
- The drift alarm's new comparison is: `live balance − Σ_i(trade-sum_i + owned_uPnL_i) − unattributed flows`, with the $0.01 threshold and 2-cycle confirmation unchanged.

**Display PnL per strategy, in priority order:**

1. For a shared-wallet member: `initial_capital_i + Σ(trades.realized_pnl − trades.exchange_fee) + Σ(funding_i) + owned_unrealized_pnl_i` (where `owned_unrealized_pnl_i` is the on-chain uPnL of the positions member `i` owns, split on shared coins as today).
2. For a non-shared-wallet strategy: same formula; should agree with the existing modeled `PortfolioValue` when the ledger is complete.
3. The `clearinghouseState` total becomes a **drift alarm only** — when the trade-sum path disagrees with the live balance beyond the threshold, the alarm fires with per-coin attribution as today. The capital-weight split is removed from the display calculation; capital weight no longer enters the display number for shared-wallet members.

**Preconditions on the `trades` table:**

- Every on-chain fill that affects a tracked position lands as a row, regardless of who placed the order. Sources include: scheduler-placed orders (open, TP, SL, manual-open, manual-close, `placeManualProtectionInline`); reconciler-detected external closes (peer closes, owner SL/TP, external partial, all-TP-tiers dust, orphan auto-close, manual-flatten); circuit-breaker force-closes; operator UI/CLI closes. The known gap to close: the no-mark-price reconcile branch must record a row (defer or retry the price lookup) instead of dropping the close.
- Each captured row carries the most accurate value the exchange API provides:
  - `exchange_order_id` — from the order placement, persisted on close rows.
  - `price` — from the fill (`userFills[].px` VWAP or per-aggregator mirror on OKX), not from a mid/orderbook snapshot.
  - `realized_pnl` — per the gross convention above. Never from a stale `mark_price` reference.
  - `exchange_fee` — from `userFills[].fee`, summed across partial fills of the same OID. `hl_user_fills.py` already has the helpers; the gap is the call sites that stamp `0` when the lookup misses (the modeled fee is used for PnL math but not stamped).
- One fill, one row. No double-booking when a CB close and a reconciler close compete for the same fill (the existing #758 TP-fill fix is the template).

**Backfill:**

- A `go-trader backfill trade-ledger --apply` (mirroring `go-trader backfill hl-fees --all --apply` from #591/#602) walks the existing `trades` history, rewrites the lossy columns from `userFills`, and performs the net→gross convention migration. Idempotent; safe to re-run; documented in the post-update agent protocol so future updates know to run it after a `git pull`.

## Acceptance criteria

- [ ] Every on-chain fill that affects a tracked position lands in `trades` exactly once, regardless of placement source; the no-mark-price reconcile branch no longer drops the row.
- [ ] For HL close rows, `exchange_fee` is sourced from `userFills` (with the indexer-lag retry budget already in `lookup_fill_fee_by_oid`) at insert time, not on a later cycle. The fraction of HL close rows with `exchange_fee = 0` drops to near zero for any fill the indexer surfaces within the retry window; the remainder is backfilled by the new command.
- [ ] The gross PnL convention is implemented consistently: new rows store gross `realized_pnl` + populated `exchange_fee`; historical rows are migrated by the backfill; no sum ever mixes conventions; the #698 guard is re-scoped accordingly.
- [ ] Funding payments are booked as ledger rows attributed by ownership share; deposits/withdrawals are tracked per wallet; the drift alarm compares the live balance against the trade-sum path (threshold and confirmation cadence unchanged) without persistent false fires from funding.
- [ ] `display_value_i − initial_capital_i − Σ(trades.realized_pnl − trades.exchange_fee) − Σ(funding_i) − owned_uPnL_i = 0` within sub-cent rounding, for every shared-wallet strategy on every cycle — excluding cycles where cash was mutated outside the ledger (`backfill --reset-cash`, operator initial-capital edits), which re-baseline.
- [ ] The capital-weight split is removed from the display path (`reconcileSharedWalletMemberValues` / `reconcileSharedWalletDisplayValues`). The helper, if retained for the drift comparison, is documented as unused for display.
- [ ] Non-shared-wallet strategies: displayed value matches the existing modeled path when the ledger is complete; any divergence surfaces as a drift signal, not a silent change.
- [ ] `go-trader backfill trade-ledger [--all|--strategy <id>] [--apply] [--dry-run]` exists, replays the lossy-column repair and convention migration for any historical row, is idempotent, and is mentioned in the post-update agent protocol entry for this PR.
- [ ] One fill, one row: a CB force-close of a position that the reconciler also detected does not produce two `trades` rows; existing #758 TP-fill attribution rules extend to this case.

## Out of scope

- Risk PnL (kill switch, drawdown, peak, `capital_pct` sizing) — still uses the modeled `PortfolioValue`, unchanged. (A separate issue can address that if desired.)
- Backtester parity — out of scope unless this becomes user-visible in backtests.
- OKX shared-wallet path (#920 also covers it) — fold in as a follow-up if the HL path lands cleanly; same shape.
- Cross-strategy attribution of opens when more than one strategy could claim a position. (#619/#620 / "shared-coin" math already handles this; verify the trade-ledger path respects those ownership rules.)

## Related

- #918, #920 — exchange-authoritative shared-wallet reconciliation + drift alarm (the path this replaces for display).
- #698 — `ClosedPnLGross` guard (the convention this issue deliberately revisits).
- #585, #589 — closed issues about `exchange_fee=0` rows; the underlying gap is still present, this issue consolidates the fix.
- #758 — TP-fill double-booking (template for "one fill, one row" enforcement).
- #591, #602 — `go-trader backfill hl-fees` (template for the new backfill command).
- #669, #671 — trade-row schema for `exchange_order_id` / `tp_tiers_json` / `stop_loss_atr_mult` (already in place).
- #844, #870 — `trailing_tp_ratchet` and default TP ladder; relevant to the "TP-fired close" attribution rules.

---

richkuo:
## Summary

Implements #954: Hyperliquid shared-wallet **display PnL** is now derived from the local trades ledger — `initial_capital + Σ(ledger deltas) + owned uPnL` — instead of the #918 capital-weight balance split. An idle member's display value no longer inherits drift from an active peer's trading, and the wallet balance becomes a **pure drift alarm** (baseline-anchored at adoption, re-anchored by backfill).

### Gross PnL convention (`scheduler/trade_pnl.go`, schema)
- `trades` gains `pnl_gross` + `fee_source` columns. Rows written at/after #954 store **pre-fee** `realized_pnl` with `exchange_fee` always stamped (`fee_source` = `userfills` | `modeled`); legacy rows keep net semantics.
- Every consumer reads through `tradeNetPnL` / `tradeNetPnLSQL` / `tradeLedgerDeltaSQL` (W/L stats, UI, alerts, leaderboard inputs) so the two conventions cannot mix — the #698 gross-vs-net trap, generalized. Cash, RiskState, and `closed_positions` stay net everywhere.

### Ledger completeness preconditions
- **No-mark-price external closes book a row** at AvgCost (zero gross PnL, modeled fee) instead of silently dropping the close from the ledger (4 reconciler branches).
- **One-fill-one-row**: full closes dedupe on (strategy, OID) across the reconciler and circuit-breaker force-close racing over the same on-chain fill; partial legs of one OID remain legitimate.
- **Live flips apportion the single real fill fee** across close+open legs by quantity share (was: full real fee on the close + an extra modeled fee on the open — overcharging the cash book each flip and permanently drifting the ledger). Both legs stamp the shared OID.

### Funding + non-trade flows (`scheduler/wallet_ledger.go`)
- Go-side HTTP fetch of `userFunding` + `userNonFundingLedgerUpdates` per shared wallet (outside the state lock); ingestion under the risk-phase write lock **before** the display reconcile.
- Funding on an owned coin → `trade_type='funding'` rows split by virtual-qty share (dual dedup: DB + in-memory). Orphan funding, deposits, withdrawals, class/internal/sub-account/vault transfers → `wallet_transfers` with signed perp-balance effect; unknown kinds recorded at \$0 with a WARN so drift surfaces them.
- Watermarks advance only past durably-persisted events; first contact initializes to now (history belongs to the drift baseline, not the ledger).

### Display path (`scheduler/shared_wallet_reconcile.go`)
- HL wallets route through `reconcileHLWalletViaLedger`; drift = balance − Σvalues − transfers − baseline. Any DB failure falls back to the #918 split for the cycle (rows stay populated, WARN logged). OKX keeps the split. Existing `SharedWalletDriftTracker` (\$0.01, 2-cycle confirm) unchanged.

### `backfill trade-ledger` (`scheduler/backfill_trade_ledger.go`)
- Migrates legacy net rows to gross (booking-time fee stamped, provenance recorded), then trues fee/price/PnL up to HL `userFills` (fee apportioned across **all** legs of an OID, `closedPnl` across close legs only), replays cash + `closed_positions` with the same SIGHUP-divergence gate as `hl-fees`, and resets drift baselines on `--apply`. Idempotent. `backfill hl-fees` now skips gross rows (`gross_convention_row`).

## Out of scope (per issue)
Risk PnL (kill switch / circuit breakers keep modeled `PortfolioValue`), the OKX ledger path, backtester parity.

## Testing
- `go test ./...` green (full scheduler suite), `go vet` clean, `gofmt` clean.
- New `shared_wallet_ledger_test.go`: convention helpers, SQL ledger sums, ledger member-value math (orphans, idle-member independence, baseline anchoring), full DB-backed reconcile cycle (baseline set → new drift → reset re-anchor), nil-DB split fallback, `signedPerpFlowUSD` table, funding split + dedup + orphan routing, watermark advancement + replay idempotency, gross booking + OID dedup at the close site, planner migration/true-up/idempotency/flip-OID apportioning, ApplyTradeLedgerPlan round-trip.
- Updated legacy tests that pinned net-convention or old flip-fee policy to the new invariants (net asserted via `tradeNetPnL`, cash still moves net).
- Python: `pytest` 1279 passed; `fetch_hl_user_fills.py` py_compile clean.

## Operator note
After deploy, run `./go-trader backfill trade-ledger --all --apply` (daemon stopped) to migrate history and make the drift baseline meaningful. Documented in SKILL.md (commands, dedicated section, Post-Update Reference entry).

Closes #954

---


richkuo:
@claude

richkuo:
Addressed all review findings in 32855a2:

1. **Funding watermark durability (Needs Fixing)** — `ingestFundingEvent` now holds the watermark whenever a funding row is not durably persisted: it checks the eager-insert outcome after `RecordTrade` AND treats an in-memory dedup hit on a still-unpersisted row as held (so the watermark releases only once SaveState lands the row). Already-persisted co-owners dedup-skip on retry; `tradeRecorder==nil` batch-persist contexts still advance. Tests cover mid-split failure + recovery without double-booking, the end-to-end watermark hold/advance, and the no-eager-persist context.
2. **`value` in the rewrite predicate (Recommended 1)** — added `math.Abs(nv.Value-t.Value) > 1e-9`; test pins a row whose only stale column is `value` (rewritten, then idempotent, fully-correct rows untouched).
3. **Scoped baseline reset (Recommended 2)** — `ResetWalletLedgerBaselines` replaced with per-wallet `ResetWalletLedgerBaseline(platform, account)`; `--apply` resets only wallets with a repaired member (perps + same-account live manual membership, matching the reconcile), and plans with zero changes skip Apply entirely. Test: repaired wallet cleared, co-resident wallet's baseline + offset preserved, no-op apply resets nothing.
4. **Withdraw fee semantics (Requires Human Review)** — kept `-(usdc+fee)` (Chainstack's withdraw3 reference states the fee is "deducted from your Hyperliquid balance, not from the withdrawal amount"; other community sources conflict and official docs are ambiguous). Documented the assumption in `signedPerpFlowUSD` with the failure signature: if `usdc` already includes the fee, drift steps by exactly +$1 per withdrawal — please confirm against a real `userNonFundingLedgerUpdates` withdraw sample; fix is dropping the fee term.

Full `go test ./...` green; gofmt/vet clean.

---

richkuo:
@claude

richkuo:
Needs Updates

### Needs Fixing

1. **A mid-batch insert failure loses every unprocessed event that shares a millisecond with an already-processed one — the watermark advances past the failed event.**
`ingestWalletLedgerEvents` (scheduler/wallet_ledger.go:353-364, funding; 380-387, transfers) breaks on failure but then sets the watermark to `maxTime + 1`. When the failed event's `Time` equals `maxTime` (a prior same-ms event succeeded), the watermark jumps past the failed event and it is never re-fetched — permanent ledger shortfall, the exact loss the hold-on-failure logic exists to prevent. This is realistic: two coins fund at the same hourly tick, and HL emits two transfer legs at one timestamp (the `transferDedupID` comment at wallet_ledger.go:311-314 says so itself). Fix: on failure, clamp the new watermark to `min(maxTime+1, failedEvent.Time)` in both loops — re-reading earlier same-ms successes is absorbed by the dedup keys.
**Invariant:** A watermark must never advance to or past the timestamp of an event that was fetched but not durably processed.
**Must survive:** (a) [ok@t, fail@t] — watermark stays ≤ t and the failed event is re-fetched, while the succeeded one dedup-skips; (b) [ok@t, fail@t+5] — watermark still advances to t+1 (no regression to never-advancing); (c) the funding-persist-hold path (`ingestFundingEvent` returning false on an unpersisted row) hits the same clamp.

2. **The boot-time Details-parse migration overwrites #954 zero-gross close rows with NET PnL while `pnl_gross=1` stays set, double-subtracting the fee.**
`backfillTradeCloseFlags` (scheduler/db.go:589, 600) targets `is_close = 1 AND realized_pnl = 0 AND details LIKE '%PnL%'` with no convention gate. The PR's no-mark-price external closes book at AvgCost with `RealizedPnL=0`, `pnl_gross=1`, and Details like `"External close @ mark, PnL: $-0.18 (fee $0.18)"` (portfolio.go close Details format) — exactly matching the scan. On the next restart the migration writes the parsed NET value (−0.18) into `realized_pnl` while the row stays gross, so `tradeNetPnL` yields −0.36 (fee subtracted twice) and the ledger sum corrupts. Exact-breakeven gross closes hit the same path. Fix: add `AND COALESCE(pnl_gross, 0) = 0` to both statements — the migration is legacy-rows-only by definition.
**Invariant:** Legacy-convention migrations must never rewrite rows already written under the gross convention; `pnl_gross=1` rows are owned exclusively by the booking sites and `backfill trade-ledger`.
**Must survive:** (a) a zero-gross AvgCost close row across two restarts — `realized_pnl` stays 0; (b) a genuine legacy row (`pnl_gross=0`, `realized_pnl=0`, PnL token) is still migrated; (c) a gross close with non-zero PnL is untouched (already excluded by `realized_pnl = 0`, but the regression test should pin it).

---

richkuo:
Both findings fixed in 478024c:

1. **Same-ms watermark skip** — both ingest loops now track the failed event's timestamp and clamp the new watermark to `min(maxTime+1, failedAt)`, so a failure sharing a millisecond with a processed sibling is re-fetched (dedup absorbs the sibling's re-read). Tests: [ok@t, fail@t] holds at t then recovers with no loss/double-book; [ok@t, fail@t+5] still advances to t+1; composes with the funding persist-hold path.
2. **#455 migration vs gross rows** — both statements in `backfillTradeCloseFlags` now carry `AND COALESCE(pnl_gross, 0) = 0`. Test: a gross zero-PnL AvgCost close survives two migration runs untouched while a legacy `realized_pnl=0` row is still repaired.

Full `go test ./...` green; gofmt clean.

---

richkuo:
@claude rereview

richkuo:
Resolved the Requires Human Review item empirically in af41c45 — no human judgment call needed after all, because the info endpoint is public and real ledger samples settle it:

**Withdraw semantics: current code is CORRECT.** Queried `userNonFundingLedgerUpdates` for 12 active mainnet accounts (332 withdraw entries). The `usdc` values are overwhelmingly round-requested-amounts-minus-$1 with `fee: "1.0"` — e.g. `usdc=1999999.0`, `999999.0`, `24999.0`, `999.0`, `9.0`. So `usdc` is the NET (post-fee) amount and the fee is debited on top: total account debit = `usdc + fee` = the round requested amount, exactly what `-(usdc + parseHLFloat(d.Fee))` books. Comment updated from UNVERIFIED to verified-with-evidence.

The same investigation surfaced and fixed three adjacent gaps (validated against the nktkas/hyperliquid SDK schema + the live samples):
1. **`vaultWithdraw` booked $0 silently** — it has no `usdc` field at all (`requestedUsd`/`commission`/`closingCost`/`netWithdrawnUsd`); the handler returned 0 with ok=true, i.e. invisible drift instead of the designed WARN. Now credits `netWithdrawnUsd`.
2. **`send` was unmapped** (730 occurrences across the sampled accounts) — core USDC sends now map like `internalTransfer` (inbound `+amount`, outbound `-(amount+fee)`); non-USDC sends are spot-side $0; dex-routed USDC stays unmapped-and-visible rather than guessed.
3. **Noise/correctness on remaining kinds** — `liquidation` maps to $0 explicitly (its balance impact arrives via fills the reconciler already books; a flow row would double-count), `rewardsClaim` credits USDC rewards, and spot/staking kinds (`gossipPriorityGasAuction` — 3063 entries on one sampled account, `deployGasAuction`, `cStakingTransfer`) map to $0 without the unmapped WARN.

Table test extended to every mapped kind plus the dex-routed unmapped case. Full `go test ./...` green.

---

richkuo:
@claude rereview

richkuo:
Addressed the remaining finding in e76df13: `reconcileHLWalletViaLedger` now honors the `found` flag from `GetWalletLedgerState`. An absent watermark row (fetch init failed that cycle) falls back to the capital-weight split and writes nothing — the row remains owned by `fetchWalletLedgerEvents`, which anchors watermarks at `now` before any baseline is set. Regression test `TestReconcileSharedWalletDisplayValues_HLMissingLedgerStateFallsBack` covers all three must-survive cases (absent row → no row originated + fallback values; re-init next cycle → baseline anchors with watermarks intact; existing happy-path test now asserts the baseline upsert preserves watermarks).

---

richkuo:
@claude rereview

Issue #103, opened 2026-04-03
Bug: HTF filter incorrectly applied to delta_neutral_funding strategy


richkuo:
## Description

The `init.go` wizard applies `HTFFilter = true` to all non-options strategies, including `delta_neutral_funding`. The HTF trend filter suppresses signals based on price trend direction, which is irrelevant to a funding-rate harvest strategy. This incorrectly suppresses valid funding entries when price is in a downtrend (often the most profitable time to collect positive funding).

## Fix

Skip HTF filtering in `check_hyperliquid.py` when `strategy_name == "delta_neutral_funding"`, or add a per-strategy `htf_filter_compatible` flag.

## Origin

Found during code review of PR #74 (HTF filter) + PR #73 (delta-neutral funding). Confidence: 87.

PR #970, opened 2026-06-11
feat(strategies): funding_skew + funding-rate history in backtester (#960)

### LINKED ISSUE #960: New strategy: funding-rate aware perps entry (prereq: funding data in backtester)
Part of #955. Audit verdict (#956): confirmed but **blocked** — the backtester has no funding-rate data, so `delta_neutral_funding` is untestable and a funding-aware entry can't be validated.

## Scope

1. **Prereq:** funding-rate history in the backtest data path (HL `info.meta_and_asset_ctxs()` live; historical funding via HL API/fetcher cache) so funding-aware strategies are backtestable.
2. **Strategy:** funding-rate aware perps entry — bias/veto entries when funding is extreme (crowded side), complementing `delta_neutral_funding`.

## Validation bar (from #955)

Registry entry + `PLATFORM_ORDER`; `knownShortNames` + defaults in `init.go`; `DEFAULT_PARAM_RANGES` in `backtest/optimizer.py`; signal-value tests; backtest report with in/out-of-sample split in the PR body; bidirectional → `bidirectionalPerpsStrategies`. Must beat the median incumbent on out-of-sample Sharpe and DD-adjusted return. No look-ahead.

---

richkuo:
Closes #960

**Draft: borderline miss on the #955 performance bar** (details below). The funding-data infrastructure half of this issue is complete, tested, and useful independent of the strategy verdict.

## Part 1 — funding-rate history in the backtest path (the #960 prereq)

- `platforms/hyperliquid/adapter.py`: `get_funding_history_range(symbol, start_ms, end_ms)` — paginates past the ~500-record (~20 days hourly) single-call cap of `funding_history`, dedupes on time, stops on no-progress. Verified live: 8,787 hourly BTC records over 12 months.
- `shared_tools/storage.py`: `funding_rates` table (idempotent `CREATE TABLE IF NOT EXISTS`) + `store_funding_rates`/`load_funding_rates`.
- `shared_tools/funding_fetcher.py`: `load_cached_funding(coin, start, end)` (cache-first, edge-tolerant refetch) and `attach_funding_column(df, funding)` — merge_asof **backward**, so a bar never sees a future snapshot; handles tz-naive (binance) and tz-aware indexes, ns-normalized keys.
- `backtest/run_backtest.py`: attaches the column for `FUNDING_COLUMN_STRATEGIES` in single and optimize modes; loud `[WARN]` when funding is missing (the strategy is fail-safe flat, which would silently zero a backtest).

This also makes `delta_neutral_funding` backtestable in a follow-up (it reads scalar params today).

## Part 2 — `funding_skew` strategy

Rolling z-score of HL hourly funding marks positioning extremes; entries require price confirmation so funding alone is never enough: long when crowded shorts (z ≤ −z_entry, rate below a noise floor) and close > EMA; short (futures variant) when crowded longs and close < EMA. Exits on crowding resolution (|z| inside z_exit) or EMA flip; missing funding exits and blocks entries. Live path: `check_hyperliquid.py` passes paginated `funding_records`, aligned in-core with the same backward-only rule (`funding_records` vs column parity is tested). HTF-filter skip extended to funding strategies (#103).

## Wiring checklist

- [x] Futures registry `@register` + `PLATFORM_ORDER` — `--list-json` byte-identical for existing strategies, both shims
- [x] `knownShortNames` (`fskew`) + perps/futures default lists + `bidirectionalPerpsStrategies` (`scheduler/init.go`)
- [x] `DEFAULT_PARAM_RANGES` (`backtest/optimizer.py`)
- [x] 12 strategy tests (pinned entries/exits, records/column parity, backward-only alignment, tz-naive parity, noise floor, look-ahead perturbation incl. funding), 7 fetcher tests (cache hit/refetch/alignment), 5 adapter pagination tests
- [x] Full pytest 1303 passed; `go build` + `go test ./...` green

## Performance — borderline miss

Defaults selected on in-sample only (z_entry 2.0, confirm_ema 40). Bench: -0.76 median incumbent OOS Sharpe (#963 harness).

| Framing | IS mean Sharpe | OOS mean Sharpe |
|---|---|---|
| Long-only (plain path, comparable to incumbents) | -0.96 | **-1.23** |
| Bidirectional, engine path + default `tiered_tp_atr` close | +0.03 | **-0.81** |

Long-only structurally mismeasures it: buying short-squeezes in a 12-month bear is the wrong half of the design, while the short half (crowded-long breakdowns) is only visible on the engine path. **Caveat (per review):** the engine path reads `signal` transitions as open-side requests, so a long-exit `-1` landing after the close evaluator already flattened opens an unconfirmed short — the engine-path numbers (+0.03 IS / -0.81 OOS) mix genuine confirmed shorts with these exit-artifacts and should not be read as clean evidence for the short half. The strategy's own z/EMA exits govern only the plain long/flat path (docstring now states exit ownership per path). Given funding extremes are episodic, six months of OOS is thin evidence either way.

**Recommendation:** merge the data layer regardless; paper-trade the perps variant rather than judging it on this window, or iterate (e.g. funding-percentile gate instead of z-score) before un-drafting.

---


richkuo:
@claude

richkuo:
Both findings addressed.

1. **Cache refetch:** added a `funding_coverage` ledger (new `funding_coverage` table + `load/store_funding_coverage` in `shared_tools/storage.py`) recording the API range actually fetched; `load_cached_funding` now decides hits against it instead of the stored rates, and `run_backtest.py` passes the data window's real end (`df.index[-1]`). Must-survive cases tested: historical window months before now (`test_cache_hit_survives_elapsed_wallclock`), repeat runs after elapsed wall-clock (same), late-listed coin no longer refetching forever (`test_late_listed_coin_cached_after_first_fetch` — coverage records the requested start, not the first record). A partial pagination never claims the tail (`test_partial_fetch_does_not_claim_tail_coverage`).

2. **Exit ownership / engine-path contamination:** `funding_skew` docstring now states exit ownership per path (close evaluator owns exits on live HL perps + engine path; the z/EMA exits govern only the plain long/flat path, and a flat-landing exit transition can be misread as an unconfirmed entry — inherent to transition-signal bidirectional strategies). PR body's performance section now carries the same caveat: the engine-path numbers are not clean evidence for the short half; per-side judgment uses the plain path long-only, and the recommendation remains paper-trade before allocation.

Also merged `origin/main` (mtf_confluence, supertrend fix, multi-trailing peers); conflicts resolved keeping both strategies. 1326 pytest passed, `go test ./...` ok, `--list-json` parity intact (futures adds only `funding_skew`).

richkuo:
@claude rereview

PR #974, opened 2026-06-11
feat(hl-reconcile): active operator alerts for persistent shared-coin SL gaps (#971)

### LINKED ISSUE #971: Add active alerting or auto-heal for persistent HL SL reconciliation gaps
## Context
The single-owner fail-closed path (#964) leaves a Hyperliquid shared-coin SL reconciliation gap when a partial drop cannot be confirmed by exact stop-loss OID. PR #965 additionally added the multi-owner branch that leaves a gap when there are more than one SL owner and no confirmed SL fill (`scheduler/hyperliquid_balance.go:1150-1156`). Both are safer than guessing via `hl_sync_external`, but they can leave a phantom virtual position as a `ReconciliationGap` when user-fills lookup misses or returns the wrong OID.

## Problem
The residual phantom *position* — not the ephemeral gap record, which is recomputed each cycle — can continue to feed drawdown and kill-switch math across cycles. Today this is surfaced through `/status` and `[WARN]` logs, but there is no active operator alert (notifier/DM) specifically for a stuck SL reconciliation gap.

Note: the reconcile pass already re-runs the user-fills lookup every cycle (the fill resolver is rebuilt per call) and auto-heals — booking the SL close and clearing the phantom — once a later cycle returns an exact-OID-confirmed fill (the `confirmedSLFills` path). So a *transient* miss self-heals; a gap is only truly stuck when the OID can never be confirmed, in which case auto-heal cannot fire without violating the fail-closed invariant.

## Proposed direction
- Add active notifier alerting for persistent SL reconciliation gaps after a small confirmation window. Model the tracker on `SharedWalletDriftTracker` (#918): in-memory per-coin streak, N-cycle confirmation, throttle + re-alert ratio, recovery notice. Specify the miss/timeout/error failure policy.
- Re-evaluate the auto-heal item. The per-cycle reconcile already heals on a later exact-OID confirm, so a new auto-heal step is largely redundant; only include it if there is a distinct case (e.g. aging out or clearing a permanently-unconfirmable gap) that the fail-closed invariant still permits.
- Preserve the fail-closed invariant: do not guess an owner or book an SL close unless exact OID-confirmed fills explain the residual within tolerance.

## Ownership / placement
- Owner: the reconcile pass (`reconcileHyperliquidAccountPositions`) — it sees all strategies and holds `mu`.
- Tracker medium: in-memory field on `AppState` keyed by coin (the gap record itself is ephemeral, so the confirmation-window streak must live in a separate persistent-in-memory structure).
- Emit via the existing `pendingAlerts` flush + `MultiNotifier`, not a new notification path.

## References
- Follow-up from PR #965 review.
- Single-owner fail-closed: #964. Multi-owner no-confirmed-fill gap branch: #965.
- Related tests: `TestReconcileSharedCoin_Detector2_UnconfirmedFillLeavesGap`, `TestReconcileSharedCoin_Detector2_WrongOIDInUserfillsLeavesGap`.

---
LLM: GPT-5 | high | Harness: GitHub follow-up issue from PR review

richkuo:
## Summary
Adds active operator alerting for persistent Hyperliquid shared-coin reconciliation gaps. When a shared-coin position partially drops but the reconciler cannot confirm the residual via an exact stop-loss/TP OID (user-fills miss or wrong OID), it fails closed (#964/#965): no owner is guessed, no SL close is booked — leaving a `ReconciliationGap` and a phantom virtual position that keeps feeding drawdown/kill-switch math. Previously this was visible only via `/status` and stderr `[WARN]`; there was no active alert.

## What changed
- **`scheduler/hl_reconcile_gap_alerts.go`** — `HLReconcileGapTracker` (in-memory, per-coin, resets on restart). A gap must persist **3 consecutive reconcile cycles** before the first owner DM, then re-throttles (re-alert only on a materially-changed residual, every 10th cycle, or hourly), with a one-shot recovery notice when it clears or the coin stops being shared. Modeled on `SharedWalletDriftTracker` (#918).
- **`scheduler/main.go`** — calls `reportHLReconcileGaps(...)` as a separate main-loop step right after the reconcile pass (analogous to `reportSharedWalletDrift`), reading the already-populated `state.ReconciliationGaps`.

## Design notes
- **Alerting only.** It never books, guesses an owner, or mutates a position — the fail-closed invariant is fully preserved.
- **Placed outside `reconcileHyperliquidAccountPositions`** (not inside it) to mirror the existing drift-alert pattern and to avoid contaminating the 24 reconcile tests' exact-DM-count assertions via the global tracker.
- **Auto-heal intentionally dropped.** As surfaced during issue validation, the reconciler already heals a gap within a later cycle the moment a user-fills lookup confirms the exact OID (the `confirmedSLFills` / Detector paths re-run every cycle with a freshly-built fill resolver). A transient miss self-heals; only a permanently-unconfirmable gap persists — for which alerting, not healing, is the correct response.
- **Confirmation window = 3** (vs the drift tracker's 2): a recorded gap already means this cycle's lookup failed, so 3 absorbs brief HL indexer lag while still surfacing a genuinely stuck gap within a few cycles.

## Tests
`scheduler/hl_reconcile_gap_alerts_test.go` — confirmation window, transient self-heal, throttle/re-alert, recovery, vanished-coin sweep, nil-notifier safety, sub-tolerance noise. Full scheduler suite green; build clean.

Closes #971

---

### DIFF
--- a/scheduler/hl_reconcile_gap_alerts.go
+++ b/scheduler/hl_reconcile_gap_alerts.go
@@ -1,0 +1,263 @@
+package main
+
+import (
+	"fmt"
+	"math"
+	"os"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+)
+
+// #971: Active operator alerting for persistent Hyperliquid shared-coin
+// reconciliation gaps.
+//
+// When a shared coin's on-chain position drops but the reconciler cannot
+// confirm the residual via an exact stop-loss/TP OID (user-fills miss or a
+// wrong OID), it fails closed (#964/#965): it does NOT guess an owner or book
+// an SL close, leaving a `ReconciliationGap` and a phantom virtual position
+// that keeps feeding drawdown/kill-switch math. The reconciler already
+// auto-heals such a gap WITHIN a later cycle the moment a user-fills lookup
+// confirms the exact OID (the confirmedSLFills / Detector paths re-run every
+// reconcile with a freshly-built fill resolver). So a transient miss self-heals
+// silently; this layer only surfaces a gap that stays stuck — the lookup never
+// confirms — so an operator can reconcile it manually. It deliberately does NOT
+// book or guess anything: the fail-closed invariant is preserved.
+
+// hlReconcileGapTolerance is the absolute signed-qty drift below which a
+// shared-coin reconciliation gap is treated as float noise rather than a real
+// stuck gap. Mirrors the 1e-6 qty epsilon used throughout the reconciler.
+const hlReconcileGapTolerance = 1e-6
+
+// hlReconcileGapAlertThreshold is the number of CONSECUTIVE reconcile cycles a
+// coin's gap must persist before the first operator alert fires. A recorded gap
+// already means THIS cycle's user-fills lookup could not explain the residual
+// (the reconciler books and clears the gap in-cycle the moment a lookup
+// confirms the exact OID), so a gap surviving several consecutive lookups is
+// genuinely stuck — a permanently missing or wrong OID — not the brief HL
+// indexer lag that clears once the fill is indexed. Three absorbs a couple of
+// lagged cycles while still surfacing a real stuck gap within a few cycles.
+const hlReconcileGapAlertThreshold = 3
+
+// hlReconcileGapRealertRatio gates re-alerts on an already-alerted coin to a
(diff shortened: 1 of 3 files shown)


richkuo:
@claude

PR #976, opened 2026-06-11
feat(strategies): regime_adaptive_htf — HTF-classified composite-regime fades (#973)

richkuo:
Closes #973. Part of #955 / follow-up to #967 (#958).

## What the redesign found

#973 hypothesized regime labels become tradeable when classification is slower than execution, via three mechanisms. All three were built and benchmarked independently on the #963 incumbent harness (six datasets: BTC/ETH/SOL × 1h/4h, binanceus fees, long-only single mode, IS since 2025-06-10, OOS since 2026-01-01):

| Mechanism | Verdict | OOS mean Sharpe (isolated) |
|---|---|---|
| 1. HTF classification window (`htf_factor`) | **Shipped** (default 6) — but only paired with selectivity (below) | -2.86 naively; -0.32 in final design |
| 2. Confirmation hysteresis (`confirm_buckets`) | **Shipped** (default 2) — worth +0.3–0.8 OOS Sharpe across the htf grid | -1.74 isolated |
| 3. Transition-as-signal (`trend_entry="transition"`) | **Rejected as default** — by the time a slow confirmed label flips, the move is exhausted; entering on the flip buys tops | -2.07 isolated, -3.06 combined |

The decisive finding is **selectivity, not label speed**: #967's `regime_adaptive` trades 50–80×/window on 1h; at ~0.3% round-trip cost the churn explains most of its bleed. Passing incumbents (`mean_reversion_pro`, `mtf_confluence`) trade 0–8×/window. The shipped default is therefore **fade-only in confirmed true ranges**: z-recovery fades fire only when the confirmed HTF label is `ranging_quiet`/`ranging_volatile`; clean-trend entry modes stay sweepable behind `trend_entry` but default off.

HTF labels also fix #967's choppy-fade hack for real: a same-timeframe big-move label almost always covers the z-recovery bar (which forced #967 to map choppy trends to fades), but a 6×-slower bucket label doesn't flip on one native recovery bar — so ranging-only fades stay reachable (`fade_labels="all_mr"` restores the #967 mapping for comparison). The #967 slow-trend drift veto is retained unchanged.

## Performance — clears the bar

Bar recomputed per the #963 protocol on this window (per-dataset median of the audit top-8 incumbents re-run on the identical harness): **Sharpe -0.724, DDadj -0.490**.

Defaults (`htf_factor=6, confirm_buckets=2, period=14, mr_entry_z=2.0`, fade-only), long-only single mode:

[...]

PR #1002, opened 2026-06-12
feat(strategies): M4 regime-profile allocation — slow regime switch between two validated profiles (#998)

### LINKED ISSUE #998: M4 — regime-profile allocation: slow regime switch between two validated profiles (#977)
Child of the #977 umbrella.

## Definition — regime-profile allocation

Run **two validated profiles of one strategy** with a slow drift/regime switch deciding which is active, instead of shipping one compromise default. Reference case: #976's `regime_adaptive_htf` — fade-only wins the grind-year protocol window, breakout+drift-confirm dominates 2023-2025H1 trending years; the windows are disjoint by regime, and M1 step 8 deliberately stops at "ship the protocol-window winner, expose the alternative via `DEFAULT_PARAM_RANGES`". M4 is the follow-through: make the alternative reachable automatically.

## What it needs (why this is not just tooling)

Unlike M2/M3/M5 this requires a **live mechanism**, not only backtest work:

1. A slow regime/drift classifier deciding the active profile (candidates: the existing multi-window regime store (#879) with a long-window spec, or the #976 drift-confirm signal) — switching must be slower than the profiles' own signals or it becomes curve-fitting the boundary (M1 step 8's warning).
2. A config surface for per-profile param sets on one strategy entry, plus switch hysteresis (flat-only switching, like the existing SIGHUP rules for regime fields).
3. Backtest parity: the backtester must replay the same switch rule, or the feature is HL-live-only and rejected at `--config` load like its siblings (`regime_directional_policy`, `regime_window_divergence`).
4. Validation: each profile must independently clear the M1 bar on its own regime's windows; the switched composite must beat the better single profile on the full multi-window table — otherwise ship the single profile and close as a negative finding.

## Consumers

#985 (donchian_breakout regime-gate), #981 (mean_reversion_pro), and `regime_adaptive`/`regime_adaptive_htf` (the #976 two-profile case this generalizes).

---

richkuo:
## Summary

Implements **M4 regime-profile allocation** (#998): run two validated open-strategy param profiles of one strategy and let a slow long-window regime classifier (the #879 store) decide which is active, instead of shipping one compromise default. Generalizes the #976 `regime_adaptive_htf` two-profile case (fade-only in grind windows, breakout+drift-confirm in trending years).

Closes #998.

## Mechanism (HL perps only, live + paper)

1. New optional `regime_profile_allocation` block — `window`, `profiles` (label → profile), `param_sets` (exactly two), `confirm_bars`, `initial_profile`. **No config-version bump** (additive/optional). Validated at load: HL-perps + `regime.enabled` gating (`config.go`); param_sets count, label coverage, window existence (`regime_window_spec.go`).
2. Switch is **hysteretic and flat-only**: commits only after the desired profile persists for `confirm_bars` **closed bars** and the strategy is flat. While a position is open the active profile is frozen to `pos.OpenProfile` (hold-on-transition); the counter keeps accruing so the first flat bar commits. Pure state machine `resolveRegimeProfile`.
3. The active profile's params merge **over** the loop-local `open_strategy.params` **before** the check subprocess, so the profile shapes the signal itself. Rides the existing `--strategy-refs` JSON — **no new CLI flag, no `probeArgv` change** (symmetric deploy).
4. Hysteresis counts **closed bars** (regime bundle `BarTime`), so live cadence matches the per-bar backtester exactly.

## Persistence & lifecycle

- `positions.open_profile` (frozen for the position's life) and `strategies.active_profile` (flat-switch state across restarts); the pending counter re-arms on restart — a restart can only **delay** a switch, never fast-track one.
- SIGHUP: reshape blocked while open; applied + state-reset when flat.
- Surfacing: `/status` + trade DMs.

## Backtest parity

The switch rule is a pure function of closed-bar OHLCV, so the backtester **replays it** (NOT rejected at `--config` load like `regime_window_divergence`). `backtester.py` selects per-bar among `signal__<profile>` columns driven by a shifted `_profile_label`; `run_backtest.py --config` and `eval_windows.py --profile-allocation` build the per-profile legs + label series. Look-ahead invariants preserved (signal + label both `shift(1)`).

## Validation

The full protocol (each profile clears the M1 bar on its own regime's windows; the switched composite beats the better single profile on the full 5-window × 6-dataset table) is the operator's research step that decides ship vs negative finding. Preliminary check on 2024 BTC/USDT 1h:

1. fade single — Sharpe 0.01
2. trend single — Sharpe 1.12
3. switched composite — Sharpe 1.12 (the long-window classifier selects the trend profile throughout the trending window, exactly matching the better single profile)

confirming the switch tracks regime and picks the regime-appropriate profile.

## Tests

- **Go:** state-machine table tests (hysteresis, flat-only, open-freeze-then-commit, empty-label freeze, cold start), config validation, hot-reload `EqualForReload`, ValidateConfig integration, full LoadConfig-from-file, DB round-trip of `open_profile`/`active_profile`.
- **Python:** switcher state machine, engine replay (active-profile selection, no-switch-when-fade), missing-column guards.
- Full Go suite and full Python suite (1180) pass.

## Notes for reviewers

- `closeStrategyOwnedKeys` is **not applicable** — this is an open-strategy-level feature, not a close evaluator.
- Profiles swap `open_strategy.params` only; `close_strategy`, stops, leverage, direction are shared between profiles.

---
Created with LLM: Opus 4.8 | high | Harness: commit-push-pr

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -234,6 +234,91 @@ def _open_action_from_signal(signal: int) -> str:
     return "none"
 
 
+def _parse_profile_allocation(alloc: Optional[dict]) -> Optional[dict]:
+    """Validate and compact a regime_profile_allocation block for the engine
+    (#998). Returns None when unset; raises ValueError on a malformed block so
+    a misconfigured --config fails loudly rather than silently single-profile.
+
+    The compact form mirrors the Go RegimeProfileAllocation: profiles (label ->
+    profile name), param_sets (profile -> params), confirm_bars, initial_profile.
+    The window key and per-profile signal computation live in run_backtest.py;
+    the engine only replays the switch over the supplied ``signal__<profile>``
+    columns.
+    """
+    if not alloc:
+        return None
+    profiles = dict(alloc.get("profiles") or {})
+    param_sets = dict(alloc.get("param_sets") or {})
+    confirm_bars = int(alloc.get("confirm_bars") or 0)
+    initial_profile = str(alloc.get("initial_profile") or "").strip()
+    if len(param_sets) != 2:
+        raise ValueError(
+            f"regime_profile_allocation.param_sets must define exactly 2 "
+            f"profiles (the M4 two-profile model), got {len(param_sets)}"
+        )
+    if confirm_bars < 1:
+        raise ValueError("regime_profile_allocation.confirm_bars must be >= 1")
+    if initial_profile not in param_sets:
+        raise ValueError(
+            f"regime_profile_allocation.initial_profile={initial_profile!r} "
+            f"is not a param_sets profile {sorted(param_sets)}"
+        )
+    for lbl, prof in profiles.items():
+        if prof not in param_sets:
+            raise ValueError(
+                f"regime_profile_allocation.profiles[{lbl!r}]={prof!r} is not "
+                f"a param_sets profile {sorted(param_sets)}"
+            )
+    return {
+        "profiles": profiles,
+        "param_sets": param_sets,
+        "confirm_bars": confirm_bars,
+        "initial_profile": initial_profile,
+        "names": sorted(param_sets),
+    }
+
+
+class _ProfileSwitcher:
+    """Per-bar flat-only, confirm_bars hysteresis profile switch — the exact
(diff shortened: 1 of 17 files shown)


richkuo:
@claude

Issue #990, opened 2026-06-11
vwap_rejection_st: short-leg baseline now runnable, churn screen first (M1)


richkuo:
Part of #977 (Methodology M1, short-leg baseline now unblocked by #989) / #978. Audit source: #956 / #975. M5 source: #999 / #1003.

## Why

vwap_rejection_st is **short-only and still lacks a real short-leg verdict**. The original audit compare mode was long/flat only, so it produced 0 trades and an "n/a (short-only, unmeasured)" verdict; that evidence gap remains until the short leg is actually run. Same research queue as bear_pullback_st (#989), but the shared short-leg measurement path has now landed: `backtest/eval_windows.py --direction short` models `signal=-1` as short open and `signal=+1` as short close.

A caution flag: its long-side cousin vwap_reversion was the **worst strategy in the registry** (-59.5%, 1,504 trades, pure fee churn). The rejection variant must prove it does not share the churn gene before any tuning work.

## Current tooling state

- **M1 short-leg baseline is runnable now**: use `uv run --no-sync python backtest/eval_windows.py --strategy vwap_rejection_st --registry futures --direction short` on the audit-identical datasets/windows.
- **M5 churn screen still needs short-aware execution**: the committed `fee_audit.py` report is long/flat and still labels vwap_rejection_st as `unscreened_short`. Add/adapt a focused net-vs-gross screen that threads `direction="short"` through the same `run_leg` path before tuning.
- **Do not route this through optimize mode yet**: `run_backtest.py --mode optimize --direction short` is rejected because the walk-forward warmup seeder is still long-only.

## Focus

1. **First baseline (M1 steps 1-2)**: audit-identical datasets + protocol OOS window against the per-window incumbent-median bar, using `--registry futures --direction short`.
2. **Churn screen first (M5 spirit)**: before any tuning, check trade count x fee drag on the short leg. If it trades like vwap_reversion, the fee arithmetic alone may decide the verdict.
3. **Only then decide follow-up scope**: bottom-tier/churn result means deprecate; mid-tier or better can graduate to an M1 application child under #977.

## Verdict gate

Bottom-tier baseline or vwap_reversion-class churn → deprecate. Mid-tier or better → M1 application child under #977.

## Validation bar

#955 criteria on the measurable short leg; `--list-json` byte-identical unless the implementation intentionally changes registry output.

---
Created with LLM: Fable 5 | high | Harness: Claude Code

richkuo:
[...]

Issue #977, opened 2026-06-11
[C15] Strategy improvement initiative: methodology index (M1–M5) and validation bar


richkuo:
## Problem

#955 raised the bar for *new* strategies, but there is no repeatable process for improving *existing* ones. The #973 → PR #976 work produced one and validated it end-to-end (`regime_adaptive` baseline OOS mean Sharpe -1.63 → new `regime_adaptive_htf` at -0.32, clearing the incumbent-median bar). This umbrella codifies that process as **Methodology M1** and establishes the structure for adding further methodologies over time.

**This is the global umbrella for the strategy-improvement initiative.** Each methodology lives in its own child issue (definition, blockers, deliverables, reference application); per-strategy applications are tracked in the child umbrella #978 (#979-#993, one per strategy). This issue holds the index, the methodology → strategy mapping, and the validation bar.

## Methodology children

- [x] #995 — **M1: multi-window incumbent-relative validation.** The 8-step protocol (baseline reproduction → incumbent-median bar → mechanism isolation → plateau → held-out windows) + the shared `backtest/eval_windows.py` harness. Reference application already merged (#973 → PR #976); harness merged in PR #994.
- [x] #996 — **M2: close-stack co-optimization.** Joint open+close param sweeps; close-param optimization wired into the optimizer (the gap #944 deferred). Merged in PR #1000. Unblocks #983/#984.
- [x] #997 — **M3: exit-quality refinement.** Holding-time/excursion diagnostics (`backtest/exit_diagnostics.py`) + `time_stop`/`atr_stop`/`zscore_target` close evaluators. Merged in PR #1001.
- [x] #998 — **M4: regime-profile allocation.** Slow regime/drift switch between two validated profiles of one strategy (generalizes #976's fade-only vs breakout+drift finding). Merged in PR #1002.
- [x] #999 — **M5: fee-aware selectivity audit.** Registry-wide trade-count x fee-drag screen with a gross-vs-net salvage test (`backtest/fee_audit.py`, report at `docs/research/fee-audit-m5.md`). Merged in PR #1003.

[...]

PR #1004, opened 2026-06-12
feat(backtest): #989 short-leg measurement path + first bear_pullback_st baseline

### LINKED ISSUE #989: bear_pullback_st: build short-leg measurement, then first baseline (M1, blocked)
Part of #977 (Methodology M1, blocked on tooling) / #978. Audit source: #956 / #975.

## Why

bear_pullback_st is **short-only and has never been measured**: the audit's compare mode is long/flat only, so it produced 0 trades and an "n/a (short-only, unmeasured)" verdict. We have zero evidence either way — in a year where buy-and-hold lost 44-61%, a working short strategy should have shone, and we can't see whether it did.

## Focus

Measurement prerequisite first, then a standard M1 baseline:

1. **Short-leg measurement path.** #951 already gave `Backtester` direction/invert modeling for `--config` runs; extend the audit/eval harness (`backtest/eval_windows.py` from #977's tooling deliverable) so short-entry strategies are evaluated on their short leg rather than silently producing 0 trades. This unblocks vwap_rejection_st (#990) and the short legs of every bidirectional strategy at the same time — build it once.
2. **First baseline (M1 steps 1-2)**: audit-identical datasets + protocol OOS window, incumbent-median bar. Note the bar itself comes from long-leg incumbents in a bear market — interpret with that asymmetry in mind, per-window recomputation as always.

## Verdict gate

Bottom-tier baseline → deprecate. Mid-tier or better → M1 application child under #977.

## Validation bar

#955 criteria once measurable; `--list-json` byte-identical.

---

richkuo:
## Summary

Builds the #989 short-leg measurement path and lands the first-ever bear_pullback_st baseline. The plain signal path was structurally long/flat, so short-only and bidirectional strategies scored zero trades on the audit/M1 harness; this unblocks #990/#992 and every M5 `unscreened_short` verdict.

- **Backtester short/flat plain path:** `direction="short"` with no close evaluator now mirrors the long/flat path exactly — signal −1 opens a short with full notional (same mechanics as the engine path's short open), +1 buys it back. Standalone fixed/trailing/pct stops mirror above the entry; the trailing stop walks a low-water mark and only tightens; `starting_long` is rejected on this path. The open/close engine path is unchanged.
- **Loader guards relaxed precisely:** `direction="short"` without close refs is now modelable and accepted (`run_backtest.load_strategy_config`, `eval_windows.validate_candidate`); `"both"` stays rejected on the plain path — one signal cannot open one side and close the other.
- **`eval_windows.py --direction long|short|both`** selects the candidate leg on the audit-identical harness.

## First baseline: bear_pullback_st (M1 steps 1–2, futures registry, short leg)

Raw short/flat leg (audit-identical semantics): protocol OOS **PASS**, beating the long-leg incumbent bar 6/6 on Sharpe and DDadj — but the strategy emits entries only (−1, never +1), so every leg is ONE short held to end-of-window. The PASS is short-and-hold in a bear market, and the held-out bull years are catastrophic (2023 SOL 4h: −433%). Low evidentiary value on its own.

Live-like leg (default close stack: `tiered_tp_atr` + 1.5×ATR SL, engine path): protocol **FAIL**, but marginal — OOS mean Sharpe −0.76 vs bar −0.75, DDadj −0.30 vs −0.49 (beats DDadj, misses Sharpe by a hair); bull-year losses are controlled (−3% to −29% vs the raw leg's blowups); 2025H1 passes. Not bottom-tier — the entry has some signal that the default exit stack bleeds (1h legs churn 53–127 trades).

| window | raw short/flat | live-like close stack |
|---|---|---|
| is | PASS | FAIL |
| oos (protocol) | **PASS** (Sharpe 2.35 vs −0.72) | **FAIL** (Sharpe −0.76 vs −0.75) |
| 2023 | FAIL (−0.18 vs 1.36) | FAIL (−1.13 vs 1.46) |
| 2024 | FAIL | FAIL |
| 2025H1 | PASS | PASS |

Per the #989 verdict gate this is mid-tier, not deprecate: recommend an M1 application pass (exit-quality knobs from M3, regime gating) tracked under #977/#978.

## Test plan

- 18 new tests in `backtest/tests/test_backtester_short_leg.py`: round-trip PnL sign and fees, N-close→N+1-open fill alignment (entry and stop), stop mirroring above entry (fixed ATR / trailing low-water / pct), trail-never-loosens, regime gate, invert_signal composition, long-path regression guard, engine-path precedence with close refs, `starting_long` rejection.
- Updated the loader-guard tests (short now accepted, both still rejected) in `test_strategy_refs_641.py` / `test_eval_windows.py`.
- Full Python suite: 1516 passed.

Closes #989

---


richkuo:
@claude fable

richkuo:
Review addressed in 1740723.

**Needs Fixing 1 (bust-account sign flip):** guarded all four open blocks — plain long/short AND engine long/short — with `cash > 0`, not just the reported short-open sites: the same state reaches the engine LONG block via direction=both after a blown short, and the plain long block carries the guard so the invariant holds by construction. Entries from a bust account now skip until end of data. Regression tests cover each must-survive case: consecutive re-entry attempts after blowup (cash does not compound), the engine path via a `time_stop` force-close, the inverse side (long open from negative cash), and the exact-wipeout `cash == 0` boundary (no zero-share dangling trade). Full suite: 1521 passed.

**Create Follow-up Issue 1 (negative-equity metric inversion):** filed as issue 1005 with the liquidation-floor / robust-returns / degenerate-flag options.

The posted bear_pullback_st baselines are unaffected: the raw legs hold one trade per window (no re-entry), and the close-stack legs run with a 1.5×ATR stop that keeps cash positive throughout.

---

richkuo:
@claude fable rereview

richkuo:
@claude fable rereview

richkuo:
Review addressed in 7aca206 — `--direction` now reaches every mode, with all three must-survive cases covered:

1. **single + `--config` supplying `direction`** — rejected loudly instead of setdefault (a CLI flag silently losing to the config is the same wrong-leg class; mirrors the existing `--close-strategy`-alongside-`--config` rejection).
2. **`--direction both` in single mode with no close refs** — rejected at the CLI before any data load, AND `Backtester.run` now rejects `direction="both"` on the plain single-leg path as a backstop for API callers that bypass the loaders.
3. **compare/multi fan-out** — `run_all_strategies`/`run_multi_asset` gained a `direction` passthrough; `main()` forwards it. `--sweep-close --direction both` in optimize mode still passes the guard (the grid supplies the engine path).

8 new tests (per-mode CLI threading, both-rejection before any run, config collision, sweep-close passthrough, Backtester backstop). Full suite: 1529 passed.

---

richkuo:
@claude rereview

richkuo:
Review addressed in a9fa986. Optimize mode now rejects `--direction short` loudly with the true reason — the walk-forward warmup seeder (`warmup_exit_long_entry`) is long-only and would carry a phantom long into the short run — pointing to `--mode single` / `eval_windows.py --direction short`. Must-survive coverage:

1. **`--sweep-close --direction short` (headline)** — rejected at the CLI before any data fetch; `walk_forward_optimize` rejects it too as a backstop for API callers.
2. **close-evaluator-only stacks + short** — rejected wholesale at the optimizer (the seeder limitation applies regardless of close stack); tested with both fixed close refs and a close-only grid, so the phantom-long path is unreachable.
3. **long path byte-identical** — guards only add raises for short/both; existing long-direction optimize tests pass unchanged.

Also: the stale "plain path cannot open shorts" justification is corrected everywhere (the remaining stack guard narrows to `both` with the single-leg-path reason), `--direction both` against a grid containing no-close baselines is rejected at the CLI instead of tracebacking mid-optimize, and the mocked sweep test that never reached the guard is replaced with a close-only `--close-stacks-json` end-to-end threading test plus real rejection tests. Full suite: 1532 passed.

---

richkuo:
@claude rereview

PR #1007, opened 2026-06-13
fix(backtest): #1005 liquidation floor — stop return/Sharpe sign inversion on negative equity curves

### LINKED ISSUE #1005: Backtester: return/Sharpe metrics invert sign when the equity curve goes negative
From the PR #1004 review (Create Follow-up Issue item).

`_calculate_metrics` computes per-bar returns with `equity.pct_change()`. When the equity curve crosses into negative territory — reachable for stop-less short legs that lose more than 100% (the #989 short/flat harness measures exactly these, e.g. bear_pullback_st 2023 SOL 4h at −433%) — `pct_change` divides by a negative base, so a recovery reads as a negative return and a deepening blowup reads as a positive one. Sharpe (the protocol verdict metric) and volatility are corrupted for the affected stretch; total return and max DD are anchored at `initial_capital` and stay correct in sign.

Pre-existing for engine-path shorts (negative equity means an unmodeled liquidation) and orthogonal to the #1004 entry guard, which only stops NEW entries from a bust account.

Options to evaluate:

1. Floor the equity curve at a liquidation point (equity clamped at 0 once bust; mirrors a real account) — simplest and economically honest.
2. Compute per-bar returns robustly for non-positive bases (e.g. delta over `abs(prev)` or over `initial_capital`).
3. Flag the run (`liquidated: true`) so harness consumers (eval_windows, fee_audit) can treat the leg as degenerate rather than trust its Sharpe.

Whatever the choice, eval_windows/fee_audit verdicts on blown legs must not silently rank a deeper blowup above a shallower one.

---

richkuo:
Closes #1005

## Problem

`_calculate_metrics` computes per-bar returns with `equity.pct_change()`. When the equity curve goes negative — reachable for stop-less short legs losing more than 100% (the #989 short/flat harness measures exactly these, e.g. bear_pullback_st 2023 SOL 4h at −433%) — `pct_change` divides by a negative base: a deepening blowup reads as a **positive** per-bar return and a recovery as negative. Sharpe (co-verdict metric in eval_windows), Sortino, and volatility were corrupted for the affected stretch; total return and max DD were already anchored at `initial_capital` and stayed sign-correct.

## Fix (issue options 1 + 3 combined)

A real account is dead at zero:

- **Liquidation floor** (`backtester.py::_calculate_metrics`): detect the first bar where equity ≤ 0 and sticky-floor the curve at 0 from that bar onward (no resurrection if the short later recovers). Return and DD floor at −100%, the bust bar contributes a −100% return, post-bust bars drop out as NaN. Deeper and shallower blowups now tie instead of inverting — verdicts can never rank a deeper blowup above a shallower one.
- **`liquidated` flag** (additive results key) threaded through the harness so blown legs are never silent:
  - `eval_windows`: `leg_from_results` propagates it; report rows get a `LIQ` marker and the verdict line counts liquidated legs (`liquidated_legs` in `score_candidate`; verdict logic unchanged — floored metrics already rank blown legs honestly).
  - `fee_audit`: `screen_leg`/`aggregate_strategy` carry `liquidated`/`n_liquidated`; the markdown report gains a "Liquidated legs" section only when any leg blew up. `salvage_verdict` needs no change — a floored −100% gross already lands on `deprecate`.
  - `format_results` prints a LIQUIDATED banner.

Option 2 (robust returns over `abs(prev)`) was rejected: it produces statistically fictional Sharpe values for an economically dead account and still lets a blown leg carry a plausible-looking Sharpe into verdicts.

Out of scope: force-closing the simulated position at bust (sim behavior, not metrics) and `final_capital` (stays raw cash; the flag tells the story).

## Verification

- Real-data repro of the issue's example: `eval_windows.py --strategy bear_pullback_st --registry futures --direction short --windows 2023` — SOL 4h leg now reads return −100%, Sharpe −2.26 (was −433% with corrupted Sharpe), `LIQ` marker and `[3 liquidated leg(s)…]` verdict note shown.
- 13 new tests in `backtest/tests/test_metrics_liquidation.py`: sign-inversion regression (deepening blowup must read negative Sharpe), deeper-never-ranks-above-shallower tie, sticky floor (no resurrection), zero-equity boundary, healthy-curve metrics byte-identical (positive control), end-to-end stop-less short blowup on the #989 path, flag propagation + report markers in eval_windows, count + conditional markdown section in fee_audit, legacy dicts without the key default safely.
- Full suite: 1545 passed.

---

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -2308,6 +2308,20 @@ class Backtester:
         # Calmar ratio
         calmar = annual_return / abs(max_drawdown) if max_drawdown != 0 else 0
 
+        # Liquidation risk-adjusted floor (#1005): when the sticky floor leaves
+        # <2 surviving returns (a leg busting within 1-2 bars — the post-bust
+        # NaN tail drops out), the variance guards above collapse Sharpe/
+        # Sortino/volatility to a NEUTRAL 0.0. That reads a dead account as
+        # "fine" and ranks a fast blowup ABOVE a slow one — re-inverting the
+        # exact axis this issue fixed. Floor every blown leg to the annualized
+        # total-loss magnitude (−100% return) so all deaths tie below any
+        # surviving leg, mirroring the −100% floor already applied to
+        # return/DD. Uniform (not path-dependent) so two busts at different
+        # bars tie instead of the earlier one out-ranking the later.
+        if liquidated:
+            sharpe = sortino = -ann_factor
+            volatility = ann_factor
+
         return {
             "total_return_pct": round(total_return * 100, 2),
             "annual_return_pct": round(annual_return * 100, 2),
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -2232,6 +2232,20 @@ class Backtester:
         equity = equity_df["equity"]
         ann_factor = math.sqrt(periods_per_year(timeframe))
 
+        # Liquidation floor (#1005): a stop-less short losing >100% drives
+        # equity negative, and pct_change over a negative base inverts return
+        # signs (a deepening blowup reads as a positive return, a recovery as
+        # negative), corrupting Sharpe/Sortino/volatility. A real account is
+        # dead at zero — floor the curve at 0 from the first bust bar onward
+        # (sticky: no resurrection if the position later recovers) and flag
+        # the run so harness consumers (eval_windows, fee_audit) can surface
+        # it. Post-bust bars contribute 0/0 = NaN returns, dropped below.
+        liquidated = bool((equity <= 0).any())
+        if liquidated:
+            bust_pos = int(np.argmax(equity.values <= 0))
+            equity = equity.copy()
+            equity.iloc[bust_pos:] = 0.0
+
         # Anchor return + drawdown at initial_capital so seeded runs (where
         # equity[0] reflects the starting_long mark-to-market, not the true
         # pre-trade balance) don't distort the baseline. For non-seeded runs
@@ -2307,6 +2321,7 @@ class Backtester:
             "total_trades": total_trades,
             "avg_win_pct": round(avg_win * 100, 2),
             "avg_loss_pct": round(avg_loss * 100, 2),
+            "liquidated": liquidated,
         }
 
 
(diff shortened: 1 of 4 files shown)


richkuo:
@claude fable

richkuo:
@claude

richkuo:
Needs Updates

### Needs Fixing
1. **Early-bust legs report Sharpe/Sortino/volatility = 0.0 instead of negative, re-inverting the exact signal #1005 set out to fix.**
`backtest/backtester.py::_calculate_metrics` (Sharpe ~L2252, Sortino ~L2261, volatility ~L2298): after the liquidation floor truncates `daily_returns` to the path-to-death, a leg that busts within 1–2 bars leaves ≤1 surviving return sample, so the `len(daily_returns) > 1` guard zeroes Sharpe/Sortino/volatility. A blown account then reads as *neutral* (0.0), not bad — and because Sharpe is a co-verdict metric in `eval_windows`, a leg busting at bar 2 out-ranks one busting at bar 4 on the Sharpe axis. The new tests only use 4–5-bar curves (≥2 samples) so they miss this. `total_return`/`max_drawdown`/`ddadj` still floor to −100%, so pass/fail is backstopped — but the Sharpe surface the PR claims to fix is still inverted for early busts. Fix: when `liquidated`, force Sharpe/Sortino to a negative sentinel (or compute over the floored series without the `>1` short-circuit) and assert volatility is non-zero; add 1-bar and 2-bar bust tests.
**Invariant:** a liquidated leg must never report a Sharpe/Sortino ≥ that of a non-liquidated leg, regardless of how few bars preceded the bust.
**Must survive:** a 2-bar bust (single surviving return sample); a bar-1 bust (zero surviving samples → empty-series / div-by-zero path); two liquidated legs busting at different bars — the earlier bust must never out-rank the later on Sharpe.

### Recommended Optional
1. **`liquidated_legs` is counted over `rows` while every adjacent metric uses `scored`.**
`backtest/eval_windows.py:175`: the count includes liquidated legs with no matching incumbent bar (excluded from `scored`), so it can exceed `scored_datasets` and reference legs absent from the displayed means. If counting all blowups is intended, keep it and say so in the comment; otherwise count over `scored` for consistency.
**Invariant:** operator-facing counts in a score dict are computed over the same leg set as the metrics they sit beside, unless explicitly documented otherwise.
**Must survive:** a candidate where a liquidated leg has no incumbent bar (rows ⊋ scored); a candidate where all legs are scored (count must be identical under either definition).

richkuo:
@claude rereview

richkuo:
@claude rereview

PR #1008, opened 2026-06-13
fix(shared-wallet): classify perps force-closes as 'perps' not 'futures'

richkuo:
## Summary

`forceCloseAllPositions` labeled any `pos.Multiplier>0` close as `futures` (the legacy OkxPerps/TopStep convention). HL perps and OKX perps carry `pos.Multiplier=1` as the canonical perps PnL valuation value (#254/#497 — NOT a contract multiplier), so every perps circuit-breaker / kill-switch close was recorded with the wrong `trade_type` label (`futures` instead of `perps`).

**Correction (was overstated in an earlier draft):** this label is cosmetic for accounting. `tradeLedgerDeltaSQL` (`trade_pnl.go:48`) computes a row's ledger contribution from `is_close` / `pnl_gross` / `realized_pnl` / `exchange_fee` only — it never reads `trade_type`. The sole `trade_type` filtering anywhere (`db.go:1695`, `db.go:1770`) excludes `scale_in`/`funding` from open-counting, not from value sums. So relabeling `futures`→`perps` leaves every `LedgerByID` sum, `displayStrategyValue`, and drift number byte-identical. This PR does **not** by itself change the drift math.

**What actually caused the live drift** (hyperliquid/0x736dD10faf3CaE88199Dab551dBe0713fe68cc86: Σ member value $2024.89 vs real balance $1149.76, $875.13 raw reconciliation diff): the rowid-54 `realized_pnl` corruption ($1074.52 vs the source-of-truth `closed_positions` value of $0.22) plus phantom close rows. Those were repaired by the out-of-band `state.db` mitigation below — not by the code in this diff. The mechanism that booked a force-close leg's `realized_pnl` ~4884× its `closed_positions` row is **not yet root-caused** (see Follow-up).

## What this PR changes (observability + correctness-of-label only)

- **Add `classifyPositionTradeType` helper**: HL perps / HL manual / OKX perps with `pos.Multiplier=1` → `perps`; TopStep / legacy futures keep `futures`; `pos.Multiplier=0` → `spot`. Makes future perps force-close rows carry an accurate label.
- **Wire `forceCloseAllPositions` through the helper.** PnL branch is unchanged (qty × multiplier × price_delta works for both perps and futures); only the label string differs.
- **Add `TestForceCloseAllPositions_TradeType_PerpsVsFutures`** pinning the per-platform label mapping.
- **Make `formatSharedWalletDriftAlert` show BOTH the raw reconciliation diff** (memberSum - balance, consistent with the numbers above it) **AND the post-baseline drift**, so an operator can distinguish new drift from legacy data baked into the adopted baseline.

## Live mitigation (applied to scheduler/state.db — not in this diff)

- Reclassify the 12 `futures` phantom trades on hl-rmc-eth-live / hl-tema-eth-live / manual-eth as `perps` (correct label, no PnL/ledger change).
- Fix trade rowid 54 `realized_pnl`: 1074.52 → 0.216 (matches the source-of-truth `closed_positions` row and the Details string 'PnL: $0.22'). Only rowid 54 had a `realized_pnl` vs `closed_positions` mismatch; the other 11 phantom trades' PnL was internally consistent. **This edit — not the relabel — is what removed the inflation.**
- Reset `wallet_ledger_state.baseline_set` to 0; new baseline re-anchors on the next reconcile at **+$199.17** (the remaining legacy cash adjustments, ~17% of balance — non-trivial but no longer dwarfed by the $1074 bug).

## Follow-up (root cause, not addressed here)

The source that booked rowid-54's `realized_pnl=1074.52` against a `closed_positions` row of 0.22 is unaddressed; if it is code rather than a one-off, drift can recur. Tracked in #1009 (perps close-leg quantity/PnL corruption) — acceptance criterion: a close leg's `realized_pnl` must reconcile with its `closed_positions` row across all close paths (signal + force-close), with a regression test.

## Test plan

- [x] `go test ./scheduler -run TestForceCloseAllPositions_TradeType_PerpsVsFutures`
- [x] `go test ./scheduler -run TestReconcile` (full reconcile + drift surface)
- [x] Deployed to live (pid 39970); cycle 119311 reports Total value $1148.16, no SHARED-WALLET DRIFT alerts after baseline re-anchor

### DIFF
--- a/scheduler/risk.go
+++ b/scheduler/risk.go
@@ -1133,10 +1133,13 @@ func forceCloseKillSwitchPositions(s *StrategyState, sc StrategyConfig, prices m
 // for circuit-breaker / kill-switch close records. HL perps and OKX perps
 // carry pos.Multiplier=1 (#254/#497 perps PnL valuation convention — NOT a
 // contract multiplier), so the legacy "Multiplier>0 → futures" classifier
-// mislabels every perps force-close as "futures" and leaks phantom PnL into
-// the #954 shared-wallet trade ledger. TopStep/CME futures keep pos.Multiplier
-// as the real contract multiplier; that is the only branch where "futures"
-// is correct.
+// mislabels every perps force-close as "futures". This is an operator-facing
+// label fix only: tradeLedgerDeltaSQL (trade_pnl.go) keys on
+// is_close/pnl_gross/realized_pnl/exchange_fee and never reads trade_type, so
+// the label does NOT affect any ledger sum — relabeling here changes what an
+// operator sees (Discord/leaderboard/audit), not the #954 ledger math.
+// TopStep/CME futures keep pos.Multiplier as the real contract multiplier;
+// that is the only branch where "futures" is correct.
 func classifyPositionTradeType(s *StrategyState, pos *Position) string {
 	if pos == nil {
 		return "spot"
@@ -1169,8 +1172,9 @@ func forceCloseAllPositions(s *StrategyState, prices map[string]float64, logger 
 		// PnL branch is the same for perps (Multiplier=1) and futures
 		// (Multiplier=contract size) — qty*multiplier*price_delta. Only the
 		// trade_type LABEL differs by venue, classified via
-		// classifyPositionTradeType so perps force-closes no longer leak into
-		// the "futures" trade_ledger bucket.
+		// classifyPositionTradeType so perps force-closes carry an accurate
+		// operator-facing label. The label does not feed any ledger sum
+		// (tradeLedgerDeltaSQL ignores trade_type); it is display-only.
 		tradeType := classifyPositionTradeType(s, pos)
 		if pos.Multiplier > 0 {
 			if pos.Side == "long" {
--- a/scheduler/risk.go
+++ b/scheduler/risk.go
@@ -1129,6 +1129,32 @@ func forceCloseKillSwitchPositions(s *StrategyState, sc StrategyConfig, prices m
 	forceCloseAllPositions(s, prices, logger)
 }
 
+// classifyPositionTradeType maps a position to the correct trade_type label
+// for circuit-breaker / kill-switch close records. HL perps and OKX perps
+// carry pos.Multiplier=1 (#254/#497 perps PnL valuation convention — NOT a
+// contract multiplier), so the legacy "Multiplier>0 → futures" classifier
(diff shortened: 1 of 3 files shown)


richkuo:
Needs Updates

### Recommended Optional
1. **The PR description's causal claim — that the `futures` mislabel "inflated per-strategy ledger sums" — is not supported by the ledger code and should be corrected.**
`trade_pnl.go:48` `tradeLedgerDeltaSQL` keys only on `is_close`, `pnl_gross`, `realized_pnl`, and `exchange_fee`; it never reads `trade_type`. The only `trade_type` filtering anywhere (`db.go:1695`, `db.go:1770`) excludes `scale_in`/`funding` from open-counting, not from value sums. So relabeling a perps force-close from `futures` to `perps` leaves every `LedgerByID` sum, `displayStrategyValue`, and drift number byte-identical — there is no "futures trade_ledger bucket." The real inflation was the rowid-54 `realized_pnl` corruption (1074.52 vs true 0.22) plus phantom rows, fixed by the out-of-band `state.db` edit, not by this diff. Reword the Summary so a future operator does not read this PR as the fix for the drift incident. **(Done — PR body updated.)**
Invariant: a PR's stated root cause must match what its code actually changes.
Must survive: the inverse case (a reader assuming drift is now fixed and skipping root-cause work); a future trade_type-based ledger filter being added (would change this analysis — re-verify if so).

### Create Follow-up Issue
1. **The source of the rowid-54 `realized_pnl=1074.52` corruption is unaddressed; tracked in #1009 rather than a new issue.**
This PR's code only corrects the `trade_type` label going forward and the drift-alert wording; the value corruption that actually caused the $875 raw diff was repaired by manual DB surgery not in the diff. #1009 (perps close-leg quantity/PnL corruption on direction reversal) is the right class of bug and is now the tracked home for this finding. Note it is not a confirmed match for rowid 54 — #1009's uncapped `closeQty = fillQty` overstates the close leg ~2× and lives on the *signal* close path, whereas rowid 54 was ~4884× on the *force-close* path (`forceCloseAllPositions`). Acceptance criterion added to #1009: a close leg's `realized_pnl` must reconcile with its `closed_positions` row across all close paths (signal, partial, reversal flip, and force-close), with a regression test.

richkuo:
@claude

richkuo:
@claude rereview

PR #1011, opened 2026-06-14
feat(backtest): #988 funding data + carry booking for delta_neutral_funding

### LINKED ISSUE #988: delta_neutral_funding: add funding data to the backtester, then first baseline (M1, blocked)
Part of #977 (Methodology M1, blocked on tooling) / #978. Audit source: #956 / #975 (candidate verdict: BLOCKED).

## Why

The audit could not evaluate delta_neutral_funding at all: **0 trades, because the backtest path feeds it no funding data**. The strategy is untested rather than disproven — it cannot clear (or fail) its own gate until the harness can both feed it per-bar funding and book the carry.

Two root causes (validated against `main`):

1. **No funding wired to this strategy in the backtest path.** `run_backtest.py:27` only attaches the per-bar `funding_rate` column for `FUNDING_COLUMN_STRATEGIES = {"funding_skew"}` — delta_neutral_funding is excluded.
2. **The strategy is structurally non-backtestable as written.** `delta_neutral_funding_strategy` (registry.py:957-978) reads **scalar** params (`current_funding_rate`/`avg_funding_rate_7d`, defaulting to `0.0`) and emits a signal **only on the last bar** (`result.iloc[-1, ...]`). With the scalar default `0.0` it early-returns flat; even with a non-zero scalar it can produce at most one signal at the final bar. Attaching a per-bar column changes nothing until the function itself is rewritten to consume it.

## Focus

Tooling + strategy rewrite first, then a standard M1 baseline:

1. **Wire delta_neutral_funding into the existing funding-column path.** The fetch/cache/align infra already exists from #960 (`shared_tools/funding_fetcher.py`: `load_cached_funding`, `attach_funding_column`, `get_funding_history_range`; HL hourly funding cached in SQLite, `merge_asof` backward). Scope here is: add `delta_neutral_funding` to `FUNDING_COLUMN_STRATEGIES` (run_backtest.py:27) so the per-bar `funding_rate` column is attached — **not** building new caching.
2. **Rewrite `delta_neutral_funding_strategy` to consume the per-bar column.** Compute the rolling 7d funding average from the attached `funding_rate` column and emit **per-bar** signals, instead of the current scalar / last-bar-only decision (registry.py:961-977). **Parity constraint:** the live path injects scalar funding (`check_hyperliquid.py:141-148`, via `adapter.get_funding_rate` → `info.meta_and_asset_ctxs()`, adapter.py:637) — keep that path working or unify the column+scalar inputs so live and backtest stay consistent.
3. **Book funding accrual in the engine.** This strategy's PnL is carry, not direction. `backtest/backtester.py` has **zero** funding awareness today — it books only price PnL. The engine must charge/credit funding on the held notional per holding period (look-ahead-safe, backward-merged like ATR), or the result is meaningless.
4. **First baseline (M1 steps 1-2)**: with funding flowing and accrual booked, run the compare harness + protocol OOS window for its first real numbers against the incumbent-median bar. Keep the audit-identical harness shape so results stay comparable.

## Verdict gate

Bottom-tier baseline → deprecate. Mid-tier or better → open an M1 application child under #977. No tuning before the baseline exists.

## Validation bar

#955 criteria once measurable; `--list-json` byte-identical.

---

richkuo:
Closes #988

## Problem

The audit could not evaluate `delta_neutral_funding` (0 trades). Validated against `main`, the cause was two-fold, not just "no funding data":

1. The backtest path attached a per-bar `funding_rate` column only for `funding_skew` (`run_backtest.py`), never delta.
1. The strategy was structurally un-backtestable: it read **scalar** funding params and emitted a signal **only on the last bar** (`registry.py`), so even with a column attached it produced at most one terminal signal.
1. The engine booked only price PnL — `backtester.py` had zero funding awareness, so the strategy's entire carry edge was invisible.

## Changes

- **Strategy (`registry.py`)** — when a per-bar `funding_rate` column is present (backtest), compute the trailing 7d-average funding and emit a **per-bar** short/exit signal series; the 7d window is inferred from index spacing so it's the same ~7 days on 1h and 4h bars, with full-window warmup gating early signals. The **live/paper scalar path is preserved unchanged** (no column → last-bar decision). No registered params added, so `--list-json` stays byte-identical.
- **Plumbing (`run_backtest.py`, `funding_fetcher.py`)** — add `delta_neutral_funding` to `FUNDING_COLUMN_STRATEGIES` (reuses the #960 fetch/cache) and attach a new **timeframe-correct** `funding_accrual` column: the sum of hourly funding over each bar's interval, backward-looking (preserves the look-ahead invariant; a 4h bar sums its ~4 hourly events where the signal snapshot would take one).
- **Engine (`backtester.py`)** — auto-detect `funding_accrual` and book carry each bar against the position carried in: `funding_cash = -position * mark * rate` (a short receives positive funding, a long pays it). Opt-in via column presence, so no other strategy is touched; `funding_skew` intentionally excluded to preserve its established M1 baseline. Surfaces `total_funding_pnl` in results.

## Tests

- Accrual column: 1h per-bar, 4h interval-sum, right-closed boundary, total-equals-held invariant, empty-funding.
- Strategy series path: warmup gating, hysteresis hold band, column-overrides-live-scalar, valid signals; existing scalar tests unchanged.
- Engine booking: short collects/pays, long mirror, one-bar open lag, NaN-safe, no-column no-op, flat no-op.
- Full Python suite green (1569 passed; 19 new tests across the three areas).

## Follow-up (not in this PR)

The first M1 baseline run itself — `eval_windows.py --strategy delta_neutral_funding --registry futures --direction short` on the protocol windows — needs network funding fetch; this PR lands the tooling + carry model that unblocks it. Per the verdict gate in #988: bottom-tier → deprecate; mid-tier or better → open an M1 application child.

---

### DIFF
--- a/backtest/backtester.py
+++ b/backtest/backtester.py
@@ -1382,6 +1382,14 @@ class Backtester:
         )
         active_profile = ""
 
+        # #988: carry strategies (delta_neutral_funding) attach a per-bar
+        # `funding_accrual` column — the total funding rate accrued over the bar.
+        # When present, book it each bar against the position carried into the
+        # bar. Auto-detected so no construction-site plumbing is needed; other
+        # strategies' frames have no such column and are unaffected.
+        book_funding = "funding_accrual" in df.columns
+        total_funding_pnl = 0.0
+
         for i, (idx, row) in enumerate(df.iterrows()):
             fill_price = row["open"] if has_open else row["close"]
             mark_price = row["close"]
@@ -1409,6 +1417,22 @@ class Backtester:
             # trail_from_here path inside _maybe_apply_sl_after already seeds
             # the HWM at the partial-close fill price. See #715.
             sl_after_just_applied = False
+
+            # #988: book funding carry on the position carried into this bar,
+            # over this bar's interval, before marking equity to the close. A
+            # long pays funding when the rate is positive, a short receives it:
+            #   funding_cash = -position * mark * rate
+            # (position is signed: + long, - short), which lands in `cash` and
+            # flows into the close-marked equity below. Newly opened positions
+            # this bar start accruing next bar; closed positions stop after
+            # their last full bar — the standard discrete one-bar convention.
+            if book_funding and position != 0:
+                accrual = row.get("funding_accrual", 0.0)
+                accrual = float(accrual) if accrual == accrual else 0.0  # NaN→0
+                if accrual != 0.0:
+                    funding_cash = -position * mark_price * accrual
+                    cash += funding_cash
+                    total_funding_pnl += funding_cash
 
             equity = cash + position * mark_price
             equity_curve.append({"date": idx, "equity": equity})
@@ -1981,6 +2005,7 @@ class Backtester:
             "end_date": str(df.index[-1]),
             "initial_capital": self.initial_capital,
             "final_capital": round(final_equity, 2),
+            "total_funding_pnl": round(total_funding_pnl, 4),
             "params": open_ref.get("params") or params or {},
             "open_strategy": open_ref,
             "close_strategies": [dict(r) for r in self._close_refs],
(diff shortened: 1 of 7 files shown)


richkuo:
@claude review

PR #1013, opened 2026-06-14
fix(#1012): extension-based .db protection in gitignore + update.sh rsync

### LINKED ISSUE #1012: Protect all .db files: generic gitignore + update exclusion by extension
## Problem

State databases must never be committed or overwritten by an update. Today the protection exists but is enumerated by specific filename rather than by extension, so a new or renamed `.db` file silently falls outside both safeguards.

**`.gitignore`** lists individual DB paths, not a generic glob:
- `trading_bot.db`, `scheduler/state.db`, `*.db-wal`, `*.db-shm`, `*.db.lock`
- No bare `*.db` — a DB at any other path (e.g. a per-instance `db_file`, a future second DB) is trackable and can be committed by accident.

**`scripts/update.sh`** (`run_rsync_from`, `update_resolve_db_exclude`) excludes only the configured `db_file` path plus `trading_bot.db*`:
- `--exclude="${db_excl}*"` resolves a single `db_file` from `config.json`.
- A `.db` not named in config (stale instance DB, manually placed file) is **not** excluded and gets `--delete`d or overwritten by the rsync sync.

## Proposed change

Make DB protection extension-based and defense-in-depth:

1. **`.gitignore`** — add a generic `*.db` (keep the existing WAL/SHM/lock globs). Confirm no currently-tracked file is a `.db` (none today: `git ls-files | grep '\.db'` is empty).
2. **`scripts/update.sh`** — add `--exclude='*.db'` `--exclude='*.db-wal'` `--exclude='*.db-shm'` `--exclude='*.db.lock'` to `rsync_excludes`, in addition to the existing config-resolved `db_excl` (keep that as a belt-and-suspenders for non-`.db`-suffixed paths).
3. Verify the `--ff-only` pull path: git won't clobber gitignored/untracked files on fast-forward, but document this so the guarantee is explicit rather than incidental.

## Acceptance

- A `.db` file at an arbitrary path is ignored by git and survives both `--rsync-from` and `pull --ff-only` updates unchanged.
- Existing named exclusions remain (no regression for `trading_bot.db` / configured `db_file`).

richkuo:
## Summary
State databases were protected by enumerated filename only, so a renamed or per-instance `.db` at an arbitrary path silently fell outside both safeguards — trackable by git and `--delete`'d/overwritten by `--rsync-from`. This makes DB protection extension-based and defense-in-depth.

## Changes
1. **`.gitignore`** — add generic `*.db` (keep existing named entries + WAL/SHM/lock globs). No currently-tracked `.db` (`git ls-files | grep '\.db'` is empty).
2. **`scripts/update_helpers.sh`** — new pure helper `update_db_rsync_excludes` emitting the static `*.db` / `*.db-wal` / `*.db-shm` / `*.db.lock` globs (unanchored → rsync matches at any depth).
3. **`scripts/update.sh`** — `run_rsync_from` appends those globs; the config-resolved `db_excl` stays as belt-and-suspenders for non-`.db`-suffixed paths. Inline comment documents that `pull --ff-only` never touches untracked/ignored files (guarantee made explicit, not incidental).
4. **`scripts/test_update_helpers.sh`** — regression test asserting the full glob set, unanchored form, and that `*.db` deliberately does **not** cover `.db-wal`/`.db.lock` (why all four globs are required).

## Verification
- `bash -n` clean; `test_update_helpers.sh` → `OK`; Go `TestUpdateShell*`/`TestUpdateHelpers*` → `ok`.
- `git check-ignore` matches a `.db` at nested arbitrary paths.
- Functional rsync test: stray `.db`/`.db-wal`/`.db.lock` at root **and** nested `instances/i2/per_instance.db` all survived `--delete`, while source code synced normally.

## Acceptance
- ✅ A `.db` at an arbitrary path is git-ignored and survives both `--rsync-from` and `pull --ff-only`.
- ✅ Existing named exclusions (`trading_bot.db` / configured `db_file`) remain — no regression.

Closes #1012

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 __pycache__/
 *.pyc
 *.pyo
+# State databases — never commit. Generic *.db plus the SQLite sidecar/lock
+# globs so a renamed or per-instance DB at any path stays protected (#1012).
+*.db
 trading_bot.db
 *.db-wal
 *.db-shm
--- a/scripts/test_update_helpers.sh
+++ b/scripts/test_update_helpers.sh
@@ -69,4 +69,19 @@ assert_eq "$(update_should_sweep_proc go-trader /opt/go-trader '')" \
 assert_eq "$(update_should_sweep_proc go-trader '' /opt/go-trader)" \
     "" "unreadable proc cwd -> spare"
 
+# --- #1012: extension-based DB rsync excludes --------------------------------
+db_globs=$(update_db_rsync_excludes)
+assert_eq "$db_globs" $'*.db\n*.db-wal\n*.db-shm\n*.db.lock' \
+    "db rsync excludes emit the full .db family, one glob per line"
+# Globs must be unanchored (no leading slash) so rsync matches at any depth.
+if printf '%s\n' "$db_globs" | grep -q '^/'; then
+    echo "FAIL: db rsync globs must be unanchored (no leading slash)" >&2
+    exit 1
+fi
+# Each suffix is distinct: *.db must NOT cover *.db-wal / *.db.lock (different
+# trailing chars), which is why all four globs are required.
+case "stale_instance.db" in *.db) ;; *) echo "FAIL: *.db should match stale_instance.db" >&2; exit 1;; esac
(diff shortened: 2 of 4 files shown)


richkuo:
@claude

PR #1018, opened 2026-06-15
[codex] Fix paper regime trailing SL snapshot

### LINKED ISSUE #1015: Paper trailing_stop_atr_regime SL never arms: per-cycle position snapshot omits regime label
## Bug

Paper-mode HL perps positions using `trailing_tp_ratchet_regime` + `trailing_stop_atr_regime` can remain open with `stop_loss_trigger_px=0.0`, leaving them without the synthetic paper trailing SL.

Observed on v0.78.1 paper instances:

```text
=== go-trader-paper-hl-eth ===
hl-stochrsi-eth-60|short|stop_loss_trigger_px=0.0|23.8 hrs open
hl-vwap-eth-60|short|stop_loss_trigger_px=0.0|21.8 hrs open
hl-bb-eth-60|short|stop_loss_trigger_px=0.0|17.8 hrs open
... (27 positions, most with trigger_px=0.0)
```

## Corrected current-code diagnosis

Validated against current `origin/main` at `c5d9c6a`.

The original `LoadConfig` / scalar `stop_loss_atr_mult` theory is not correct for current main:

- `LoadConfig` intentionally skips `default_stop_loss_atr_mult` when `trailing_stop_atr_regime` is configured. That block is a stop owner, not an omitted stop field. See `scheduler/config.go:950`.
- `trailing_stop_atr_regime` is mutually exclusive with `stop_loss_atr_mult` — validation rejects combining them. See `scheduler/regime_atr.go:832`. So the fix must NOT auto-apply scalar `stop_loss_atr_mult` on top of `trailing_stop_atr_regime`.
- `EffectiveStopLossPct` returns `0` for ATR/regime trailing owners at order-placement time (`scheduler/config.go:650`), but paper trailing SL does not depend on that helper.

## Root cause

The per-cycle perps paper snapshot omits the position regime label, so the paper trailing walker can never resolve the `trailing_stop_atr_regime` distance.

- The paper trailing walker is gated by `effectiveTrailingStopPct(sc, hlPosSnapshot)` and invoked with the same snapshot: `scheduler/main.go:1746` and `scheduler/main.go:1752`.
- `hlPosSnapshot` is built at `scheduler/main.go:1724` as `&Position{AvgCost, EntryATR, PostTPTrailingATRMult}` — it carries **no `Regime` and no `RegimeWindows`** (the Phase-1 RLock capture at `scheduler/main.go:1447` reads `pos.EntryATR` etc. but never the regime fields).
- For a `trailing_stop_atr_regime` owner, `effectiveTrailingStopPct` requires a non-empty position regime label: `scheduler/hyperliquid_trailing_stop.go:118`. With the snapshot's `Regime=""` / `RegimeWindows=nil`, `positionATRRegimeLabel` returns `""` (`scheduler/regime_multi_window.go:449`), so the helper returns `0` at `scheduler/hyperliquid_trailing_stop.go:119`.

Result: `effectiveTrailingStopPct` returns `0` on **every** cycle (including HOLD cycles), so the paper walker never arms a trigger and `stop_loss_trigger_px` stays `0.0`. Positions only gain a trailing SL once a TP tier stamps `PostTPTrailingATRMult` (`scheduler/hyperliquid_trailing_stop.go:86`), which takes the no-regime path — matching the observation that only positions that hit ~2x ATR profit ever close.

Note: the same snapshot also gates the live walker at `scheduler/main.go:1772`. Live still rests an SL at open via `armTrailingStopAtOpenNow` (`scheduler/hyperliquid_open_trailing.go:47`, which uses the real position with regime), so live is protected — but the live walker may not re-walk/ratchet the trigger pre-TP for regime owners. Worth confirming alongside the fix.

## Secondary gap (not the operative cause here)

The protection maintenance blocks are gated on `result.Signal == 0` while a position is open (`scheduler/main.go:1725`, `:1734`, `:1743`, `:1746`, `:1809`). A strategy that keeps emitting a same-side nonzero signal (later a no-op via `PerpsOrderSkipReason`) skips maintenance for that cycle.

This is a real gap, but it is NOT what produces the reported symptom: even on a pure HOLD cycle the paper walker returns `0` because of the snapshot defect above. The reported positions ran 17-24 cycles (60m timeframe, 17-24h open), so HOLD cycles almost certainly occurred and still did not arm. Fixing the signal gating alone would NOT arm a `trailing_stop_atr_regime` SL — the walker still returns `0` without the regime label.

## Expected behavior

Open HL paper positions with `EntryATR` + regime stamped and no `StopLossTriggerPx` should arm the synthetic paper trailing SL on the cycle after open. Missing `EntryATR` or missing position regime should produce a clear no-op warning/alert path rather than silently leaving `stop_loss_trigger_px=0.0`.

## Proposed fix direction

Preserve the current stop-owner invariant (do not apply scalar `stop_loss_atr_mult` when `trailing_stop_atr_regime` is present).

Primary:

- Capture the open position's `Regime` and `RegimeWindows` in the Phase-1 RLock (alongside `hlEntryATR` at `scheduler/main.go:1447`) and set them on `hlPosSnapshot` (`scheduler/main.go:1724`), so `effectiveTrailingStopPct` resolves the regime ATR distance for both the paper walker (`:1746`) and the live walker (`:1772`).

Secondary (optional hardening, only if needed after the primary fix):

- For an already-open position with a same-side no-op signal, treat it like HOLD for protection maintenance, or run the maintenance walkers independent of `result.Signal` while guarding real close/flip/open/scale-in paths.

Regression tests should cover:

1. Paper HL perps, open long/short, `trailing_stop_atr_regime` + `trailing_tp_ratchet_regime`, `EntryATR` + regime stamped, `StopLossTriggerPx=0`, HOLD (`Signal==0`) cycle -> paper trailing trigger is armed (this currently fails due to the snapshot defect).
2. (If the secondary fix is taken) same-side no-op signal on an open position -> maintenance still runs; real close/flip signals and allowed scale-ins do NOT run stale maintenance against the pre-update position.
3. Missing `EntryATR` or missing stamped position regime remains a no-op with an operator-visible warning/log.

## Environment

- Observed on go-trader v0.78.1 paper instances
- Config: `default_stop_loss_atr_mult: 1.0`, per-strategy `trailing_stop_atr_regime` with regime values set to `2.0`
- Close strategy: `trailing_tp_ratchet_regime` with `use_defaults: true`

---
Reference only: issue updated after validation by GPT-5 Codex.

richkuo:
## Summary

- Add a Hyperliquid protection position snapshot helper that carries regime labels, regime windows, post-TP trailing state, and the frozen risk anchor.
- Use that snapshot for the lock-free HL perps trailing/fixed protection walkers, including after trailing TP ratchet refresh.
- Add a regression proving the old incomplete snapshot cannot arm a regime trailing paper SL, while the fixed snapshot resolves the regime and arms at the frozen-anchor trigger.

## Root Cause

Paper-mode `trailing_stop_atr_regime` protection used `hlPosSnapshot`, but that snapshot only carried `AvgCost`, `EntryATR`, and `PostTPTrailingATRMult`. Without `Regime`/`RegimeWindows`, `effectiveTrailingStopPct` returned `0`, so the paper trailing walker never armed `stop_loss_trigger_px` on HOLD cycles.

Closes #1015

## Validation

- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'TestRunHyperliquidTrailingStopPaper|TestArmTrailingStopAtOpenNowRegime|TestEffectiveTrailingStopPct_ATRMult'`

LLM: GPT-5 Codex | high | Harness: go test ./...

### DIFF
--- a/scheduler/hyperliquid_stop_loss_test.go
+++ b/scheduler/hyperliquid_stop_loss_test.go
@@ -1798,6 +1798,40 @@ func TestRunHyperliquidTrailingStopPaper(t *testing.T) {
 	}
 }
 
+func TestRunHyperliquidTrailingStopPaper_RegimeSnapshotArms(t *testing.T) {
+	regimeBlock := &RegimeATRBlock{TrendRegime: map[string]RegimeATREntry{
+		"trending": {ATR: 2.0},
+		"ranging":  {ATR: 1.0},
+	}}
+	sc := StrategyConfig{
+		ID:                    "hl-regime-paper",
+		Platform:              "hyperliquid",
+		Type:                  "perps",
+		RegimeATRWindow:       "medium",
+		TrailingStopATRRegime: regimeBlock,
+	}
+	pos := &Position{
+		AvgCost:         2100,
+		EntryATR:        50,
+		RiskAnchorPrice: 2000,
+		Regime:          "ranging",
+		RegimeWindows:   map[string]string{"medium": "trending"},
+	}
+
+	incomplete := &Position{AvgCost: pos.AvgCost, EntryATR: pos.EntryATR}
+	_, gotTrig, gotBreach, gotPx := runHyperliquidTrailingStopPaper(sc, "long", incomplete, 2000, 0, 0)
+	if gotTrig != 0 || gotBreach || gotPx != 0 {
+		t.Fatalf("incomplete snapshot unexpectedly armed: trig=%v breach=%v px=%v", gotTrig, gotBreach, gotPx)
+	}
+
+	snapshot := hyperliquidProtectionPositionSnapshot(pos)
+	gotHW, gotTrig, gotBreach, gotPx := runHyperliquidTrailingStopPaper(sc, "long", snapshot, 2000, 0, 0)
+	if !approxEq(gotHW, 2000) || !approxEq(gotTrig, 1900) || gotBreach || gotPx != 0 {
+		t.Fatalf("regime snapshot paper trail = (hw=%v trig=%v breach=%v px=%v), want (2000, 1900, false, 0)",
+			gotHW, gotTrig, gotBreach, gotPx)
+	}
+}
+
 // #532: paper-mode trailing stop close must operate only on the strategy's
 // own virtual position. Two strategies on the same symbol with independent
 // StrategyState maps must remain isolated when one breaches.
--- a/scheduler/hyperliquid_trailing_stop.go
+++ b/scheduler/hyperliquid_trailing_stop.go
@@ -15,6 +15,29 @@ var (
 	hlTrailingUpdateLocks   = make(map[string]*sync.Mutex)
 )
 
+func hyperliquidProtectionPositionSnapshot(pos *Position) *Position {
+	if pos == nil {
+		return nil
+	}
+	snap := &Position{
+		AvgCost:                  pos.AvgCost,
+		EntryATR:                 pos.EntryATR,
+		RiskAnchorPrice:          pos.RiskAnchorPrice,
+		Regime:                   pos.Regime,
+		RegimeWindows:            cloneStringMap(pos.RegimeWindows),
+		RegimeAppliedLabel:       pos.RegimeAppliedLabel,
+		RegimePendingLabel:       pos.RegimePendingLabel,
+		RegimePendingCount:       pos.RegimePendingCount,
+		PostTPTrailingATRMult:    nil,
+		SLAdjustedTiersProcessed: pos.SLAdjustedTiersProcessed,
+	}
+	if pos.PostTPTrailingATRMult != nil {
+		v := *pos.PostTPTrailingATRMult
+		snap.PostTPTrailingATRMult = &v
+	}
+	return snap
+}
+
 func lockHyperliquidTrailingUpdate(symbol string) func() {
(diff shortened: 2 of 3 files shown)


richkuo:
@claude

richkuo:
Optional nits (cosmetic, none block merge):

1. **Drop the explicit `PostTPTrailingATRMult: nil` in the struct literal** — it's the zero value and is immediately overwritten by the conditional copy below it; keep only the `if pos.PostTPTrailingATRMult != nil` block.
2. **Add a doc comment to `hyperliquidProtectionPositionSnapshot`** — peer helpers in this file are documented; one line on why it exists (carries regime/anchor so the lock-free walker can resolve `trailing_stop_atr_regime`) would age well.
3. **Note the deliberately-broad field set** — `RegimeAppliedLabel`/`RegimePendingLabel`/`RegimePendingCount`/`SLAdjustedTiersProcessed` aren't read by the trailing/fixed-ATR walkers; harmless and future-proof, but a comment that this is an intentional full-protection snapshot clarifies intent.

---
Validated with LLM: Opus 4.8 | high | Harness: Claude Code

richkuo:
@claude sonnet rereview

Issue #1021, opened 2026-06-16
vol_momentum: static M1 complete; deprecate default listing or defer to regime gate


richkuo:
Part of #977 (Methodology M1) / #978. Original source: M5 fee-audit table (`docs/research/fee-audit-m5.md`, PR #1003), not the original audit #956.

## Status

The static M1 work for `vol_momentum` is complete. PR #1024 added the research note at `docs/research/vol_momentum_1021.md` with:

1. Focused M5 long-leg and short-leg fee-audit reruns.
2. Default long/short M1 protocol + held-out results.
3. OOS plateau sweeps over `entry_threshold`, `exit_threshold`, `eff_entry`, and `eff_exit`.
4. Held-out validation of the OOS-pass static-threshold candidates.

This issue is no longer asking for another short-leg screen or another static threshold sweep.

## Baseline That Triggered This Issue

The committed PR #1003 M5 row remains the validation target for the original concern: futures `vol_momentum`, long leg only, 146 trades, 24.9/yr, **-0.27% gross** / **-3.81% net**, verdict `unscreened_short`.

That row was the closest-to-breakeven traded `unscreened_short` entry in the committed screen, and `vol_momentum` already ships a bidirectional futures path via `allow_short=True`.

## Static M1 Findings

From `docs/research/vol_momentum_1021.md`:

| direction | IS | OOS | 2023 | 2024 | 2025H1 | held-out |
|---|---|---|---|---|---|---|
| long | PASS | **FAIL** | FAIL | FAIL | FAIL | 0/3 |
| short | PASS | **PASS** | FAIL | FAIL | PASS | 1/3 |

Interpretation:

- The long leg fails protocol OOS on current cache.
- The short leg is fee-healthy and clears the judged OOS window, but it is a regime-dependent bear-window edge.
- The short leg fails the 2023/2024 bull held-outs and can liquidate in strong bull regimes.
- Static threshold tuning did not produce a deployable plateau: the long OOS-pass pocket fails every held-out window, and the stricter short candidate also fails held-out stress.

## Decision

Do not change `vol_momentum` registry defaults from static threshold tuning. Do not iterate further on static threshold sweeps under this issue.

The static M1 verdict is: move `vol_momentum` to the deprecation/default-hidden path unless #1027 proves a regime-aware policy that survives the held-out bull windows.

## Remaining Action

Choose one of these paths:

[...]

PR #1003, opened 2026-06-12
feat(backtest): #999 M5 fee-aware selectivity audit — registry-wide trade-count × fee-drag screen

richkuo:
## What

Adds `backtest/fee_audit.py`, the M5 fee-aware selectivity screen from the #977 umbrella. For every registered strategy it runs each (dataset × window) leg **twice** on the `eval_windows.py` harness — net (binanceus platform fee + 5 bps slippage) and gross (`commission_pct=0.0, slippage_pct=0.0`) — to isolate fee drag, then applies the salvage test:

- `deprecate` — gross ≤ 0; no selectivity filter can save a strategy with no positive zero-fee edge.
- `graduate_m1` — gross > 0, net ≤ 0; a real edge exists under the churn, so the strategy graduates to an M1 application with "raise selectivity" as the mechanism.
- `healthy` — net > 0; edge already survives fees.
- `no_trades` — never fired on the audit slices.

## How

- `eval_windows.run_leg` gains keyword-only `commission_pct`/`slippage_pct` passthroughs (default `None` keeps the M1 harness byte-identical) plus an additive `span_days` leg key for trade-count annualization. Forwarding `commission_pct=None` resolves to the platform fee exactly as before; gross runs pass `0.0`.
- All aggregation/ranking/rendering logic is pure (`salvage_verdict`, `aggregate_strategy`, `rank_rows`, `trades_per_year`) and unit-tested without data access, mirroring `exit_diagnostics.py`.
- One bad strategy never aborts the registry-wide run: per-leg errors are captured and reported, not raised. `hold` (always flat) is skipped; futures-only names are appended for `--registry both`.

## Result

Report committed at `docs/research/fee-audit-m5.md` (ranked by fee drag). The bottom-tier churners reproduce the #956 audit — `vwap_reversion` 1,467 trades, `parabolic_sar` 1,301, `macd` 1,231, all `deprecate`. Four strategies graduate to M1 (positive gross, dead net): `regime_adaptive`, `tema_cross`, `breakout`, `regime_adaptive_htf`. The M5-tagged candidates `tema_cross_bd`/`triple_ema_bidir`/`sma_crossover`/`atr_breakout` all `deprecate` (gross negative).

## Verification

- `uv run --no-sync python -m pytest backtest/tests/test_fee_audit.py` — 18 pass (verdict matrix, annualization guards, error exclusion, ranking, markdown, gross-zeroes-friction e2e).
[...]

PR #1024, opened 2026-06-16
Document vol_momentum M1 findings and deprecate default listing

richkuo:
## Summary
- Add a research note for #1021 covering long and short leg fee audits, M1 protocol results, and OOS plateau sweeps for `vol_momentum`.
- Record that the long leg fails held-out stress, the short leg is regime-dependent, and the strategy should move to the default deprecation list.
- Capture reproduction commands and the validation target for the committed PR #1003 M5 row.

## Testing
- Not run (not requested)


richkuo:
@claude

PR #1026, opened 2026-06-16
feat(#1025): backtester parity for regime_directional_policy

### LINKED ISSUE #1025: Backtester parity for regime_directional_policy (per-cycle regime→direction resolver)
## Problem

The bar-level backtester **loud-rejects** `regime_directional_policy` at `backtest/run_backtest.py:270` (#779/#822). The live policy resolves inside `runHyperliquidCheck` after the check returns and `result.Regime` is known — it mutates the loop-local strategy config (`scheduler/main.go:2949`, `scheduler/regime_directional_policy.go:308`) then applies signal inversion. The replay engine has no equivalent per-cycle hook, so the reject is deliberate: without it the config would silently backtest the *static* `direction`/`invert_signal` and diverge from live.

## Why it matters

Any strategy using a live regime→direction policy is unvalidatable before risking real money — you can only reason about it live. This blocks future **bidirectional** regime-directional perps work (encode "short in trending_down, long otherwise" once, then prove it out of sample). It is not motivated by any single strategy today (see the #1021 note below).

## Proposed approach

Add a per-cycle regime→direction/invert resolver hook into `Backtester.run` that mirrors `regime_directional_policy.go`:

1. When flat, resolve direction/invert from the bar's regime, respecting the look-ahead invariant — gate reads the N-1 regime, same as existing `allowed_regimes` gating.
2. When a position is open, inherit the policy it was opened under (`pos.Regime`); block new entries opposite the held side; always run close evaluators.
3. Replace the loud reject in `load_strategy_config` with the backtested path once parity is proven.

## Critical constraint (do not skip)

Parity is the entire reason the reject exists. A half-correct simulator reintroduces exactly the silent divergence the reject prevents — **worse than no backtest**. So the deliverable must include:

1. `backtest/parity_diff.py` coverage proving the resolver matches the live decision layer cycle-by-cycle.
2. A look-ahead regression in `backtest/tests/test_backtester_lookahead.py` (resolver must read closed-bar regime, never the in-flight bar).
3. `--list-json` byte-identical (no registry change).

## Scope and related work

Scope this issue to `regime_directional_policy` only.

- `regime_window_divergence` (rejected at `backtest/run_backtest.py:285`, #907/#942) is the **same entry-direction resolver gap** — it mutates direction per cycle off a short/medium regime split. Land it as a related follow-up reusing this hook.
- `tiered_tp_atr_live_regime_dynamic` (rejected at `backtest/run_backtest.py:318`, #843) is a **different root cause** — dynamic TP/SL regime hysteresis and on-chain protection-sync (`scheduler/regime_dynamic_tp_sl.go:64`, `scheduler/hyperliquid_protection.go:570`), not entry direction. It needs a separate dynamic-close/protection parity effort; do not fold it into this resolver.

## Note: not a blocker for #1021

The `vol_momentum` decision does **not** need this. The long leg is dead, so the salvage is short-only-except-in-bull, already expressible with existing, already-modeled flags: `--direction short` (`backtest/run_backtest.py:781`, `backtest/backtester.py:1813`, regression `backtest/tests/test_backtester_short_leg.py:124`) + `allowed_regimes` excluding the bull label. This issue is a deliberate parity investment for future bidirectional regime-directional work, not a #1021 dependency.

---

richkuo:
## Summary
Replaces the config-load **rejection** of `regime_directional_policy` with a bar-level replay path, so backtests now mirror the HL-live per-regime direction/invert override instead of silently falling back to the static config.

- `load_strategy_config` requires `regime.enabled=true` for the policy and threads the regime params (`period`, `adx_threshold`, enabled) through to `Backtester`.
- `Backtester` normalizes the already-validated `trend_regime` shape (Go owns vocabulary validation) and resolves the effective `direction`/`invert_signal` per bar, frozen to the open position's regime while a position is held.
- `direction='both'` on the plain-signal path is rejected (needs a close evaluator), matching the live geometry.

## Parity boundary (unchanged)
`regime_window_divergence` (#907) and `tiered_tp_atr_live_regime_dynamic` (#843) remain rejected — no bar-level parity path. The `regime_divergence.go` comment is updated to point at #1025 as the reason the divergence override is still the outlier.

## Docs / tests
- Updated `parity_diff.py`, `AUDIT.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `SKILL.md`.
- Added direction/invert and parity-diff regression tests.

Closes #1025

---
Created with LLM: gpt5.5 | xhigh | Harness: Codex


richkuo:
@claude

richkuo:
@claude rereview

PR #1028, opened 2026-06-16
refactor(ci): extract @claude review contract to prompt file + fix verdict logic

richkuo:
## Summary
Refactors the `@claude review` output contract and fixes logic flaws in how the verdict is decided.

The 600-word `--append-system-prompt` string was inlined in `claude.yml` (fragile to YAML quoting, duplicated in two CLAUDE.md files). It now lives in a single source of truth: `.github/prompts/pr-review-format.md`. A new **Compose review prompt** workflow step flattens it into `claude_args` and fails the run if the file contains a double quote (which would break arg quoting).

## Contract fixes
1. **Verdict keys off blocking sections only.** Previously `Needs Updates` fired whenever *any* finding survived — even though `Recommended Optional`, `Create Follow-up Issue`, and `Requires Human Review` are non-blocking. A mergeable PR with one optional nit read as "do not merge". Now: `Needs Updates` iff ≥1 item under `Needs Fixing` or `Requires Human Review`; otherwise `LGTM` (may carry non-blocking sections).
2. **Materiality filter → trivia only.** The old "drop anything that wouldn't change the merge decision" contradicted the existence of a `Recommended Optional` section (optional items by definition don't change the merge decision). Substantive non-blocking findings are now routed, not dropped.
3. **LGTM precondition.** `LGTM` is a bare-token auto-merge trigger; an early-exit/partial run could emit it with no proof of coverage. Now `LGTM` requires inspecting the full diff and checking CI, else it downgrades to a `Requires Human Review` item.
4. **Safety carve-out.** Findings touching money, data integrity, security, or an auto-protective mechanism (kill switch, circuit breaker, stop-loss, reconciliation, fill/position accounting) are never dropped by the materiality filter; uncertain ones go to `Requires Human Review`.

## Notes
- The upstream `code-review` plugin's `<80` confidence gate is unchanged (not cleanly editable without forking). The safety carve-out mitigates at the comment stage but cannot recover a finding the plugin dropped earlier — full closure would require forking the plugin.
- Project `CLAUDE.md` updated to point at the prompt file and reflect the new semantics.

## Verification
- `claude.yml` parses (ruby/psych).
- Compose step simulated end-to-end: valid single-line `claude_args`, balanced quotes, model id (`claude-opus-4-8[1m]`) intact, double-quote guard fires correctly.

---


richkuo:
@claude

PR #1033, opened 2026-06-17
ci: drop ccusage token/cost line from Claude comment footer

richkuo:
## Summary
Removes the `ccusage` block from the "Append LLM footer" step in `claude.yml`. The footer now strips any stale footer Claude wrote and appends `LLM: <model> | <effort> | Harness: <harness>` directly — no `npx ccusage@latest` fetch on each run.

## Why
The `ccusage` call only enriched the footer with a best-effort token/cost line. It was wrapped to fail safe (error → `[]` → empty `COST` → footer appended without the line), so it had no functional role. Dropping it removes a per-run npm fetch; the authoritative model/effort/harness attribution is unchanged.

## Verification
- `grep` confirms no remaining `ccusage` / `COST` / `USAGE` / `TODAY` references.
- Workflow YAML parses clean.

---

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -251,16 +251,7 @@ jobs:
           # the real --effort value, so it guesses). The workflow is authoritative.
           BODY=$(printf '%s' "$BODY" | python3 .github/scripts/strip_llm_footer.py)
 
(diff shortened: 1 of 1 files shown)


richkuo:
@claude sonnet

richkuo:
@claude

PR #1036, opened 2026-06-17
[codex] Fix Claude workflow error detection

richkuo:
## Summary

- Fail the Claude workflow when the Claude Code execution output reports `is_error: true`, including API overload failures that the action wrapper otherwise marks successful.
- Record the PR head SHA before Claude runs so the follow-up `CLAUDE.md` revision step only runs when Claude actually changes the branch.

## Root Cause

The Claude Code action can complete its GitHub step successfully even when the execution result contains an API error such as `529 Overloaded`. Separately, PR `issue_comment` runs initially check out `main`, while the action later moves to the PR branch, so comparing the original `main` SHA to the post-action PR SHA falsely detected commits.

## Validation

- `jq` smoke checks for JSONL and JSON-array Claude execution output shapes
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/claude.yml"); puts "yaml ok"'`
- `git diff --check`
- `actionlint .github/workflows/claude.yml`

LLM: GPT-5 | high | Harness: actionlint + jq smoke

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -133,7 +133,20 @@ jobs:
       - name: Record HEAD sha before Claude runs
         if: steps.verify_invocation.outputs.invoked == 'true'
         id: pre_claude
-        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+        env:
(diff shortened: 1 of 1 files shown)


richkuo:
@claude

PR #1047, opened 2026-06-18
Keep trailing stop ratcheting under a latched circuit breaker (#1046)

### LINKED ISSUE #1046: Bug: circuit breaker blocks trailing stop loss ratcheting for existing positions
### Summary

When a per-strategy circuit breaker fires, the entire strategy loop iteration is skipped — including the trailing stop loss update logic. For a position that **remains open** while the CB is latched, the trailing SL stops ratcheting for the duration of the latch (up to 24h), leaving a stale stop-loss order on the exchange that doesn't lock in favorable price movement.

**Scope — this only bites positions that survive the latch.** On a single-strategy coin the CB force-closes the position when it fires (`forceCloseAllPositions`, `risk.go:1438-1439`), so there is nothing left to ratchet. The position only survives on **shared coins**: `shouldForceCloseAllPositionsOnCircuitBreaker` returns `!hyperliquidCircuitBreakerHasSharedCoin` (`risk.go:949-950`), so force-close is skipped and the virtual position stays in state. The on-chain drain is reduce-only and proportional on shared coins (`main.go:1240-1244`), so a residual can remain open too. That surviving residual is what goes unprotected.

### Expected behavior

Circuit breaker should only prevent **new positions from opening**. Existing position management — trailing stop ratcheting, TP updates, close-signal evaluation — should continue operating normally on any position that remains open while CB is active.

### Actual behavior

`CheckRisk` returns `(false, RiskReasonCircuitBreakerActive)` while CB is latched (`risk.go:1374-1376`). The first-fire cycle returns `RiskReasonMaxDrawdownExceeded` (`risk.go:1441-1442`); every subsequent cycle logs `circuit breaker active`. The per-strategy loop in `main.go` then `continue`s (`main.go:1546`, immediately after the `CheckRisk` call at `main.go:1535`), skipping the entire loop body — no check script, no trailing stop evaluation, no TP ratchet.

Note the CB is *designed to close* the position, not keep it open: on fire it both enqueues a reduce-only close (`setHyperliquidCircuitBreakerPending`, `risk.go:1433`, drained by `runPendingHyperliquidCircuitCloses` at `main.go:1244`) and force-closes virtual state on non-shared coins. The bug is that on shared coins, where neither path flattens the position, nothing then manages the surviving position's stop-loss.

### Evidence

A live perps short on a shared coin (multiple live HL strategies on the same symbol):

1. CB fired when perps margin drawdown exceeded `max_drawdown_pct` (price moved unfavorably above entry)
2. CB latched for 24h (`risk.go:1432`); force-close was skipped because the coin is shared, so the position stayed open
3. For ~20 hours while CB was active:
   - `stop_loss_high_water_px` (`Position.StopLossHighWaterPx`) stayed stale (price had moved significantly in favor)
   - `stop_loss_trigger_px` (`Position.StopLossTriggerPx`) never ratcheted down
   - Exchange stop-loss order sat untouched
   - Log every cycle: `[WARN] Risk block: circuit breaker active`
4. After manually clearing CB and restarting, the very first cycle immediately ratcheted the trailing SL to the current mark and placed a new trigger oid

All trailing-SL / ratchet / mark-fetch sites sit inside the loop body, after the `continue`:
- `runHyperliquidTrailingStopUpdate` — perps `main.go:1780`, manual `main.go:2092`
- `applyTrailingTPRatchet` (`trailing_tp_ratchet_regime`) — perps `main.go:1725`, manual `main.go:2077`
- `runHyperliquidCheck` (close-signal evaluation + fresh mark) — `main.go:1704`

### Impact

- **Stop-loss orders go stale** — they don't lock in favorable price movement, so a rally back to the original (stale) trigger causes a larger realized loss than necessary
- **TP ratchets also freeze** — `trailing_tp_ratchet_regime` doesn't run while CB is active
- Extended periods of unprotected favorable movement on live shared-coin positions

### Proposed fix

Move the trailing stop / TP ratchet / position-management logic **outside** the `CheckRisk` gate in the strategy loop, scoped to positions that remain open. Specifically:

1. `CheckRisk` must continue to block **new trade signals** (Signal != 0 → no open/add) — this is the safety guarantee that must not regress.
2. The trailing SL update (`runHyperliquidTrailingStopUpdate`) and TP ratchet evaluation should run **regardless of CB state** as long as there's an open position.
3. The check script (`runHyperliquidCheck`) must still run when CB is active — both to evaluate close signals and because the trailing SL update needs a **fresh mark** from it. Only new-entry signals should be suppressed.

Reconcile with the existing CB close machinery: the management path must not race the reduce-only drain (`runPendingHyperliquidCircuitCloses`) or re-arm an SL on a position the drain is about to flatten. The case that genuinely needs ongoing management is the shared-coin residual the CB deliberately leaves open; define the miss/failure policy for it (e.g. skip the ratchet when a pending close is in flight for the coin). Scope note: trailing-SL ratcheting is HL perps + manual only — non-HL platforms have no equivalent path.

Alternatively, split the risk check into two outcomes:
- `canOpen: false` when CB is active (blocks new positions)
- `canManage: true` always (allows existing position management)

### Environment

- go-trader v0.80.0
- Perps margin drawdown signal (`max_drawdown_pct` configured per strategy)

---
Updated with LLM: Opus 4.8 | xhigh | Harness: validate-issue

richkuo:
Closes #1046

## Problem

A latched per-strategy circuit breaker returns `(false, RiskReasonCircuitBreakerActive)` from `CheckRisk` (`risk.go:1374-1376`), and the per-strategy loop `continue`s past the entire body (`main.go:1546`). That skips the `Signal==0` position-management blocks — the trailing-SL update, TP ratchet, and protection sync — so an open position's stop-loss stops ratcheting for the whole latch window (up to 24h), leaving a stale trigger that fails to lock in favorable movement.

The gap bites **shared-coin HL perps** positions specifically: on those, the CB skips force-close (`shouldForceCloseAllPositionsOnCircuitBreaker` returns `!hyperliquidCircuitBreakerHasSharedCoin`, `risk.go:949-950`), so the position survives the latch instead of being flattened — and then sits unprotected.

## Fix

When a **latched** CB blocks an **HL perps** strategy that still has an **open position**, fall through in manage-only mode instead of skipping, then force `result.Signal = 0` right after the HL check returns.

Forcing the hold signal provably suppresses every position-changing path:
- scale-in and live execute gate on `result.Signal != 0` → skipped
- `executePerpsSignalWithLeverage` returns `(0, nil)` immediately on `signal == 0` (`portfolio.go:1116-1118`) → paper/deferred open is a no-op

…while the existing `Signal==0` blocks (trailing-SL update, TP ratchet, protection sync) run normally.

**Safety:** the change only *adds* protective management under a latched CB; it removes nothing that previously ran (the whole loop was skipped before), and `Signal=0` can never open a position. So it is strictly safer than current behavior and cannot regress entry-blocking.

## Scope decisions

- **Latched CB only.** The first-fire cycle (`max drawdown exceeded` / `5 consecutive losses`) is excluded — that is the cycle that force-closes / enqueues the reduce-only drain, so position state is mid-transition. Management resumes the next cycle, i.e. ~the entire latch window.
- **Live shared-coin HL perps only.** Manual is exempt from `CheckRisk` (`risk.go:1365-1367`); futures/OKX have no in-loop HL trailing-SL walker. Paper does **not** benefit: the shared-coin force-close skip is live-only (`hyperliquidCircuitBreakerHasSharedCoin` gated on `hyperliquidIsLive`, `risk.go:939`), so paper positions are force-closed on the first-fire cycle and none survive the latch to manage. Paper shared-coin coverage would require a separate change.
- The notional-cap skip is bypassed under manage-only (the position is never grown anyway).

## Changes

- `risk.go` — new pure helper `circuitBreakerPermitsManagement(reason, platform, type, posQty)`.
- `main.go` — manage-only fall-through + `Signal=0` force in the HL check branch.
- `risk_test.go` — 10-case table test for the gate.

## Testing

- `go build .` clean, `gofmt` clean.
- Full scheduler Go test suite passes (`go test .`).
- New `TestCircuitBreakerPermitsManagement` covers: latched-CB open perps → manage; no position / negative qty → skip; first-fire reasons (plain + formatted) → skip; OKX/futures/spot → skip; empty reason → skip.

---
Updated with LLM: Opus 4.8 | xhigh | Harness: Claude Code

### DIFF
--- a/scheduler/main.go
+++ b/scheduler/main.go
@@ -1538,16 +1538,31 @@ func main() {
 						cbSnapshot = snapshotPerStrategyCircuitBreaker(stratState, prices)
 					}
 					mu.Unlock()
+					// #1046: a latched per-strategy circuit breaker on an HL perps
+					// strategy with an open position falls through in manage-only
+					// mode rather than skipping — the dispatch below forces the
+					// signal to hold (0), which suppresses every entry/add/flip/
+					// close path while still running the Signal==0 trailing-SL/TP
+					// management so a stranded (e.g. shared-coin) position keeps
+					// ratcheting its stop-loss. All other blocks skip as before.
+					cbManageOnly := false
 					if !allowed {
 						notifyPerStrategyCircuitBreakerWithSnapshot(sc, cbSnapshot, reason, pv, totalPV, stateDB, notifier, killSwitchFired)
 						logger.Warn("Risk block: %s (portfolio=$%.2f)", reason, pv)
-						logger.Close()
-						lastRun[sc.ID] = time.Now()
-						continue
+						if circuitBreakerPermitsManagement(reason, sc.Platform, sc.Type, hlPosQty) {
+							cbManageOnly = true
+							logger.Info("Circuit breaker latched — suppressing new entries but continuing trailing-SL/TP management for open position (#1046)")
+						} else {
+							logger.Close()
+							lastRun[sc.ID] = time.Now()
+							continue
+						}
 					}
 
-					// #42: Notional cap blocks new trades for this strategy.
-					if notionalBlocked {
+					// #42: Notional cap blocks new trades for this strategy. Under
+					// manage-only the position is never grown anyway, so let the
+					// trailing-SL/TP management run instead of skipping.
+					if !cbManageOnly && notionalBlocked {
 						logger.Warn("Notional cap exceeded — skipping new trades")
 						logger.Close()
 						lastRun[sc.ID] = time.Now()
@@ -1703,6 +1718,15 @@ func main() {
 							}
 						} else if result, signalStr, price, ok := runHyperliquidCheck(&sc, prices, hlPosCtx, cfg.Regime, notifier, logger); ok {
 							prices[result.Symbol] = price
+							// #1046: circuit breaker latched — force hold so no entry/
+							// add/flip/close executes (every execution path below gates
+							// on Signal != 0, and executePerpsSignalWithLeverage returns
+							// early on signal==0). The Signal==0 trailing-SL/TP-ratchet
+							// management blocks below still run, keeping the open
+							// position's stop-loss ratcheting through the latch window.
+							if cbManageOnly {
+								result.Signal = 0
+							}
 							// #879: single-source regime — read the global store for this
 							// strategy's signature instead of the check output, and point
 							// result.Regime at it so stamp-at-open inside execute* shares it.
--- a/scheduler/risk.go
+++ b/scheduler/risk.go
@@ -1354,6 +1354,28 @@ const (
 	RiskReasonConsecutiveLosses    = "5 consecutive losses"
 )
 
+// circuitBreakerPermitsManagement reports whether a CheckRisk block should still
+// run existing-position management (trailing SL ratchet, TP ratchet, protection
+// sync) for an open position instead of skipping the strategy outright. A
+// per-strategy circuit breaker exists to block NEW entries; it must not freeze
+// the stop-loss on a position that is already open — e.g. a shared-coin residual
+// the CB cannot force-close (shouldForceCloseAllPositionsOnCircuitBreaker is
+// false when the coin is shared), which then sits with a stale trailing SL for
+// the whole latch window and fails to lock in favorable movement (#1046).
+//
+// Scoped to the latched reason and to HL perps: only that path runs the
+// trailing-SL/TP-ratchet walker. Manual strategies are exempt from CheckRisk
+// entirely (returns allowed early), and other platforms/types have no equivalent
+// in-loop SL ratchet, so they keep the plain skip. The first-fire cycle (reason
+// "max drawdown exceeded" / "5 consecutive losses") is deliberately excluded:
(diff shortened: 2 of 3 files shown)


richkuo:
@claude review

richkuo:
Addressed the Recommended Optional finding (doc-only, no code change).

The scope claim that paper HL perps benefit was wrong. The shared-coin force-close skip is live-only — `hyperliquidCircuitBreakerHasSharedCoin` is gated on `hyperliquidIsLive` (`risk.go:939`), so for paper `shouldForceCloseAllPositionsOnCircuitBreaker` returns true and the position is flattened on the first-fire cycle. By the latched cycle `posQty` is 0 and `circuitBreakerPermitsManagement` returns false, so no paper position ever reaches manage-only.

Updated the PR description: scope is now **live shared-coin HL perps only**, with a note that paper shared-coin coverage would require a separate change. Code is unchanged — it was already correct and safe.

---
Updated with LLM: Opus 4.8 | xhigh | Harness: Claude Code

PR #1049, opened 2026-06-18
feat(#1048): per-strategy circuit-breaker enable/disable flag

### LINKED ISSUE #1048: [C55] Per-strategy circuit-breaker enable/disable flag
**Complexity: 55/100** — scope: medium (config field, `CheckRisk` gate, validation, init/wizard, hot-reload, tests); risk: high (disables an auto-protective mechanism on live money); uncertainty: medium (the `CheckRisk` gate placement is behavior-determining — see Proposal).

## Problem
The circuit breaker (CB) in `CheckRisk` (`scheduler/risk.go:1384`) applies to every non-`manual` strategy and makes **no paper/live distinction**. Its two triggers are:

1. Drawdown > the strategy's `max_drawdown_pct` (fire arm `risk.go:1452`).
2. 5 consecutive losses — **hardcoded** threshold (fire arm `risk.go:1469`), no config knob.

There is no way to turn CB off for a given strategy. The drawdown arm can be neutered by setting `max_drawdown_pct` very high, but the 5-consecutive-losses arm cannot be disabled at all.

(Line numbers reflect current `main` after #1046, which inserted `circuitBreakerPermitsManagement` immediately above `CheckRisk`; the manual exemption is now `risk.go:1387` and the latch check `risk.go:1396`.)

## Requirement
**Every strategy gets an explicit on/off switch for its circuit breaker, regardless of live or paper. It defaults to ON.** No live/paper guard — the operator owns the choice for any strategy. **The flag is hot-reloadable** (SIGHUP), including while a position is open.

## Proposal
Add an opt-out flag on `StrategyConfig` using the `*bool` nil-default pattern (nil/missing → enabled, explicit `false` → disabled). The closest precedent is the root-`Config` field `NotifyTPSLFills *bool` (`config.go:150`), but it lives on the **root config** — `StrategyConfig` has no `*bool` field today (its booleans are plain `bool`: `htf_filter`, `invert_signal`, `allow_scale_in`), so this would be the **first per-strategy `*bool`**. The pattern transfers cleanly and keeps omitted configs safe (CB stays on by default).

- Field: `CircuitBreaker *bool` (json `circuit_breaker`). Nil/true → CB active (current behavior, the default); explicit false → CB disabled for that strategy. **Applies uniformly to live and paper — no platform/live gating.**
- Gate in `CheckRisk`: when explicitly disabled, suppress **both** fire arms — guard the two `return false` fire points (drawdown arm `risk.go:1452`, consecutive-loss arm `risk.go:1469`) with the disabled check and fall through to the existing `return true, ""`. **Do NOT early-return at the top of the function like the `sc.Type == "manual"` exemption (`risk.go:1387`).** That placement is wrong here for two reasons: (1) it runs *before* the peak update and drawdown computation, so `CurrentDrawdownPct`/peak would never update — contradicting the display requirement below; (2) it runs *before* the latch check (`risk.go:1396`), so a strategy already latched would be allowed to trade again immediately on disable, bypassing the drain of an in-flight circuit close. Keep `rolloverDailyPnL` and the latch check intact; gate only the two fire points, so `CurrentDrawdownPct`/peak still compute for display and any pending/latched close still winds down.
- Unknown-key guard: **automatic** — `knownStrategyConfigKeys()` (`config_unknown_keys.go:16`) derives accepted keys from `StrategyConfig` json tags via reflection, so adding the field registers it. No manual registration step.

## Hot reload
The flag **must be hot-reloadable via SIGHUP, including while a position is open** — do NOT add it to `validateHotReloadCompatible`/`validateHotReloadStateCompatible` block lists (`config_reload.go`); a new field defaults to allowed. Because the gate sits *below* the latch check (`risk.go:1396`), the transitions are:
- **on→off while open:** CB stops *new* fires from the next cycle. An already-latched CB stays latched until its window expires, and any already-pending circuit close still drains (no new fire) — the latch check and the out-of-band pending-close drain are untouched by the flag.
- **off→on while open:** CB resumes evaluating on the next cycle and may fire immediately if the strategy is already past a threshold — this is intended (operator re-arming protection). Confirm a re-enable does not double-fire or mis-handle stale `ConsecutiveLosses`.

## Considerations
- **Config version:** an optional nil-default field needs no migration or version bump (precedent: #661 `NotifyTPSLFills`, #866). Stays at `CurrentConfigVersion=15`.
- **#1046 interaction:** `circuitBreakerPermitsManagement` (#1046, `risk.go`) keys off `reason == RiskReasonCircuitBreakerActive` to keep the trailing-SL/TP ratchet running under a *latched* CB. Since the flag gates only the fire points (not the latch), this path is unaffected — add a regression test that disabling the flag does not change behavior for an already-latched HL perps position.
- **Operator visibility:** when CB is disabled for a strategy, surface it (status UI / startup log) so it's not silently unprotected — this matters more given disable is allowed on live money.
- **Init/wizard:** expose the toggle in `InitOptions`/wizard/`generateConfig` so new configs can set it; absence keeps the safe default. (Note: `NotifyTPSLFills` is not wired into `init.go`, so there's no existing `*bool` wizard precedent to copy — this is net-new wizard wiring, still straightforward.)

## Tests
- `CheckRisk` with `circuit_breaker:false` does **not** fire on drawdown > `max_drawdown_pct` **nor** on 5 consecutive losses — on both a live and a paper strategy.
- nil and explicit true still fire on both arms (regression).
- With the flag disabled, `CurrentDrawdownPct`/peak are still computed (display not suppressed), and a pre-existing latched CB / pending circuit close is unaffected.
- Hot-reload: `circuit_breaker` toggles on↔off while a position is open without being rejected by the hot-reload validators; off→on with the strategy already past a threshold fires on the next cycle; on→off suppresses new firing.
- #1046 regression: disabling the flag on an already-latched HL perps strategy leaves `circuitBreakerPermitsManagement` behavior unchanged.

---
LLM: Opus 4.8 | xhigh | Harness: validate-issue

richkuo:
## Summary
Adds a per-strategy opt-out for the circuit breaker (CB). Today CB fires for every non-`manual` strategy with no off-switch: the drawdown arm can be neutered via a high `max_drawdown_pct`, but the hardcoded 5-consecutive-losses arm cannot be disabled at all. This adds `circuit_breaker *bool` on `StrategyConfig` (json `circuit_breaker`) — nil/missing → enabled (the safe default), explicit `false` → disabled. Applies uniformly to live and paper (CheckRisk has no platform/live gating). Closes #1048.

## Behavior
- **Gate** (`risk.go`): when disabled, both firing arms are suppressed (`r.CurrentDrawdownPct > MaxDrawdownPct` AND `ConsecutiveLosses >= 5`). The gate sits **below** the latch check and the drawdown computation: an already-latched CB and any pending circuit close still drain, and `CurrentDrawdownPct`/peak still update for display. `manual` stays exempt via the existing early return.
- **Hot reload** (`config_reload.go`): the toggle is hot-reloadable via SIGHUP **including while a position is open**. Mechanically this required adding `CircuitBreaker` to the `strategyRestartShape` allow-mask (otherwise the immutable-fields `DeepEqual` would flag a toggle as restart-required) plus an apply block so the new value is actually written; no state-compat guard. Disabling suppresses **new** fires only; re-enabling resumes evaluation next cycle and may fire immediately if already past a threshold.
- **Visibility** (`inspect_cmd.go`): `cb=off` in the `#704` startup summary line, plus text and JSON `inspect` surfaces, so a strategy trading without the auto-protective halt isn't silently unprotected (matters since disable is allowed on live money).
- **Init** (`init.go`): `DisableCircuitBreaker` on `InitOptions` is honored by `generateConfig` for non-manual strategies; the interactive wizard keeps the safe default (no "disable safety" prompt).
- Unknown-key guard is reflection-derived, so the field auto-registers. No config-version bump (optional nil-default field, precedent `NotifyTPSLFills`).

## Tests
`CheckRisk` both arms × {live, paper} with `circuit_breaker:false` (no fire) vs nil/true (fire, regression); display drawdown still computed when disabled; an already-latched CB still blocks when disabled; hot-reload toggle on↔off while a position is open is accepted and applied; restart-shape neutrality; `CircuitBreakerEnabled()` default; unknown-key acceptance; `generateConfig` stamping (non-manual only); startup-summary `cb=off`. Full Go suite green; `init --json` + `inspect` smoke confirms end-to-end.

Closes #1048

---
https://claude.ai/code/session_01MCPJBFkFgrx2keUm65Q72n

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -242,6 +242,18 @@ func (c *Config) NotifyTPSLFillsEnabled() bool {
 	return *c.NotifyTPSLFills
 }
 
+// CircuitBreakerEnabled reports whether the per-strategy circuit breaker is
+// active. Nil pointer (missing field) defaults to true so existing configs keep
+// the auto-protective behavior without an explicit opt-in; an explicit false
+// disables both firing arms in CheckRisk (drawdown and consecutive-loss). Safe
+// on a nil receiver (treated as enabled). (#1048)
+func (sc *StrategyConfig) CircuitBreakerEnabled() bool {
+	if sc == nil || sc.CircuitBreaker == nil {
+		return true
+	}
+	return *sc.CircuitBreaker
+}
+
 // ParseSummaryFrequency converts a summary_frequency value to a duration.
 // Returns -1 to mean "use legacy default", 0 to mean "every channel run", or a
 // positive duration when caller should post every duration. An unrecognized
@@ -347,6 +359,7 @@ type StrategyConfig struct {
 	CapitalPct              float64                  `json:"capital_pct,omitempty"`     // 0-1; dynamic capital = wallet_balance * capital_pct (overrides capital)
 	InitialCapital          float64                  `json:"initial_capital,omitempty"` // fixed starting balance for PnL display (never overwritten by capital_pct)
 	MaxDrawdownPct          float64                  `json:"max_drawdown_pct"`
+	CircuitBreaker          *bool                    `json:"circuit_breaker,omitempty"`            // #1048 — per-strategy circuit-breaker opt-out. Nil/missing → enabled (the safe default); explicit false disables BOTH firing arms in CheckRisk (drawdown > max_drawdown_pct AND 5 consecutive losses), uniformly for live and paper (no platform/live gating). Hot-reloadable via SIGHUP including while a position is open: disabling only suppresses NEW fires — an already-latched CB and any pending circuit close still drain. No effect on type=manual (exempt from CheckRisk). Read via CircuitBreakerEnabled(), never directly.
 	IntervalSeconds         int                      `json:"interval_seconds,omitempty"`           // per-strategy override (0 = use global)
 	HTFFilter               bool                     `json:"htf_filter,omitempty"`                 // higher-timeframe trend filter
 	InvertSignal            bool                     `json:"invert_signal,omitempty"`              // HL perps/manual only: flip BUY<->SELL on a non-zero signal before execution (HOLD/0 is never flipped). Lets inverse variants reuse the same open/close refs. Composes with Direction — invert runs in the Go layer before direction interprets the resulting sign (e.g. direction="short" + invert_signal=true opens short on raw-BUY triggers, distinct from plain direction="short" which opens on raw-SELL). Rejected outside HL perps/manual.
--- a/scheduler/config_reload.go
+++ b/scheduler/config_reload.go
@@ -59,6 +59,17 @@ func applyHotReloadConfig(cfg, next *Config, state *AppState, notifier *MultiNot
 				ss.RiskState.MaxDrawdownPct = ns.MaxDrawdownPct
 			}
 		}
+		// #1048: per-strategy circuit-breaker toggle is hot-reloadable always,
+		// including while a position is open (no state-compat guard). Disabling
+		// only suppresses NEW fires from the next cycle; an already-latched CB
+		// and any pending circuit close continue to drain. Re-enabling resumes
+		// evaluation next cycle and may fire immediately if already past a
+		// threshold — intended re-arming. The next CheckRisk reads sc.CircuitBreaker
+		// directly, so no state mutation is needed here.
+		if !boolPtrEqual(sc.CircuitBreaker, ns.CircuitBreaker) {
+			addChange("strategy[%s].circuit_breaker: %s -> %s", sc.ID, formatCircuitBreaker(sc.CircuitBreaker), formatCircuitBreaker(ns.CircuitBreaker))
+			sc.CircuitBreaker = ns.CircuitBreaker
+		}
 		if sc.CapitalPct == 0 && sc.Capital != ns.Capital {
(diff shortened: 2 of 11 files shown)


richkuo:
@claude review

richkuo:
Addressed the optional finding (ab234a2). When `circuit_breaker:false` and a halt threshold is crossed, `CheckRisk` now emits a one-shot WARNING via `logger.Warn` at that cycle — stating plainly that there is NO circuit breaker and that it is a warning only (nothing is closed; trading continues). Deduped to once per suppression episode (package-level `sync.Map`, mirroring the trailing-stop missing-EntryATR throttle), naming the breached arm(s). The throttle clears eagerly whenever the breaker is enabled (so the enabled+breached genuine fire still alerts via the normal path) and when the breach clears, so a later re-disable warns afresh. Tests cover warn-once/dedup, re-enable-clears + genuine-fire, re-disable-warns-again, and breach-clear reset.

Kept it at WARN severity (not CRITICAL) per the requirement that this is a warning, not a halt.

---
LLM: Opus 4.8 | xhigh | Harness: Claude Code

richkuo:
@claude sonnet rereview

richkuo:
@claude sonnet rereview

PR #1052, opened 2026-06-19
feat(#1050): manual-update-sl / manual-cancel-sl in-place stop-loss CLI

### LINKED ISSUE #1050: Add `manual` subcommand for atomic position interventions
## Problem

When operating go-trader interactively, several manual interventions are commonly needed on open positions: ratcheting a stop-loss, updating a take-profit tier, cancelling protective orders, opening/closing positions for a "manual" strategy that just tracks them.

The existing flat subcommands cover **open** (`manual-open`), **close** (`manual-close`), **scale-in** (`manual-add`), and **cancel a resting limit order** (`manual-cancel`). What is missing today is in-place protection editing on an already-open position — there is no command to **update an SL trigger**, **update a TP tier**, or **cancel an on-chain SL/TP**. To do those today an operator must:

1. Locate the correct Python adapter (varies by exchange)
2. Load environment variables from `.env` (path varies by deployment)
3. Shell out via Python with the right args
4. Parse JSON response to extract the new OID
5. Open `state.db` separately and `UPDATE` the position row

This is error-prone (env path mismatches, wrong script path, DB drift if a step is skipped) and violates the principle that manual interventions should go through the same atomic pipeline as automated ones. It also creates a reconciliation gap: if any step fails partway, the on-exchange order and the local `state.db` row diverge.

> **Validation note (corrects the original draft):** the original "Restart the service so in-memory state reloads from disk" step is **not** how the existing tooling works and should not be reintroduced. `manual-open` already avoids a restart by **queuing a `PendingManualAction`** that the running daemon drains on its next cycle (`scheduler/main.go:579`, `scheduler/manual.go:380`). A direct `UPDATE positions` from a separate CLI process is the *wrong* fix here: the daemon's in-memory `AppState` is authoritative and its next `SaveState` (`scheduler/db.go:942`) overwrites the row. New manual ops must follow the same queue-and-drain pattern, not write the positions table directly.

## Proposed solution

A set of new manual subcommands for in-place protection editing on open positions:

```
go-trader manual-update-sl  --strategy=X --symbol=Y --trigger=Z
go-trader manual-update-tp  --strategy=X --symbol=Y --tier=N --trigger=Z
go-trader manual-cancel-sl  --strategy=X --symbol=Y
go-trader manual-cancel-tp  --strategy=X --symbol=Y --tier=N
```

(Naming kept flat to match the existing `manual-open` / `manual-close` / `manual-add` / `manual-cancel` style rather than introducing a nested `manual <op>` group. If a nested group is preferred, the four existing commands should be migrated too so the surface stays consistent — that is a larger, separate change.)

All share the established `manual-open` pipeline:
- exchange call executed in the CLI process via the existing helpers (`RunHyperliquidUpdateStopLoss`, `RunHyperliquidCancelOrder`, `placeManualProtectionInline`)
- the result is **enqueued as a typed `PendingManualAction`** (new action kinds: `update-sl` / `update-tp` / `cancel-sl` / `cancel-tp`); the daemon adopts it into in-memory state on its next cycle — **no direct `state.db` positions UPDATE, no restart**
- kill-switch + circuit-breaker guards and owner/peer (shared-coin) guards mirrored from `manual-open`
- atomic semantics: cancel-then-place, or update-or-error, no half-states; on a queue-insert failure after the on-chain order succeeds, attempt the same cleanup path `manual-open` uses

This complements the existing `manual-open` pattern and extends it to in-place protection edits.

A `manual`-type strategy that only tracks operator-placed positions is the main beneficiary. Note these strategies *are* reconciled per cycle (`reconcileHyperliquidAccountPositions`, owner/peer-scoped and non-destructive) — the genuinely uncovered case is drift in an **SL/TP trigger price** the operator changed on-chain, which the reconciler does not auto-correct (it books only exact OID-confirmed fills).

## Scope

Additive — no behavior change to the existing pipeline. Larger than a thin CLI wrapper, though: each new op needs (a) a CLI handler, (b) a new `PendingManualAction` action kind plus any new fields, (c) daemon-side `applyManualAction` + `drainPendingManualActions` logic, (d) kill-switch/CB + owner/peer guards, (e) tests for the queue-and-drain path, and (f) docs. The `~50–150 LOC` estimate in the original draft is optimistic once the daemon-side adoption and guards are counted.

## Out of scope

- Multi-leg / batched updates in a single call (can be added later)
- Interactive mode / TUI (CLI only for now)
- A nested `manual <op>` command group (would require migrating the four existing flat commands)

## Agent guidance

Once implemented, agents operating the bot should prefer the `manual-*` subcommands over manual Python adapter + sqlite3 workflows. The repo's `AGENTS.md` / `CLAUDE.md` should be updated to document this rule, mirroring the existing `manual-open` guidance.

---

---

richkuo:
## Summary

Adds two manual subcommands to edit a tracked position's stop-loss in place, replacing the error-prone "find the Python adapter → load `.env` → shell out → parse JSON → `UPDATE state.db`" flow the issue describes:

- `./go-trader manual-update-sl <id> --trigger N [--symbol Y] [--dry-run]` — cancel-then-place the on-chain SL at a new trigger
- `./go-trader manual-cancel-sl <id> [--symbol Y] [--dry-run]` — cancel the resting on-chain SL

## Design

Both perform the exchange call in the CLI, then **queue a `PendingManualAction`** (`update-sl` / `cancel-sl`) the running daemon drains into in-memory state on its next cycle — **never a direct `positions` UPDATE, and no restart**. This matches the established `manual-open` pipeline: a separate CLI process writing the positions table directly would be clobbered by the daemon's next `SaveState`. Reuses the existing `pending_manual_actions` schema, so there is no DB migration.

This corrects the issue's original sketch (which proposed an in-process `state.db` UPDATE + restart) — see the validation note on the issue.

## Safety

- **`manualSLAutoManaged` hard-reject:** when an ATR/regime `stop_loss_atr_mult` or a trailing close would re-pin the operator's trigger on the next protection-sync cycle, the command refuses with guidance. Only strategies opted out of auto-SL (`stop_loss_atr_mult: 0`, no trailing) can manage the SL manually — otherwise the edit silently bounces back.
- **Immediate-fill guard:** `update-sl` rejects a trigger on the wrong side of the current mark (best-effort; a failed mark fetch warns but does not block — Python's `StopLossFilledImmediately` is the final arbiter).
- Same kill-switch / pending-circuit-breaker guards and owner checks as `manual-open`.
- SL ops record **no trade**; `manualActionRecordsTrade` keeps `drainPendingManualActions`' `sendTradeAlerts` tail slice aligned.
- Queue-insert-failure paths warn loudly; `cancel-sl` flags the position as on-chain-naked until reconcile.

## Scope decision

Built `update-sl` + `cancel-sl` only. `update-tp` / `cancel-tp` were deferred: there is no per-tier trigger stored on `Position` (only `TPOIDs`) and the protection sync re-derives tier triggers from ATR geometry, so TP-tier editing needs new storage + sync-suppression — a larger, separate change.

## Tests

`scheduler/manual_sl_test.go` covers the auto-managed guard (ATR/trailing → reject, no-auto / missing-ATR → allow), the immediate-fill predicate, the trade-recording predicate, and both `applyManualAction` cases (mutation, owner-mismatch, missing-position). Full Go suite + `go vet` pass.

## Docs

`CLAUDE.md` / `AGENTS.md` (symlinked) guardrail line and `SKILL.md` manual-trading section updated.

Closes #1050

---


richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

PR #1057, opened 2026-06-19
feat(#1055): --all auto-discovers deployments from systemd WorkingDirectory

### LINKED ISSUE #1055: update.sh --all scan-root model can't auto-discover deployments in non-sibling paths
### Summary

`scripts/update.sh --all` globs `$scan_root/go-trader-*/`, where `scan_root` defaults to the **parent dir of the source repo** (`dirname "$repo_root"`) — it is *not* a hardcoded `/root/go-trader-*` path (`update.sh:591-624`). The default only matches deployments that are direct siblings of the source clone. Deployments elsewhere (e.g. `/root/.openclaw/workspace/go-trader-*` while the source is `/root/go-trader`) don't match the default, so `--all` fails at preflight.

There **is** an existing escape hatch: `--update-all-root <dir>` / `GO_TRADER_UPDATE_ALL_ROOT=<dir>` overrides `scan_root`, so `./scripts/update.sh --all --update-all-root /root/.openclaw/workspace --restart` already works for the reproduction below. The real gap is twofold: (1) the default doesn't auto-discover non-sibling layouts, and (2) the single-glob model can't span deployments scattered across **different** parent dirs at all, even with the override. systemd auto-discovery solves both.

### Reproduction (default invocation, no override)

```bash
# Source at /root/go-trader
# Deployments at /root/.openclaw/workspace/go-trader-{live,paper-*}
$ cd /root/go-trader && ./scripts/update.sh --all --restart
[update] FAIL phase=preflight: no directories matching /root/go-trader-*/ (batch root: /root)
```

(Verified: message format matches `update.sh:624`; default `scan_root=/root` because the source is `/root/go-trader`.)

### Impact

When `update.sh --all` fails and the operator doesn't know about `--update-all-root`, they fall back to manual `rsync -a --delete /root/go-trader/ <deployment>/`. The source tree contains only `scheduler/config.example.json` — never the real runtime `scheduler/config.json` (gitignored; `.gitignore:17`, only `config.example.json` tracked). `--delete` then removes the real `config.json` from every deployment, and the next restart fails with `Failed to load config: open scheduler/config.json: no such file or directory`.

The script's own `--rsync-from` mode is safe here — it excludes `scheduler/config.json` (#790, `update.sh:399,424`). The data loss comes from the *manual* rsync that bypasses those excludes, not from `update.sh` itself.

This is exactly how I lost all 7 deployment configs during the v0.81.0 deploy on 2026-06-19 — `update.sh --all` failed (I didn't pass `--update-all-root`), I rsynced manually, every service crashed. Recovery required restoring `scheduler/config.json` from a separate pre-v0.77 backup and re-applying user-level overrides (TP3 trailing, strategy swaps) that weren't in the backup.

### Suggested fix

Make `update.sh --all` auto-discover deployments from systemd units, keeping the existing glob (with `--update-all-root`) as a fallback. The systemd unit already has the canonical `WorkingDirectory`:

```bash
discover_deployments_from_systemd() {
    # NOTE: confirm the actual unit naming on deployments — instance units of the
    # template are `go-trader@<name>.service` (CLAUDE.md: systemd/go-trader@.service),
    # which the `go-trader-*.service` glob below does NOT match. Adjust the pattern
    # (e.g. add `go-trader@*.service`) after checking `systemctl list-units`.
    for unit in $(systemctl list-units 'go-trader-*.service' 'go-trader@*.service' --no-legend --plain | awk '{print $1}'); do
        systemctl show "$unit" -p WorkingDirectory --value
    done | sort -u
}
```

Feed the (non-empty) results into the **existing** per-deployment loop, which already:
- skips any discovered dir lacking `scheduler/config.json` (`update.sh:634`) — preserve this guard;
- re-invokes `update.sh` per dir with the child args.

Skip empty `WorkingDirectory` lines (`systemctl show --value` returns empty for unset units; `sort -u` would otherwise emit a blank entry). Keep the legacy `$scan_root/go-trader-*/` glob as the fallback when no systemd units are found, so existing single-parent setups and the `--update-all-root` override are unchanged.

### Acceptance criteria

- `update.sh --all` auto-discovers deployment dirs from systemd `WorkingDirectory` (resolve `list-units` vs `list-unit-files` and the `go-trader-*` vs `go-trader@*` glob during impl).
- Layout-independent: works for `/root/go-trader-*`, `/root/.openclaw/workspace/go-trader-*`, and deployments split across different parent dirs.
- Discovered dirs still pass through the existing missing-`config.json` skip guard (`update.sh:634`); empty `WorkingDirectory` entries are dropped.
- Same rsync safety as `--rsync-from` mode (#790) — never touch `scheduler/config.json`.
- Existing flows unchanged: `update.sh --restart`, `--rsync-from <dir>`, and `--all --update-all-root <dir>` / `GO_TRADER_UPDATE_ALL_ROOT`.

---

richkuo:
Closes #1055

## Problem

`scripts/update.sh --all` globbed only `$scan_root/go-trader-*/` (scan_root = parent of the source repo by default). Deployments living under a different parent (e.g. `/root/.openclaw/workspace/go-trader-*` while the source is `/root/go-trader`) were invisible, so `--all` failed at preflight. Operators then fell back to manual `rsync -a --delete`, which wipes the gitignored `scheduler/config.json` from every deployment — the v0.81.0 config-loss incident.

Note: the existing `--update-all-root` / `GO_TRADER_UPDATE_ALL_ROOT` override already solved the single-parent repro, but couldn't span deployments scattered across different parent dirs, and required the operator to know about it.

## Change

`--all` now auto-discovers deployment dirs from loaded go-trader systemd units' `WorkingDirectory` — the canonical, layout-independent source. Unit patterns cover the primary unit (`go-trader.service`), plain per-deployment units (`go-trader-*.service`), and template instances (`go-trader@*.service`).

Fallback chain (all preserved):
1. An explicit `--update-all-root` / `GO_TRADER_UPDATE_ALL_ROOT` pins the legacy glob and skips auto-discovery — existing flow unchanged.
2. When `systemctl` is absent (macOS dev), no go-trader units are loaded, or discovery yields nothing, fall back to the `$scan_root/go-trader-*/` glob.
3. The existing missing-`config.json` skip guard and per-dir `[[ -d ]]` guard are untouched, so dirs without a real deployment are still skipped. `scheduler/config.json` is never touched.

## Tests

Pure helpers (`normalize_systemd_deployment_dirs`, `update_systemd_unit_globs`, `discover_deployment_dirs_from_systemd`) live in `update_helpers.sh` and are unit-tested in `test_update_helpers.sh`:
- Normalizer: trailing-slash collapse, drop empty/relative, de-dupe, layout-independence.
- Unit globs cover primary / plain / template-instance units.
- Stubbed-`systemctl` full pipeline (list-units → show → normalize), including a unit with unset `WorkingDirectory` (dropped) and dirs in unrelated parents.
- `systemctl`-absent fallback yields empty.

Dedup uses a newline accumulator (no associative arrays) for bash 3.2 compatibility. All helper tests + the three Go wrapper tests (`TestUpdateShell*`, `TestUpdateHelpers*`) pass; shellcheck clean apart from pre-existing infos.

---


richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude review

PR #1060, opened 2026-06-19
feat(#1056): move runtime config out of the deploy tree (systemd StateDirectory + migration)

### LINKED ISSUE #1056: Move scheduler/config.json outside the deployment tree to eliminate rsync footgun
### Motivation

`scheduler/config.json` is a runtime-only file (~8–91 KB depending on number of strategies) that is **never** committed to git. It lives inside the deployment working directory, which is also the path that `rsync -a --delete <source>/ <deployment>/` syncs into from `/root/go-trader`. Since the source tree only ships `scheduler/config.example.json`, `--delete` will silently remove the real config on every sync.

This is a structural footgun, not a workflow bug — the file simply shouldn't be in a path that any rsync/git/cleanup operation could reach. It already bit during the v0.81.0 deploy on 2026-06-19 when manual rsync killed all 7 deployment configs.

> **Validation note:** `update.sh`'s own `--rsync-from` path already excludes `scheduler/config.json` (`scripts/update.sh:399`), so the footgun is specifically **manual / ad-hoc** `rsync -a --delete`, not `update.sh`. The proposal still stands — moving the file out of the tree closes the manual-rsync / `git clean` / `rm -rf` hole that `update.sh`'s exclude list does not cover.

### Proposal

Move runtime config out of the deployment tree and pass its path explicitly via the existing `--config` flag (the binary already supports it: `go-trader --config <path>`, `main.go:89`; `agent-info` and `inspect` accept `--config` too).

**New canonical layout** (prefer `/var/lib/go-trader` over `/etc` — the daemon **rewrites** this file at runtime, which is `/var/lib` state semantics, not `/etc` static-admin-config semantics):

```
/var/lib/go-trader/
  live/config.json
  paper-hl-bnb/config.json
  paper-hl-btc/config.json
  paper-hl-eth/config.json
  paper-hl-hype/config.json
  paper-hl-sol/config.json
  paper-testing/config.json
```

**Systemd unit changes** (current units run `ProtectSystem=strict` and whitelist only `.../scheduler` as writable — see `systemd/go-trader@.service`):

```ini
[Service]
WorkingDirectory=/root/.openclaw/workspace/go-trader-live   # code checkout only
ExecStart=/root/.openclaw/workspace/go-trader-live/go-trader --config /var/lib/go-trader/live/config.json
EnvironmentFile=/root/.openclaw/workspace/go-trader-live/.env
# REQUIRED: the daemon writes config in place, so the new config DIRECTORY
# (not just the file) must be writable under ProtectSystem=strict.
ReadWritePaths=/var/lib/go-trader/live
```

**Critical: config is mutated in place.** The daemon writes `config.json` via temp-file + `os.Rename` **in the config's own directory** on several paths — config migration write-back on load/SIGHUP (`config_migration.go:225`), the UI tuner (`ui_tuner.go:735`), and Discord mutating commands / POST `/config` (`discord_mutating_commands.go:588`). So the **directory** must be daemon-writable. Under `ProtectSystem=strict`, `/etc` is read-only and the current `ReadWritePaths=.../scheduler` no longer covers the config — every runtime config write (migration, UI, Discord) fails with `EROFS` unless the new dir is added to `ReadWritePaths` in both units.

The working directory becomes a pure code checkout. `rsync`, `git clean`, `trash`, accidental `rm -rf`, etc. can no longer reach the config. Backups are obvious (one directory, one purpose).

### Migration

1. Move existing `scheduler/config.json` files to `/var/lib/go-trader/<name>/config.json` (one-shot — this is live config data; back up first, a botched move or missing `ReadWritePaths` leaves a daemon unable to persist config)
2. Update each systemd unit's `ExecStart` to pass `--config /var/lib/go-trader/<name>/config.json` **and** add `ReadWritePaths=/var/lib/go-trader/<name>`
3. Symlink `scheduler/config.json -> /var/lib/go-trader/<name>/config.json` for a transition period so existing scripts (and `update.sh`'s deployment-dir guard, below) keep working
4. After a release cycle, drop the symlink — but only once `update.sh` no longer keys off `scheduler/config.json` (see below)

### Backwards compatibility

- `--config <path>` already exists, so binaries don't need to change
- `update.sh` / `inspect` / `agent-info` already accept `--config` (defaulting to `scheduler/config.json`); ad-hoc runs without `--config` break if the file is absent, so they need the new default/fallback or an explicit flag
- One-cycles, ad-hoc runs from the working dir need to either pass `--config` explicitly or have the binary look one level up at the symlink

### Acceptance criteria

- `scheduler/config.json` is not a file the deployment owns directly — it's a symlink or absent
- `rsync -a --delete` from source cannot destroy the live config (any `--delete` operation is safe)
- All runtime config writes (migration, UI tuner, Discord `/config`) still succeed under `ProtectSystem=strict` — i.e. the new config dir is in `ReadWritePaths`
- All existing systemd units work without code changes (just the `ExecStart` + `ReadWritePaths` lines)
- `update.sh` is updated for the new path — beyond dropping its config-exclude rule, it reads `scheduler/config.json` for **deployment detection** (`update.sh:652,666` — refuses to run without it) and **db-exclude resolution** (`update.sh:367`); both must learn the new path before the transition symlink is dropped

### Related

- Companion fix: `update.sh --all` should auto-discover deployments via systemd (#TBD)
- Both changes make the deployment safer against accidental data loss

---

richkuo:
Closes #1056.

## Problem
`scheduler/config.json` is gitignored runtime state that lived inside the deployment tree — the same path a manual `rsync -a --delete <source>/ <deploy>/` syncs into. Since source ships only `config.example.json`, `--delete` silently removes the live config. Move it out of the tree so no rsync / `git clean` / `rm` can reach it.

## Changes
- **systemd units** (`go-trader.service`, `systemd/go-trader@.service`): launch with `--config /var/lib/go-trader[/%i]/config.json` and add `StateDirectory=go-trader[/%i]`. `StateDirectory` is the FHS-correct, systemd-native way to create that directory writable under `ProtectSystem=strict` — the daemon **rewrites config in place** (migration write-back `config_migration.go:225`, UI tuner `ui_tuner.go:735`, Discord/POST `/config` `discord_mutating_commands.go:588`) via atomic temp+rename, so the *directory*, not just the file, must be writable. `WorkingDirectory` stays the code checkout (scripts/.venv/state DB resolve relative to it).
- **`scripts/migrate-config-out-of-tree.sh`** (new): idempotent one-shot — moves config to the canonical path, leaves a transition symlink at `scheduler/config.json`, refuses to clobber an existing target (incl. a broken symlink), `--dry-run`, and prints the unit edits to apply. Does **not** touch live units.
- **`update.sh`**: unchanged behavior through the transition symlink — deployment detection (`:652`), db-exclude resolution (`:367`), and `--all` discovery (`:634`) all follow the symlink; the rsync exclude still protects it from `--delete`. Not-found message now points at the new layout.
- **`update_config_migration_state` helper + tests**: classifies `regular` / `symlink` / `missing`; a **dangling** symlink reports `symlink` (checked first) so a re-run never re-migrates and clobbers the live config pointer. New script added to the bash-syntax test.
- **Docs**: CLAUDE.md, README, SKILL.md.

## Design notes
- **Transition symlink** chosen over teaching `update.sh`/binary the real path now: lowest risk, `update.sh` keeps working with no logic change. Dropping the symlink later (acceptance-criterion step 4) is a deliberate follow-up.
- **`/var/lib/go-trader`** over the issue's `/etc/go-trader`: the file is runtime-mutated state, which is `/var/lib` semantics, not static `/etc` config (and `/etc` is read-only under `ProtectSystem=strict` anyway).

## Verification
- Full Go suite green (`go test ./...`); `update_helpers` bash tests green; `bash -n` clean on all scripts.
- Migration script functionally tested in a sandbox: dry-run leaves the file untouched; real run moves + symlinks (config readable through the link); re-run is an idempotent no-op.

## Operator action (live deploys)
Run `scripts/migrate-config-out-of-tree.sh --instance <name>` per deployment, then apply the printed `ExecStart`/`StateDirectory` edits + `systemctl daemon-reload` + restart.

---

https://claude.ai/code/session_01HxbTLMVNf9DjJthKyYJq6B


richkuo:
@claude review

richkuo:
Addressed all three Recommended Optional findings in 5a989fb.

1. **Config-write race in the migrate→restart window** — confirmed: writes do `os.Rename(tmp, configPath)` and `rename(2)` over a symlink replaces the symlink itself, re-creating an in-tree config and losing the write. The migration script now **refuses to run while this deployment's daemon is live** (matched by working dir == deploy_dir via the existing `update_should_sweep_proc` predicate; best-effort, needs `/proc`), with `--force` to override after a manual stop. Printed steps + README/CLAUDE.md reordered to stop-first / start-after.
1. **Base-aware writable directive** — new `update_config_writable_directive` prints `StateDirectory` for `/var/lib` bases and `ReadWritePaths=<dir>` for any other `--base`, so the printed unit edits are valid for every accepted base.
1. **Instance-name escapes** — new `update_validate_instance_name` rejects `.`, `..`, leading-dash, empty, and disallowed chars (the old bare char-class let `..`/`-live` through).

Both new helpers are pure and unit-tested, including the adversarial `..`/`.`/`-live`/slash/space and non-`/var/lib`-base cases. Full Go suite + bash syntax + helper tests green.

---

richkuo:
@claude review

richkuo:
Addressed in bc8fe13.

**Daemon-refusal ordering vs idempotency** — confirmed: the refusal ran before `update_config_migration_state`, so a re-run on an already-migrated live deployment (symlink present, daemon running) errored instead of no-opping. Moved the refusal to **after** the state case block, so the `symlink` (no-op exit 0) and `missing` (error) cases return first and the refusal is reached only when `state == regular` (a mutating `mv`/`ln` is actually impending). Nothing is clobbered when the path is already a symlink, so no stop is required.

Added an end-to-end regression test: an already-migrated symlinked deployment exits 0 as a no-op with the daemon-refusal code present and mutates nothing (covers the live-migrated re-run, converge-loop, and dangling-symlink cases). Helper tests + bash syntax green.

---

richkuo:
@claude review

PR #1063, opened 2026-06-19
feat(#1062): configurable regime window display in cycle summary

### LINKED ISSUE #1062: Configurable regime window display in summary
### Problem
`formatStrategyRegimeDisplay` (`scheduler/regime_window_spec.go:395`) renders **all** regime windows in the Discord/cycle summary, including scalar windows (`long`, `medium`, `short`) and composite windows (`composite_long`, `composite_medium`, `composite_short`).

This produces verbose output like:
```
composite_long=ranging_directional [composite]; composite_medium=ranging_directional [composite]; composite_short=trending_down_choppy [composite]; long=ranging [adx]; medium=ranging [adx]; short=trending_down [adx]
```

The scalar (non-composite) windows add noise without much value — the composites already aggregate the regime signal.

### Proposed
Add a config option to control which regime windows appear in the summary display. Something like:

```json
"regime": {
  "display_windows": ["composite_long", "composite_medium", "composite_short"]
}
```

- If `display_windows` is set, `formatStrategyRegimeDisplay` only renders windows whose names are in the list
- If unset/empty, current behavior (show all windows) is preserved
- Alternatively: a boolean `regime.display_composite_only` could toggle the scalar windows off

### Location
- `scheduler/regime_window_spec.go:395` — `formatStrategyRegimeDisplay()`
- The loop at line 410 iterates `ss.RegimeWindows` alphabetically and formats each as `name=label [classifier]`

### Acceptance
- New config field controls which windows render in the summary
- Default behavior unchanged (backward compatible)
- Scalar windows (`long`, `medium`, `short`) can be hidden without removing them from the regime calculation

richkuo:
## Summary

Adds optional `regime.display_windows` (`[]string`) to filter which regime windows render in the Discord/cycle summary, addressing the verbose all-windows output (composite + scalar) described in #1062.

Display-only: the filter never affects regime calculation, gating, or persisted state — only what `formatStrategyRegimeDisplay` prints.

## Behavior

- **Empty/omitted** → render every window (legacy behavior, backward compatible).
- **Set** → only windows whose name matches render (case/space-insensitive via `normalizeRegimeWindowKey`). Lets operators show composite-only, scalar-only, or any subset.
- **Set but no populated window matches** → falls back to the single primary regime string (same fallback as the multi-window-disabled path).
- **Blank-only entries** (`""`, `"   "`) are ignored, so a stray empty string can't collapse the summary to show-nothing.

Example — composite-only:
```json
"regime": { "display_windows": ["composite_long", "composite_medium", "composite_short"] }
```

## Notes

- No `config_version` bump: the strict unknown-key guard (`config_unknown_keys.go`) is scoped to `strategies[]` only; the top-level `regime` block tolerates forward-compat keys (same precedent as #1048 `circuit_breaker`).
- Window names are runtime/Python-determined (composite_* aren't in `regime.windows`), so the field is a pure display filter with no strict config-load validation — a typo simply doesn't match (and falls back to the primary regime string).
- A "none/off" mode was intentionally not added per discussion; a non-matching list yields the primary regime string, not a blank line.

## Tests

Added 5 cases to `regime_window_spec_test.go`: show-all default, composite-only, case-insensitive match, no-match fallback, blank-entry safety. Full Go suite + `go vet` pass.

Closes #1062

---


richkuo:
@claude review

richkuo:
Addressed both Recommended Optional findings in 51c734c.

1. Typo guard — implemented at config load (`validateRegimeWindowsConfig`), not as a runtime warn. `display_windows` names must be `regime.windows` keys; a typo now rejects, and `display_windows` set without `regime.windows` rejects too. Note on the review's must-survive case 3 ("valid `composite_*` names absent from `regime.windows` must NOT warn"): that premise doesn't hold against the code — `check_regime.compute_regime_bundle` keys window snapshots by `windows_spec` names only, so `composite_*` are operator-chosen window keys, not Python-synthesized. A `composite_*` not in `regime.windows` is therefore a genuine typo and is correctly rejected. The render-time fallback to the primary regime string stays for the valid-but-unpopulated-this-cycle case.

2. Hot-reload — `regimeConfigEqualIgnoringDisplayWindows` masks the field from the restart-required compare; `applyHotReloadConfig` applies it. Display-only change applies with an open position; a compound change (display_windows + any other regime field) still rejects without mutating; clearing reverts to render-all.

Tests cover both classes (valid/typo/mixed/blank/requires-windows; display-only-applies/compound-rejects/clear-reverts). Full Go suite + vet green.

---

richkuo:
@claude review

PR #1072, opened 2026-06-19
fix(#1067): backtester --config falls back to args[0] for open strategy name

### LINKED ISSUE #1067: [C35] init-generated configs omit open_strategy.name → run_backtest.py --config rejects them
**Complexity: 35/100** — scope: small (init config emission + backtester `--config` reader); risk: low (backtest-only tooling, no live/money path); uncertainty: low-medium (the fix locus is a genuine choice)

## Summary

Configs produced by `./go-trader init` (and `init --json`) describe each strategy in the legacy `args[0]` form — e.g. `args: ["momentum", "BTC", "1h"]` — and do **not** populate an `open_strategy.name` field. The backtester's `--config` path (`backtest/run_backtest.py`, the #951/#1025 migrated-config reader) requires `open_strategy.name` and rejects these configs at load:

```
ValueError: <config>: strategy 'momentum-btc' has no open_strategy.name; the migrated config should always populate it.
```

So an operator cannot take a freshly-`init`-generated config and backtest a strategy from it via `--config` without first hand-editing it into the migrated v15 shape. Backtesting via the `--strategy` CLI flag (no `--config`) is unaffected.

## Repro

1. `./go-trader init --json '{"Assets":["BTC"],"EnableSpot":true,"SpotStrategies":["momentum"],"SpotCapital":1000}' --output /tmp/mom.json`
1. `uv run --no-sync python backtest/run_backtest.py --config /tmp/mom.json --strategy momentum-btc --symbol BTC/USDT --timeframe 1h --mode single`
1. → `ValueError: ... has no open_strategy.name`

Affects all strategies, not specific to any one (surfaced while building the `atr_band_revert` strategy; orthogonal to that work).

## Resolution options

1. **Emit `open_strategy.name` in `generateConfig`** so init configs are born in the migrated v15 shape. Most correct long-term (single canonical config shape), but touches every platform loop / config consumer and needs a round-trip test.
1. **Fall back to `args[0]` in the backtester `--config` reader** when `open_strategy.name` is absent, mirroring how the live daemon derives the open strategy from args. Smaller, backtest-local blast radius.

Option 2 is lower-risk; option 1 is more correct. Decide the locus before implementing, and add a regression test that round-trips an `init`-generated config through `run_backtest.py --config`.

---

richkuo:
## Summary

`./go-trader init` (and `init --json`) emit each strategy in the positional **args-form** — `args[0]` is the concept name, `open_strategy.name` is empty — and stamp the file `config_version=15`. The v13→v15 migration only backfills `open_strategy.name` for **pre-v13** files, so an init-stamped v15 config reaches the backtester `--config` reader name-less and was hard-rejected:

```
ValueError: <config>: strategy 'momentum-btc' has no open_strategy.name; the migrated config should always populate it.
```

The **live daemon** runs these configs fine — every consumer resolves the open strategy via `effectiveOpenStrategy` (`scheduler/strategy_composition.go:51`), which prefers `open_strategy.name` and **falls back to `args[0]`**. The backtester `load_strategy_config` reader was the lone path lacking that fallback, so backtest diverged from live on a config the daemon accepts.

## Fix (Option 2 — parity with live)

`backtest/run_backtest.py` `load_strategy_config` now mirrors `effectiveOpenStrategy`: prefer `open_strategy.name`, else fall back to `args[0]`, and raise only when **neither** exists. The returned `open_strategy.name` uses the resolved value.

Chose Option 2 over Option 1 (emit `open_strategy.name` in `generateConfig`) because the live runtime already tolerates the args-form everywhere via the fallback — so Option 2 fixes init configs **and** any hand-edited/legacy args-form config, while Option 1 alone would leave that broader class still rejected (migration never re-touches a v15-stamped file).

## Tests

5 regression tests added to `backtest/tests/test_strategy_refs_641.py`:
1. Round-trips the verbatim `go-trader init` shape (`args:[name,…]`, `open_strategy:{"name":""}`) through `load_strategy_config` → resolves `args[0]`.
2. Missing `open_strategy` key entirely → falls back to `args[0]`.
3. Explicit `open_strategy.name` wins over `args[0]` (precedence parity).
4. Whitespace-only name → falls back (TrimSpace parity).
5. Neither name nor args → still raises.

Verified: reproduced the original failure against a real `init --json` config, confirmed it now backtests; full backtest suite green (598 passed).

Closes #1067

---

https://claude.ai/code/session_01CnfoggdxDbHXCmA277WPbU

### DIFF
--- a/backtest/run_backtest.py
+++ b/backtest/run_backtest.py
@@ -377,11 +377,29 @@ def load_strategy_config(config_path: str, strategy_id: str,
     for sc in cfg.get("strategies", []) or []:
         if sc.get("id") != strategy_id:
             continue
-        open_ref = sc.get("open_strategy") or {}
-        if not isinstance(open_ref, dict) or not open_ref.get("name"):
+        open_ref = sc.get("open_strategy")
+        if not isinstance(open_ref, dict):
+            open_ref = {}
+        # #1067: mirror the live daemon's open-strategy resolution
+        # (effectiveOpenStrategy, strategy_composition.go): prefer
+        # open_strategy.name, else fall back to the positional args[0] strategy
+        # arg. `go-trader init` emits the args-form (args[0]=concept name) with an
+        # empty open_strategy.name, and the v13->v15 migration only backfills
+        # open_strategy.name for pre-v13 files — so an init-stamped v15 config
+        # (and any hand-edited args-form config) reaches here name-less. The live
+        # daemon runs these fine via this same args[0] fallback, so the backtester
+        # must resolve the identical name instead of rejecting, or backtest and
+        # live silently diverge on a config the daemon accepts.
+        open_name = str(open_ref.get("name") or "").strip()
+        if not open_name:
+            args_list = sc.get("args")
+            if isinstance(args_list, list) and args_list:
+                open_name = str(args_list[0] or "").strip()
+        if not open_name:
             raise ValueError(
-                f"{config_path}: strategy {strategy_id!r} has no open_strategy.name; "
-                f"the migrated config should always populate it."
+                f"{config_path}: strategy {strategy_id!r} has neither "
+                f"open_strategy.name nor a positional args[0] strategy arg to "
+                f"resolve the open strategy from."
             )
         regime_cfg = cfg.get("regime") or {}
         if not isinstance(regime_cfg, dict):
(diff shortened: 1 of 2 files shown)


richkuo:
@claude review

Issue #1074, opened 2026-06-19
[C70] Wire a gate-passing regime model into the live classifier (parity-safe)


richkuo:
**Complexity: 70/100** — scope: Python live inference path (causal-decoder relocation + possible enriched on-cycle features) + Go composite-spec/model schema + fail-closed validation + parity tests; risk: the live regime classifier feeds SL/TP/entry-gate labels and the Go↔Python parity surface, mitigated by an opt-in model block whose absence preserves the current hand-rule byte-for-byte; uncertainty: no gate-passing model exists yet and the enriched-vs-canonical schema of an eventual winner is unknown.

> **Status (2026-07-02): dormant — no shippable model exists, but the promotion gate is now reachable.** #1080 and #1081 are closed, but they delivered gate harnesses and evidence runs, not a gate-passing model. The #1095 enriched-feature bake-off (PR #1168) abstained every candidate verdict, and #1177 recorded that the incumbent hand-rule's own OOS forward-volatility separation did not replicate (p=0.105/0.113 at n_perm=1799 vs the original knife-edge 10/201 ≈ 0.0498 at n_perm=200). **#1211 (PR #1215) resolved this:** a 24-cell window×asset re-measurement confirmed the incumbent's separation does not replicate family-wide (11/24 significant at α/24), so `gate_verdict` was revised to candidate-self-v2 semantics — `ship = separation_ok and stability_ok`, with the `incumbent_trustworthy` veto dropped. Candidates now promote on their own significant separation + non-inferiority + stability gain, so the gate no longer abstains every candidate. A gate-passing model still has to be produced (blockers 2–4 below). #1078, #1082, and #1085 remain complete as before.

## Blocked On
Do **not** start live wiring until all of the following hold; otherwise the current composite hand-rule remains the live classifier:

1. ~~**A trustworthy incumbent baseline**~~ — **RESOLVED by #1211 (PR #1215).** The #1177 downgrade was resolved via a deliberate gate-semantics revision: the `incumbent_trustworthy` hard-gate was dropped and candidates now promote on their own evidence (candidate-self-v2). Verdicts no longer abstain on incumbent non-significance; blockers 2–4 still stand.
2. **A candidate model clears the forward-volatility separation/stability gate** out-of-sample (the #1080/#1095 gate pipeline; nothing has passed to date — no bake-off has been re-run since the #1211 semantics change; that re-run is tracked in #1218).
[...]

PR #1086, opened 2026-06-20
feat(#1082): bounded-window ADX re-validation gate for regime-model promotion

### LINKED ISSUE #1082: [C35] Bounded-window ADX re-validation before regime-model promotion
**Complexity: 35/100** — scope: small (a comparison harness + go/no-go check); risk: low (read-only); uncertainty: low-medium (size of the live-vs-fit ADX drift is unknown until measured).

## Problem
The offline fit computes Wilder ADX over the **full** eval window (`composite_feature_matrix` → `compute_regime`, recursive warm-up seeded at the window start), while the **live** regime check runs on a bounded fetched lookback. Wilder ADX is recursive, so the same bar can receive a different ADX live vs in the fit — a model fit on full-history ADX may not reproduce live.

**Note (effective ADX period):** the composite classifier's ADX sub-component is period-capped at `COMPOSITE_ADX_PERIOD_CAP = 14` (`shared_tools/regime.py:48`, applied in both `composite_feature_matrix` and `compute_regime_composite`), independent of the fit's `period=48`. The Wilder recursion therefore decays at `(13/14)^k` and is near-fully forgotten well within the live `--ohlcv-limit 200` lookback, so the realistic warm-up drift may be near-zero. The harness must measure drift on the **effective period-14** ADX (not period-48) and the go/no-go threshold must be set against that — otherwise it over-expects drift.

## Goal
Re-validate a fitted model's forward-volatility separation under **bounded-window** ADX (matching the live lookback) before any promotion, and quantify the drift.

## Definition of done
A bounded-window ADX comparison harness + a documented go/no-go check; promotion (#1074) is gated on it.

## Relationships
Relates #1078, #1080; gates #1074. One PR.

---

richkuo:
## Summary
Adds a **live-faithful** re-validation harness that gates promoting a fitted regime model into the live classifier (#1074) on a re-score under **bounded-window ADX**.

The offline fit (`regime_calibrate.fit_on_window` → `composite_feature_matrix` → `compute_regime`) seeds recursive Wilder ADX at the **full eval-window start**. Live (`shared_scripts/check_regime.py`, `--ohlcv-limit` default 200, a fresh subprocess per cycle) re-runs over a **bounded** fetch, so the same bar's recursive ADX can differ live vs in the fit — a full-window-fitted model may not reproduce live.

## What it does
`backtest/regime_bounded_window_validate.py`:
- **Faithful per-bar live reproduction** — for each scored bar, runs exactly what one live cycle does for that bar: the real `composite_feature_matrix` (bounded ADX inside) over the trailing `lookback` slice, then `forward_filter_labels` (model) / `compute_regime_composite` (incumbent) over that slice, last bar.
- **Drift quantification** — ADX abs/rel drift (mean/median/p95/max/corr), per-bar label agreement + a confusion map, on the bars valid in both views.
- **Go/no-go gate (#1074 is blocked on it)** — re-runs `regime_calibrate.gate_verdict` on the **bounded-window** labels of both arms; promote iff the gate still `ship`s under bounded ADX **AND** full-vs-bounded model label agreement ≥ `--agreement-threshold` (0.95). Surfaces the full-window verdict to flag a passes-full/fails-bounded **regression**, plus an optional `--lookback-sweep`. CLI exits non-zero when promotion is blocked.

## Findings during the build
- **Wilder ADX is causal** → bounded == full *exactly* once the bounded window is armed (> 2·`COMPOSITE_ADX_PERIOD_CAP` = 28 bars; live's `--min-bars 30` guarantees arming). Drift is therefore *only* lost warm-up and decays at `((cap−1)/cap)^k`, so at the live lookback of 200 it is typically small — the harness **measures** it rather than assuming. (Synthetic check: mean ADX drift L=50 → 2.35 pts, L=150 → 0.002.)
- Only the `adx` feature is path-dependent (`standard_atr` is a simple rolling mean, not Wilder), but the harness recomputes the **whole** feature vector per window so it stays faithful if that ever changes.
- The full-window (fit) view is *cold* at the window's first ~2·period bars while the bounded view is *warm* there — the real live asymmetry; drift/agreement are measured on the intersection so the fit's cold warm-up bars cannot masquerade as drift.

## Tests
`backtest/tests/test_regime_bounded_window.py` (15 tests): causality invariant (bounded == full when the window spans the series), drift decay vs lookback, faithful per-bar reproduction == one-shot live call, drift stats, and the go/no-go gate logic (ships / fails-gate / low-agreement / regression / default threshold). Pure `validate_frames` core unit-tested without the data loader.

Full suite: **1765 passed**.

> Note: the real-data CLI path (`validate` → `load_cached_data`) was not exercised end-to-end — no price cache in the dev worktree. The pure core is fully tested with synthetic OHLCV + a fitted HMM.

## Docs
Entry added to `docs/ARCHITECTURE.md` (Backtest harnesses) and a guardrail line in `CLAUDE.md` (Backtest).

Closes #1082

---

https://claude.ai/code/session_01QCdkkYihicBvkZJw2Q2FLb


richkuo:
@claude review

richkuo:
Addressed all three Recommended-Optional findings (commit 0243ff0):

1. **Vacuous agreement on near-empty intersection** — `go_no_go` now takes the model label-drift dict and **fail-closes** below `--min-agreement-bars` (default 30, mirrors `check_regime --min-bars 30`); `label_drift_stats`' `agreement=1.0`-on-zero can no longer clear the gate. Tests: `test_go_no_go_blocks_on_insufficient_comparable_bars`, `test_go_no_go_zero_comparable_bars_vacuous_agreement_blocks`.
2. **Incumbent period mismatch** — hand-rule arm is scored at `--incumbent-period` (= `regime_diagnostics.DEFAULT_PERIOD`, now a shared constant imported not re-declared) while the model arm uses its fit period, so the full-window verdict equals `regime_calibrate`'s gate even when `period != 48`. Test: `test_validate_frames_scores_incumbent_at_its_own_period`.
3. **`--lookback-sweep` always exited 0** — sweep exit is now worst-case via `_sweep_blocked` (non-zero if ANY swept lookback blocks promotion). Test: `test_sweep_blocked_worst_case_over_lookbacks`.

Full suite: 1769 passed. Smoke confirms the floor trips and periods/agreement/comparable-bars are surfaced. Docs updated.

---

richkuo:
@claude review

richkuo:
Addressed the two Recommended-Optional findings in `603a5a1`.

1. **In-sample re-score guard.** `validate` now resolves the model's `fitted_on` stamp against the validation `(symbol, timeframe, window)`. A confirmed in-sample re-score (all three match) hard-blocks promotion in `go_no_go`; an unstamped model warns by default or blocks under the new `--require-provenance` flag. Both surface a `provenance` report block and a stderr line, so an in-sample re-score is never silently promoted. The confirmed-overlap case is always refused regardless of the flag.

2. **Gate-primary-horizon guard.** A `--horizons` set omitting the gate's primary horizon is now rejected up front (`parser.error` at the CLI, `ValueError` in `validate_frames`) instead of raising a deep `KeyError` mid-run. The required horizon is derived by introspecting `gate_verdict`'s `primary` default, not a second hardcoded `4`, so it tracks the gate.

`go_no_go.promote` is now the single negation of `blocking_reasons`. 9 new tests (28 total in the harness suite, all green); ARCHITECTURE.md + CLAUDE.md updated.

---

richkuo:
@claude review

richkuo:
Addressed both Recommended-Optional findings in `5540ba6`.

1. **In-sample guard now detects date-range overlap, not just window-name equality.** `_provenance_status` resolves both the fit and validation windows to their `eval_windows.WINDOWS` date ranges and flags `in_sample` on any overlap (`_windows_overlap`, half-open `[start, end)` with strict `<` so the shared `is`/`oos` boundary stays disjoint), for the same `(symbol, timeframe)` only. So `is`-fit validated on the held-out `2025H1` (~3 weeks shared) now blocks; `is` vs `2023`/`2024` stays clean; and a fit-window name absent from `WINDOWS` is treated as unverifiable (warn, or block under `--require-provenance`) rather than silently OK or crashing.

2. **`validate()` + `main()` are now covered.** Added monkeypatched-loader tests of the real-data glue and CLI exit-code contract: in-sample model blocks (exit 1), disjoint OOS model promotes (exit 0), unstamped model blocks only under `--require-provenance`, `--horizons` missing the gate primary rejects up front, and the `--lookback-sweep` worst-case exit. `gate_verdict` is stubbed to ship in these so the gate's own statistics don't make the promote path flaky — the glue under test is the provenance→go_no_go wiring, not the gate's separation math (covered by the `go_no_go` unit tests).

9 new tests (37 total in the harness suite, all green); ARCHITECTURE.md + CLAUDE.md updated.

---

richkuo:
@claude review

PR #1089, opened 2026-06-20
fix(#1088): throttle shared-wallet drift WARN log + drop cycles%10 re-alert

### LINKED ISSUE #1088: [C25] Shared-wallet drift alert spam: log WARN every reconcile cycle + cycles%10 re-alert preempts hourly back-off
## Problem

The shared-wallet drift alert in `scheduler/shared_wallet_drift_alerts.go` spams excessively. Observed: **620 `[WARN]` log lines in 6 hours** for a stable $0.88 drift on `hyperliquid/0x736dD10faf3CaE88199Dab551dBe0713fe68cc86`. That is ~1 line per 35s — the rate at which the per-cycle reconcile runs (the WARN fires once per main-loop reconcile, not per strategy check).

### 1. stdout `[WARN]` fires every cycle unconditionally

In `reportSharedWalletDrift()` (`shared_wallet_drift_alerts.go:271`):
```go
fmt.Printf("[WARN] shared-wallet %s drift $%+.2f ...\n", ...)
if !shouldNotify || notifier == nil || !notifier.HasBackends() {
    continue
}
```

The `fmt.Printf` is **outside** the `shouldNotify` gate (`:273`). Only the Discord/DM notification is throttled. The log line fires every reconcile cycle in which drift exceeds tolerance, regardless of throttle state — this is the 620-lines-in-6h spam.

### 2. `cycles%10 == 0` re-alert cadence ignores the hourly back-off

The throttle in `Record()` re-alerts on (`shared_wallet_drift_alerts.go:171`):
```go
case e.cycles%10 == 0:
    shouldNotify = true
```

`cycles` increments every reconcile cycle, so for a stable (no material change) already-alerted drift the `cycles%10` case fires a Discord/DM alert every 10th cycle. At the observed ~35s reconcile cadence that is roughly every 6 minutes. The hourly back-off (`now.Sub(e.lastNotifiedAt) >= time.Hour`, `:173`) is effectively never reached: it sits **after** the `cycles%10` case in the same switch, so the 10-cycle rule always trips first. The 10-cycle cadence was likely designed assuming much longer cycle intervals.

## Proposed Fix

1. **Dedup the `[WARN]` log on unchanged drift** (or apply a minimum log interval, e.g. once per minute) — preferred over gating it entirely behind `shouldNotify`, which would suppress all per-cycle visibility of a worsening-but-sub-threshold drift.
2. **Remove the `cycles%10` case** so the existing hourly back-off at `:173` governs persistent stable drift (or replace `cycles%10` with an explicit ≥30-minute time-based back-off). No new timer infrastructure is needed — the hourly case already exists; it is simply being preempted.

## Environment
- Deployment: `go-trader-live` on `openclaw-hetzner-2`
- Reconcile cadence: ~35s (620 WARN lines / 6h); the ~2.5s figure is the strategy check interval, not the reconcile/main-loop cadence
- Observed drift: $-0.88 stable (Σ members $1129.54 vs balance $1128.66)

---
Created with LLM: GLM-5.2 | unknown | Harness: zai

richkuo:
## Problem

The shared-wallet drift alarm spammed: **620 `[WARN]` lines in 6h** for a stable $0.88 drift. Two causes in `scheduler/shared_wallet_drift_alerts.go`:

1. The stdout `[WARN]` line fired every reconcile cycle — it sat **outside** the `shouldNotify` gate, so only the Discord/DM notification was throttled.
2. The `cycles%10 == 0` re-alert case preempted the hourly back-off (it sits earlier in the same switch and `cycles` increments every cycle), firing operator alerts on a fixed cycle count instead of the intended ~hourly cadence.

## Changes

- **Log throttle:** `Record` now also returns `shouldLog`, throttled per wallet — logs at episode onset, on any cycle that fires an operator alert, then at most once per `sharedWalletDriftLogInterval` (1 min). `reportSharedWalletDrift` gates the `[WARN]` `Printf` on it. The onset/alert log lines are preserved, so the stdout trail still correlates with Discord/DM alerts; only the per-cycle repetition of an unchanged stable drift is suppressed.
- **Re-alert cadence:** removed the `cycles%10` switch case. A stable, already-alerted drift now re-alerts on the existing hourly back-off only — no new timer infrastructure; the hourly case was simply being preempted.

## Tests

- Updated the 22 `Record` call sites in `shared_wallet_reconcile_cycle_test.go` to the new 3-return signature.
- Added `TestSharedWalletDriftTracker_LogThrottledPerInterval` — onset logs, alert cycle logs, intra-interval cycles do **not** log, post-interval cycle logs again.
- Added `TestSharedWalletDriftTracker_StableDriftRealertsHourlyNotEveryTenth` — drives 40 sub-hour stable cycles asserting no re-alert (the old `%10` would have fired at 10/20/30/40), then confirms the hourly back-off re-alerts.
- Full `go test ./...` (scheduler) passes.

Closes #1088

---

https://claude.ai/code/session_01YQsxXPDK1gpn4e2XgVxsZC

### DIFF
--- a/scheduler/hl_reconcile_gap_alerts.go
+++ b/scheduler/hl_reconcile_gap_alerts.go
@@ -43,15 +43,25 @@ const hlReconcileGapAlertThreshold = 3
 // hlReconcileGapRealertRatio gates re-alerts on an already-alerted coin to a
 // material change in the residual qty (relative to the qty at the LAST
 // notification), so a gap whose residual drifts slightly (a peer partial fill,
-// a small re-open) does not spam; combined with the every-10th-cycle / hourly
-// back-off below.
+// a small re-open) does not spam; combined with the hourly back-off below. The
+// same ratio (anchored to the last LOGGED residual) gates the stdout log.
 const hlReconcileGapRealertRatio = 0.10
+
+// hlReconcileGapLogInterval is the heartbeat ceiling for the stdout [WARN] gap
+// line per coin (#1088 sibling fix): once a gap is persisting, a STABLE,
+// unchanging residual re-logs at most once per this interval. The stdout line
+// previously fired every reconcile cycle a gap persisted (unconditionally,
+// outside the alert gate) — the identical per-cycle log spam #1088 fixed in the
+// drift reporter. Aligned to the hourly notification back-off so a persistent
+// stable gap's stdout cadence matches its alert cadence; a materially-changed
+// residual (hlReconcileGapRealertRatio) or any cycle that fires an alert logs
+// immediately regardless, preserving onset, worsening, and alert visibility.
+const hlReconcileGapLogInterval = time.Hour
 
 // hlReconcileGapEntry is one slot in the per-coin gap tracker.
 type hlReconcileGapEntry struct {
 	// cycles counts CONSECUTIVE over-tolerance reconcile cycles for this coin.
-	// It is the duration shown in operator messages and the base of the
-	// every-10th-cycle cadence.
+	// It is the duration shown in operator messages.
 	cycles int
 	// alerted marks that the confirmation window has been crossed and an alert
 	// has fired, so subsequent cycles re-throttle instead of re-confirming.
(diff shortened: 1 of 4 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-20T05:54:28Z) in 8fdca24.

### Resolved (Recommended Optional)
1. **The 1-minute log interval only halved the WARN volume.**
Confirmed against `shared_wallet_drift_alerts.go` — at the ~35s reconcile cadence the flat interval passed roughly every other cycle (~310 lines/6h, ~50% reduction). Replaced the plain interval throttle with a materially-changed log gate anchored to the last *logged* drift (reusing `sharedWalletDriftRealertRatio`) plus an hourly heartbeat aligned to the notification back-off. A stable drift now collapses to onset + alert cycles + hourly heartbeat; a worsening sub-threshold drift still logs on each material move, preserving the per-move visibility #1088 deliberately kept by throttling rather than gating the log behind `shouldNotify`. New `TestSharedWalletDriftTracker_WorseningDriftLogsWithinInterval` proves a worsening drift logs with `notify=false` (driven through a churning orphan set to hold the wallet in the confirmation window); `LogThrottledPerInterval` rewritten for the hourly heartbeat.

### Folded into this PR (was Create Follow-up Issue)
1. **The sibling `HLReconcileGapTracker` carried the same `%10`-preempts-hourly bug — fixed here rather than deferred.**
Per maintainer direction, folded in instead of filing a separate issue. Validated `hl_reconcile_gap_alerts.go`: `:116` (now removed) `case e.cycles%10 == 0:` sat ahead of the hourly case `:118` with `e.cycles` incrementing every cycle, re-DMing the operator ~every 6 min. It *also* had an unconditional per-cycle stdout `[WARN]` at `:216` — the identical log spam #1088 fixed in the drift reporter. Applied the symmetric fix: removed the `%10` case so the hourly back-off governs, and gated the stdout line behind a new `shouldLog` with the same onset / material-change / hourly-heartbeat throttle. `Record` now returns `(shouldNotify, shouldLog, cycles)`; all call sites migrated. New regressions `TestHLReconcileGapTracker_StableGapRealertsHourlyNotEveryTenth` and `TestHLReconcileGapTracker_LogThrottled`.

Full scheduler `go test ./...` green; build clean.

---

richkuo:
@claude review

richkuo:
@claude rereview

Issue #1104, opened 2026-06-24
[C60] Switch HL shared-wallet drift alarm onto the cash-flow journal


richkuo:
**Complexity: 60/100** — scope: small–medium (basis switch + fail-closed gate + env kill-switch + snapshot-bounded ingestion + baseline-offset retirement, plus tests); risk: high (this is the live money drift-alarm path); uncertainty: low (depends only on child 1 having baked clean on real shadow data).

Parent epic: #1100. Phase 2 of 4 (HL). Depends on #1100-child-1 (journal + shadow). **Do not switch the alarm until the shadow journal shows clean journal-vs-ledger agreement on live data.**

## Problem

With the HL cash-flow journal running in shadow (child 1), the drift alarm still fires off the trade-ledger basis, so modeled-fee / fallback-price / model-only force-close rows keep reading as residual drift. The journal is only valuable once it actually drives the alarm.

## Goal

Promote the journal from shadow to the HL shared-wallet drift-alarm basis: the alarm fires on `accountValue` vs the journal's expected-equity, with internal trade rows kept for per-strategy attribution/display only. OKX and TopStep stay entirely on their existing path (separate children).

## Approach / acceptance criteria

- Switch the HL drift basis to the journal when **usable**: baseline anchored, all three streams fetched this cycle, not `incomplete`. Otherwise **fall back to the trade-ledger drift**, preserving the drift streak (fail-closed).
- **Snapshot-bounded ingestion:** book journal events only up to the `accountValue`/uPnL snapshot instant; defer in-flight fills (settled between the balance snapshot and the journal fetch) to next cycle so they don't read as a full cycle of false drift during active trading. Existing 2-cycle confirmation absorbs indexer-lag transients.
- **Operator kill-switch** `GO_TRADER_CASHFLOW_JOURNAL_ALARM=0/off/false/no` forces the legacy trade-ledger basis with no redeploy (registered in `agent-info`).
- Log the journal-vs-ledger comparison every cycle so both bases stay visible after the switch.
- **Baseline-offset retirement (#1100 phase 6):** the journal's adoption baseline is migration-only (anchored once, never re-zeroed); `baseline_offset_usd` now only silences the fail-closed fallback, not the steady-state alarm.
- Tests: snapshot-cutoff deferral, usability gating (first contact / stream-miss / incomplete), basis switch (enabled+usable overrides; disabled/not-usable/nil keep trade-ledger; other wallets untouched), env kill-switch parser.

[...]

Issue #1103, opened 2026-06-24
[C65] HL shared-wallet cash-flow journal + shadow reconciliation


richkuo:
**Complexity: 65/100** — scope: medium (new `cashflow_journal` + `cashflow_journal_state` tables, a journal ingest module, reuse of the existing HL Go fetchers, tests); risk: medium (foundation of a money/data-integrity reconciliation path, but shadow-only so no live alarm/display/member-value change this phase); uncertainty: low–medium (HL `userFillsByTime` / `userFunding` / `userNonFundingLedgerUpdates` are already plumbed).

Parent epic: #1100. Phase 1 of 4 (HL). Must land and bake before #1100-child-2 (alarm switch).

## Problem

Hyperliquid shared-wallet drift reconstructs account equity from the bot's own trade ledger plus exchange uPnL (`member_value = initial_capital + Σ tradeLedgerDelta + owned_uPnL`), then compares that derived value to exchange `accountValue`. Every modeled fee, fallback fill price, and model-only force-close / kill-switch row becomes residual drift. The fix (per #1100) is to make the wallet **total** come from exchange events, with internal rows used only for per-strategy attribution.

## Goal

Build a durable, exchange-sourced cash-flow journal per HL shared wallet and run its expected-equity calculation **in shadow** beside the live trade-ledger drift path, logging the delta every cycle. Zero behavior change — no alarm, display, or member-value change. This de-risks the alarm switch (child 2) by proving the journal agrees with reality on live data first.

## Approach / acceptance criteria

- New `cashflow_journal` (per-event rows: platform, account, time_ms, kind, amount_usd, coin, closed_pnl_gross, fee_usd, UNIQUE dedup_id) + `cashflow_journal_state` (per-wallet cursors + adoption baseline) tables; idempotent `IF NOT EXISTS` migrations.
- Incremental three-stream fetch (fills + funding + transfers) reusing the existing Go fetchers; durable per-stream cursors/watermarks; per-event dedup (`INSERT OR IGNORE` on UNIQUE dedup_id); halt-cursor-on-persist-failure.
- Equity equation: `expected = baseline_accountValue + Σ(settled-cash deltas) + (uPnL_now − uPnL_baseline)`; fill settled delta = `closedPnl_gross − fee`.
[...]

Issue #1100, opened 2026-06-24
[C90][Epic][DONE] Event-source shared-wallet reconciliation from the exchange cash-flow feed (drift is structural, not a missing fee column)


richkuo:
> **✅ COMPLETED — tracking epic, all four phases shipped and merged.** The exchange-sourced cash-flow journal now backs HL, OKX, and TopStep shared-wallet reconciliation; the HL drift alarm keys off the journal expected-equity (`scheduler/cashflow_journal.go`, `scheduler/okx_cashflow_journal.go`, `scheduler/topstep_cashflow_journal.go`; alarm `formatSharedWalletJournalDriftAlert` in `scheduler/shared_wallet_drift_alerts.go`). Closing this epic. The detailed analysis below is retained as the design rationale.

## Sub-issues

- [x] #1103 — **HL cash-flow journal + shadow reconciliation** (phase 1). Build the exchange-sourced journal; run expected-equity in shadow beside the live trade-ledger drift; zero behavior change. *(PR #1101 — base of the stack; merged)*
- [x] #1104 — **Switch the HL drift alarm onto the journal** (phase 2). Promote journal → alarm basis with fail-closed fallback, env kill-switch, snapshot-bounded ingestion. Only after #1103 bakes clean on live data. *(PR #1107 — stacked on #1101; merged)*
- [x] #1105 — **Extend to OKX** (phase 3). Per-venue feed verification + reconciled-field choice (`totalEq` equity vs `availBal` cash). Independent of #1106. *(PR #1108 — merged)*
- [x] #1106 — **Extend to TopStep** (phase 4). Enable a TopStep shared-wallet snapshot first, then the journal. Independent of #1105. *(PR #1109 — merged)*

**Seam logic:** 1→2 is the shadow→promote boundary — never trust an alarm basis that hasn't been watched on live data. 3 and 4 are independent per-venue feed work that can land in parallel once the HL pattern (1+2) is proven. **HL behavior must stay untouched while OKX/TopStep land.**

---

**Complexity: 90/100** — scope: large (exchange-sourced journal + per-venue incremental sync + reconciliation rewrite + attribution decoupling across HL/OKX/TopStep); risk: high (money / data-integrity reconciliation path); uncertainty: medium (per-venue feed completeness, cursor semantics, OKX/TopStep cash-flow endpoints unverified).

## Correction to the original framing

[...]

PR #1108, opened 2026-06-24
feat(#1105): OKX shared-wallet cash-flow journal (shadow)

### LINKED ISSUE #1105: [C75] Extend cash-flow journal reconciliation to OKX shared wallets
**Complexity: 75/100** — scope: medium-high (NEW OKX cash-flow fetchers as Python/ccxt subprocesses, event-kind mapping, per-venue cursor semantics, probe argv, tests, two languages); risk: high (money/data-integrity reconciliation path on a second venue); uncertainty: high (OKX feed completeness, exact reconciled balance field, and cursor/pagination semantics are unverified — this is the gating work).

Parent epic: #1100. Phase 3 of 4. Independent of TopStep (#1100-child-4); both unblock once the HL journal pattern (children 1–2) is proven. **Leave HL behavior untouched.**

## Problem

The cash-flow journal currently reconciles Hyperliquid only (hard `key.Platform != "hyperliquid"` guard in `scheduler/cashflow_journal.go`). OKX shared wallets still use the capital-weight split (#918) and never moved to the trade-ledger basis (#954) — `scheduler/shared_wallet_reconcile.go:444-447` routes HL through the ledger path and OKX through `reconcileSharedWalletMemberValues` (capital-weight). So OKX's total reconciliation remains an APPROXIMATION (split semantics / orphan-uPnL), not exact reconciliation against real exchange money. #1100 step 5 explicitly gates OKX on per-venue feed verification before it can join exact reconciliation.

## Goal

Extend the exchange-sourced journal contract to OKX shared wallets, reusing the HL journal table/cursor/fail-closed machinery, after verifying OKX's feeds and reconciliation field.

- **Verify before building:** confirm OKX's fill, cash-flow (funding/transfer/deposit/withdrawal), and equity feeds, their pagination/cursor semantics, and the exact balance field the OKX fetcher reports today. **These feeds do not exist in the repo yet** — only `fetch_okx_balance.py` (equity) and `fetch_okx_positions.py` (uPnL) are wired; there is no OKX fills/funding/ledger fetcher anywhere.
- **The OKX cash-flow fetchers must be NEW Python/ccxt subprocess scripts** (`fetch_my_trades` for fills, `fetch_ledger`/bills for funding/transfers) wired through `executor.go` like the existing `RunOKXFetchPositions`/`RunOKXFetchBalance`, threaded into `probeArgv` + startup checks. **They cannot reuse HL's in-process Go fetchers** (`hyperliquid_fills.go` / `wallet_ledger.go` POST directly from Go); OKX integration is Python-subprocess/ccxt only. This widens the deploy surface — a Go/Python deploy at different SHAs fails the startup probe.
- **Name the reconciled field explicitly:** OKX reports `eq` (equity = cash + uPnL, via ccxt `balance["total"]["USDT"]` — `platforms/okx/adapter.py:239-247`), the same totalEq-style equity field HL's `accountValue` uses. Reconcile against `eq`, or deliberately switch OKX to a cash-only invariant — do NOT mix `totalEq`-style equity and `availBal`-style cash semantics. The equity equation must name the exact API field being reconciled.
- Reuse the journal table, durable cursors, per-event dedup, halt-on-persist-failure, snapshot-bounded ingestion, and the `incomplete`/fail-closed gate from children 1–2 (the DB layer is already keyed `(platform, account)`); add OKX-specific event-kind mapping and dedup IDs (OKX bill/ledger types → journal kinds).
- **Name the fail-closed fallback explicitly:** when the OKX journal is unusable, OKX fails closed to the **capital-weight split** (its only existing basis), NOT the trade-ledger basis — OKX has no trade-ledger basis to fall back to (unlike HL).
- Same gross-PnL discipline: venue-reported closed/realized PnL is attribution-only unless OKX's reconciliation contract explicitly defines it as the settled balance delta and tests prove the fee convention.
- Per-cycle journal-vs-existing-basis shadow log for OKX before flipping its alarm; same kill-switch coverage. **Recommended sequencing:** mirror HL's two-step — a shadow phase (OKX journal lands and runs in production, validating feeds, no alarm change; like #1103) before flipping the OKX alarm onto the journal (like #1104).
- Tests for the OKX equity equation, event-kind mapping, cursor advance/halt, and usability gating. HL and TopStep paths unchanged.

Updated: Opus 4.8 | high | Harness: validate-issue

richkuo:
## What & why

Phase 3a of #1100. Extends the exchange-sourced cash-flow journal from Hyperliquid to **OKX shared wallets in SHADOW mode**. The OKX wallet TOTAL is reconstructed from OKX's account-bills feed and reconciled against `eq`, logged next to the live capital-weight drift every cycle — but it **never drives the OKX alarm**, which stays on the #918 capital-weight split. Flipping the OKX alarm onto the journal is **Phase 3b**, gated on this shadow log proving the OKX feed/field in production first (mirrors HL's #1103 shadow → #1104 flip two-step). **HL and TopStep paths untouched.**

This scoping matches the issue's "per-cycle shadow log before flipping its alarm" criterion and keeps a money-path reconciliation on a second venue, with feed semantics that can't be verified offline, fully fail-closed until an operator validates it.

## Design

Reuses the platform-generic journal machinery built for HL (children 1–2): the `cashflow_journal` / `cashflow_journal_state` tables (keyed by `(platform, account)`), durable cursor, adoption baseline, per-event dedup, halt-on-persist-failure, snapshot-bounded ingestion, the `incomplete` fail-closed latch, and the shared equity equation. **No schema change.**

**Deliberate divergence from HL's three-stream shape.** HL reconstructs settled cash from three streams (fills + funding + transfers) and computes each fill's delta as gross-PnL minus the fee actually charged. OKX exposes a **single** settled-cash-flow source — the account-bills feed — where every balance change carries `balChg`, the signed settled-cash delta already netting trade PnL, fees, funding, transfers, deposits and withdrawals. So OKX needs no per-fill fee arithmetic: `balChg` IS the settled delta. Venue `pnl`/`fee` are attribution metadata only, never summed (same gross-PnL discipline as HL). Because OKX is single-stream, only the `FillsSinceMs` cursor is used (funding/transfers unused).

**Reconciled field is OKX `eq`** (ccxt `balance["total"]["USDT"]`, equity = cash + uPnL — the same field `fetch_okx_balance.py` reports and the #918 split already reconciles). The journal is **USDT-settlement only**: a non-USDT bill with non-zero `balChg`, or an unmapped bill `type`, latches the journal `incomplete` (fail closed) so the shadow log flags it for operator review before any Phase-3b flip. `balChg` is still booked for an unknown USDT type (authoritative) so the running drift surfaces it.

**Fail-closed fallback for OKX is the #918 capital-weight split** — OKX has no trade-ledger basis to fall back to (unlike HL). Unchanged this phase.

## Changes

- `platforms/okx/adapter.py`: `get_account_bills(since_ms)` — pages OKX bills via ccxt `fetch_ledger`, oldest-first, with a safety page cap (`capped` → not-usable that cycle); `_normalize_okx_bill` pure shaper.
- `shared_scripts/fetch_okx_bills.py`: thin ccxt-bills→JSON pump (`--since-ms`); all mapping/summing/fail-closed logic lives in tested Go.
- `scheduler/okx_cashflow_journal.go`: bill kind mapping, `balChg` settled-delta, dedup, fetch/ingest/reconcile, and the shadow logger (never mutates a `driftResult`).
- `scheduler/executor.go`: `RunOKXFetchBills` + `parseOKXBillsOutput` (same 5-case matrix as the other OKX fetch parsers).
- `scheduler/main.go`: per-cycle shadow call after the HL journal block, bounded to the eq/uPnL snapshot.

## Verification

- `go build` (binary, with ldflags), `go vet`, `gofmt -l` clean.
- `go test ./...` (full suite) and `go test -race` on the journal/executor: pass.
- New Go units: settled-delta convention (USDT/unknown-type/non-USDT), dedup, ingest cursor-advance + halt-on-persist-failure + snapshot cutoff, incomplete latch, non-USDT fail-closed, reconcile usability gating (clean/capped/fetch-error), shadow no-mutate, parser 5-case.
- Python (42 passed): adapter bills pagination / overlap-dedup / cap-truncation / no-forward-progress guard / paper-mode raise, normalizer shaping, and the fetcher-script JSON envelope on success + error paths.

## Notes / limits

- ccxt `fetch_ledger` reads OKX `/account/bills` (~7-day horizon). Steady per-cycle operation is well inside it; after a multi-day outage older bills fall outside the window and surface as journal drift in the shadow log (visible before any Phase-3b flip).
- The fetcher is a runtime subprocess (not a probed check script), so a Go/Python SHA skew degrades to a fail-closed fetch error (shadow logs it, OKX alarm unaffected) rather than a startup probe failure.

Closes #1105

LLM: Opus 4.8 | high | Harness: work-on-issue

### DIFF
--- a/platforms/okx/adapter.py
+++ b/platforms/okx/adapter.py
@@ -31,6 +31,23 @@ def _bill_float(value) -> float:
         return 0.0
 
 
+def _okx_usdt_cash_balance(info):
+    """Pull the USDT ``cashBal`` from a raw OKX fetch_balance ``info`` payload
+    (``data[0].details[]``). Returns a float, or None when the field is absent
+    or unparseable. Pure — unit-tested without a live exchange."""
+    try:
+        details = ((info or {}).get("data") or [{}])[0].get("details") or []
+    except (AttributeError, IndexError, TypeError):
+        return None
+    for d in details:
+        try:
+            if str(d.get("ccy") or "").upper() == "USDT":
+                return float(d.get("cashBal"))
+        except (TypeError, ValueError):
+            return None
+    return None
+
+
 def _normalize_okx_bill(entry: dict) -> dict:
     """Normalize one ccxt fetch_ledger entry into the #1105 journal bill shape.
 
@@ -296,6 +313,39 @@ class OKXExchangeAdapter:
         except (TypeError, ValueError):
             return 0.0
 
+    def get_account_equity_and_upnl(self) -> Tuple[float, float]:
+        """Return a COHERENT ``(equity, unrealized_pnl)`` snapshot from ONE
+        ``fetch_balance`` call, for the #1105 cash-flow journal.
+
+        ``equity`` is the USDT account value (ccxt ``total["USDT"]`` == OKX
+        ``eq`` — the same field ``get_account_balance`` returns and the #918
+        capital-weight split reconciles). ``unrealized_pnl`` is derived as
+        ``eq - cashBal`` for USDT from the SAME response, so eq and uPnL are one
+        atomic snapshot: in the journal's equity equation the uPnL term then
+        cancels exactly against ``eq``, leaving residual drift = settled-cash
+        reconstruction error only. Pairing eq with a uPnL sampled from a SEPARATE
+        ``fetch_positions`` call (a different instant) would inject intra-cycle
+        uPnL jitter — dollars on a leveraged position — into the shadow drift,
+        the very signal Phase 3b must evaluate.
+
+        Falls back to uPnL ``0.0`` when the USDT ``cashBal`` is absent/unparseable
+        (the shadow drift then conservatively carries uPnL — visible, never a
+        false "clean"). Only available in live mode; raises in paper mode.
+        """
+        if not self._is_live:
+            raise RuntimeError(
+                "get_account_equity_and_upnl requires live mode (set OKX_API_KEY, OKX_API_SECRET, OKX_PASSPHRASE)"
+            )
+        bal = self._exchange.fetch_balance()
+        total = bal.get("total") or {}
+        try:
+            eq = float(total.get("USDT") or 0.0)
+        except (TypeError, ValueError):
+            eq = 0.0
+        cash_bal = _okx_usdt_cash_balance(bal.get("info"))
+        upnl = (eq - cash_bal) if cash_bal is not None else 0.0
+        return eq, upnl
+
     def get_account_bills(self, since_ms: int = 0, page_limit: int = 100,
                           max_bills: int = 10000) -> Tuple[list, bool]:
         """Fetch OKX account bills (settled cash-flow events) since ``since_ms``
@@ -321,6 +371,13 @@ class OKXExchangeAdapter:
         older bills fall outside it and surface as journal drift in the shadow
         log (visible to the operator before any Phase-3b alarm flip).
 
+        Ordering assumption: this pages FORWARD by timestamp and relies on ccxt's
+        unified ``fetch_ledger`` returning entries ascending-from-``since`` (ccxt
+        ``parse_ledger`` sorts by timestamp). That contract — and that the
+        millisecond ``ts`` matches the live feed — is offline-unverifiable; confirm
+        the settled-Σ tracks eq to ~0 over a multi-cycle live window before any
+        Phase-3b alarm flip.
+
         Only available in live mode; raises RuntimeError in paper mode.
         """
         if not self._is_live:
@@ -330,30 +387,40 @@ class OKXExchangeAdapter:
         collected = {}
         cursor = int(since_ms or 0)
         capped = False
-        # Page FORWARD from the cursor: ccxt fetch_ledger returns unified entries
-        # ascending and `since` filters from that point; advance `since` past the
-        # newest entry each page until the feed is exhausted (short page) or the
-        # safety cap is hit. The page-loop bound is a belt-and-suspenders guard
-        # against a non-advancing cursor.
+        # Page FORWARD from the cursor, advancing the `since` watermark with a
+        # one-millisecond OVERLAP (cursor = page's last ts, NOT last_ts + 1) so a
+        # bill that shares the boundary millisecond but fell beyond the page cut
+        # is re-fetched on the next page — the `collected` dedup absorbs the
+        # re-read. Advancing past last_ts would permanently skip such a same-ms
+        # bill (a standing offset in the settled sum, not a transient miss). The
+        # page-loop bound caps total iterations.
         for _ in range(max(1, max_bills // max(1, page_limit)) + 2):
             page = self._exchange.fetch_ledger(code=None, since=cursor, limit=page_limit) or []
             if not page:
                 break
+            before = len(collected)
             for entry in page:
                 bill = _normalize_okx_bill(entry)
                 key = bill["bill_id"] or f"{bill['type']}:{bill['ts_ms']}:{bill['trade_id']}"
                 collected[key] = bill
+            added = len(collected) - before
             if len(collected) >= max_bills:
                 capped = True
                 break
             if len(page) < page_limit:
+                break  # short page → feed exhausted
+            page_last_ts = max((int(e.get("timestamp") or 0) for e in page), default=cursor)
+            if page_last_ts <= cursor and added == 0:
+                # A FULL page that neither advanced the timestamp nor yielded any
+                # new bill: a single-millisecond block larger than page_limit, which
+                # timestamp paging cannot step through without dropping bills. Fail
+                # closed — the Go journal treats `capped` as not-usable for the
+                # cycle rather than silently advancing past the undelivered tail.
+                capped = True
                 break
-            last_ts = max((int(e.get("timestamp") or 0) for e in page), default=cursor)
-            if last_ts <= cursor:
-                break  # no forward progress — stop rather than loop forever
-            cursor = last_ts + 1
+            cursor = page_last_ts  # overlap on the boundary ms (dedup absorbs it)
         bills = sorted(collected.values(), key=lambda b: b["ts_ms"])
-        if capped:
+        if len(bills) > max_bills:
             bills = bills[:max_bills]
         return bills, capped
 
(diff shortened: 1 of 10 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-24T08:55:33Z) in 90d67e8. Both findings were `Recommended Optional` (non-blocking); both confirmed against the code and implemented.

### Fixed
1. **Coherent OKX eq/uPnL snapshot (eliminates intra-cycle uPnL jitter).** The journal now sources eq and uPnL from a single `fetch_balance` read, with `uPnL = eq − cashBal` (`adapter.py` `get_account_equity_and_upnl` + `_okx_usdt_cash_balance`; `fetch_okx_balance.py` emits `unrealized_pnl`; `OKXBalanceResult.UnrealizedPnL`; `defaultOKXEquitySnapshot`; `main.go` journal call). Rather than approximate-coherent (pairing eq with the balance-`info` `upl` field), I derived `uPnL = eq − cashBal` because it makes the equity equation's uPnL term cancel against eq exactly — residual drift reduces to `cashBal − baseline_cashBal − Σ balChg` (pure settled-cash error), and the journal no longer needs the `fetch_positions` read at all. `balance` keeps its #360 meaning so the #918 split and kill-switch-clear paths are unchanged. Removed the now-dead `sumOKXAccountUPnL`.

2. **Same-ms bill pagination no longer skips a boundary-millisecond bill.** Paging advances with a one-ms overlap (`cursor = page's last ts`, not `last_ts + 1`; the `collected`/DB dedup absorbs the re-read), so a same-ms bill split across the page cut is re-fetched (`adapter.py:get_account_bills`). A full page that neither advances the timestamp nor yields a new bill (a single-ms block larger than `page_limit`) now fails closed via `capped=True` — the Go journal already treats `capped` as not-usable — rather than silently dropping the tail. This is stronger than the suggested overlap alone, which left the >`page_limit` same-ms case unhandled. The ccxt ascending-order assumption is documented as live-verify-before-3b (offline-unverifiable, as flagged).

### Note
While addressing these I found the adapter-level `get_account_bills` / `_normalize_okx_bill` tests had been dropped from the original commit (the file was at HEAD with zero bills tests). Re-added the full suite — updated for overlap paging — covering same-ms straddle (captured), same-ms-overflow (capped/fail-closed), `max_bills` cap, dedup, plus the new `_okx_usdt_cash_balance` / `get_account_equity_and_upnl` coverage. Full Go suite + `-race`, vet, gofmt, py_compile, and 53 OKX Python tests pass.

---
Created with LLM: Opus 4.8 | high | Harness: fix-pr-review

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-24T09:17:09Z) in 62a0a05.

### Fixed
1. **Capped bills fetch no longer advances the journal cursor past truncated same-ms bills.** Confirmed: `ingestOKXCashflowJournalEvents` advanced `FillsSinceMs` to `maxTime+1` unconditionally — `res.Capped` only gated `Usable`, not the cursor (`scheduler/okx_cashflow_journal.go:283`). Both cap paths can split a same-ms group at `maxTime` (single-ms-overflow returns the first `page_limit` of a `>page_limit` block; `max_bills` `bills[:N]` can cut mid-millisecond), so the kept siblings were booked while the truncated ones sat behind an advanced watermark — a standing offset in `SumCashflowJournal` that would survive into Phase 3b. Fix: when `res.Capped`, advance only to `maxTime` (pass `maxTime-1` to `advanceCashflowCursor`) so the boundary ms is re-read next cycle; the `okxBillDedupID` / `INSERT OR IGNORE` dedup absorbs the re-read, and a genuinely un-pageable `>page_limit` same-ms block pins fail-closed at that ms (Usable stays false) rather than corrupting the sum. `advanceCashflowCursor` still clamps to the current watermark, so it never moves backwards.

Added Go coverage for all three must-survive cases: `TestOKXCashflowJournalCappedAdvancesOnlyToMaxTime` (non-capped → `maxTime+1`; capped → `maxTime`) and `TestOKXCashflowJournalCappedSameMsSiblingNotStranded` (a capped all-same-ms cycle keeps the cursor on that ms; a later cleared cycle re-reads it, books the previously-truncated sibling, and does not double-count the deduped re-reads). Full Go suite + `-race`, vet, gofmt, and 53 OKX Python tests pass.

Note: the `max_bills` truncation keeps the OLDEST `bills[:N]` and drops the newer tail, so `maxTime` is the kept boundary and everything dropped is `≥ maxTime` — re-reading from `maxTime` next cycle recovers the full dropped tail, not just same-ms siblings.

---
Created with LLM: Opus 4.8 | high | Harness: fix-pr-review

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-24T09:35:31Z) in f1e51ed.

### Fixed
1. **`get_account_bills` no longer returns `capped=False` on an incomplete fetch when the page-loop budget is exhausted.** Confirmed: the loop bound `max_bills // page_limit + 2` was sized for non-overlap paging, but the one-ms overlap re-fetch (commit 90d67e8) makes net-new-per-page < `page_limit` under same-ms clustering, so a dense backlog can exhaust the budget before the feed drains and before the `max_bills` cap trips — the loop fell through with `capped=False`, and the Go journal advanced the cursor to `maxTime+1`, stranding the tail (`platforms/okx/adapter.py:397`). Fix: a `for…else` clause sets `capped=True` when the loop completes without a feed-drained/cap `break` (`adapter.py` `get_account_bills`). Every drained exit (empty/short page) and cap exit already `break`, so `else` runs only on genuine budget exhaustion; the Go side then re-reads forward from `maxTime` across cycles and the backlog drains incrementally instead of corrupting `SumCashflowJournal`.

Tests for all three must-survive cases: `test_loop_budget_exhaustion_reports_capped` (400-bill / 2-per-ms backlog, budget 12 → `capped=True`, incomplete prefix), `test_drained_distinct_ts_not_spuriously_capped` (drained distinct-ts feed exits on a short page → `capped=False`, no spurious cap), and Go `TestOKXCashflowJournalCappedNewerTailNotStranded` (a capped prefix with a newer-ms tail keeps the cursor at `maxTime`; a later cleared cycle re-reads and books the tail without double-counting). Full Go suite + `-race`, vet, gofmt, py_compile, and 55 OKX Python tests pass.

This is the same "never advance past an unfetched bill" invariant from the prior two reviews, now closed on the loop-budget path: `capped=True` ⟺ the returned set is not the complete contiguous feed from the cursor.

---
Created with LLM: Opus 4.8 | high | Harness: fix-pr-review

richkuo:
@claude review

PR #1109, opened 2026-06-24
feat(#1106): TopStep shared-wallet cash-flow journal (shadow)

### LINKED ISSUE #1106: [C75] Extend cash-flow journal reconciliation to TopStep shared wallets
**Complexity: 75/100** — scope: large (TopStep has no shared-wallet balance fetcher enabled yet, so the equity-snapshot side must be built before the journal can reconcile, plus the journal ingest, cursors, tests); risk: high (money/data-integrity reconciliation path on a venue with no current shared-wallet reconciliation); uncertainty: high (TopStep fill/cash-flow/equity feed availability and semantics unverified).

Parent epic: #1100. Phase 4 of 4. Independent of OKX (#1100-child-3); both unblock once the HL journal pattern (children 1–2) is proven. **Leave HL behavior untouched.**

## Problem

TopStep is recognized as a shared-wallet candidate but has **no shared-wallet balance fetcher enabled** (`scheduler/shared_wallet.go`, `scheduler/shared_wallet_reconcile.go`), so it does not participate in shared-wallet reconciliation at all today. #1100 step 5 gates TopStep on verifying its feeds before it can join the exact-reconciliation journal.

## Goal

Bring TopStep into exchange-sourced journal reconciliation: first enable a TopStep shared-wallet equity/balance snapshot, then extend the journal contract to its fill/cash-flow feeds, reusing the HL machinery.

- **Verify + enable the snapshot side first:** confirm TopStep exposes a shared-wallet equity/balance endpoint and wire a fetcher (prerequisite the other venues already had).
- **Verify the event feeds:** confirm TopStep's fill, cash-flow, and (if applicable) funding/transfer feeds and their cursor/pagination semantics.
- **Name the reconciled field explicitly:** equity vs settled-cash — the equity equation must name the exact field; do not mix equity and cash semantics.
- Reuse the journal table, durable cursors, per-event dedup, halt-on-persist-failure, snapshot-bounded ingestion, and the `incomplete`/fail-closed gate from children 1–2; add TopStep-specific event-kind mapping.
- Same gross-PnL discipline; venue-reported realized PnL is attribution-only unless the reconciliation contract defines it as the settled delta and tests prove the fee convention.
- Per-cycle journal-vs-existing-basis shadow log before flipping any TopStep alarm; same kill-switch coverage.
- Tests for the TopStep snapshot fetcher, equity equation, event-kind mapping, cursor advance/halt, and usability gating. HL and OKX paths unchanged.

richkuo:
Phase 4 of #1100. Brings TopStep into the exchange-sourced cash-flow journal as a **SHADOW** reconciler, reusing the HL (#1103/#1104) and OKX (#1105) machinery. **HL and OKX paths are untouched.**

## What this does

- **Enables the snapshot side first** (the issue's gating prerequisite): registers `topstep` in `platformsWithSharedWalletBalanceFetcher` so `detectSharedWallets` surfaces TopStep shared wallets. The coherent `(equity, uPnL)` snapshot is owned by a dedicated main-loop block (`defaultTopStepEquitySnapshot`), mirroring OKX — `defaultSharedWalletFetcher` intentionally has no topstep case.
- **No display mutation.** TopStep stays display-skipped in `reconcileSharedWalletDisplayValues` (no position source → `default: continue`), so enabling the fetcher does not activate any capital-weight split. A balance-fetch failure leaves `walletBalances[tsKey]` unset, so `computeTotalPortfolioValue` falls back to the member-PV sum — no false-zero equity that could trip a kill switch.
- **Single fill stream, HL-like settled-cash convention.** Each fill's settled-cash delta is GROSS realized PnL − commission (`cashflowFillSettledDelta`); the gross value is retained for attribution only and is never summed into equity (same #698 discipline). Reconciled field is named explicitly: **TopStep account EQUITY** (settled USD cash + unrealized PnL).
- **Fail-closed classification.** An unmapped fill `kind` books its authoritative gross−fee delta but latches the journal `incomplete` so a future Phase-4b alarm flip fails closed.
- Reuses the durable cursor, per-event dedup, halt-on-persist-failure, snapshot-bounded ingest, and the shared equity equation.

## ⚠️ Endpoint contracts are unverified (read before promoting)

TopStep has no balance/fills feed today, and the existing live adapter targets a speculative `X-API-Key` / `GET /v1/account/*` shape that does **not** match the real TopStepX/ProjectX gateway. The new `get_account_equity_and_upnl` / `get_account_fills` methods and `fetch_topstep_{balance,fills}.py` mirror that existing (unverified) contract. **These endpoint shapes MUST be confirmed against the live venue before any Phase-4b alarm flip.** This is safe to land now because the journal is **shadow-only** (logs, never drives an alarm) and inert unless 2+ live TopStep strategies share one account — which is not the case today.

## Verification

- `go build ./scheduler` ✅ · full `go test ./scheduler/...` ✅
- `platforms/` pytest: 318 passed; new topstep adapter feed tests: 21 passed ✅
- `py_compile` of both fetchers ✅; both emit a clean error envelope + exit 1 when not live ✅
- New Go journal unit tests: settled-delta, dedup, baseline anchor, ingest+expected-equity, snapshot bound, capped-cursor discipline, unclassified→incomplete, persist-failure halt, reconcile usability, non-key reject, shadow no-mutate, parsers. Existing `detect-shared-wallets` and fetcher-map tests updated for the new fetcher.

Closes #1106

---

### DIFF
--- a/platforms/topstep/adapter.py
+++ b/platforms/topstep/adapter.py
@@ -207,6 +207,79 @@ class TopStepExchangeAdapter:
             })
         return positions
 
+    def get_account_equity_and_upnl(self):
+        """Return ``(equity, unrealized_pnl)`` for the configured account, in USD.
+
+        Used by the shared-wallet cash-flow journal (#1106). ``equity`` is the
+        account value (settled cash + unrealized PnL); ``unrealized_pnl`` is
+        ``equity − cashBalance`` from the SAME balance read so the journal
+        reconciles a coherent equity/uPnL snapshot with no intra-cycle jitter
+        (mirrors OKX ``get_account_equity_and_upnl``). Live mode only.
+
+        NOTE: the ``/v1/account/balance`` shape mirrors this adapter's existing
+        (unverified) TopStepX REST contract; confirm against the live venue
+        before any Phase-4b alarm flip (#1106).
+        """
+        if not self.is_live:
+            raise RuntimeError("get_account_equity_and_upnl requires live mode")
+        resp = self._session.get(
+            f"{API_BASE_URL}/v1/account/balance",
+            params={"accountId": self._account_id},
+            timeout=10,
+        )
+        resp.raise_for_status()
+        data = resp.json()
+        equity = float(data.get("equity", 0))
+        # cashBalance is the settled cash; fall back to "balance" then to equity
+        # (uPnL 0) so a feed without an explicit cash field degrades to cash==equity
+        # rather than mis-stating uPnL.
+        cash = float(data.get("cashBalance", data.get("balance", equity)))
+        return equity, equity - cash
+
+    def get_account_fills(self, since_ms=0, limit=1000):
+        """Return ``(fills, capped)`` — settled trade fills at or after
+        ``since_ms``, oldest first (#1106).
+
+        Each fill: ``{"fill_id", "ts_ms", "symbol", "kind", "realized_pnl",
+        "fee"}``. ``realized_pnl`` is GROSS (0 for entry fills); ``fee`` is the
+        commission charged. ``capped`` is true when the page limit was hit (the
+        Go journal treats that cycle as not-usable). Live mode only.
+
+        NOTE: the ``/v1/account/fills`` shape mirrors this adapter's existing
+        (unverified) TopStepX REST contract; confirm against the live venue
+        before any Phase-4b alarm flip (#1106).
+        """
+        if not self.is_live:
+            raise RuntimeError("get_account_fills requires live mode")
+        resp = self._session.get(
+            f"{API_BASE_URL}/v1/account/fills",
+            params={
+                "accountId": self._account_id,
+                "sinceMs": int(since_ms),
+                "limit": int(limit),
+            },
+            timeout=15,
+        )
+        resp.raise_for_status()
+        raw = resp.json().get("fills", []) or []
+        fills = []
+        for f in raw:
+            try:
+                ts_ms = int(f.get("timestamp", f.get("ts_ms", 0)))
+            except (TypeError, ValueError):
+                continue
+            fills.append({
+                "fill_id": str(f.get("id", f.get("fill_id", ""))),
+                "ts_ms": ts_ms,
+                "symbol": (f.get("symbol") or "").upper(),
+                "kind": (f.get("kind") or "").lower(),
+                "realized_pnl": float(f.get("realizedPnl", f.get("realized_pnl", 0)) or 0),
+                "fee": float(f.get("fee", f.get("commission", 0)) or 0),
+            })
+        fills.sort(key=lambda x: x["ts_ms"])
+        capped = len(raw) >= int(limit)
+        return fills, capped
+
     # ─────────────────────────────────────────────
     # Yahoo Finance helpers (paper mode)
     # ─────────────────────────────────────────────
--- a/platforms/topstep/adapter.py
+++ b/platforms/topstep/adapter.py
@@ -38,6 +38,25 @@ CONTRACT_SPECS = {
 }
 
 
+def _normalize_topstep_fill(f):
+    """Shape one raw TopStep fill into the cash-flow journal record, or None if
+    its timestamp is unparseable (#1106). ``realized_pnl`` is GROSS; ``fee`` is
+    the commission. Tolerant of both camelCase venue keys and the snake_case
+    journal keys."""
+    try:
+        ts_ms = int(f.get("timestamp", f.get("ts_ms", 0)))
+    except (TypeError, ValueError):
+        return None
+    return {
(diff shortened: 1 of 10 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-24T10:30:27Z) in 3160910.

### Fixed
1. **Silent-zero TopStep equity into the live kill-switch path** — confirmed: a `200` with a shape-mismatched body coerced `equity` to `0`, returned success, set `walletBalances[tsKey] = 0`, and `computeSubsetPortfolioValue` (`scheduler/shared_wallet.go:257`) consumed the *present* zero with no member-PV fallback or peak freeze. Fixed at the source so every consumer is covered: `defaultTopStepEquitySnapshot` now routes through `validatedTopStepEquity` (`scheduler/topstep_cashflow_journal.go:170`), which treats `equity <= 0` or NaN as a fetch MISS (returns an error) → `walletBalances[tsKey]` stays UNSET → member-PV fallback + peak freeze. The Python `get_account_equity_and_upnl` also raises when the `equity` field is absent (`platforms/topstep/adapter.py`) so the error envelope carries a clear cause instead of a silent `0`. Must-survive (a) missing `equity` → error; (b) present non-positive → miss; (c) good positive → unchanged. New `TestValidatedTopStepEquity` + `test_equity_and_upnl_missing_equity_raises`.

### Fixed (recommended optional)
1. **Fills `capped` detection under a sub-limit venue page cap** — confirmed the naive `len(raw) >= limit` could mark a truncated backlog `Usable`. `get_account_fills` now pages FORWARD by the `sinceMs` cursor with a one-ms overlap + `fill_id` dedup until the feed is DRAINED (a short page), mirroring `OKXExchangeAdapter.get_account_bills`. `capped` is True only when the feed is NOT proven drained (`max_fills` cap, a single-ms block larger than `page_limit`, or iteration-budget exhaustion via `for…else`). Must-survive (a) short page with more server-side → not-usable; (b) full page then drained → forward progress, deduped; (c) zero fills → usable, drift ~0. New `test_get_account_fills_drains_across_pages` + `test_get_account_fills_single_ms_overflow_caps`.

Verification: `go build` + full `go test ./scheduler/...` pass; topstep adapter pytest 23 passed; `py_compile` clean. HL and OKX paths untouched.

---

richkuo:
@claude review

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-24T19:50:52Z) in 76b260f.

### Fixed
1. **Unverified TopStep equity driving the all-platform kill switch on the wrong-but-positive path** — confirmed and fixed at the root. The prior commit's `"topstep": true` in `platformsWithSharedWalletBalanceFetcher` (`scheduler/shared_wallet.go:96`) is what pulled TopStep into `detectSharedWallets` → `computeTotalPortfolioValue` (`main.go:1041`) → `CheckPortfolioRisk` (`main.go:1059`), so a wrong-but-positive `/v1/account/balance` equity could crater `totalPV` behind only the `>0` guard. Reverted that entry, restoring byte-for-byte pre-PR per-strategy member-PV portfolio-risk behavior. The shadow journal now detects its own grouping via `detectTopStepSharedWallet` (`shared_wallet.go`), independent of the kill-switch `sharedWallets` map, and the coherent equity/uPnL snapshot is fed **directly** to `reconcileTopStepCashflowJournal` (`main.go:1158`) — never written into `walletBalances`. TopStep equity is fully off the live portfolio/kill-switch path until Phase 4b verifies the feed.

This also covers all three must-survive cases: (a) wrong-but-positive equity *below* true value cannot lower `totalPV` (TopStep no longer contributes to it); (b) inverse — equity *above* true value cannot ratchet the portfolio peak (it never reaches `CheckPortfolioRisk`); (c) the #243 double-count avoidance is preserved unchanged for HL/OKX, and TopStep reverts to the pre-PR per-strategy sum.

I chose the reviewer's first option (gate TopStep out of the live path) over sanity-bounding against the member-PV sum: bounding still admits an unverified feed onto a money/auto-protective path during a shadow phase, and adds a tolerance knob with its own failure modes. Gating it out entirely is strictly safer and is the correct shape for "shadow" — the journal logs, nothing live moves. As a bonus it removes a latent regression the bounding option would not have: leaving `walletBalances[tsKey]` unset while TopStep stayed in `sharedWallets` would have set `usedFallback` every cycle, freezing the whole-portfolio peak ratchet (`shared_wallet.go:261`). Override if you'd prefer the sanity-bounded variant instead.

Verification: `go build` + full `go test ./scheduler/...` pass (30.2s); new `TestDetectTopStepSharedWallet` covers 2+-live-shared / single-not / paper-not and confirms TopStep stays excluded from `detectSharedWallets` simultaneously. HL/OKX paths untouched.

---

richkuo:
@claude review

Issue #1110, opened 2026-06-25
[C50] Alert on trailing TP ratchet tier trigger


richkuo:
**Complexity: 50/100** — scope: medium (ratchet helper, perps/manual call sites, config, alert tests); risk: medium (auto-protective stop-loss path); uncertainty: low-to-medium (delivery boundary and SL-trigger semantics need care).

### Summary

When a `trailing_tp_ratchet` / `trailing_tp_ratchet_regime` tier clears and tightens the trailing stop, send a proactive operator alert with the ratchet details.

### Current Behavior

The ratchet state change is durable but not proactively surfaced. When a mark clears a ratchet tier, `applyTrailingTPRatchetToPosition` updates `pos.PostTPTrailingATRMult` and `pos.SLAdjustedTiersProcessed`, and logs the tighten event. The operator can later see the state through persisted DB fields (`post_tp_trailing_atr_mult`, `sl_adjusted_tiers_processed`), the exchange SL order after the trailing walker replaces it, or Discord position summaries (`Ratchet: N/M | Trail: ...`), but there is no immediate Discord/Telegram notification when a tier clears.

### Proposed Alert Payload

When a ratchet tier clears, send an alert containing:

- **Strategy ID** and **symbol**
- **Side** (long/short)
- **Tier index cleared** (for example, `Tier 1 of 3`)
- **ATR multiple triggered** (for example, `1.5x ATR`)
- **Price at trigger** (mark price that cleared the tier)
- **Entry/risk-anchor price** and **entry ATR**
- **Profit at trigger** (ATR multiples and USD)
- **Previous trail mult** -> **new trail mult** (for example, `2.0x -> 1.5x ATR`)
- **Intended trailing SL trigger price** computed from the current high-water mark and new trail multiple; label it as intended/computed unless the later trailing-stop update confirms the exchange replacement
- **High-water mark** used for the computed trigger
- **Next ratchet tier**, when present (ATR multiple, trail-after value, and trigger price)
- **Regime label** used to resolve `trailing_tp_ratchet_regime` tiers
- **Position regime stamped at open**, when distinct/relevant

### Example Alert

```text
[hl-rmc-eth-live] ETH long — Ratchet Tier 1/3 cleared
  Triggered at: 1.5x ATR ($1,611.98) | Mark: $1,619.25
  Entry: $1,579.50 | ATR: $21.65 | Profit: 1.83x ATR (+$34.85)
  Trail tightened: 2.0x -> 1.5x ATR
[...]

PR #1113, opened 2026-06-25
fix(#1111): use IsConfigured for pre-resolution regime-stop checks

### LINKED ISSUE #1111: [C40] trailing_stop_min_move_pct rejects trailing_stop_atr_regime (IsZero-before-resolution); same bug skips the stop_loss↔trailing regime mutex
**Complexity: 40/100** — scope: small-medium (two predicate sites across `config.go` + `regime_atr.go`, plus tests); risk: medium (startup AND SIGHUP-reload config-load path; one instance is an auto-protective stop-config mutex that silently fails open); uncertainty: low (root cause confirmed against current `main`, both instances traced).

### Summary

Several config-load-time validators read a regime ATR block (`trailing_stop_atr_regime` / `stop_loss_atr_regime`) via `IsZero()` **before** the block has been resolved, so they misjudge a block the operator did configure. Two confirmed instances of one root cause:

1. **False reject** — `trailing_stop_min_move_pct` is rejected when the strategy uses `trailing_stop_atr_regime` (the originally reported symptom).
2. **False accept** — the `stop_loss_atr_regime` ⟷ `trailing_stop_atr_regime` mutual-exclusivity check never fires, so a config setting **both** is silently accepted.

### Root cause (shared)

`RegimeATRBlock.UnmarshalJSON` (`scheduler/regime_atr.go:87`) captures **only** raw JSON into `b.raw`; it does not populate the typed `TrendRegime` / `UseDefaults` fields. Those are filled only by `ResolveSurface` / `ResolveSurfaceWithLabels` (`scheduler/regime_atr.go:124`), which runs inside `validateRegimeATRConfig` — invoked **last** in `validateConfig` (`scheduler/config.go:2000`). `IsZero()` (`scheduler/regime_atr.go:182`) returns `!UseDefaults && len(TrendRegime)==0`, which is `true` for an unresolved-but-raw-populated block; its own doc comment says callers running before `ResolveSurface` MUST use `IsConfigured()` (`scheduler/regime_atr.go:192`). Any validator reading a regime block via `!IsZero()` *before* line 2000 therefore treats a configured block as absent.

Reachable on both config-load paths: startup (`scheduler/main.go:102`) and SIGHUP hot-reload (`scheduler/main.go:527`) — both call `LoadConfig` → `validateConfig`.

### Instance 1 — `trailing_stop_min_move_pct` false reject

Error:
```
strategy[hl-rmc-eth-live]: trailing_stop_min_move_pct requires trailing_stop_pct > 0, trailing_stop_atr_mult > 0, or trailing_stop_atr_regime
```

The min-move check at `scheduler/config.go:1845` computes:
```go
regimeTrail := sc.TrailingStopATRRegime != nil && !sc.TrailingStopATRRegime.IsZero()
```
At line 1845 the block is still unresolved, so `IsZero()` is `true`, `regimeTrail` is `false`, and the check wrongly rejects.

Reproduction:
```json
{
  "id": "hl-rmc-eth-live",
  "type": "perps",
  "trailing_stop_atr_regime": {
    "trend_regime": {
      "trending_up": {"atr_multiple": 2.0},
      "trending_down": {"atr_multiple": 2.0},
      "ranging": {"atr_multiple": 2.0}
    }
  },
  "trailing_stop_min_move_pct": 0.1
}
```
Result: validation fails, service won't start (and SIGHUP reload of the same edit is rejected).

### Instance 2 — `stop_loss_atr_regime` ⟷ `trailing_stop_atr_regime` mutex false accept

The two regime stop blocks are mutually exclusive, but the **only** enforcement is `scheduler/regime_atr.go:806`, gated by `scheduler/regime_atr.go:805`:
```go
if sc.TrailingStopATRRegime != nil && !sc.TrailingStopATRRegime.IsZero() {
    errs = append(errs, "... stop_loss_atr_regime is mutually exclusive with trailing_stop_atr_regime")
}
```
This check sits inside the `stop_loss_atr_regime` branch, which resolves *its own* block at `regime_atr.go:783` but reads the **trailing** block before *it* is resolved (resolution at `regime_atr.go:814`). So `IsZero()` is `true` and the mutex error is skipped. The trailing branch (`regime_atr.go:818` onward) has no symmetric `stop_loss_atr_regime` check, and there is no count-based fallback elsewhere — so a config with both is accepted. (The sibling mutex checks at `regime_atr.go:791-803` are unaffected: they compare against *scalar* `!= nil` fields, which need no resolution; only the regime-vs-regime check at 805 does.)

Reproduction: a strategy with both `stop_loss_atr_regime` and `trailing_stop_atr_regime` blocks loads without error.

Severity: low at runtime (the stop-owner priority order picks one, no double-arm), but it defeats the "pick one" guard the validator is meant to enforce — an auto-protective / stop-config-integrity gap.

## Fix

Swap both pre-resolution sites to the raw-aware, nil-safe predicate `IsConfigured()` (`scheduler/regime_atr.go:192`) — already the convention in the defaults loop (`scheduler/config.go:971`, `:1012`):

- `scheduler/config.go:1845`: `regimeTrail := sc.TrailingStopATRRegime.IsConfigured()`
- `scheduler/regime_atr.go:805`: `if sc.TrailingStopATRRegime.IsConfigured() {`

`IsConfigured()` returns `true` when `b.raw` is non-empty, so it recognizes an operator-supplied block before resolution; it nil-guards internally (the explicit `!= nil` is no longer needed).

**Class-killer alternative (optional, higher risk):** hoist a single regime-surface resolution pass to the top of `validateConfig` so every downstream check sees resolved blocks and `IsZero()` is valid everywhere. Riskier — `ResolveSurface` mutates blocks in place and emits parse errors, so resolution and error-surfacing must not double-run. The predicate swap is the minimal, convention-matching fix and is recommended; the hoist is the durable way to prevent the next instance.

**Scope note:** these two are the only config-load-time (pre-resolution) `!IsZero()` reads on a regime block. `config.go:671` and `:1110` also read `!IsZero()` but live inside `EffectiveStopLossPct` and the (currently unreferenced) `hasHyperliquidStopLossOwnership`; their only live callers are runtime/post-resolution (`main.go:3307`, `scale_in.go:163`). All other `!IsZero()` sites (`hyperliquid_trailing_stop.go`, `manual_sl.go`, `scale_in.go`, `trailing_tp_ratchet.go`, `trade_protection_snapshot.go`, `inspect_cmd.go`, `regime_unified.go`) run at runtime, after resolution, where `IsZero()` is correct.

## Tests

Add `LoadConfig`-level tests (full JSON round-trip, not hand-built structs) for both instances. Existing tests cover scalar min-move and regime parsing **separately**, not the combined load path (`scheduler/hyperliquid_stop_loss_test.go:964`, `scheduler/regime_atr_composite_test.go:107`).

Must survive (test the class, not just the examples):
1. **Instance 1 accept** — `perps` strategy with `trailing_stop_atr_regime` + `trailing_stop_min_move_pct`, no scalar trailing field, loads without error.
2. **Instance 1 inverse** — neither scalar trailing nor `trailing_stop_atr_regime` set → `trailing_stop_min_move_pct` still rejected (the validator must not become a no-op).
3. **Instance 2 reject** — strategy with both `stop_loss_atr_regime` and `trailing_stop_atr_regime` → rejected as mutually exclusive.
4. **`use_defaults` form** — `trailing_stop_atr_regime: {"use_defaults": true}` (raw present, `TrendRegime` empty pre-resolution) is accepted for instance 1 and still triggers the instance-2 mutex.
5. **Reload path** — instances 1 and 2 behave identically through the SIGHUP `LoadConfig` path (`main.go:527`), not only first startup.

---
LLM: Opus 4.8 | xhigh | Harness: Claude Code

richkuo:
## Summary

Fixes a config-load-time bug class: two validators read a regime ATR block (`trailing_stop_atr_regime` / `stop_loss_atr_regime`) via `!IsZero()` **before** `validateRegimeATRConfig` resolves it. `RegimeATRBlock.UnmarshalJSON` captures only raw JSON, and `IsZero()` returns `true` on an unresolved-but-configured block, so each read treated a configured block as absent.

Two confirmed instances of one root cause:

1. **False reject** (the reported symptom) — `trailing_stop_min_move_pct` was rejected when the strategy used `trailing_stop_atr_regime`, blocking the service on startup (`main.go`) and on SIGHUP reload.
2. **False accept** — the `stop_loss_atr_regime` ⟷ `trailing_stop_atr_regime` mutex (whose sole enforcement lives in `regime_atr.go`) silently never fired when both were set, because it read the trailing block before that block was resolved.

## Fix

Swap both pre-resolution sites to the raw-aware, nil-safe `IsConfigured()` — already the convention in the defaults loop (`config.go:983/1024`):

- `scheduler/config.go` — min-move trailing-mode check
- `scheduler/regime_atr.go` — the regime-stop mutex gate

`IsConfigured()` is purpose-built for pre-resolution checks (true when `raw` is non-empty; nil-guards internally). All other `!IsZero()` sites run at runtime, after resolution, and are unaffected.

## Tests

New `scheduler/regime_iszero_resolution_test.go`, constructing blocks raw-only to reproduce the production pre-resolution state (a resolved struct never exercises the bug — which is why it shipped):

1. instance 1 accept — explicit `trend_regime` and `use_defaults` shapes
1. instance 1 inverse — no trailing mode still rejected (fix is not a no-op)
1. instance 2 — both regime stops rejected as mutually exclusive
1. instance 2 control — a single regime stop does not trip the mutex

Verified the new tests fail without the fix and pass with it; full `go test ./...` for `scheduler/` passes.

Note: the issue body's line numbers were against the pre-#1110 commit; on current `main` the sites are `config.go:1857` and `regime_atr.go:805`.

Closes #1111

---
LLM: Opus 4.8 | xhigh | Harness: Claude Code

### DIFF
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -1100,31 +1100,6 @@ func loadConfig(path string, skipLiveCredentialChecks bool) (*Config, error) {
 func normalizeHyperliquidPeerStopLosses(strategies []StrategyConfig) {
 }
 
-// hasHyperliquidStopLossOwnership reports whether sc would place a reduce-only
-// HL trigger at any point in its lifecycle. It is the peer-conflict predicate
-// rather than the runtime price-% predicate: TrailingStopATRMult arms an
-// initial trigger only on the cycle after the position opens (EntryATR must be
-// stamped first), so EffectiveStopLossPct returns 0 at order-placement time.
-// Peer validation must still treat that strategy as the trigger owner.
-func hasHyperliquidStopLossOwnership(sc StrategyConfig) bool {
-	if EffectiveStopLossPct(sc) > 0 {
-		return true
-	}
-	if sc.TrailingStopATRMult != nil && *sc.TrailingStopATRMult > 0 {
-		return true
-	}
-	if sc.StopLossATRMult != nil && *sc.StopLossATRMult > 0 {
-		return true
-	}
-	if sc.StopLossATRRegime != nil && !sc.StopLossATRRegime.IsZero() {
-		return true
-	}
-	if sc.TrailingStopATRRegime != nil && !sc.TrailingStopATRRegime.IsZero() {
-		return true
-	}
-	return false
-}
-
 // hyperliquidPeerStrategyErrors returns validation messages for HL wallet
 // strategies that share a coin but disagree on MarginMode or exchange Leverage (#491/#619).
 // Returns an empty slice when no peer conflicts exist.
--- a/scheduler/config.go
+++ b/scheduler/config.go
@@ -1854,7 +1854,11 @@ func validateConfig(cfg *Config, skipLiveCredentialChecks bool) error {
 			}
 			// #870: the regime ratchet owns its trail via trailing_stop_atr_regime
 			// rather than the scalar trailing_stop_atr_mult, so accept that too.
-			regimeTrail := sc.TrailingStopATRRegime != nil && !sc.TrailingStopATRRegime.IsZero()
+			// #1111: use IsConfigured (raw-aware), NOT !IsZero() — this check runs
+			// before validateRegimeATRConfig resolves the raw block, and IsZero()
+			// reports true on an unresolved-but-configured block (see its doc), so
+			// !IsZero() would wrongly reject a strategy that did set the regime trail.
+			regimeTrail := sc.TrailingStopATRRegime.IsConfigured()
 			if fixedTrailingPct <= 0 && atrMult <= 0 && !regimeTrail {
 				errs = append(errs, fmt.Sprintf("%s: trailing_stop_min_move_pct requires trailing_stop_pct > 0, trailing_stop_atr_mult > 0, or trailing_stop_atr_regime", prefix))
 			}
--- a/scheduler/regime_atr.go
+++ b/scheduler/regime_atr.go
@@ -802,7 +802,12 @@ func validateRegimeATRConfig(cfg *Config) []string {
 				if sc.TrailingStopATRMult != nil {
 					errs = append(errs, fmt.Sprintf("%s: stop_loss_atr_regime is mutually exclusive with trailing_stop_atr_mult", prefix))
 				}
-				if sc.TrailingStopATRRegime != nil && !sc.TrailingStopATRRegime.IsZero() {
+				// #1111: IsConfigured (raw-aware), NOT !IsZero() — the trailing block
+				// is resolved further below in the `if sc.TrailingStopATRRegime != nil`
+				// branch, so here it is still unresolved and IsZero() would report it
+				// absent, silently skipping this (sole) mutex check when both regime
+				// stops are set.
+				if sc.TrailingStopATRRegime.IsConfigured() {
 					errs = append(errs, fmt.Sprintf("%s: stop_loss_atr_regime is mutually exclusive with trailing_stop_atr_regime", prefix))
 				}
 				if sc.Platform != "hyperliquid" || sc.Type != "perps" {
--- a/scheduler/regime_iszero_resolution_test.go
+++ b/scheduler/regime_iszero_resolution_test.go
@@ -1,0 +1,149 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// #1111 regression suite. Several config-load-time validators read a regime ATR
+// block (trailing_stop_atr_regime / stop_loss_atr_regime) via IsZero() BEFORE
+// the block has been resolved. RegimeATRBlock.UnmarshalJSON captures only raw
+// JSON; the typed TrendRegime/UseDefaults fields are populated later, by
+// validateRegimeATRConfig. IsZero() returns true on an unresolved-but-configured
+// block, so a pre-resolution !IsZero() read treats a configured block as absent.
+// The blocks below are constructed raw-only (no resolved TrendRegime) to
+// reproduce that exact production state — a struct with TrendRegime already set
+// would not exercise the bug, which is why it shipped.
+
+const (
+	minMoveRequiresErr = "trailing_stop_min_move_pct requires"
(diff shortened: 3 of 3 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-25T01:01:24Z — LGTM + 1 Recommended Optional) in 1892130.

### Resolved (Recommended Optional)
1. **`hasHyperliquidStopLossOwnership` same-bug-class landmine** — deleted the function (was `scheduler/config.go:1103-1126`). Re-validated the reviewer's two-option call against the code and chose **delete** over swap-to-`IsConfigured()`+test:
   - Confirmed zero callers anywhere in the repo (only its own def + doc comment); `go build`/`go vet`/full `scheduler` test suite stay green after removal.
   - It is orphaned leftover from the pre-#601 peer stop-loss *normalization* design, which #601 superseded with per-strategy reduce-only protection orders (its sibling `normalizeHyperliquidPeerStopLosses` is the no-op'd remnant of the same old approach). It is not pending wiring.
   - Deleting closes the same-bug-class landmine completely — a function that doesn't exist can't reintroduce the bug — and avoids preserving dead code that encodes a stale, pre-#601 notion of "stop ownership" that a future caller shouldn't grab as-is.
   - **Alternative rejected:** swap both reads to `IsConfigured()` and add a regression test — keeps dead code alive and pins behavior nothing uses (maintenance debt + a misleading "this is live" signal). Override this if you'd rather keep+fix it for an imminent peer-ownership use; if peer stop-ownership validation is wanted, it's a fresh, in-scope-of-its-own-issue check built against the now-documented `IsConfigured()` convention.

The two original #1111 fixes (min-move false-reject, regime-stop mutex false-accept) are unchanged from the LGTM'd commit.

---

richkuo:
@claude review

PR #1117, opened 2026-06-25
feat(#1115): default manual close to trailing_tp_ratchet_regime when regime enabled

### LINKED ISSUE #1115: [C72] Default manual close to trailing_tp_ratchet_regime (conditional fallback to tiered_tp_atr_live)
**Complexity: 72/100** — scope: medium (config resolution, default-block synthesis, validation, manual-open inline protection path, tests); risk: high (manual SL/TP auto-protective path — money; the change opens a new naked-position window at manual-open that must be closed); uncertainty: medium (per-classifier default trail block design; regime-key resolution at manual-open CLI time; two-mode default behavior).

## Problem

Manual strategies (`type=manual`, HL perps) currently auto-default their close evaluator to `tiered_tp_atr_live` with a 1.5× scalar ATR stop-loss and tiered take-profits (`scheduler/config.go:1011-1046`). We want manual closes to use a regime-adaptive trailing take-profit ratchet (`trailing_tp_ratchet_regime`) instead, so the trail distance adapts per volatility regime.

`trailing_tp_ratchet_regime` cannot be an *unconditional* default: its validation requires `regime.enabled=true` (`scheduler/trailing_tp_ratchet.go:375`) AND a `trailing_stop_atr_regime` block (`:311`), forbids the scalar `trailing_stop_atr_mult` (`:314`), and forbids the scalar `stop_loss_atr_mult` that manual defaults to today (`:333-334`). (`:294` is the HL-perps/manual-only platform check, which manual satisfies.) Forcing it everywhere would break every manual strategy on a regime-less config.

## Goal

Make `trailing_tp_ratchet_regime` the manual default **when regime is enabled and a default per-regime trail block can be resolved**, falling back to the current `tiered_tp_atr_live` otherwise. Non-breaking; the chosen close must be observable (logged), never silently divergent in a protection path.

## Approach

1. **Conditional default selection** in the manual-defaults resolution loop (`config.go:1011`):
   - If `cfg.Regime` is enabled AND a default `trailing_stop_atr_regime` block can be synthesized for the active classifier vocabulary, set `CloseStrategy = trailing_tp_ratchet_regime` and attach that trail block.
   - Else keep `tiered_tp_atr_live` (today's behavior).
2. **Synthesize a default `trailing_stop_atr_regime` map** keyed to the active classifier's labels (adx 3-label vs composite — `shared_tools/regime.py` vocab; resolve the per-window labels via `regimeLabelsForStrategyWindow`, validate via `validateStrategyRegimeVocabulary`, `regime_window_spec.go:273`). **The composite vocab is no longer a flat 7-label set** — `#1059` (`4f1e987`) split the ranging ratchet ladder into quiet/volatile/directional substates, so the synthesized map must key to that expanded vocabulary, not a hardcoded 7-label list. Make the per-regime trail mults operator-tunable via `manual_defaults` (mirror the existing `stop_loss_atr_mult` / `tp_tiers` knobs in `ManualDefaultsConfig`, `config.go:178-192`).
3. **Suppress the conflicting scalar SL default** when the ratchet-regime close is selected — `trailing_tp_ratchet*` forbids `stop_loss_atr_mult`. The existing manual SL default at `config.go:1024` **already self-suppresses** when a regime trail block is present (its predicate includes `!sc.TrailingStopATRRegime.IsConfigured()`), so the integration point is **ordering**: attach the synthesized `trailing_stop_atr_regime` block *before* line 1024 runs and the scalar default opts out automatically. Do not add a parallel suppression branch.
4. **manual-open inline protection** (`scheduler/manual.go`) — this is where the change introduces a NEW naked-position window that must be closed:
   - Inline SL arming (`manual.go:346`) currently uses the scalar `effectiveATRMult` (`manual.go:301-303`). Under the ratchet-regime path `sc.StopLossATRMult` is nil (forbidden), so `effectiveATRMult==0` and **no inline SL is armed** — a regression from today's non-naked manual-open. The live (daemon) path arms the regime trail via `armTrailingStopAtOpenNow` (`hyperliquid_open_trailing.go:47`) keyed on `pos.Regime`, but that runs on the daemon cycle and is **never called by the manual-open CLI subprocess**, leaving the position naked for up to one interval.
   - Fix: have manual-open resolve the current regime at CLI time via the path it already uses (`runHyperliquidCheck(&sc, …, cfg.Regime, …)`, `manual.go:1261`), index the synthesized block with `resolveRegimeATR` to get the initial per-regime trail mult, and arm the initial trail inline — reusing the `armTrailingStopAtOpenNow` primitive so the inline and deferred arming can't drift. The resolved regime label should be stamped/frozen so the position's later trail re-derives the same key.
   - Inline TP placement (`manual.go:374`) is gated on `strategyUsesTieredTPATRClose` — correctly skips for `trailing_tp_ratchet*` (no on-chain TP). No change needed there; the SL arming above is what guarantees the position opens protected.
   - **Paper manual:** `trailing_tp_ratchet` regime arming is live-only (no on-chain SL on paper). Define what a paper manual strategy defaults to and how it behaves — either keep `tiered_tp_atr_live` on paper, or select the ratchet-regime close but document that its trail is only enforced live.
5. **Observability**: emit an INFO line at resolution stating which close was selected and why (regime on/off), matching the existing `default_stop_loss_atr_mult` INFO style (`config.go:986`).
6. **Docs**: update CLAUDE.md manual-defaults notes + `docs/ARCHITECTURE.md` close-evaluator section.

## Acceptance criteria

- Manual strategy on a regime-enabled config (adx and composite) defaults to `trailing_tp_ratchet_regime` with a valid `trailing_stop_atr_regime` block; passes `validateStrategyRegimeVocabulary` and the `trailing_tp_ratchet*` validators.
- Manual strategy on a regime-disabled config defaults to `tiered_tp_atr_live` exactly as today (no behavior change).
- No scalar `stop_loss_atr_mult` is attached when the ratchet-regime close is selected (no validation conflict) — verified via the existing `config.go:1024` self-suppression, ordering preserved.
- `manual-open` opens with an armed initial SL under the ratchet-regime path, **resolved inline in the CLI subprocess** (not deferred to the next daemon cycle); no naked-position gap.
- Paper manual behavior under a regime-enabled config is defined and tested (no on-chain arming attempted; documented default).
- Explicit operator `close_strategy` / stop-field overrides still win in both modes.
- INFO log states the selected close and the regime-on/off reason.
- Go tests cover both modes, both classifier vocabularies (including the `#1059` composite substates), the override-wins case, the paper-vs-live split, and the no-naked-SL invariant at manual-open.

---

richkuo:
## Summary

Closes #1115. Manual (`type=manual`, HL) strategies with no explicit `close_strategy` now default to the regime-adaptive trailing take-profit ratchet (`trailing_tp_ratchet_regime`) **when regime detection is enabled and a default per-regime opening-trail block resolves** for the active classifier vocabulary — otherwise they keep today's `tiered_tp_atr_live` + 1.5×ATR scalar SL, so regime-less configs are unchanged.

## What changed

- **Conditional default selection** (`config.go:resolveManualRatchetRegimeTrailBlock`): picks `trailing_tp_ratchet_regime` iff regime is enabled, no explicit `close_strategy`/stop field is set, and every active classifier label (ADX-3 or composite-7) maps onto the `regimeATRDefaults.Trailing` baseline. It synthesizes a `trailing_stop_atr_regime` `use_defaults` block and attaches it **before** the scalar manual-SL default — so that default self-suppresses via its existing `!TrailingStopATRRegime.IsConfigured()` guard (the regime block becomes the SL owner; no validation conflict). Selection is INFO-logged.
- **Operator-tunable**: `manual_defaults.trailing_stop_atr_regime` supplies the per-regime opening trail in place of the baseline; cloned per-strategy (`cloneRegimeATRBlock`) and resolved against each strategy's labels at `validateConfig`.
- **No naked window at `manual-open`** (`manual.go:resolveManualRatchetRegimeOpeningTrail`): for the ratchet-regime path (scalar SL absent), it reads the live regime at CLI time via `runHyperliquidCheck`, resolves the per-regime opening trail (`resolveRegimeATR`), and feeds it as `effectiveATRMult` so the existing post-fill SL arm places the initial trailing SL inline. The daemon's trailing walker re-pins from `pos.Regime` next cycle, so a CLI-vs-daemon regime mismatch self-heals. Regime-read failure falls back to a protective `1.5×ATR` SL (warned), still non-naked.

## Acceptance criteria

- ✅ Regime-enabled (ADX **and** composite) → `trailing_tp_ratchet_regime` with a valid `trailing_stop_atr_regime` block; passes `validateStrategyRegimeVocabulary` + the `trailing_tp_ratchet*` validators.
- ✅ Regime-disabled → `tiered_tp_atr_live` exactly as today.
- ✅ No scalar `stop_loss_atr_mult` attached when ratchet-regime is selected.
- ✅ Explicit `close_strategy` / stop-field overrides win in both modes.
- ✅ INFO log states the selected close + the regime on/off reason.
- ✅ `manual-open` opens with an armed initial SL under the ratchet-regime path (inline, not deferred).

## Tests

`scheduler/manual_ratchet_regime_default_test.go` — regime-disabled (unchanged), regime ADX (3-label) and composite (7-label) selection + validation + positive opening trail per label (no-naked invariant), explicit-`close_strategy`-wins, explicit-stop-field-keeps-tiered, `manual_defaults` override, and no-aliasing across strategies.

## Verification

`go build .`, `go test .` (full suite + `-race`), `go vet`, `gofmt` — all clean.

## Note for existing deployments

This changes the manual close **default** for regime-enabled configs. Operators with open manual positions who rely on the default should set `close_strategy` explicitly **before** upgrading if they want continuity across a restart (config defaults are re-derived at load and can't see open positions). SIGHUP reloads on the same binary are idempotent (no spurious close-evaluator flip).

---

https://claude.ai/code/session_011nvFQ7NVYdvScByfbc6sRP


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-25T04:19) in a407aab. Verdict was LGTM; both items below were Recommended Optional and are now implemented + tested (`go build`, `go vet`, `gofmt`, full `scheduler` suite + `-race` clean).

### Fixed
1. **No-naked fallback now has a pure, unit-tested helper.** Split `resolveManualRatchetRegimeOpeningTrail` into the impure `resolveManualRatchetRegimeLabel` (spawns the regime subprocess) and the pure `manualRatchetOpeningTrailOrFallback(block, label)` (`manual.go:1283-1303`), which **never returns ≤0** — per-regime trail when the label resolves a positive distance, else the `1.5×ATR` fallback. The arming site (`manual.go:314-329`) consumes `(mult, fellBack)`. New test `TestManualRatchetOpeningTrailOrFallback` covers all three "Must survive" branches: (a) empty label / regime-read failure → fallback, (b) label with no/zero trail → fallback, (c) good label → per-regime trail — each asserting a strictly-positive armed mult. `--record-only` is the outer `!*recordOnly` guard at the call site (arming skipped entirely).

2. **Close-evaluator drift now alerts the operator.** The daemon's manual close-eval emits a once-per-(strategy,symbol) owner DM when an open position opened under a tiered-TP close (carries resting on-chain TP OIDs) is now under the ratchet default after an upgrade+restart (`main.go:2222-2240`, detection `manualCloseEvaluatorDriftedFromTPs` in `trailing_tp_ratchet.go:682-702`, deduped via `manualCloseEvaluatorDriftWarned`). Pure detection helper unit-tested (`TestManualCloseEvaluatorDriftedFromTPs`). This satisfies the invariant via the "operator is alerted" arm.

### Resolved judgment calls (the reviewer left the choice open: warn vs. auto-cancel)
1. **Chose the non-destructive warning over auto-cancelling the stale TPs.** Auto-cancelling on-chain reduce-only orders off a *config-default flip* (not an explicit operator action) is exactly the surprising auto-mutation on a money path that correctness/safety rules say to avoid — and it would silently remove profit-taking exits the operator may still want. The warning gives the operator the decision (pin `close_strategy=tiered_tp_atr_live` and restart, or cancel the TPs on the HL UI to fully adopt the ratchet). Override if you'd prefer auto-reconciliation. *Rejected:* (a) cancel-on-adoption — destructive, off a default flip; (b) startup-only check — the daemon (config + state) is the right layer and per-cycle dedup fires reliably right after restart.

**Confirmed stale-TP reconciliation path (as requested):** the resting TPs are **not** permanently orphaned — `manual-close` cancels `pos.TPOIDs` via `extraCancelOIDs` (`main.go:2303`), and HL auto-cancels reduce-only orders when the trailing SL flattens the position. The only residual is a TP firing mid-life under a now-let-it-ride config, which the alert surfaces. The SIGHUP-while-open flip stays fail-closed by the `trailing_stop_atr_regime` mode-change guard (`config_reload.go`), unchanged.

---

richkuo:
@claude review

PR #1122, opened 2026-06-25
Add ratchet fallback normalization for manual opens

richkuo:
## Summary
- Raise the manual default stop-loss fallback to 2.0x ATR and add a separate ratchet fallback resolver that never returns zero
- Persist a one-shot normalize marker through state and pending manual actions so fallback manual opens can widen once to the configured regime trail
- Update trailing-stop logic and tests to cover the new fallback normalization path and zero-opt-out behavior

## Testing
- Added config helper coverage for the new fallback resolver and zero scalar opt-out handling
- Added trailing-stop update coverage for one-shot widen behavior
- Updated existing config and storage-path expectations to match the new manual default


richkuo:
Superseded by #1123, which uses the issue-specific branch codex/issue-1121-manual-ratchet-fallback based directly on origin/main and excludes unrelated commit 1f1b633.

PR #1123, opened 2026-06-25
fix(#1121): normalize manual ratchet fallback stops

### LINKED ISSUE #1121: [C70] manual-open ratchet fallback SL ignores config and first re-pin won't widen
**Complexity: 70/100** — scope: medium (manual-open path, trailing walker, config resolver, new persisted Position flag + SQLite migration, lock-free snapshot threading, tests); risk: high (live HL protective-stop placement — floors complexity ≥60); uncertainty: medium (part-2 normalize-once state mechanism to design; live incident evidence is external to repo validation).

## Summary

`go-trader manual-open` can arm a hardcoded **1.5x ATR** fallback stop-loss for a `trailing_tp_ratchet_regime` manual strategy when the live regime label cannot be resolved at open time. The fallback path does not use the operator's manual defaults or the configured per-regime trail block, and the daemon's next-cycle trailing walker may not normalize it because widening an already-resting stop is treated as an unsafe loosening move.

Result: a manual position can remain on a tighter or looser initial SL than the configured ratchet trail, while the warning says the daemon will re-pin next cycle.

## Validated Current Behavior

`defaultManualStopLossATRMult` is hardcoded to `1.5` in `scheduler/manual.go`.

`manualRatchetOpeningTrailOrFallback(sc.TrailingStopATRRegime, label)` returns the per-regime trail only when `label` resolves to a positive entry in the block. Empty/unresolved labels, missing labels, zero entries, or nil block all return the package constant `defaultManualStopLossATRMult` **directly** — this pure helper takes only `(block, label)` and has no `*Config`, so it cannot honor `manual_defaults.stop_loss_atr_mult` (which `resolveManualStopLossATRMult` reads).

That multiplier is fed into the inline SL arm after fill:

- long: `fillPrice - mult * entryATR`
- short: `fillPrice + mult * entryATR`

The CLI queues the resulting `StopLossOID` and `StopLossTriggerPx`; `applyManualAction` adopts both into `Position`, so the daemon walker sees an existing trigger rather than an unarmed position.

The live manual walker calls `runHyperliquidTrailingStopUpdate` with `forceResize` only for scale-in sizing. There is no first-normalize mode for a fallback SL.

The shared trailing update logic only replaces an existing trigger when the move is favorable:

- long: candidate trigger must be higher than current trigger
- short: candidate trigger must be lower than current trigger

So for a short, a wider configured trail produces a higher trigger and is skipped as a loosening move. The warning text `daemon re-pins per-regime next cycle` overpromises for this case.

## Config Surface

There is already a manual config surface:

- `manual_defaults.stop_loss_atr_mult` controls scalar manual SL defaults.
- `manual_defaults.trailing_stop_atr_regime` controls the synthesized per-regime ratchet trail block for default manual ratchet strategies.

The bug is that the unresolved-regime fallback path bypasses these defaults and returns the package constant directly. A new `StrategyConfig.manual_open.default_sl_atr_mult` field does not match the repo's current config model unless per-strategy fallback control is intentionally added as a broader design change.

Preferred config direction: keep the fallback under `manual_defaults` or derive it from the existing `manual_defaults.trailing_stop_atr_regime` block. If a new knob is needed, make it a clearly named `manual_defaults` sibling for the ratchet fallback, validate it as strictly positive, and keep the no-naked-position invariant.

## Fix Scope

0. Change the hardcoded `defaultManualStopLossATRMult` constant from `1.5` to `2.0`. Update the `config.go` doc comment and any test fixtures pinned to `1.5`.
1. Make the unresolved-regime ratchet fallback config-driven through the existing manual defaults shape. The new absent-config fallback is `2.0`; `manual_defaults` may override it. This requires `manualRatchetOpeningTrailOrFallback` to take a `*Config` (or to resolve the manual-defaults value before the pure call), since today it cannot read `manual_defaults`.
2. Add a first-normalize path for CLI-armed ratchet fallback SLs. Concretely: freeze a `normalize-once` marker on `Position` at the manual-open fallback arm (new persisted field + SQLite column + idempotent migration), then expose it to the walker path that consumes it — the manual trailing walker reads it directly from `Position` (`main.go` ~2274 passes `pos`), while any shared perps/paper implementation carries it through the existing position snapshot helpers (`main.go` ~1933) as needed. Allow **exactly one** widening cancel+replace toward the configured trail, and clear the marker on the first confirmed normalize. Invariant: the one-shot marker must be durable and visible wherever `runHyperliquidTrailingStopUpdate` decides whether to permit a one-time widening normalize. A no-flag approach (detect mismatch and correct) is unsafe because it would loosen the stop every cycle.
3. Preserve the existing favorable-only trailing behavior after normalization; do not allow ordinary later cycles to loosen a ratcheted stop.
4. Update the warning text so it only promises re-pin when the daemon actually has enough state to normalize.
5. Add tests for short widen, long widen, currentTrigger=0, scale-in `forceResize`, normal favorable ratchet movement, unresolved-label fallback config resolution, and the one-shot normalization (first cycle widens, second cycle reverts to favorable-only).

## Live Reference

Operator evidence from `manual-eth` short opened 2026-06-25 11:59:40 UTC:

- Strategy: `manual-eth`, type=manual, HL perps, ETH, 30m
- Entry: $1,631.90, ATR $7.81
- Configured `trailing_stop_atr_regime.trend_regime.*.atr_multiple = 2.0`
- Open armed fallback SL at $1,643.60 using 1.5x ATR
- 2.0x target would have been $1,647.52, wider for the short
- Daemon cycles did not re-pin before the fallback SL fired
- Closed by SL at 2026-06-25T12:35:03, qty 0.613 @ $1,643.60, realized -$7.61

This live evidence was not independently verified from repo state during issue validation; the code paths above explain how the reported behavior can occur.

---

LLM: GPT-5 | high | Harness: validate-issue/code inspection
Updated with LLM: Opus 4.8 | high | Harness: validate-issue

richkuo:
Closes #1121

Summary:
- Make the absent manual SL fallback 2.0x ATR and use manual_defaults.stop_loss_atr_mult as the manual-open ratchet fallback when positive.
- Persist a one-shot RatchetFallbackNormalizePending marker through pending manual actions and positions.
- Allow the trailing walker to widen exactly once to the configured regime trail, then clear the marker on confirmed replacement/immediate fill.

Tests:
- env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'Test(LoadConfigManualDefaultsStopLossATRMultTo2Point0|LoadConfigManualDefaultsStopLossATRMultOverride|LoadConfigManualDefaultsStopLossATRMultZeroOptsOut|ConfigResolveManualHelpersFallback|ConfigResolveManualHelpersFromConfig|ConfigResolveManualRatchetFallbackIgnoresZeroScalarOptOut|ManualRatchetOpeningTrailOrFallback|RunHyperliquidTrailingStopUpdate_RatchetFallbackNormalizeWidensOnce|ApplyTrailingStopUpdateResult_RestingReplacement|ApplyTrailingStopUpdateResult_CancelWithoutRestClearsStaleOID|PendingManualActionOpenFieldsRoundtrip|ApplyManualAction_OpenSetsProtectionFields|DefaultManualStopLossATRMult|RatchetFallbackNormalizePendingPersistsRoundTrip)'

LLM: GPT-5 | high | Harness: targeted go test


richkuo:
@claude review

PR #1126, opened 2026-06-26
feat(#1124): split ranging_directional into _up/_down composite labels

### LINKED ISSUE #1124: [C75] Split ranging_directional into explicit ranging_directional_up / ranging_directional_down labels
### Current behavior
The composite classifier emits `ranging_directional` as a single label regardless of whether the drift is bullish or bearish. In `shared_tools/regime.py:map_composite_label` (lines 159-172), `return_eff`'s sign only gates the **trending** split (`up = return_eff > 0` → `trending_up_*`); when there is no decisive net move but ADX is high, `ranging_directional` is returned with **no** sign baked into the label. The directional bias is resolved at evaluation time from `snapReturnEff`:

```go
func regimeLabelBias(label string, snapReturnEff float64) divergenceBias {
    // ...
    case "ranging_directional":
        if snapReturnEff > 0 { return biasBullish }
        if snapReturnEff < 0 { return biasBearish }
        return biasNeutral
}
```
(`scheduler/regime_divergence.go:240-258`; `snapReturnEff` is read from `RegimeSnapshot.Metrics["return_eff"]` per window at `regime_divergence.go:357-364`.)

This means a `ranging_directional` regime can be up, down, or neutral (when `return_eff == 0`) depending on the window's `return_eff`, but the label itself doesn't tell you which. Operators have to check the metrics payload to know the direction. Note: because `ranging_directional` fires when `big_move` is **false** (no decisive net move), `return_eff` is frequently near zero — so the `== 0` neutral case is not an edge case, it is a common path.

### Why this matters
- The trending family already splits cleanly: `trending_up_clean`, `trending_up_choppy`, `trending_down_clean`, `trending_down_choppy` (`regime.py:167-168`) — the label carries the direction.
- The ranging family keeps direction implicit: `ranging_directional` is the only label whose direction is resolved at runtime (`regime_divergence.go:246-257`; `ranging_quiet`/`ranging_volatile` resolve neutral).
- For config blocks that want different behavior per sub-regime (e.g. different SL/TP geometry for ranging-up vs ranging-down), the operator needs to write custom logic or inspect `return_eff` themselves.

### Proposed change
Emit `ranging_directional_up` when `return_eff > 0`, `ranging_directional_down` when `return_eff < 0`, and keep emitting the bare `ranging_directional` label when `return_eff == 0` (neutral). Keeping the bare label as the producer-side fallback makes back-compat the SSoT's responsibility rather than a config resolver's, and preserves a valid label for the common near-zero-return case.

This would make the ranging family consistent with the trending family.

### Impact
The label producer is a single source of truth — `shared_tools/regime.py:map_composite_label` — imported by both the live `check_regime.py` path and the backtester (e.g. `backtest/regime_vol_model.py:214`). All consumers below must move in lockstep in one PR (dual-language Go + Python surface).

- **producer (Python, SSoT):** `map_composite_label` (`regime.py:172`) emits the three labels; `VALID_LABELS_COMPOSITE` (`regime.py:51-59`) must accept `ranging_directional_up`/`_down`; bare `ranging_directional` stays valid.
- **config vocabulary:** `validateStrategyRegimeVocabulary` / `regimeLabelsForStrategyWindow` (`scheduler/regime_window_spec.go:273-296`) must accept the new labels in `allowed_regimes`, `regime_directional_policy`, and `*_atr_regime` keys, or config validation rejects them.
- **code (Go divergence):** `regimeLabelBias` (`regime_divergence.go:240`) collapses the three labels to the same bias (up→bullish, down→bearish, bare→neutral via `snapReturnEff`).
- **code (Go ratchet default-ladder) — SAFETY-CRITICAL:** `ratchetCloseDefaultGroup` (`scheduler/trailing_tp_ratchet.go:123-132`) is an explicit switch over `ranging_quiet`/`ranging_volatile`/`ranging_directional`. New `_up`/`_down` labels would NOT match and fall through to `regimeCloseDefaultGroup` (`regime_atr.go:257-258`, `HasPrefix("ranging")`→`"ranging"`), which is **absent** from `ratchetTierGroupDefaults` (`trailing_tp_ratchet.go:84`) → `defaultTrailingRatchetTiersForRegime` returns nil (`trailing_tp_ratchet.go:138-150`) → **silent never-arm of the auto-protective ratchet exit (money path)**. Either add explicit `ranging_directional_up`/`_down` ladder keys to `ratchetTierGroupDefaults` + the switch, OR gate the resolver so they map to the existing `ranging_directional` ladder. The shared `regimeCloseDefaultGroup` is unaffected (already collapses to `"ranging"`).
- **code (Python mirror):** `regime_close_default_group` in `shared_strategies/close/trailing_tp_ratchet.py:87` mirrors `ratchetCloseDefaultGroup` and must move with it (dual-language SSoT).
- **code (Go ATR defaults):** `regimeATRDefaults.Trailing` (`regime_atr.go:228-239`) is an explicit keyed map; it needs `ranging_directional_up`/`_down` entries or a fallback, else `trailing_stop_atr_regime` resolves ATR=0 for those labels.
- **code (Go fleet-default rendering):** `regimeTPFleetDefaultLabelsByGroup` (`regime_atr.go:281`) lists labels per group for inspect-time rendering; add the new labels to the `ranging` group.
- **backtest parity:** `backtest/regime_vol_model.map_latent_to_names` (`regime_vol_model.py:214`) and the #1058 7-state composite backtester consume `map_composite_label`; both must see the new labels. Update `backtest/tests/test_backtester_composite_regime.py` and `test_backtester_composite_regime_wiring.py` (currently assert the 7-label set).
- **config:** Any `tp_tiers` or `trend_regime` blocks that reference `ranging_directional` continue working as a fallback/catch-all (bare label still emitted on `return_eff == 0`).

### Back-compat / failure policy
- Bare `ranging_directional` remains a valid, emitted label (the `return_eff == 0` case), so existing configs keyed on it do not silently no-op.
- Any consumer that maps the new `_up`/`_down` labels must either add explicit entries or fall back to the bare `ranging_directional` ladder — never silently miss, since a missed ratchet ladder = never-armed protective exit.

### Reference
Live evidence: ETH composite_short at 22:55 UTC shows `ranging_directional` with `return_eff=-0.0944` (bearish), but the label itself is direction-neutral. The same label was bullish earlier today when `return_eff` was positive.

---
Updated with LLM: GLM 5.2 | 0.5 | Harness: Cursor

richkuo:
## Summary

Splits the composite classifier's `ranging_directional` label into direction-carrying
`ranging_directional_up` / `ranging_directional_down`, with the bare `ranging_directional`
retained only for the exactly-zero-drift case. This makes the ranging family carry direction
in the label like the trending family (`trending_up_*` / `trending_down_*`) already does, so
operators can gate and configure per sub-regime without reading `return_eff` from the metrics
payload.

The producer `map_composite_label` (`shared_tools/regime.py`) is the single source of truth,
imported by both the live `check_regime` path and the backtester (`compute_regime_composite`,
`regime_vol_model.map_latent_to_names`). All consumers move in lockstep in this one PR across
the dual-language Go + Python surface.

## Changes

- **Producer + vocabulary:** `map_composite_label` emits the three labels by `return_eff` sign;
  `VALID_LABELS_COMPOSITE` (Python) and `regimeLabelsForClassifier` (Go) accept the two new labels.
- **Divergence bias** (`regimeLabelBias`): `_up` → bullish, `_down` → bearish; the bare label still
  resolves via `snapReturnEff` for back-compat with stale snapshots.
- **Ratchet default ladder — SAFETY-CRITICAL** (`ratchetCloseDefaultGroup` / `ratchet_close_default_group`):
  `_up`/`_down` map to the existing `ranging_directional` scale-out ladder (geometry is direction-agnostic).
  Without this they would fall through to the shared `"ranging"` group, which has **no** ratchet ladder
  → `defaultTrailingRatchetTiersForRegime` returns nil → **silent never-arm of the auto-protective ratchet
  exit** (the failure mode the issue flagged). Verified by test.
- **ATR trailing defaults** (`regimeATRDefaults.Trailing` / `REGIME_ATR_DEFAULTS_TRAILING`): explicit
  `_up`/`_down` entries at ATR 1.0 so they inherit the tight `ranging_directional` opening trail rather
  than the wider 2.0 `"ranging"` family fallback via `mapRegimeToBaselineFamily`.
  *(Validation note: the issue described this as resolving to ATR=0; in fact the prefix fallback yields the
  2.0 ranging-family value — a value drift, not a never-arm. The ratchet ladder is the genuine never-arm.)*
- **Fleet-default rendering** (`regimeTPFleetDefaultLabelsByGroup`): new labels added to the `ranging` group.

## Back-compat: fail-closed migration (design decision)

Per the chosen "full per-direction" posture, the new labels join the composite classifier vocabulary, so
they are valid for `allowed_regimes`, `regime_directional_policy`, `*_atr_regime` keys, and
`regime_profile_allocation` — and, by that same vocabulary, **required** in explicit (non-`use_defaults`)
composite `trend_regime` / profile-allocation blocks. A pre-#1124 explicit composite block now errors at
load with `missing required regime labels: ranging_directional_up, ranging_directional_down` rather than
silently no-opping (consistent with the repo's existing "must be exhaustive — no silent fallback"
invariant). `use_defaults` blocks auto-expand and are unaffected. No shipped config example contains such a
block, so nothing in-repo needs migrating.

## Out of scope

The `regime_adaptive` / `regime_adaptive_htf` open strategies re-implement label mapping for their own
internal state machine and emit the bare `ranging_directional` as a diagnostic `ra_label` column. They do
not feed the gating / ratchet / divergence / ATR consumers, so their state machine is left unchanged (the
bare label remains valid).

## Verification

- `go build .` + `go test .` (full scheduler suite) — pass.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — pass.
- New tests: producer `_up`/`_down`/zero cases; divergence bias for the explicit labels; ratchet ladder +
  ATR trailing parity to the `ranging_directional` substate (asserts non-nil, never-arm guard); composite
  exhaustiveness fixtures updated to 9 labels across Go and Python.

Closes #1124

LLM: Opus 4.8 | high | Harness: work-on-issue

### DIFF
--- a/backtest/tests/test_backtester_composite_regime_wiring.py
+++ b/backtest/tests/test_backtester_composite_regime_wiring.py
@@ -613,6 +613,8 @@ _COMPOSITE_SL_BLOCK = {"trend_regime": {
     "ranging_quiet": {"atr_multiple": 1.2},
     "ranging_volatile": {"atr_multiple": 1.5},
     "ranging_directional": {"atr_multiple": 1.4},
+    "ranging_directional_up": {"atr_multiple": 1.4},
+    "ranging_directional_down": {"atr_multiple": 1.4},
 }}
 _ADX_SL_BLOCK = {"trend_regime": {
     "trending_up": {"atr_multiple": 2.0},
--- a/scheduler/manual_ratchet_regime_default_test.go
+++ b/scheduler/manual_ratchet_regime_default_test.go
@@ -156,12 +156,13 @@ func TestManualDefault_RegimeComposite_SelectsRatchetRegime(t *testing.T) {
 		t.Fatalf("StopLossATRMult = %v, want nil", *sc.StopLossATRMult)
 	}
 	block := sc.TrailingStopATRRegime
-	if block == nil || len(block.TrendRegime) != 7 {
-		t.Fatalf("trailing_stop_atr_regime must resolve to 7 composite labels, got %#v", block)
+	if block == nil || len(block.TrendRegime) != 9 {
+		t.Fatalf("trailing_stop_atr_regime must resolve to 9 composite labels, got %#v", block)
 	}
 	for _, label := range []string{
 		"trending_up_clean", "trending_up_choppy", "trending_down_clean",
 		"trending_down_choppy", "ranging_quiet", "ranging_volatile", "ranging_directional",
+		"ranging_directional_up", "ranging_directional_down",
 	} {
 		if v, ok := resolveRegimeATR(*block, label); !ok || v <= 0 {
 			t.Fatalf("opening trail for composite label %q = (%g, %v), want positive", label, v, ok)
--- a/scheduler/regime_atr.go
+++ b/scheduler/regime_atr.go
@@ -236,6 +236,12 @@ var regimeATRDefaults = struct {
 		"ranging_quiet":        {ATR: 1.0},
 		"ranging_volatile":     {ATR: 1.0},
 		"ranging_directional":  {ATR: 1.0},
+		// #1124: directional-drift substates inherit the tight ranging_directional
+		// opening trail (1.0). Explicit entries keep parity with the bare label —
+		// without them mapRegimeToBaselineFamily would fall the _up/_down labels
+		// back to the wider "ranging" family (2.0), silently loosening the trail.
+		"ranging_directional_up":   {ATR: 1.0},
+		"ranging_directional_down": {ATR: 1.0},
 	},
 }
 
@@ -278,7 +284,7 @@ var regimeTPTierGroupDefaults = map[string][]hlProtectionTier{
 var regimeTPFleetDefaultLabelsByGroup = map[string][]string{
 	"clean":   {"trending_up_clean", "trending_down_clean"},
 	"choppy":  {"trending_up", "trending_down", "trending_up_choppy", "trending_down_choppy"},
-	"ranging": {"ranging", "ranging_quiet", "ranging_volatile", "ranging_directional"},
+	"ranging": {"ranging", "ranging_quiet", "ranging_volatile", "ranging_directional", "ranging_directional_up", "ranging_directional_down"},
 }
 
 // defaultRegimeBlockForSurface returns the baseline trend_regime map for a
--- a/scheduler/regime_atr_composite_test.go
+++ b/scheduler/regime_atr_composite_test.go
@@ -59,8 +59,8 @@ func TestValidateRegimeATRConfig_CompositeStopLossExplicit(t *testing.T) {
 		t.Fatalf("composite stop_loss_atr_regime must validate, got: %v", errs)
 	}
 	block := cfg.Strategies[0].StopLossATRRegime
-	if got := len(block.TrendRegime); got != 7 {
-		t.Fatalf("block must be populated with all 7 composite labels, got %d: %v", got, block.TrendRegime)
+	if got := len(block.TrendRegime); got != 9 {
+		t.Fatalf("block must be populated with all 9 composite labels, got %d: %v", got, block.TrendRegime)
 	}
 	// Runtime SL resolution must succeed for a composite label (proves the
 	// authoritative pass populated the composite vocabulary, not ADX).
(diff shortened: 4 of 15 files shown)


richkuo:
@claude review

PR #1129, opened 2026-06-26
fix(#1128): lazy-init HL Exchange and exempt transient 429 script failures

### LINKED ISSUE #1128: [C70] HL adapter eager Exchange init causes regime 429 failures
## Summary

The per-cycle regime store spawns `shared_scripts/check_regime.py` once per distinct (platform, symbol, timeframe, windows-spec) signature. For Hyperliquid signatures (e.g. `hyperliquid/ETH/30m`), the script's `_load_adapter("hyperliquid")` → `ExchangeAdapter()` constructor instantiates the HL SDK's `Exchange(...)` whenever `HYPERLIQUID_SECRET_KEY` is set. The SDK `Exchange` embeds a fresh `Info` without the cached `meta`/`spot_meta` kwargs, so its constructor POSTs `spotMeta` + `meta` to `/info` on every adapter init — even when `adapter._build_info` already loaded them from `/tmp/hl_meta.json`. When HL's CloudFront CDN rate-limits this IP with a 429, the constructor raises:

```
RuntimeError: Failed to initialize Hyperliquid Exchange client: (429, None, 'null', None, {'Content-Type': 'application/json', 'Content-Length': '4', ..., 'Server': 'nginx/1.22.1', 'X-Cache': 'Error from cloudfront', 'Via': '1.1 ...cloudfront.net (CloudFront)', ...})
```

The script emits the structured-error JSON, exits 1, and the scheduler's `scriptFailureTracker` counts it as a script error. After 3 consecutive cycles of failures the **`SIGNAL SCRIPT FAILING`** operator DM fires (synthetic ID `regime[hyperliquid/<sym>/<tf>]`), which is what surfaced this in logs on 2026-06-26 around 05:03 UTC. This does **not** disarm or reset the portfolio kill switch — that path is unrelated (`CheckPortfolioRisk` drawdown latch + manual/auto reset in `main.go`).

## Why the meta cache (#768) only partially helps

`META_CACHE_PATH = /tmp/hl_meta.json` (60-min TTL) short-circuits `adapter._build_info` → `Info(..., meta=..., spot_meta=...)` so the adapter's primary `Info` skips its two-call init storm on cache hit. Subprocesses of the **same** systemd instance share that `/tmp` (`PrivateTmp=true` isolates per service instance, not per subprocess — see `adapter.py:40-41`). So peer subprocesses within one `go-trader-live` unit can reuse a warm cache.

The remaining gap: `Exchange(...)` is still constructed eagerly in `__init__` without passing those cached blobs, so the SDK builds a **second** embedded `Info` that re-fetches `spotMeta` + `meta` every spawn. On a cold cache or concurrent burst (up to `pythonSemaphore=4` parallel spawns), multiple subprocesses can all miss before the first write lands. Total `/info` pressure scales roughly with (strategies-due + distinct-regime-signatures) per cycle, each paying at least the Exchange-embedded `Info` init cost when live creds are present.

## Why this breaks regime (operator-alert impact)

`check_regime.py`'s regime calc only needs OHLCV (`adapter.get_ohlcv` → `self._info.candles_snapshot(...)`). It does NOT need the trading `Exchange` client. The current adapter constructor builds `Exchange(...)` unconditionally when a secret is set, so a 429 on the embedded `Info`'s `spotMeta`/`meta` fetch fails the whole adapter init — even though OHLCV alone would suffice. That becomes a regime-bundle failure for the cycle (signature fails open per #879), then escalates to the 3-strike `scriptFailureTracker` alert. An armed portfolio kill switch stays armed; the harm here is spurious operator noise and fail-open regime gating, not kill-switch disarm.

## Repro

1. Live trader running `go-trader-live` with HL strategies and regime enabled for an HL symbol/timeframe (e.g. ETH 30m).
2. From a single IP, run enough scheduler cycles to trip CloudFront's `/info` rate limiter (observed: ~5+ HL adapter inits/sec for several cycles is sufficient).
3. Observe the `SIGNAL SCRIPT FAILING [regime[hyperliquid/<sym>/<tf>]]` alert with `Failed to initialize Hyperliquid Exchange client: (429, ...)`.

### Part 1 — Lazy-init `Exchange` on OHLCV-only paths (adapter SSOT)

In `platforms/hyperliquid/adapter.py` (covers `check_regime.py`, `check_hyperliquid.py`, and every other HL subprocess consumer):

- Keep `self._info = self._build_info(...)` eager — needed for OHLCV and read-only calls — with cached `meta`/`spot_meta` on hit.
- **Defer** `Exchange(...)` until the first call that needs it (`market_open`, `market_close`, `update_leverage`, etc.). Trading methods already gate on `self._exchange` (`adapter.py:751+`).
- When `Exchange` is eventually constructed, pass the same cached `meta`/`spot_meta` into `_HLExchange(...)` so it does not spawn a second uncached `Info` (SDK supports these kwargs).
- Wrap deferred construction in a per-adapter lock so concurrent trading calls in one subprocess don't stampede the same SDK init.
- On `Exchange(...)` init failure (rate limit, network), cache the negative result for a short in-process TTL (e.g. 30s) so subsequent trading calls in the same subprocess fast-fail without re-hitting CloudFront; OHLCV paths remain usable.

No separate per-cycle pre-warm subprocess is required if the adapter defers `Exchange` and shares cached meta into it.

### Part 2 — Don't let transient 429s trip the 3-strike script-failure alert

In `scheduler/script_failure_alerts.go` ( **`scriptFailureTracker` / `notifyScriptFailure`** — not kill-switch reset logic):

- Classify 429 / rate-limit / CloudFront `X-Cache: Error from cloudfront` responses as **TRANSIENT** and exempt them from the consecutive-failure count that triggers the first `SIGNAL SCRIPT FAILING` DM at 3 strikes. Still log at WARN so operators see HL throttling.
- Decide explicitly whether transient exempt means: (a) never increment the streak, or (b) increment but never alert — prefer (a) so recovery clears cleanly.
- Optionally retry once with backoff on the regime subprocess path before recording a failure (Go-side or Python-side).

## Test plan

- Unit test: `ExchangeAdapter.__init__` with live creds + warm meta cache constructs `self._info` and leaves `self._exchange` as `None`; no `Exchange` `/info` calls on init (mock SDK).
- Unit test: `get_ohlcv` works with `_exchange is None`; first `market_open` lazily constructs `Exchange` with cached `meta`/`spot_meta` passed through.
- Unit test: 429 during deferred `Exchange` init does not fail `__init__`; adapter remains OHLCV-usable; trading calls get structured error until backoff TTL expires.
- Unit test: concurrent lazy `Exchange` init in one subprocess only hits SDK once (lock).
- Integration: N concurrent `check_regime.py` spawns for the same HL signature — with warm `/tmp/hl_meta.json`, observe zero `spotMeta`/`meta` fetches per spawn when `Exchange` is deferred; at most one cold-cache fetch wave on first cycle after TTL expiry.
- Regression: existing live-mode tests for `market_open` / `market_close` still pass.

## Environment

- go-trader live deployment (`go-trader-live.service`)
- 4 strategies due per cycle, ~3s cycle cadence → ~5+ HL adapter inits/sec from one IP
- systemd `PrivateTmp=true` on the live unit (intentional — `/tmp` cache scope is per **systemd instance**, shared across that instance's subprocesses)
- HL rate-limit response observed at `https://api.hyperliquid.xyz/info` via CloudFront (`HEL51-P2`)

---

Updated with LLM: Composer | high | Harness: Cursor

richkuo:
## Summary
- Defer Hyperliquid SDK `Exchange` construction until the first trading call; regime/signal subprocesses keep using eager cached `Info` for OHLCV only (#1128).
- Pass cached `meta`/`spot_meta` into lazy `Exchange` init to avoid a second uncached embedded `Info`.
- Add 30s in-process backoff after Exchange init failure; OHLCV paths stay usable.
- Exempt HL 429 / rate-limit / CloudFront throttle errors from `scriptFailureTracker` streak increments (still WARN-logged).

## Test plan
- [x] `pytest platforms/hyperliquid/test_adapter.py platforms/hyperliquid/test_meta_cache.py` (92 passed)
- [x] `go test -run 'TestScriptFailure|TestNotifyScriptFailure|TestFormatScript' ./scheduler`

Closes #1128

LLM: Composer | high | Harness: Cursor

### DIFF
--- a/platforms/hyperliquid/adapter.py
+++ b/platforms/hyperliquid/adapter.py
@@ -16,6 +16,7 @@ import math
 import os
 import sys
 import tempfile
+import threading
 import time
 from decimal import Decimal, ROUND_DOWN
 
@@ -30,6 +31,10 @@ TESTNET_URL = "https://api.hyperliquid-testnet.xyz"
 # call entirely (#768).
 META_CACHE_PATH = "/tmp/hl_meta.json"
 META_CACHE_TTL_S = 3600  # 60 minutes
+
+# After a lazy Exchange init failure (429, network), suppress re-attempts briefly
+# so trading calls in the same subprocess fast-fail without re-hitting CloudFront.
+_EXCHANGE_INIT_BACKOFF_S = 30
 
 # OHLCV candles are re-fetched from /info by every strategy subprocess. With
 # ~20 strategies per instance sharing a handful of asset+timeframe combos, the
@@ -416,28 +421,27 @@ class HyperliquidExchangeAdapter:
         base_url = TESTNET_URL if testnet else MAINNET_URL
         self._base_url = base_url
 
-        self._info = self._build_info(base_url, allow_cache=True)
-        self._account_address = addr
+        self._wallet = None
         self._exchange = None
+        self._exchange_lock = threading.Lock()
+        self._exchange_init_error = None
+        self._exchange_backoff_until = 0.0
+        self._cached_spot_meta = None
+        self._cached_meta = None
         # Symbols we've already refreshed meta for and still couldn't find —
         # capped at one /info refresh per missing symbol per subprocess
         # lifetime, otherwise a typo or delisted asset would re-fetch meta
         # on every order operation. (PR #769 review point 2.)
         self._sz_decimals_misses: set[str] = set()
 
+        self._info = self._build_info(base_url, allow_cache=True)
+        self._account_address = addr
+
         if secret:
-            try:
-                import eth_account
-                wallet = eth_account.Account.from_key(secret)
-                account_addr = addr or wallet.address
-                self._account_address = account_addr
-                self._exchange = _HLExchange(
-                    wallet, base_url=base_url, account_address=account_addr
-                )
-            except Exception as e:
-                raise RuntimeError(
-                    f"Failed to initialize Hyperliquid Exchange client: {e}"
-                )
+            import eth_account
+            wallet = eth_account.Account.from_key(secret)
+            self._wallet = wallet
+            self._account_address = addr or wallet.address
 
     def _build_info(self, base_url: str, allow_cache: bool):
         """Construct an SDK Info instance, using the /tmp/hl_meta.json cache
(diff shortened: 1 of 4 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-26T06:00:46Z) in `64dc690`.

### Fixed
1. **Fully exempting 429s from the failure streak reintroduces #829 blind spot** — transient failures now accumulate on a separate `scriptFailureTransientTracker` with `scriptFailureTransientAlertThreshold=15`; brief storms stay WARN-only, sustained throttle escalates via `formatScriptFailureTransientAlert` (`scheduler/script_failure_alerts.go:20-24`, `218-230`). `clearScriptFailure` clears both trackers (`244-258`).
2. **`\b429\b` matches non-throttle messages** — removed bare word match; classifier now requires `(429,…)`, `HTTP/status 429`, `status_code=429`, `rate.?limit`, or `error from cloudfront` (`scheduler/script_failure_alerts.go:26-27`).

### Deferred to follow-up
(none)

---
Created with LLM: Composer | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-26T06:32:32Z) in `21371d6`.

### Fixed
1. **Fold duplicated alert-cadence state machine** — `ScriptFailureTracker.Record` now delegates to `recordScriptFailureAtThreshold` with `scriptFailureAlertThreshold`, so primary and transient trackers share one increment / signature-change / every-10th / hourly re-alert implementation (`scheduler/script_failure_alerts.go:83-85`).

---
Created with LLM: Composer 2.5 | high | Harness: Claude Code

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-26T06:41:11Z) in `8555da1`.

### Fixed
1. **Bound sustained-throttle escalation by wall-clock** — transient tracker now escalates when consecutive count ≥ `scriptFailureTransientAlertThreshold` (15) **or** elapsed time since `firstSeenAt` ≥ `scriptFailureTransientAlertMaxDuration` (75m, calibrated to 15×5m). Primary tracker unchanged (count-only). Recovery clears `firstSeenAt` via entry delete (`scheduler/script_failure_alerts.go:24-30`, `:61-66`, `:148-170`). Tests cover slow-interval wall-clock escalation, brief sub-bound storms, and post-recovery reset (`scheduler/script_failure_alerts_test.go`).

---
Created with LLM: Composer 2.5 | high | Harness: Claude Code

richkuo:
@claude sonnet rereview

richkuo:
Addressed review (claude · 2026-06-26T06:52:05Z) in `b82b7ce`.

### Fixed
1. **Exchange init backoff untested** — added three tests in `TestLazyExchangeInit`: rapid retry suppression within `_EXCHANGE_INIT_BACKOFF_S` (constructor called once), constructor retry after backoff expires, and cached exchange reuse after successful post-backoff init (`platforms/hyperliquid/test_adapter.py:930-1005`).

---
Created with LLM: Composer 2.5 | high | Harness: Claude Code

richkuo:
@claude sonnet rereview

PR #1132, opened 2026-06-29
fix(#1131): default manual-open ATR fetch to 1h when timeframe unset

### LINKED ISSUE #1131: [C20] Default manual-open ATR fetch to 1h candles when strategy timeframe is unset
**Complexity: 20/100** — scope: narrow (manual_atr.go + manual_atr_test.go); risk: low (ATR stamping only, no booking); uncertainty: none

## Problem

`fetchManualEntryATR` (`manual_atr.go:60`) guards on `sc.Timeframe == ""` and immediately returns false, falling through to the coarse heuristic ATR (`0.1 × fillPrice / leverage`, `manual_protection.go:49–53`, invoked at `manual.go:355` and `manual_limit.go:414`) rather than attempting a 1h fetch. Any manual strategy whose `timeframe` field is absent or empty silently loses the real ATR path even though 1h is the canonical default for everything else in the manual flow (the `init` wizard `init.go:1356` and the `generateConfig` generator `init.go:735–737` both default to `"1h"`).

## Goal

When `sc.Timeframe` is empty, `fetchManualEntryATR` should fall back to `"1h"` instead of failing closed. This keeps the ATR fetch attempt alive and only falls through to the heuristic if the actual network/OHLCV fetch fails.

## Approach

In `fetchManualEntryATR` (`manual_atr.go:59–81`): replace the hard guard on empty `sc.Timeframe` with a local variable that defaults to `"1h"` when the field is empty. Emit a stderr log line when the default is applied so callers can see which timeframe was used. Keep the guards on `sc.Script` and `sc.Symbol` (those have no sensible default).

The fix is centralized in the shared `fetchManualEntryATR` function, so both call sites — `manual_limit.go:407` and `manual.go:342` — inherit the new behavior with no change at the call sites.

Acceptance criteria:
- Manual-open with no `timeframe` set on the strategy config fetches ATR from 1h candles instead of jumping straight to the heuristic fallback.
- Stderr prints `[manual-open] defaulting to 1h ATR (strategy timeframe unset)` (or similar) when the default is applied.
- Update `TestFetchManualEntryATR_MissingFields` (`manual_atr_test.go:62–85`): drop the `no timeframe` subcase — empty timeframe is no longer a rejected field, so it no longer returns `ok=false` at the guard.
- Add a stubbed empty-timeframe test that asserts `runHyperliquidFetchATRFn` is invoked with `timeframe="1h"`. The empty-timeframe path MUST be exercised through the `runHyperliquidFetchATRFn` stub (as the existing `TestFetchManualEntryATR_Stub*` tests do) so the test never spawns Python — otherwise the unstubbed path would shell out to the real check script and violate the "Go CI must not depend on spawning Python" guardrail.

---
Created with LLM: Sonnet 4.6 | high | Harness: Claude Code

richkuo:
## Summary

`fetchManualEntryATR` (`manual_atr.go`) failed closed when a manual strategy's `timeframe` was unset — its guard rejected empty `sc.Timeframe`, so the manual-open dropped straight to the coarse heuristic ATR (`0.1 × fillPrice / leverage`) instead of attempting the real OHLCV fetch.

`1h` is the canonical default across the rest of the manual flow (the `init` wizard and `generateConfig` both default to `"1h"`). This change defaults an unset timeframe to `"1h"` and keeps the real fetch alive, only falling through to the heuristic on a genuine fetch failure. Guards on `sc.Script` / `sc.Symbol` are unchanged (no sensible default).

The fix is centralized in the shared `fetchManualEntryATR`, so both call sites — `manual.go` and `manual_limit.go` — inherit it with no call-site change.

## Changes

- `manual_atr.go`: drop the empty-timeframe rejection; default to `"1h"` with a stderr line `[manual-open] defaulting to 1h ATR (strategy timeframe unset)` when applied.
- `manual_atr_test.go`: remove the now-invalid `no timeframe` subcase from `TestFetchManualEntryATR_MissingFields`; add `TestFetchManualEntryATR_EmptyTimeframeDefaultsTo1h`, stubbed so it never spawns Python and asserting the fetch receives `timeframe="1h"`.

## Verification

- `gofmt -w` clean
- `go -C scheduler build .` — OK
- `go -C scheduler test .` — full suite passes (30.8s)

Closes #1131

---

### DIFF
--- a/scheduler/manual.go
+++ b/scheduler/manual.go
@@ -348,7 +348,7 @@ func runManualOpen(args []string) int {
 				} else {
 					entryATR = fetched
 					fmt.Fprintf(os.Stderr, "[manual-open] %s %s: --atr omitted; auto-fetched ATR=%.6f (period=14, %s)\n",
-						strategyID, sc.Symbol, fetched, sc.Timeframe)
+						strategyID, sc.Symbol, fetched, resolveManualATRTimeframe(sc))
 				}
 			}
 			if !fetchedOK {
--- a/scheduler/manual_atr.go
+++ b/scheduler/manual_atr.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 )
 
 // HyperliquidFetchATRResult is the JSON output from check_hyperliquid.py
@@ -57,10 +58,19 @@ var runHyperliquidFetchATRFn = RunHyperliquidFetchATR
 // and stamped ATR; if that becomes a problem, plumb a per-strategy ATR period
 // from sc.OpenStrategy.Params here.
 func fetchManualEntryATR(sc StrategyConfig) (float64, string, bool) {
-	if sc.Script == "" || sc.Symbol == "" || sc.Timeframe == "" {
-		return 0, "missing script/symbol/timeframe on strategy config", false
+	if sc.Script == "" || sc.Symbol == "" {
+		return 0, "missing script/symbol on strategy config", false
 	}
-	result, stderr, err := runHyperliquidFetchATRFn(sc.Script, sc.Symbol, sc.Timeframe, 14)
+	// 1h is the canonical default across the manual flow (the init wizard and
+	// generateConfig both default to "1h"), so an unset timeframe falls back to
+	// it rather than failing closed to the coarse heuristic ATR. Only a genuine
+	// fetch failure should drop callers to computeFallbackATR.
+	timeframe := sc.Timeframe
+	if timeframe == "" {
+		timeframe = "1h"
+		fmt.Fprintf(os.Stderr, "[manual-open] defaulting to 1h ATR (strategy timeframe unset)\n")
+	}
+	result, stderr, err := runHyperliquidFetchATRFn(sc.Script, sc.Symbol, timeframe, 14)
 	if err != nil {
 		msg := err.Error()
 		if stderr != "" {
--- a/scheduler/manual_atr.go
+++ b/scheduler/manual_atr.go
@@ -57,17 +57,27 @@ var runHyperliquidFetchATRFn = RunHyperliquidFetchATR
 // Strategies that override ATR period via params will see drift between fetched
 // and stamped ATR; if that becomes a problem, plumb a per-strategy ATR period
 // from sc.OpenStrategy.Params here.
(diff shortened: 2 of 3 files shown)


richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-29T02:51:55Z) in 0d242e2.

### Fixed
1. **Manual-open success log reported an empty timeframe when the 1h default was applied** — added `resolveManualATRTimeframe(sc)` as the single source of truth for the unset→`1h` default (`manual_atr.go`), and the success log at `manual.go:351` now prints the resolved value instead of the raw `sc.Timeframe`. Must-survive cases covered by `TestResolveManualATRTimeframe`: explicit `4h` stays `4h`, unset shows `1h`. The limit-fill path (`manual_limit.go:404`) logs no timeframe on success today, so parity is unchanged (no log added there).

### Not changed (refuted)
1. **"Empty-timeframe manual strategies are now half-supported (daemon close/protection path degraded)"** — the premise that such a config is loadable is false. `validateConfig` rejects `type=manual` with an empty timeframe at `config.go:1514`, and `validateConfig` is wired into `loadConfig` at `config.go:1225` — the path both the daemon and manual-open reach via `LoadConfig` (`manual.go:56`, `manual_limit.go:351`). Empirically, `LoadConfig` on a manual strategy with no timeframe returns `strategy[...]: type=manual requires timeframe (e.g. "1h")` and fails the load. So the args-fill gate at `config.go:1114` never sees an empty-timeframe manual strategy in production, and the degraded `runManualCloseEval` state cannot occur. The reviewer's note that this check "appears wired only into tests" is incorrect.

   The reviewer's preferred remedy — defaulting `sc.Timeframe = "1h"` in the `config.go:1106` normalization loop — would make empty-timeframe manual configs *valid* and thereby **remove an existing fail-closed safety guard**, masking a real misconfiguration instead of surfacing it. Per the repo's safety-over-cleanliness priority, the fail-closed behavior is kept. No follow-up issue filed; nothing to defer.

   (Note: this also means this PR's empty-timeframe defaulting in `fetchManualEntryATR` is defense-in-depth for direct `StrategyConfig` construction, not a production-reachable path — the loaded config always carries a non-empty timeframe. The change remains correct and harmless.)

Verification: `gofmt` clean, `go build .` OK, full `go test .` suite passes (30.2s).

---

richkuo:
@claude review

PR #1148, opened 2026-06-30
feat(#1140): add live perps force-close command

### LINKED ISSUE #1140: [C75] Add scheduler-owned perps force-close CLI path
**Complexity: 75/100** - scope: medium; risk: high (live Hyperliquid close plus scheduler state/trade reconciliation); uncertainty: medium (choose `manual-close` perps support vs a new command).

## Problem

`manual-close` is intentionally scoped to `type=manual` strategies. For live Hyperliquid `type=perps` strategies, operators need a scheduler-owned way to flatten an existing strategy-owned position without changing the strategy to `type=manual` or relying on an ad hoc exchange API call.

This comes up when a strategy's allowed direction changes and the existing position is on the old side. SIGHUP hot reload already blocks `direction` changes while a perps/manual position is open, but a restart with edited config only warns at startup and continues. The live reconciler may auto-close a sole-owner direction/regime orphan, while shared-coin cases intentionally skip auto-close to avoid touching peer exposure and require operator action.

Switching the strategy to `type=manual` is not a valid workaround for a normal perps strategy because existing perps-only fields and manual-only required fields conflict:

```text
strategy[hl-tcross-eth-live]: type=manual requires symbol (e.g. ETH)
strategy[hl-tcross-eth-live]: type=manual requires timeframe (e.g. 1h)
strategy[hl-tcross-eth-live]: margin_per_trade_usd is only supported for perps strategies (got type manual)
strategy[hl-tcross-eth-live]: regime_directional_policy is only supported for HL perps strategies (got platform="hyperliquid" type="manual")
```

Directly closing on Hyperliquid is not a complete operator workflow. For non-shared sole-owner coins, reconcile can detect the external close and book/remove the position. For shared/non-destructive cases, reconcile deliberately avoids mutating peer-sensitive exposure and can leave an operator-visible gap instead of a clean strategy-owned close path.

## Repro

1. Open a live `type=perps` Hyperliquid strategy with a long position.
2. Edit the config so the strategy now resolves to short, then restart rather than SIGHUP.
3. Startup warns about the state-vs-config gap, but the process continues.
4. Under `direction="short"` with an existing orphan long, a sell signal is skipped rather than closing and reopening; the executor leaves the long in place and asks for a manual close.
5. If the coin is shared with live peers, the reconcile orphan auto-close path is skipped by design, so the operator still needs a safe scheduler/CLI close.

## Expected

A clean operator path to close a live HL perps strategy position through the scheduler/CLI that:

- Accepts a `type=perps` strategy ID without changing the strategy type.
- Places the reduce-only close on the venue, with optional partial quantity support.
- Cancels strategy-owned SL/TP OIDs on full close while preserving shared-coin peer exposure.
- Records the close in the strategy's internal state/trade ledger with fill price, fee, OID, and realized PnL.
- Deletes or shrinks the tracked position through the scheduler-owned mutation path.
- Leaves the strategy able to reopen under the updated effective direction after the close is drained.

## Suggested Direction

Either extend `manual-close` with a perps branch or add a dedicated `force-close` subcommand. A new subcommand may be clearer because `manual-close` currently means `type=manual`; if added, wire it through command dispatch, `knownSubcommands`, `agent-info`, docs, and tests.

The implementation should reuse the existing HL close and state-adoption patterns rather than relying on the reconciler as the primary operator workflow:

- Resolve the perps symbol from the strategy args.
- Use the same close-side, shared-coin scope, and protection-OID cancellation rules as the existing manual close path.
- Queue or otherwise adopt the fill through a scheduler-owned mutation path so trade/state writes happen under the daemon's normal locking and persistence model.
- Keep owner/shared-coin guards fail-closed.

## Acceptance Criteria

- `go-trader force-close <perps-strategy-id>` or `go-trader manual-close <perps-strategy-id>` can close a live HL perps position owned by that strategy.
- A full close removes the virtual position, records a close trade, updates cash with realized PnL/fee, and clears owned protective OIDs.
- A partial close shrinks the virtual quantity and leaves appropriate protection behavior intact.
- Shared-coin behavior does not flatten peer exposure; sized closes and owner checks preserve the existing safety model.
- The command refuses unsupported strategy types/platforms with clear errors.
- Tests cover the perps command path, drain/adoption path, owner mismatch, shared-coin close scope, and command registry/agent-info updates.

## Use Case

Reverse a strategy's directional bias (long -> short or short -> long) without waiting for the natural signal cycle and without creating state/exchange drift.

---
Updated with LLM: GPT-5 | high | Harness: Codex

richkuo:
Closes #1140

## Summary

Adds `go-trader force-close <strategy-id>` for live Hyperliquid `type=perps` strategy positions.

- Validates the target is a live HL perps strategy and owns the virtual position.
- Submits the reduce-only close through the existing Hyperliquid close script.
- Queues the confirmed fill as `PendingManualAction{Action:"close"}` so scheduler-owned state/trade mutation stays in the daemon drain path.
- Records force-close trade metadata separately from manual trades and rejects paper perps drains.
- Documents the command and exposes it through `agent-info`.

## Verification

- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'Test(ApplyManualAction_PerpsForceClose|RunForceCloseQueues|KnownSubcommandsMatchDispatch|AgentInfoCommandsCoverKnownSubcommands)'`


richkuo:
@claude review

richkuo:
@claude review

richkuo:
Addressed review (claude · 2026-06-30T21:36:04Z) in 01aeed1.

### Fixed
1. **Force-close could cancel protection before a failed close** — confirmed and fixed: force-close now uses a post-fill cancellation mode (`scheduler/hyperliquid_balance.go:105`, `scheduler/executor.go:570`), and the close script only cancels protection after the fill covers the requested size (`shared_scripts/close_hyperliquid_position.py:173`); error and underfill cases are covered by regression tests (`shared_scripts/test_close_hyperliquid_position.py:435`, `shared_scripts/test_close_hyperliquid_position.py:458`).

### Recommended Optional
1. **Drain could double-book a close OID already observed by reconcile** — confirmed and fixed: pending close actions now skip accounting and quantity mutation when `strategyHasCloseTradeForOID` already sees the exchange OID, while clearing queued canceled protection on residual positions (`scheduler/manual.go:1201`; tests at `scheduler/manual_test.go:871` and `scheduler/manual_test.go:950`).

### Verification
1. `uv run --extra test pytest shared_scripts/test_close_hyperliquid_position.py -q`
2. `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test ./...`
3. `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler build .`
4. `git diff --check`

---
Created with LLM: GPT-5 | high | Harness: Claude Code

richkuo:
@claude review

PR #1153, opened 2026-06-30
fix(#1139): normalize composite regime timeframe config

### LINKED ISSUE #1139: [C75] Composite regime config: rename ATR-normalized thresholds + define regime timeframe override
**Complexity: 75/100** - scope: broad (two independently landable config changes across Go, Python, probes, docs, and tests); risk: high (regime labels feed live entry gating, close-evaluator label selection, and protection planning); uncertainty: moderate (hot-reload and backtest parity need explicit implementation contracts).

## Validation note

This issue combines two real but separable defects. Decompose before implementation:

1. **Composite threshold rename/alias.** Rename the operator-facing composite threshold keys so they match the ATR-normalized metrics they gate.
2. **Regime timeframe override.** Add a configurable regime-data timeframe without changing the strategy trading timeframe.

Both parts deliver value independently and should be implemented/tested independently unless this issue is converted into an umbrella.

### 1. `return_pct` and `range_pct` are misnamed

The composite threshold names read like price percentages, but they gate ATR-normalized efficiency metrics:

```text
return_eff = (close_end - close_start) / (atr * period)
range_eff  = (high_max - low_min) / (atr * period)
big_move   = abs(return_eff) >= thresholds["return_pct"]
wide       = range_eff >= thresholds["range_pct"]
```

Current code paths:

- Python defaults: `shared_tools/regime.py:97-102`
- Python metric computation: `shared_tools/regime.py:150-167`
- Python threshold lookup: `shared_tools/regime.py:188-197`
- Go mirror: `scheduler/regime_window_spec.go:15-31`

So `return_pct: 0.05` means "net move >= 5% of the window's total straight-line ATR travel," not "5% price return." `range_pct` has the same naming defect. `adx` and `efficiency` already match their metrics.

Important correction from validation: Go does **not** currently have an unknown-key guard for nested composite threshold keys. `RegimeWindowsMap.UnmarshalJSON` uses plain `json.Unmarshal` into `RegimeWindowSpec`, and Python merges whatever `thresholds` dict is present but `map_composite_label` only reads `return_pct` / `range_pct`. Without explicit alias parsing, new-only `return_eff` / `range_eff` keys would be silently ignored/dropped rather than reliably rejected.

### 2. Non-options regime timeframe is locked to the trading timeframe

For non-options strategies, `strategyRegimeBundleRequest` derives the regime bundle timeframe from strategy positional args:

- `strategyArgSymbolTimeframe(sc.Args)`: `scheduler/regime_store.go:248-260`
- non-options request uses that timeframe: `scheduler/regime_store.go:300-309`

There is no `RegimeConfig.Timeframe` today (`scheduler/config.go:72-87`). Options are already decoupled through `optionsRegimeTimeframe = "4h"` (`scheduler/regime_store.go:40-49`, `:281`).

The existing regime fetch layer can support the override:

- bundle key includes `Timeframe`: `scheduler/regime_store.go:59-64`
- due-strategy collection dedupes by the whole key: `scheduler/regime_store.go:316-343`
- `check_regime.py` accepts arbitrary `--timeframe`: `shared_scripts/check_regime.py:191-203`

The granularity problem is real by formula: the denominator is `ATR * period`, so a large calendar span expressed as many small candles can classify differently from the same span expressed as fewer larger candles. Do not use any unstored BTC numeric example as an acceptance fixture unless backed by a committed reproducible data source.

## Goal

Make composite regime thresholds self-describing, and let regime detection run on a configurable candle timeframe without changing the trading timeframe or silently changing ATR-based exit geometry.

### Part 1: Threshold rename and aliases

- Add canonical threshold keys `return_eff` and `range_eff`.
- Keep `return_pct` and `range_pct` as deprecated aliases.
- If old and new keys are both present, `return_eff` / `range_eff` win.
- Add explicit alias-aware parsing and unknown-key validation for composite thresholds in Go; do not rely on struct unmarshalling.
- Normalize aliases before Python merges thresholds over `_DEFAULT_COMPOSITE_THRESHOLDS`.
- Keep Go and Python defaults in sync.
- Update inspect/display text, probe JSON, docs, and strategy comments to use the canonical names.

### Part 2: Configurable regime timeframe

- Add optional top-level `regime.timeframe`.
- For non-options strategies, use `regime.timeframe` when set; otherwise keep today's fallback to `args[2]`.
- Preserve current options behavior unless the implementation explicitly defines that `regime.timeframe` also applies to options.
- Validate the override at config load using the same timeframe vocabulary accepted by data fetchers/check scripts.
- Hot reload contract: if the field is hot-reloadable, implement all three pieces:
  - mask it out of the restart-required regime comparison, similar to `display_windows`;
  - copy it in `applyHotReloadConfig` when flat;
  - reject changes while affected strategies have open positions in `validateHotReloadStateCompatible`.
- Backtest contract: `run_backtest.py --config` must honor `regime.timeframe` by computing regime labels from the override timeframe and aligning them to the trading-frame bars without look-ahead. A loud reject is acceptable only if the issue is explicitly narrowed away from backtest parity.

## ATR / close-evaluator audit

The regime label is safe to decouple, but the implementation must preserve ATR sizing semantics unless intentionally changed.

Validation found most current protection geometry uses position/trading-frame ATR, not regime-store metrics:

- on-chain protection plan uses `pos.EntryATR` and `pos.riskAnchorPrice()`: `scheduler/hyperliquid_protection.go:34-84`
- trailing stop distance uses frozen `pos.EntryATR`: `scheduler/hyperliquid_trailing_stop.go:73-164`
- check scripts pass close evaluators `market["atr"]` from the strategy dataframe, not from the regime bundle: `shared_scripts/check_hyperliquid.py:186-207`
- `tiered_tp_atr_live_regime` can use live `market["atr"]` with entry fallback: `shared_strategies/close/tiered_tp_atr_live_regime.py:1-30`

So the main expected effect of `regime.timeframe` is label selection and therefore per-regime tier/multiplier selection. It should not silently change SL/TP distance ATR source unless the code and docs say so.

## Touch points

- `shared_tools/regime.py` - defaults, alias normalization, lookup, tests
- `scheduler/regime_window_spec.go` - threshold struct/parsing, validation, display, tests
- `scheduler/config.go` - `RegimeConfig.Timeframe`, config validation
- `scheduler/regime_store.go` - non-options timeframe override routing
- `scheduler/config_reload.go` - restart/hot-reload compatibility and apply path
- `scheduler/version_probe.go` - composite probe threshold JSON
- `shared_strategies/open/regime_adaptive.py` - comments that reference old names
- `backtest/run_backtest.py` / `backtest/backtester.py` - parity for `regime.timeframe`, or explicit reject if narrowed
- `backtest/tests/test_backtester_regime.py` - add canonical-key coverage; keep alias regression
- `shared_tools/test_regime.py`, `test_regime_injection.py`, `scheduler/regime_window_spec_test.go`, and relevant regime-store/config-reload tests
- Docs: `docs/POST_UPDATE_HISTORY.md`, `docs/ARCHITECTURE.md`, `CLAUDE.md`, `AGENTS.md`

Do not touch unrelated `return_pct` names such as backtest/reporting metrics (`total_return_pct`, `annual_return_pct`) or `fee_audit.py` return fields.

## Acceptance criteria

- Existing configs using `return_pct` / `range_pct` classify identically after the change.
- New configs using `return_eff` / `range_eff` classify identically to equivalent old-key configs.
- If both old and new aliases are present, canonical keys win and the behavior is covered by tests.
- Unknown composite threshold keys are rejected explicitly instead of silently ignored.
- Python default dict and Go threshold defaults remain in sync.
- `regime.timeframe` unset produces byte-identical live regime bundle signatures to today.
- `regime.timeframe` set on a 1h trading strategy fetches regime data at the configured timeframe while entries/exits still execute on the 1h trading loop.
- Invalid `regime.timeframe` values are rejected at config load.
- Hot reload behavior is fully defined and tested: either restart-required, or flat-only apply plus open-position rejection.
- Backtest behavior is fully defined and tested: either parity support with separate regime timeframe data, or a loud config reject if the scope is explicitly narrowed.
- ATR/close-evaluator audit is documented; strategies that do not set `regime.timeframe` see no SL/TP geometry changes.
- Probe/listing/startup checks remain green.

---
Updated with LLM: GPT-5 | high | Harness: Codex

richkuo:
## Summary
- Rename composite regime thresholds to return_eff/range_eff, while accepting return_pct/range_pct as deprecated aliases with canonical-key precedence.
- Add regime.timeframe for non-options regime bundles, directional certification identity, hot reload, and backtest alignment; options keep their existing 4h path.
- Update probes, docs, and Go/Python regression coverage.

## Verification
- env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test ./...
- env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler build .
- .venv/bin/python -m pytest shared_tools backtest/tests
- .venv/bin/python -m py_compile backtest/run_backtest.py backtest/backtester.py shared_tools/regime.py shared_strategies/open/regime_adaptive.py

Closes #1139


richkuo:
@claude review

richkuo:
Addressed review (claude[bot] · 2026-06-30T22:51:46Z) in 6990829.

1. **Cross-timeframe regime alignment leaked the in-progress HTF candle onto LTF bars.** Confirmed: `_aligned_regime_columns` reindexed HTF regime columns (`load_cached_data`-indexed by candle OPEN time) directly onto the LTF trade index via `reindex(..., method="ffill")`, so an LTF bar inside a still-forming HTF candle saw that candle's label before it actually closed — the test `test_regime_timeframe_override_aligns_without_lookahead` (`backtest/tests/test_backtester_composite_regime_wiring.py:284`, pre-fix) literally asserted this leaky behavior under a name claiming the opposite. The engine's one-row `shift(1)` (`backtester.py:1438`) only protects bar-to-bar at LTF granularity and can't compensate for an HTF candle spanning multiple LTF bars. Implemented: shift the HTF columns by one row before reindexing (`run_backtest.py:273` `_aligned_regime_columns`), so a label only becomes visible once its candle closes — at the next HTF row's index — matching live's last-closed-candle `latest_regime` and this repo's documented look-ahead invariant ("regime gate reads N-1 regime"). Applied the identical fix to `_profile_label_series`'s cross-timeframe branch (`run_backtest.py:330`), which had the same bug and wasn't covered by any existing test. Rewrote the existing test to assert true no-lookahead behavior and added a new test for `_profile_label_series`. Full pytest suite (1991 tests) and `go -C scheduler test ./...` both pass.

   Alternatives rejected: (a) leaving it as the deliberate #304 M2 convention — rejected because that convention's own pattern (`_htf_trend_series`) is itself a pre-existing leak, not a deliberately-correct design, and the new #1139 code's docstring explicitly (and incorrectly) claimed look-ahead safety; (b) fixing `_htf_trend_series` in this same PR — rejected as out of scope: it predates #1139, is unrelated to the regime-timeframe feature, and changing it would shift established backtest baselines for every HTF-filter strategy, deserving its own dedicated validation pass. Filed richkuo/go-trader#1154 to track that follow-up. Override if you'd prefer `_htf_trend_series` fixed inline here instead.

---

richkuo:
@claude review

PR #1174, opened 2026-07-01
ci: harden the Claude Code workflow — comment pagination, concurrency, tool allowlist

richkuo:
## Summary

Hardens `.github/workflows/claude.yml` and its prompt/scripts following a quality review of the Claude GitHub action: three live defects fixed, plus concurrency, security, and robustness hardening.

## Changes

1. **Comment patching paginates.** Both patch steps fetched only the first page (30 comments) of the issue thread, so on long review threads the LLM footer / failure note landed on a stale claude[bot] comment. Now `gh api --paginate --slurp` piped to `jq` flattens all pages (pattern verified read-only against a live thread).
2. **Per-issue concurrency group** (`cancel-in-progress: false`) — overlapping runs on the same issue/PR could stamp each other's comments with the wrong model/effort.
3. **Narrowed tool allowlist.** `Bash(git *)`/`Bash(gh *)` become explicit read verbs plus exact-match push rules (`git push origin HEAD` and its `-u`/`--set-upstream` variants — no trailing wildcard, so a refspec like `HEAD:refs/heads/main` or `--force` does not match) and `gh pr comment`. `gh api` is dropped entirely (its `--method` flag reaches every mutating endpoint). Rationale: the reviewer reads untrusted PR content while the job token holds `contents: write` + `pull-requests: write`, so a prompt-injected diff must not be able to reach `gh pr merge` or push outside the current branch. Note: in-workflow Claude should use `gh pr view --comments` / `gh issue view` where it previously used `gh api` reads.
4. **Reviews explicitly do not gate on CI.** `pr-review-format.md` now instructs the reviewer not to gate the verdict on CI status or wait for checks — CI intentionally runs longer than the review, so a CI-gated verdict could never emit LGTM at verdict time; branch protection enforces CI separately. Accordingly no CI-reader commands (`gh pr checks`, `gh run view/list`) are allowlisted and no `actions: read` permission is granted.
5. **CLAUDE.md revisions persist.** The revise step edited CLAUDE.md in the runner workspace but nothing committed or pushed it (the skill contains no commit step), so every revision was silently discarded. The prompt now instructs commit + `git push origin HEAD`; the step also gets `--effort` (previously CLI default while the main run used the resolved effort) and a narrowed allowlist.
6. **Prompt-file guard extended** to backticks and dollar signs — `claude_args` is shell-evaluated, so those would expand/execute, not just break quoting like the already-guarded double quote.
7. **`issues: assigned` trigger dropped** — any assignment on an issue whose body mentions @claude re-ran the full job.
8. **Shared, tested patch logic + timeout.** The duplicated fetch/strip/patch bash in the footer and failure steps is extracted to `patch_claude_comment.sh` + `compose_claude_comment.py` (model-name mapping, footer, status-note stripping — idempotent across retries; the failure path now also gets the Create-PR-link footer rewrite and a pre-checkout guard). `rewrite_create_pr_link.py` is refactored into an importable function with identical CLI behavior. `.github/scripts` added to pytest `testpaths` so these tests now run in CI. `timeout-minutes: 45` bounds hung runs (recorded runs peak at 15 minutes; GitHub default is 360).

## Review round 1

Both findings of the first @claude review are addressed by commit 9f3e340: the missing `checks: read`/`statuses: read` permissions are resolved by removing the CI gate itself (intentional — see item 4), and the overbroad `git push origin HEAD*` wildcard is now exact-match (item 3).

## Verification

- `actionlint` and `shellcheck` clean.
- Full pytest suite: 2021 passed, including 27 tests in `.github/scripts` (12 new for `compose_claude_comment.py`; existing subprocess-based tests for the refactored `rewrite_create_pr_link.py` pass unchanged).
- The compose-prompt flatten + extended guard re-run locally against the edited `pr-review-format.md`.
- The workflow itself can't be exercised until merged to the default branch; the failure-note path and revise-step push are the two spots to watch on the first post-merge runs.

---
LLM: Fable 5 | xhigh | Harness: Claude Code

https://claude.ai/code/session_01DGbfne6icXKQQaqSxvtX6D


richkuo:
@claude review

---
LLM: Fable 5 | xhigh | Harness: Claude Code

richkuo:
@claude review

---
LLM: Fable 5 | xhigh | Harness: Claude Code

richkuo:
@claude sonnet review

PR #1185, opened 2026-07-02
ci(#1178): least-privilege token for the Claude review path

### LINKED ISSUE #1178: [C65] Give the Claude review path a least-privilege token
**Complexity: 65/100** — scope: medium (two-job split + `github_token` passthrough + bot-identity parameterization across workflow and scripts); risk: high (auto-protective security mechanism; review runs process untrusted, potentially prompt-injected PR content); uncertainty: high (app-token downscoping and `id-token` requirements must be confirmed empirically, and the bot-identity change has design side effects).

## Problem

`.github/workflows/claude.yml` runs both the `@claude review` and `@claude <implement>` triggers from the same job, so both share one `permissions:` block (`contents: write`, `pull-requests: write`, `issues: write`, `id-token: write`). The review path only ever needs to post a comment — it never needs to push commits or merge.

The git/gh Bash allowlist added in #1174 narrows *which* git/gh invocations are reachable during review (exact-match push rules, no `gh api`), but it is not a hard boundary: allowed interpreters (`python3`, `go`, `uv run`) can shell out to `gh`/`git` directly, bypassing the allowlist's command-string matching entirely.

**Corrected token model (this changes the fix):** the workflow passes no `github_token` to `anthropics/claude-code-action@v1`, so the agent's git/gh operations run on the **Claude GitHub App installation token** (minted via OIDC exchange — this is what `id-token: write` is for), not on the job's `${{ github.token }}`. Per the action's docs, that app token's capabilities come from the app installation (read/write Contents, Issues, Pull Requests), **not** from the job's `permissions:` block. The job token only backs `actions/checkout` and the comment-patch steps (`patch_claude_comment.sh`). So today, the push-capable credential in a review run is the app token — and narrowing the job `permissions:` alone would not restrict it at all.

## Goal

Reviews cannot reach `contents: write` capabilities (push, merge) even if a prompt-injected PR body tricks the agent into shelling out through an allowed interpreter.

- Split into two **jobs** (permissions are job-scoped; a step cannot narrow them): a review job and an implementation job, both under the existing `claude-<number>` concurrency group so comment-patching stays serialized.
- **Review job:** pass `github_token: ${{ github.token }}` to the action (the documented pattern for binding the agent to the job token) with permissions `contents: read` (required for `actions/checkout`), `pull-requests: write`, `issues: write` — no `contents: write`. Verify whether `id-token: write` can then be dropped.
- **Implementation job:** keeps the current app-token setup and `contents: write`, since it legitimately commits and pushes.
- **Bot identity:** with `github_token`, the agent's comments post as `github-actions[bot]`, not `claude[bot]`. Parameterize the bot login in `patch_claude_comment.sh` (currently hard-filters `claude[bot]`), `allowed_bots` in the workflow, and the CLAUDE.md "fetch latest bot review" jq snippet.
- **Trigger classification:** define an explicit review-vs-implement rule. Event type alone misroutes — inline `pull_request_review_comment`s can carry implementation asks ("@claude fix X"), and a bare `review` keyword match misclassifies "@claude review and fix". Misclassifying an implementation ask as review must fail closed (push rejected), never the reverse.
- Correct the now-known-wrong token-model prose: the workflow comment in `claude.yml` (the "real backstop … is the job token's scope" block) and the corresponding CLAUDE.md #1174 section.
- **Empirical verification:** on a test PR, confirm a push attempt from the review job is rejected (403); confirm the comment-patch steps still stamp the footer under the new bot identity.
- Re-run `actionlint`/`shellcheck` and the `.github/scripts` pytest suite after the change.

Raised as a follow-up from PR #1174 review round 2 (non-blocking, but flagged as `### Create Follow-up Issue` since it's a genuinely separate scope requiring its own design decision).

---

Claude-Session: https://claude.ai/code/session_01DGbfne6icXKQQaqSxvtX6D
Claude-Session: https://claude.ai/code/session_01BJFWLPaaY7KhR5SyL94dcp

richkuo:
Closes #1178

## What changed

- **`.github/workflows/claude.yml`** is now a `classify` job (invocation check, review-vs-implement mode, model/effort resolution — `permissions: {}`) plus two thin caller jobs that `uses:` the new reusable **`.github/workflows/claude-run.yml`** body, so only the call-site `permissions:` differ.
- **Review job** — `contents: read`, `pull-requests: write`, `issues: write`, and **no `id-token: write`**: the job passes `github_token: ${{ github.token }}` to the action so the agent is bound to the job-scoped token (with app auth, its push power would come from the Claude GitHub App installation token, which job permissions do not cap), and without `id-token` a prompt-injected interpreter cannot mint that app token via the OIDC exchange either. The review-mode Bash allowlist additionally drops the `git add/commit/push` rules.
- **Implement job** — unchanged capability: app auth (`id-token: write`) + `contents: write`, plus the commit-detection and CLAUDE.md-revision steps (implement-mode-only in `claude-run.yml`).
- **Fail-closed mode routing** — `pull_request_review`/`pull_request_review_comment` events always classify as review; `issue_comment`/`issues` classify as review when the code-block-stripped body contains the word "review" (so "@claude review and fix" deliberately loses push; a misrouted implement ask 403s, never the reverse).
- **Bot identity** — review-run comments post as `github-actions[bot]`, so `patch_claude_comment.sh` gained a `BOT_LOGIN` parameter (default `claude[bot]`; `claude-run.yml` sets it per mode) and the CLAUDE.md "fetch latest bot review" snippet now accepts both logins. The CLAUDE.md workflow section is rewritten to the corrected token model.

## Verification

- New regression tests `.github/scripts/test_patch_claude_comment.py` (stubbed `gh` on PATH; default login, `BOT_LOGIN` override, no-match no-op) — written first and confirmed red against the old hard-coded filter, green after.
- Full pytest suite: 2102 passed. `shellcheck` on `patch_claude_comment.sh` and `actionlint` on both workflows: clean.
- **Not yet verified live (needs post-merge test):** `issue_comment` workflows run from the default branch, so the split only takes effect after merge. Per the issue's acceptance criteria, run `@claude review` on a test PR and confirm (1) a push attempt from the review job is rejected (403), (2) the action tolerates the missing `id-token: write` when `github_token` is provided, and (3) the footer patch lands on the `github-actions[bot]` comment.

---

Claude-Session: https://claude.ai/code/session_01BJFWLPaaY7KhR5SyL94dcp


richkuo:
@claude review

richkuo:
Addressed review feedback (claude[bot] · 2026-07-02T06:56:35Z) in a30c1a8.

### Fixed
1. **Footer-patch step keyed on the shared `github-actions[bot]` identity, making "latest bot comment" a non-unique target** — the action exposes no tracking-comment-id output (checked `action.yml`: only `execution_file`, `branch_name`, `github_token`, `structured_output`, `session_id`), so the exact-id remedy isn't available; instead the tracking comment's embedded run link is used as the discriminator. `patch_claude_comment.sh:24-34` now takes `RUN_ID` and selects only comments whose body links `/actions/runs/<run_id>` (boundary-anchored, so run 22 never matches runs/222), and both patch steps in `claude-run.yml` pass `RUN_ID: ${{ github.run_id }}`. Deliberately no fallback on miss — a missing footer is benign, stamping a foreign workflow's comment is not. All three must-survive cases are regression-tested in `test_patch_claude_comment.py` (newer same-author comment from another run; no comment from this run → clean no-op; prefix-collision run ids). 34/34 `.github/scripts` tests pass; `shellcheck` + `actionlint` clean.

---

Claude-Session: https://claude.ai/code/session_01BJFWLPaaY7KhR5SyL94dcp

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (claude[bot] · 2026-07-02T07:10:44Z) in 6b55f40.

### Fixed
1. **Failure/cancellation note silent when no tracking comment exists for this run** — `patch_claude_comment.sh` now takes `ON_MISS=post` (set only by the failure step, `claude-run.yml:237-249`): on a run-id miss with a non-empty `STATUS_NOTE` it posts a NEW comment instead of exiting silently, so pre-comment failures (checkout, `uv sync`, prompt composition) always surface in the thread without misattribution to an older run's comment. The footer-patch caller keeps the deliberate no-op miss. `compose_claude_comment.py` omits an empty body so the standalone note leads the comment. All three must-survive cases covered: pre-comment failure posts new (`test_on_miss_post_creates_new_status_comment`), existing own-run comment still patches rather than posts (`test_on_miss_post_still_patches_when_own_comment_exists`), older-run comment never receives the note (run-id filter, unchanged); plus a no-STATUS_NOTE guard so a footer-only comment is never posted.
2. **Fence-strip pipeline triplicated across `verify_invocation`/`classify_mode`/`resolve_model`** — `verify_invocation` now strips once and exports the body as a step output (`claude.yml:76-89`) consumed by both downstream steps, so the invocation gate and the push-access gate read the byte-identical text by construction. Output-passing chosen over a shared script file so the `classify` job stays checkout-free at `permissions: {}`; the heredoc delimiter is randomized (`/dev/urandom`) because the body is attacker-influenced and a predictable delimiter would allow output injection. Both must-survive cases hold: a future regex edit lands in exactly one place, and any construct strips identically because downstream never re-strips.

38/38 `.github/scripts` tests pass; `shellcheck` + `actionlint` clean.

---

Claude-Session: https://claude.ai/code/session_01BJFWLPaaY7KhR5SyL94dcp

richkuo:
@claude sonnet review

PR #1192, opened 2026-07-02
fix(#1190): give the kill-switch reset DM real context

### LINKED ISSUE #1190: Kill switch reset DM lacks context (trigger reason, account, trading instance) — bare 'reply reset' misleads operator
## Problem

When the portfolio-wide kill switch fires, the DM the scheduler sends to ask the operator to reply "reset" is a single bare sentence with no provenance:

```
Kill switch active. Reply 'reset' to resume trading.
```

This kill switch is **account/process-wide** — a single `PortfolioRisk.KillSwitchActive` flag covers the whole trader instance, not any individual strategy (`scheduler/risk.go`). There is no per-strategy variant of this mechanism. The DM gives the operator no information about:

- **Why** it fired (drawdown vs. margin, the actual percentages/dollar values)
- **Which account/wallet** (e.g. the Hyperliquid address) is affected
- **Which trading instance** sent it, when multiple deployments exist (live vs. paper, or separate instances per `systemd/go-trader@.service` `%i`)
- **What state the close is in** — positions may already be flattened, or a flatten may still be retrying

## Impact

When multiple trader deployments exist (live, paper, per-asset instances), this bare DM is ambiguous and panic-inducing:

- Which deployment sent it?
- Is my account already flat, or is a close still retrying?
- Does replying 'reset' close positions / touch P&L, or does it only clear the latch?

## Current behavior (verified)

- The scheduler already builds a much richer message (`formatKillSwitchMessage`, `scheduler/kill_switch_close.go`) — drawdown/margin reason with numbers, per-platform close report, GAPS/LATCHED variants — and broadcasts it to all Discord channels (`scheduler/main.go`) every cycle the kill switch stays fired.
- The bare "Kill switch active..." text is a **separate** DM used only to solicit the 'reset' reply (`scheduler/main.go`, `AskOwnerDM`) — it does not reuse any of that already-computed context.
- Replying 'reset' only clears the `KillSwitchActive` latch (`scheduler/main.go`); it does not itself touch positions. Positions are independently flattened earlier in the same cycle by `planKillSwitchClose` (`scheduler/kill_switch_close.go`).
- There is currently **no concept of a trader-instance name/label** anywhere in `Config` (`scheduler/config.go`) — this needs to be added (e.g. a new optional config field, or a label derived from the DB/config file path), not just included in a format string.
- The Hyperliquid wallet address (`hlAddr`) is already computed in-process at the DM call site and can be surfaced directly with no new plumbing.

## Suggested Fix

Reuse the context already computed at the DM call site instead of sending a bare sentence. At minimum include:
- **Trigger reason** with numbers (already computed as `portfolioReason` / `plan.DiscordMessage`)
- **Account/wallet** (`hlAddr`, already available)
- **Trading instance** — requires adding a label (new config field or derived from the DB/config path); flag this as part of the implementation, not just message formatting
- **What 'reset' does and does not do** — it only clears the kill-switch latch; it does **not** itself close or manage positions (those are handled separately by the kill-switch close plan, which may already be confirmed-flat or still retrying — the message should say which, using the existing `plan.OnChainConfirmedFlat` signal, and must not claim positions are safely "managed by trailing stops" when a flatten is still retrying and stop-losses may already have been cancelled)

Example improved message:
```
[KILL SWITCH — ACCOUNT-WIDE] go-trader-live · Hyperliquid 0x736d...cc86
Trigger: max drawdown 12.3% > 12.0% (value=$X, peak=$Y)
Status: positions confirmed flat (or: close retrying, will retry next cycle)

Reply 'reset' to clear the kill switch latch so trading can resume next cycle.
This does NOT itself close positions — that already happened (or is retrying) above.
```

This is a single account/process-wide mechanism — there is no per-strategy version to disambiguate against.

## Related

- Circuit breaker messages already include strategy context (`CIRCUIT BREAKER [hl-tcross-eth-live] ...`, `scheduler/circuit_breaker_alert.go`) — that's a genuinely separate, already-well-labeled per-strategy mechanism, not this one.

Reference only: LLM model used: moonshot/kimi-k2.7-code

---
Updated with LLM: Sonnet 5 | high | Harness: Claude Code

richkuo:
## Summary

- The DM the scheduler sends to ask the owner to reply 'reset' after the portfolio-wide kill switch fires was a bare "Kill switch active. Reply 'reset' to resume trading." — no drawdown reason, no account identity, no trading instance, no explanation of what 'reset' does or doesn't do.
- `formatKillSwitchResetPrompt` (`scheduler/kill_switch_close.go`) now reuses the same reason/close-report context already broadcast to Discord via `formatKillSwitchMessage`, prefixes it with a trader-instance label (`killSwitchInstanceLabel`, derived from the deployed config path, e.g. `/var/lib/go-trader/<instance>/config.json`) and the Hyperliquid wallet address, and states plainly that 'reset' only clears the kill-switch latch — it never itself closes or protects a position.
- When the close plan hasn't confirmed flat (LATCHED, RETRYING), the prompt also warns that resting stop-losses may already be cancelled ahead of the flatten attempt, so the owner doesn't assume protection is still in place.
- This kill switch is account/process-wide (a single `PortfolioRisk.KillSwitchActive` flag) — there is no per-strategy variant, so the message intentionally does not attribute the trigger to a strategy.

## Test plan

- [x] `go build ./...` from `scheduler/`
- [x] `go vet ./...`
- [x] `gofmt -l` on changed files (clean)
- [x] `go test ./...` (full suite passes)
- [x] New regression tests (`TestKillSwitchInstanceLabel_*`, `TestFormatKillSwitchResetPrompt_*`) proven red on the unfixed code (compile failure without the new functions), green after the fix

Closes #1190

---

### DIFF
--- a/scheduler/kill_switch_close.go
+++ b/scheduler/kill_switch_close.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


richkuo:
@claude review

Issue #1152, opened 2026-06-30
[C68] Retune regime ratchet-tier ladders + B2 ranging TP group (deferred from #1120)


richkuo:
**Complexity: 68/100** — scope: medium (per-substate `ratchetTierGroupDefaults` + `regimeTPTierGroupDefaults` B2 ranging split, Go/Python mirrors, M6 entry-locked validation); risk: high (close/trail geometry is auto-protective and directly affects P&L); uncertainty: medium-high (bar-level OOS cannot isolate per-substate ratchet/TP geometry — needs M6 `exit_policy_ab.py` entry-locked replay per #1066).

## Context

#1120 / PR #1149 retuned only the **opening-trail** table (`regimeATRDefaults.Trailing` / `REGIME_ATR_DEFAULTS_TRAILING`). Ratchet tier ladders and the B2 ranging TP group were **deliberately left unchanged** because the bar-level harness (`regime_1120_trail_validation.py`) cannot isolate per-substate ratchet/TP geometry without M6 entry-locked replay — see `backtest/research/regime_1120_trail_validation.json:ratchet_tp_note`.

Once #1120 closes, this deferred money-path geometry work needs a tracked home.

## Scope

1. **`ratchetTierGroupDefaults`** (`scheduler/trailing_tp_ratchet.go`) — three composite ranging ladders (`ranging_quiet`, `ranging_volatile`, `ranging_directional`) currently marked as unvalidated priors in code comments. Revisit ATR trigger spacing and close fractions per substate volatility/directional character (#1120 goals 1–2).

2. **`regimeTPTierGroupDefaults` B2 ranging group** (`scheduler/regime_atr.go`) — today all composite `ranging_*` substates collapse to the same 2-rung fast exit via `regimeCloseDefaultGroup`. Evaluate substate differentiation mirroring the ratchet's three-ladder split (#1120 goal 2).

3. **Python mirrors** — `DEFAULT_RATCHET_TIERS_BY_GROUP` (`shared_strategies/close/trailing_tp_ratchet.py`) and any B2 TP defaults in `shared_strategies/close/regime_atr.py` / `post_tp_sl.py` kept in lockstep.

4. **Out of scope unless separately justified:** scalar `defaultHLProtectionTiers()` / `_DEFAULT_SCALAR_TP_TIERS` (non-regime path).

## Validation gate

[...]

Issue #1076, opened 2026-06-19
[C55] Validate the regime→direction premise behind regime_directional_policy (#779)


richkuo:
**Complexity: 55/100** — scope: medium (multi-asset/timeframe per-state validation harness, reuses #1073 diagnostics); risk: medium (the validation itself is read-only, but it gates a live long/short money path — a false "all clear" is costly); uncertainty: high (the premise may hold on other assets/timeframes; the right success metric — per-state return significance vs an economic backtest of the policy — is open).

## Context
Split from #1073 finding 5. The #1073 investigation found the 7-state composite regime classifier has **no statistically real forward-return (directional) separation** on BTC/USDT 1h: across 5 `eval_windows` × 7 horizons (1–72 bars) forward-return separation is never block-shuffle significant (0/35 tests), and per-state forward-return gaps are tiny and insignificant (e.g. oos `trending_up_choppy` gap +0.00047, p = 0.537). An in-sample-overfit k-means on the same four features fails significance too (p = 0.29–0.97) — it is not feature poverty; the directional signal is not in the data.

## Why this matters (money)
`regime_directional_policy` (#779, `scheduler/regime_directional_policy.go:5-16`) switches a live HL perps strategy long/short/inverse based on the **current** regime label (e.g. short in `trending_down`, long in `trending_up`). Its entire edge premise is regime → forward-direction. If that premise is false, live strategies using this block are choosing long vs short on noise. `allowed_regimes` directional entry-gating shares the same premise.

## Scope — what to validate
1. Per-state forward-return significance (block-shuffle + Benjamini-Hochberg FDR — reuse `backtest/regime_diagnostics.py` `per_state_significance`) across **multiple assets and timeframes**, not just BTC/USDT 1h. Does any (asset, timeframe, state) show real directional separation?
1. Economic test: does a `regime_directional_policy` config actually beat its base-direction / regime-agnostic counterpart on held-out walk-forward risk-adjusted PnL (existing backtester + `eval_windows`)? Trend-following can earn through asymmetric exits even without entry-regime forward-return separation, so the economic test — not the separation statistic alone — is the real arbiter.
1. If the premise holds nowhere: deprecate / warn on `regime_directional_policy` and regime directional entry-gating, or restrict it to configs with a demonstrated per-asset edge.

## Definition of done
[...]

PR #1212, opened 2026-07-02
feat(#1210): generalized M1–M6 auto-suggester — sweep candidates, rank gate-survivors, never auto-apply

### LINKED ISSUE #1210: [C65] Generalized M1–M6 auto-suggester: sweep candidates, rank gate-survivors, never auto-apply
**Complexity: 65/100** — scope: medium-high (new cross-harness orchestration layer spanning M1–M6 + candidate-grid generation + ranked reporting, no live-path edits); risk: moderate (suggest-only — never writes a live default, so no money path, but its output steers changes that do); uncertainty: high (generalizing over six heterogeneous harnesses with different preconditions/ordering/output contracts, and controlling multiple-comparisons so an automated sweep doesn't surface noise).

## Problem

The M1–M6 research harnesses (`backtest/eval_windows.py` M1 + `gross_edge_noise.py`, M2 joint sweep, M3 exit diagnostics, M5 fee audit, `backtest/exit_policy_ab.py` M6) grade candidates but don't orchestrate a search across each other. Two harnesses can already enumerate grids *within themselves* — M1's `--sweep P=V1,V2` cartesian plateau sweep (`eval_windows.py` `expand_sweep`) and M2's joint open×exit grid (`run_backtest.py --sweep-close` over `optimizer.DEFAULT_PARAM_RANGES`) — but there is no reusable mechanism that (a) takes one candidate space, (b) fans it out across the *relevant set* of harnesses, and (c) hands back a ranked "these survived every gate" shortlist with a sweep-wide statistical correction. Today that cross-harness search-and-grade loop is re-authored per study as a bespoke driver — `backtest/research/regime_1152_exit_retune.py` is the fullest example (candidate enumeration → parallel harness shelling → window rollup → promotion gate); simpler studies like `regime_1120_trail_validation.py` skip harness shelling entirely and drive `Backtester` inline, so even the *shape* of a study is reinvented each time.

## What this is NOT

**Not an auto-applier.** It must never write a live default (`ratchetTierGroupDefaults`, `regimeTPTierGroupDefaults`, stop fields, etc.) or open a PR that does. Auto-promotion on a statistical threshold is explicitly out of scope and unsafe: (1) it writes to the live trading path (money); (2) blind large-N candidate search directly defeats the pre-registered/incumbent-relative/both-windows gates the harnesses exist to enforce — mine enough candidates and one always clears p<.05 by luck. The #1152 study is the cautionary case: 22 disciplined candidates, zero shipped — and `mr.b2_rv_wider` actually **passed the pre-registered gate** (`candidate_beats_incumbent` under mean-reversion entries) while its squeeze-entry twin failed, so even a gate-respecting auto-promoter (not just a threshold one) would have shipped a change the human correctly held. The human makes every promotion call.

## Goal

A reusable **auto-suggester**: given a strategy + a candidate-space spec (parameter grids / ladder variants), fan out across the appropriate M-harnesses, apply each harness's own gate plus a multiple-comparisons correction across the whole sweep, and emit a ranked report of survivors with per-dataset evidence and the exact reproduction command — leaving promotion to a human.

- Extract the common driver skeleton into a reusable module: candidate enumeration → parallel harness shelling (`--jobs`) → window rollup (paired-N-weighted pooled Δ + dataset sign/sig votes) → pre-registered promotion gate. Model it on `regime_1152_exit_retune.py` (`_run_one`/`_window_rollup`/`_verdict`) so existing studies can be re-expressed on top of it.
- Candidate-space is declarative (grids/ladder variants + which harnesses + which regime gates), not hand-wired per run — **extending the existing convention**: `eval_windows.py --candidate-json` plus the per-study spec dirs under `backtest/candidates/`, not a new format.
- **Multiple-comparisons control is mandatory and explicit** — the sweep must correct for the number of candidates tested across the candidate family, **reusing `backtest/regime_stats.py` `benjamini_hochberg`** (precedent: `regime_1076_aggregate.py` applies one global correction across a battery). The report states the correction and the effective threshold; a candidate that only passes the uncorrected per-dataset test is reported as such, not as a survivor. **Correction scope:** it applies to the p-value-emitting harnesses (M1/M6 paired Wilcoxon/sign/bootstrap and the M1 step-2 permutation test); M3 exit diagnostics and M5 fee audit emit diagnostics/verdicts (`salvage_verdict`) with no p-values — those pass through as uncorrected context, clearly labeled, never counted as significance evidence.
- Output is a ranked shortlist artifact: survivors first, each with per-dataset p-values, pooled Δ/entry per window, the gate verdict, and a copy-pasteable reproduction command. No config mutation, no PR, no live-default write anywhere in the path.
- Honor each harness's preconditions/ordering (e.g. M1 step-2 `gross_edge_noise.py` noise gate runs BEFORE selectivity work; harnesses with no backtest parity — per-tick regime variants — are excluded, mirroring M6's `_REPLAYABLE_CLOSE_NAMES`).
- Per-candidate run failures follow the `regime_1152_exit_retune.py` `_run_one` precedent: a failed harness run is recorded with FAILED status and surfaced in the report, never silently dropped and never promoted.
- Tests: the rollup/gate/correction logic is pure and unit-tested (no subprocess spawning in CI), matching the repo's "extract pure helpers from subprocess wrappers" rule.

## Out of scope

Auto-promotion, config writes, PR creation, and any new deploy/argv surface. Options harnesses (M6 is spot/futures only at the registry level; perps backtest via the futures registry). New statistical methods beyond standard multiple-comparisons correction.

---

richkuo:
## Summary

Generalizes the M1–M6 research harnesses' per-study driver into one reusable **auto-suggester** (`backtest/auto_suggest.py`). Each retune (e.g. `regime_1152_exit_retune.py`) re-authored the same skeleton — candidate enumeration → parallel harness shelling → window rollup → promotion gate — from scratch. This extracts it once and adds the piece hand-rolled drivers keep omitting: a **sweep-wide multiple-comparisons correction**, so mining N candidates cannot surface a lucky p<.05 as a "survivor".

Closes #1210.

## What it does

- **Declarative spec** (`suggest.json`) extends the existing `--candidate-json` + `backtest/candidates/<study>/` convention: explicit candidate files + optional `base × sweep × gate_variants` grid + M2 close-stack grids (via `optimizer.generate_close_stack_grid`).
- **Fans each candidate across the appropriate M-harnesses**, honoring preconditions: the M1 step-2 noise gate (`gross_edge_noise.py`) runs BEFORE selectivity work (serial pre-pass primes the family cache); non-replayable M6 closes are excluded via `exit_policy_ab.candidate_is_replayable` (mirrors `_REPLAYABLE_CLOSE_NAMES`).
- **One Benjamini–Hochberg pass** (`regime_stats.benjamini_hochberg`, the #1076 precedent) over the whole candidate family = M1 step-2 permutation p + M6 paired Wilcoxon p. M3/M5 emit no p-values and pass through as clearly-labeled **uncorrected context**, never counted as significance. `--only` stamps the report EXPLORATORY (the correction is valid only over the full family).
- **Ranked shortlist artifact**: survivors first, each with per-dataset p-values, pooled Δ/entry per window, the gate verdict, and copy-pasteable reproduction commands. Failed/excluded runs stay visible, never silently dropped.

## Safety boundary (suggest-only)

Never writes a live default (`ratchetTierGroupDefaults` / `regimeTPTierGroupDefaults` / stop fields), never opens the live config, never shells `git`/`gh`, no apply/promote/PR surface. A human makes every promotion call — the #1152 study is the cautionary case (22 disciplined candidates, `mr.b2_rv_wider` even cleared the pre-registered gate, and the human still correctly held it).

## Files

- `backtest/auto_suggest.py` — the driver (pure spec/expand/argv/extract/rollup/correction/gate/rank helpers above an I/O fan-out line).
- `backtest/tests/test_auto_suggest.py` — 40 pure unit tests (no subprocess, no data access).
- `backtest/candidates/squeeze_momentum_1198/suggest.json` — demo spec re-expressing the three existing candidate files.
- `CLAUDE.md` — one § Backtest guardrail line.

## Scope note

The committed demo spec is open-kind only (m1_noise/m1/m3/m5). The M6 exit-A/B path is fully implemented and unit-tested with fixtures, but the demo does not exercise it (M6 needs a v15 `--baseline-config` fixture, kept out to keep the smoke self-contained).

## Verification

- `uv run --no-sync python -m pytest backtest/tests/test_auto_suggest.py -q` → 40 passed.
- Full `backtest/` + `shared_tools/` suite → 1271 passed (registry import-order caveat clean).
- `--dry-run` emits correct per-harness commands; a real end-to-end smoke (`--only baseline --datasets BTC/USDT:1h`) produced a ranked artifact with the baseline landing `positive_uncorrected_only` — M1 passes both protocol windows but its noise p does not survive BH at m=1, the intended conservative gate.

The Fable 5 implementation plan is posted on the issue.

---
https://claude.ai/code/session_01QS9j5Ure4nDCdMqjiPDTpB


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-02) in 7dc9bcf.

### Fixed
1. **Gated siblings sharing a noise family bypassed the BH noise gate and could promote to `survivor` on a failed permutation p** — Confirmed. `collect_family_pvalues` attaches the deduped noise test to the first sibling's `candidate_key` only, and `candidate_verdict` looked it up per-candidate via `_tests_for`, so siblings sharing a `noise_family_key` found `noise_t is None` and skipped the `positive_uncorrected_only` downgrade. Now the noise test carries its `noise_family_key` (`auto_suggest.py:498-508`) and `candidate_verdict` matches it by family across all tests (`auto_suggest.py:549-556`), so every entry in a family gets the identical BH-survival verdict. The BH family stays one noise p per family (not per sibling), so `m` is unchanged. Verified live: `baseline`, `adx_not_down`, `comp_up_family` all correctly land `positive_uncorrected_only` (previously the two gated siblings would have been `survivor`).

1. **`m5_cache` check-then-run race let two threads write the same fee-audit path, risking a torn read that aborts the run** — Confirmed. Implemented the best fix (matches the existing pattern, no locks, deterministic): the family-level M5 audit is now primed in the same serial pre-pass the noise gate uses (`ensure_m5`, `auto_suggest.py:753-770`; wired at the pre-pass `918-923`), so `run_open_entry` only reads a populated cache and no two concurrent runs share an output path. Rejected alternatives: a per-entry unique output path (wastes redundant fee_audit runs across a shared family) and a lock around the cache (serializes the slow subprocess anyway, so the pre-pass is strictly better).

### Tests
Added a 3-sibling shared-family regression (none may be `survivor` on a failed shared noise p; reorder-invariant; all promote together when it survives) and a distinct-params-families-stay-independent test. Full suite: 42 passed.

---

richkuo:
@claude review

Issue #1211, opened 2026-07-02
[C50] Restore a trustworthy incumbent baseline for the regime promotion gate (post-#1177)


richkuo:
**Complexity: 50/100** — scope: small-medium (research script + one gate function + tests + docs); risk: medium (governs every regime-model promotion decision — an auto-protective research gate — but offline-only, no live path); uncertainty: high (outcome unknown: the evidence may replicate with more power, or the gate semantics may need a deliberate redesign).

## Problem

The regime promotion pipeline cannot promote **any** candidate model, no matter how good, because the incumbent hand-rule's own evidence no longer clears the gate's significance bar:

- `gate_verdict` requires `incumbent_trustworthy = hr_p <= SIGNIFICANCE_ALPHA` and hard-gates `ship` on it, abstaining otherwise (`backtest/regime_calibrate.py:39-42`).
- The #1095 re-measurement (PR #1168) found the hand-rule's OOS forward-volatility separation is **not significant** at the resolved permutation count: p = 0.105 / 0.113 at `n_perm=1799`, versus the original knife-edge 10/201 ≈ 0.0498 at `n_perm=200` from #1080. #1177 recorded this as a documentation-only downgrade (`docs/ARCHITECTURE.md` § regime promotion).
- Consequence: every verdict in the #1095 bake-off abstained, and #1074 (live wiring) is dormant with no path to unblock. This precondition had no tracking issue — #1177 was a closed decision record, not the work item.

The in-code rationale comment (`backtest/regime_calibrate.py:33-38`) still claims the forward-volatility incumbent is "significant on 4/5 windows (PR #1077)" — stale post-#1177 for the OOS window at the higher permutation resolution.

## Goal

Restore a usable promotion gate by one of two explicit outcomes — never a silent relaxation:

1. **Evidence replicates**: a re-measurement on defensible windows/assets shows the incumbent's separation is real (the original window was underpowered), restoring `incumbent_trustworthy` honestly; or
2. **Gate semantics revised deliberately**: if the evidence does not replicate, redesign what "trustworthy baseline" means for shipping (e.g., require the candidate's own separation significance plus non-inferiority against the incumbent, rather than incumbent significance as a hard precondition), with the statistical rationale documented as a decision record.

Fail-closed abstention remains the default until one of these lands.

## Approach

[...]

PR #1094, opened 2026-06-21
feat(#1080): unsupervised volatility-state regime model (HMM/GMM/k-means bake-off)

richkuo:
Closes #1080.

## What this builds

The only regime model wired today (`regime_hmm.fit_label_anchored_hmm`) is **supervised by the hand-rule** — `regime_calibrate.fit_on_window` feeds `compute_regime_composite(...)` labels as the fit target, so it can resemble the incumbent in-sample by construction and can't independently beat it. This PR adds a genuinely **unsupervised** volatility-state model that learns states from the composite feature matrix alone, maps them to the 7 named composite labels, and is selected by a walk-forward bake-off against the #1078 forward-volatility gate.

Offline / research only. **Zero edits to any existing module** — three new files.

## The architecture (one schema, three families)

Every downstream consumer — the scorer (`regime_diagnostics.run_window`), the gate (`regime_calibrate.gate_verdict`), the #1082 bounded-window promotion harness, and the eventual #1074 live path — reads **one model-dict schema** decoded by the existing causal `regime_hmm.forward_filter_labels`. So all three families emit that **same schema**; they differ only in how the per-state Gaussian emissions are estimated. The transition table + init are always estimated empirically from the training-window state sequence, giving GMM/k-means temporal smoothing for free through the shared decoder.

- **k-means**, **GMM (diagonal-covariance EM)**, **Baum-Welch HMM** (+ Viterbi) — all hand-rolled in numpy (no sklearn/scipy/hmmlearn), matching the codebase.
- **Naming**: each learned centroid is un-standardized and run through the existing `map_composite_label` — deterministic, reuses canonical boundaries, consumes only training-window centroids (no per-bar hand-rule labels as a fit target).
- **Anti-collapse guards**: `MIN_ACTIVE_LABELS` / `MAX_LABEL_OCCUPANCY` / `MIN_TRANSITION_RATE`, **derived from the hand-rule incumbent's worst window and locked before any candidate is scored** (anti-gaming — can't be tuned to pass a favored model).
- **Bake-off**: sweeps family × K∈{2..7}, fits on in-sample, scores on held-out via `score_labels(target="volatility")` + `gate_verdict`, applies the guards on every eval window; winner = highest held-out separation among gate-passing, non-degenerate candidates. Reproducible evidence script in the #1073 style.

## Files
- `backtest/regime_vol_model.py` — library (fitters, empirical transition/init, naming, non-degeneracy, thresholds, single-family CLI).
[...]

PR #1222, opened 2026-07-03
fix(#1219): neutralize truly-dead n=0 decoder states in fitted GMM/HMM vol-regime models

### LINKED ISSUE #1219: [C25] Neutralize truly-dead n=0 decoder states in fitted GMM/HMM vol-regime models
**Complexity: 25/100** — scope: low (backtest-only, `backtest/regime_vol_model.py` + `backtest/regime_hmm.py` + tests); risk: low today (offline research path) but the fix protects the future live decode path #1074 will promote from; uncertainty: low-medium (soft-mass threshold and neutralization mechanism — exclude-from-argmax vs wide-variance park — need a small design call).

Extracted from #1074 deliverable 8 (carry-over from the #1080 review on PR #1094) so it isn't blocked by #1074's dormant gate chain — this is independently landable now.

## Problem
`fit_gmm`/`fit_hmm` take `counts` from the final argmax/Viterbi **hard** assignment and can store a component with `n=0` (`backtest/regime_vol_model.py:140-141`, `:203`; persisted as `"n"` at `:276`). The causal decoder `forward_filter_labels` decodes purely on emission mean/var + transition and never skips `n=0` states (`backtest/regime_hmm.py:64`), so such a state still competes in the decoder argmax. A negligible-soft-mass component collapsed toward the z-space origin at `var_floor` forms a sharp Gaussian that can win for bars near the global mean.

Offline this is self-correcting (a dead component worsens the candidate's score and the non-degeneracy/gate filters it out), which is why it was deferred from PR #1094. Live it matters: promote a fitted GMM/HMM model (#1074) and a truly-dead state could mislabel near-mean bars on-cycle. `fit_kmeans` already re-seeds empty clusters (`backtest/regime_vol_model.py:76-89`), so only the GMM/HMM fitters carry the gap.

## Two cases — distinguish by soft mass, not hard count
1. **Benign** — a component with real soft (responsibility/gamma) mass that simply never won a hard assignment. Its emission is legitimately soft-fitted; it must decode unchanged. Do NOT re-seed it (that would corrupt a valid emission).
2. **Truly-dead** — negligible soft mass collapsing toward the z-space origin at `var_floor`. This is the one to neutralize.

- Compute per-component **soft mass** at fit time (GMM responsibilities / HMM gammas) and use it — not the hard `n` — to classify dead components.
- Neutralize truly-dead components so they can never win the decoder argmax: either exclude them from decoding or park them with an uninformative wide-variance emission. Pick one and document why.
- Compatibility constraint: `fit_label_anchored_hmm` intentionally anchors degenerate states at origin + unit variance (`backtest/regime_hmm.py:34-36`) and must keep decoding exactly as today — any decoder-side change must not reinterpret its anchored states as dead.
- Regression tests: (a) a synthetic truly-dead component no longer wins near-mean bars; (b) a benign soft-mass `n=0` component decodes unchanged; (c) the label-anchored HMM path is byte-identical; (d) k-means path unaffected.

## Out of scope
- Live/Go wiring and model promotion (#1074).
- Gate semantics (#1211) or re-running the bake-offs.

---

richkuo:
## What & why

`fit_gmm` / `fit_hmm` (`backtest/regime_vol_model.py`) take their stored `n` from the final **hard** argmax/Viterbi assignment, so a component that wins no bar is stored with `n=0`. The causal decoder `forward_filter_labels` (`backtest/regime_hmm.py`) never reads `n` — it decodes purely on emission mean/var + transitions. A **truly-dead** component collapses toward the standardized origin at `var_floor` (numerator→0, denominator floored to `1e-10`), forming a **sharp** Gaussian that can win the decoder argmax for bars near the global mean (`z≈0`).

This is harmless offline today — dead components are rare (empirically ~0–1 in 8 fits) and self-correcting — but the #1074 live-decode promotion would inherit the mislabeling on-cycle. `fit_kmeans` already re-seeds empty clusters (#1094); only the GMM/HMM fitters carried the gap.

## Fix

Neutralize at **fit time**, classifying by **soft mass** (GMM responsibilities / HMM gamma occupancy), not the hard count:

- **Benign** — a component with real soft mass that merely never won a hard assignment is a legitimate overlapping mode; its emission is soft-fitted from real data and is left **untouched**.
- **Truly-dead** — soft mass below one effective observation (`min_soft_mass=1.0`) is a numerical ghost; it is re-anchored at the standardized origin with **unit variance** — the same uninformative emission `fit_label_anchored_hmm` already assigns its degenerate states — so any real component always out-scores it.

**Why park, not exclude-from-decode:** the decoder is shared with the label-anchored HMM, whose `n=0`/`n=1` anchored states must keep decoding. A decoder-side `n`-based skip would break that path. Parking is a pure fit-time emission edit, so the decoder and the label-anchored path stay **byte-identical**. No schema field is added; the soft mass used is exactly the responsibility/gamma that produced the stored emissions.

## Verification

`uv run --no-sync python -m pytest backtest/` — full backtest suite green (see below). New regression tests in `backtest/tests/test_regime_vol_model.py`:

1. Helper classifies by **soft mass**, not geometry/hard count (benign origin+`var_floor` component left alone; ghost parked).
2. Helper is a byte-identical no-op when all components are alive.
3. **Decoder-hijack red→green** (acceptance a): a sharp origin ghost wins every near-mean bar pre-park; the real near-mean component wins post-park.
4. A benign `n=0` component decodes unchanged (acceptance b).
5. Fitter wiring: `fit_gmm`/`fit_hmm` pass a length-k, finite, non-negative soft mass summing to `n`.
6. No-op on healthy fits (nothing parked on clean blobs).
7. Label-anchored degenerate anchor unchanged (acceptance c).
8. K-means path unaffected — no `min_soft_mass` knob, never parks (acceptance d).

Red→green confirmed: the 6 bug-demonstrating tests fail against the unfixed source; the regression guards pass both ways.

## Out of scope

Live/Go wiring and model promotion (#1074); gate semantics (#1211); re-running the bake-offs.

Closes #1219

---

### DIFF
--- a/backtest/regime_vol_model.py
+++ b/backtest/regime_vol_model.py
@@ -108,7 +108,36 @@ def _diag_logprob(z, mu, var):
     return -0.5 * (np.log(2 * np.pi * var) + diff ** 2 / var).sum(1)
 
 
-def fit_gmm(z, k, *, seed=0, iters=100, var_floor=1e-3, tol=1e-4):
+def _neutralize_dead_components(mu, var, soft_mass, *, min_soft_mass=1.0):
+    """Park truly-dead mixture/state components so they can never win the decoder argmax.
+
+    forward_filter_labels (regime_hmm.py) decodes purely on emission geometry + transitions and
+    NEVER reads the hard `n` count, so a component the fit collapsed toward the standardized origin
+    at `var_floor` forms a sharp Gaussian that can hijack near-mean (z~=0) bars at decode time.
+
+    Classify by SOFT mass (GMM responsibilities / HMM gammas), NOT the hard argmax/Viterbi count:
+      * benign  — real soft mass that merely never won a hard assignment. Its emission is legitimately
+                  soft-fitted from real data and MUST decode unchanged, so it is left untouched.
+      * truly-dead — soft mass below one effective observation (`min_soft_mass`): a numerical ghost
+                  collapsed to origin at `var_floor`. Re-anchor it at the standardized origin with
+                  unit variance — the SAME uninformative emission fit_label_anchored_hmm assigns its
+                  degenerate states (regime_hmm.py) — so any real component always out-scores it.
+
+    We PARK rather than exclude-from-decode because the decoder is shared with the label-anchored
+    HMM (whose n=0/n=1 anchored states must keep decoding); a decoder-side `n`-skip would break it,
+    whereas parking is a pure fit-time emission edit that leaves the decoder byte-identical.
+
+    Mutates mu/var in place; returns the boolean dead mask. No-op (byte-identical fit) when every
+    component clears the threshold."""
+    soft_mass = np.asarray(soft_mass, dtype=float)
+    dead = soft_mass < float(min_soft_mass)
+    if dead.any():
+        mu[dead] = 0.0    # standardized origin
+        var[dead] = 1.0   # unit variance -> maximally flat, mirrors fit_label_anchored_hmm anchor
+    return dead
+
+
+def fit_gmm(z, k, *, seed=0, iters=100, var_floor=1e-3, tol=1e-4, min_soft_mass=1.0):
     z = np.asarray(z, dtype=float)
     n = z.shape[0]
     assign0, mu, var, _ = fit_kmeans(z, k, seed=seed)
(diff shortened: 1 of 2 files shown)


richkuo:
@claude review

PR #1223, opened 2026-07-03
fix: skip #1211 incumbent-baseline smoke test before any live-data fallback fires

richkuo:
## Summary
- `test_load_cell_on_cached_data_if_available` (backtest/tests/test_regime_1211_incumbent_baseline.py) was flaky in CI: https://github.com/richkuo/go-trader/actions/runs/28640038779/job/84934073421
- Its skip guard only checked `load_cell()`'s returned status, but `load_cell()` and `run_window()` both call `data_fetcher.load_cached_data()`, which silently falls back to a **live** exchange fetch when the local OHLCV cache is empty (true on a fresh CI checkout).
- The test then issued two independent live BTC/USDT fetches a few seconds apart; market movement between them shifted the computed `kruskal_h` values outside `pytest.approx`'s tolerance (79.3134 vs 79.3116), failing CI on an unrelated push.
- Fix: check the cache directly via `storage.load_ohlcv()` (same start/end window resolution) *before* calling `load_cell` or `run_window`, and skip immediately if empty — neither live-fetch fallback can fire anymore.

## Test plan
- [x] `python -m py_compile` on the changed file
- [x] Targeted run: `pytest backtest/tests/test_regime_1211_incumbent_baseline.py -v` — 25 passed (test executes and asserts correctly against the locally populated cache, confirming the fix doesn't change behavior when cache is present)
- [x] Full `pytest backtest/tests/` — 1070 passed, 0 failed

---

### DIFF
--- a/backtest/tests/test_regime_1211_incumbent_baseline.py
+++ b/backtest/tests/test_regime_1211_incumbent_baseline.py
@@ -290,7 +290,21 @@ def test_load_cell_rejects_non_monotonic_index(monkeypatch):
 
 def test_load_cell_on_cached_data_if_available():
     """If the BTC/USDT 1h OOS cache exists, load_cell returns a scoreable cell whose kruskal_h
-    matches run_window's h4 (proving the harness mirrors the incumbent measurement exactly)."""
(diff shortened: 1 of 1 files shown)


richkuo:
@claude review

PR #1225, opened 2026-07-03
feat(#1224): per-window regime transition history, detection, and reversal alerts

### LINKED ISSUE #1224: [C55] Detect and alert on multi-window regime transitions/reversals (e.g. 30d down, 1d/3d/7d flip to up)
**Complexity: 55/100** — scope: medium (regime history table, transition detector, reversal classifier, notifier + status surfaces); risk: medium (runs inside the regime cycle, must never delay risk management); uncertainty: medium (reversal definition across windows, debounce vs detection-latency tradeoff).

## Problem

The regime pipeline computes per-window labels every cycle (`RegimeStore` bundles per (data platform, symbol, timeframe, windows-spec) signature; `shared_tools/regime.py` classifiers), but each cycle's result is discarded: the store is in-memory only and rebuilt every cycle, so there is no persisted regime history, no notion of "this window just flipped from downtrend to uptrend", and no proactive operator signal when short windows start disagreeing with a long window.

Note on windows: regime windows are **named bar-count periods** on one shared regime timeframe (`regime.windows`: name → classifier+period), not durations — an operator can *name* them `1d`/`3d`/`7d`/`30d`, and "longest" is resolved by comparing configured periods, not parsing names.

Concretely: with four windows configured (say `1d`/`3d`/`7d`/`30d` by name), the longest window may still read downtrend while the shorter windows flip to uptrend one by one — an early-reversal pattern the operator currently has to notice manually by polling `/status` or the dashboard (`/api/regime`).

**Relationship to #907 (`regime_window_divergence`):** a divergence detector already exists, but it is per-strategy, exactly two named windows (short vs medium), and surfaces only as a trade-DM line and a status field — its purpose is direction override for entries. This issue adds the complementary global market-property surface: full-key, k-of-N-window transition/reversal detection with proactive operator DMs. #907 is unchanged and coexists.

## Goal

Detect regime changes per window as fast as the data allows (i.e. within one scheduler cycle of the closed bar that flips the label) and surface cross-window reversal patterns — short windows contradicting the long-window regime — as an operator notification and a queryable status surface.

## Approach

1. **Regime history:** persist per-cycle window labels keyed by the **full bundle key + window name** — (data platform, symbol, timeframe, windows-spec, window, classifier, label, bar_time, ts) — into SQLite via an idempotent migration, so transitions and their timestamps are derivable and survive restarts. The key must be the full `regimeBundleKey`, not (symbol, window): same-symbol signatures on different platforms/timeframes/specs are distinct computations and must not collide. **Retention:** prune history older than a configurable horizon (default on the order of weeks) in the same maintenance pass that writes new rows — per-cycle × per-signature × per-window rows grow unboundedly otherwise.
2. **Per-window transition detection:** on each cycle, after the regime population wait returns (the store is stable from that point; `seal()` only fires on budget exhaustion), diff the new bundle against the last stored label per (bundle key, window); emit a transition event on change (old → new, window, bar time). Detection must key off closed-bar labels only — no intrabar look-ahead, consistent with the backtester's N-1 regime convention. First cycle after boot diffs against the last persisted row, so a restart produces no false transition.
3. **Reversal-pattern classifier:** on top of raw transitions, flag the ordered pattern "longest window says X, k shorter windows now say direction-opposing Y" (e.g. longest window trending_down while the shorter windows read trending_up / *_up composite states). Direction-opposition mapping must cover both `adx` (3-label) and `composite` (9-label, incl. `ranging_directional_up`/`_down` with the bare-covers-subs rule) classifiers.
4. **Debounce:** a window label can oscillate at the threshold boundary; require the new label to persist N consecutive cycles (configurable, default small — speed is the point) before alerting. The debounce counter is **derived from the history rows themselves** (N consecutive rows with the new label), never a separate in-memory counter, and the raw history rows are never debounced. Alert idempotence across restarts and SIGHUP comes from a **persisted alerted-marker** (e.g. an `alerted_at` column on the transition row, or a last-alerted (key, window, old→new, bar_time) record) — not in-memory state.
5. **Surfaces:** operator DM through `MultiNotifier` (new method on `MultiNotifier`, nil-guarded backends) on confirmed transitions and on reversal patterns; recent-transitions view in `/status` and the dashboard. DM sends happen on the sequential main loop after the population wait, matching the existing convention that `MultiNotifier` callers never fan out concurrently. Alerting only — this never gates entries, mutates config, or touches positions (any gating is a separate follow-on).
6. **Placement:** runs off the stable post-wait `RegimeStore` snapshot outside `mu`, after the population wait — must not extend the regime phase budget or delay risk management (the risk phase runs concurrently with population and is already complete by then); store/detector/DB failure is fail-open (log + skip cycle), matching the #879 regime fail-open convention.

## Acceptance criteria

- Per-window label history persisted in SQLite with idempotent migration, keyed by full bundle key + window; survives restart without false "transition" on first cycle after boot; retention pruning bounded and tested.
- A label flip on any configured window produces a transition event within one scheduler cycle of the closed bar, and (after debounce) exactly one operator DM — idempotent across restarts and SIGHUP via the persisted alerted-marker.
- The longest-window-down + all-shorter-windows-up scenario produces a distinct "reversal pattern" alert naming the disagreeing windows and both labels.
- Works for both `adx` and `composite` classifiers; composite directional substates map to the correct direction.
- Go tests for the transition differ, debounce, restart idempotence, retention pruning, and reversal-pattern classification (pure helpers, no Python subprocess); map iteration over windows sorted before formatting output.
- Detector failure never blocks or delays the trading loop.

---

richkuo:
Closes #1224

## What

Per-window regime transition history, detection, and cross-window reversal alerting — alerting-only, never gates entries or touches positions.

- **History:** every cycle's per-window labels persisted to SQLite (`regime_window_history`), keyed by the full bundle key (platform, symbol, timeframe, windows-spec) + window name. Raw rows are never debounced; retention pruning (hourly, default 30 days) bounds growth.
- **Transitions:** each cycle diffs the new label against the last stored row per (key, window) — a flip records a transition event within one cycle of the closed bar. After the new label persists `debounce_cycles` consecutive populations (default 2), the operator gets exactly one DM per **net** change; a flap back to the original label is consumed without a DM. The persisted `alerted_at` marker makes DMs exactly-once across restarts and SIGHUP; boot diffs against the last persisted row, so restarts never fabricate a transition.
- **Reversal pattern:** flags "longest configured window reads direction X, shorter windows read the opposing direction" (e.g. 30d `trending_down` while 1d/3d read `trending_up`). Direction mapping reuses `regimeLabelBias`, covering both `adx` (3-label) and `composite` (9-label incl. `ranging_directional_up`/`_down`) vocabularies. Deduped by a persisted per-key signature; a confirmed clear re-arms the alert. `reversal_min_opposing` (default: all shorter windows) tunes strictness.
- **Placement:** runs on the sequential main loop immediately after `regimeStoreReady()`, outside `mu`, so `MultiNotifier` sends stay serialized; every store/DB failure is fail-open (WARN + skip), matching the #879 convention — the trading loop is never blocked.
- **Config:** new optional `regime.transitions` block (`enabled`, `debounce_cycles`, `retention_days`, `reversal_min_opposing`), validated at load, hot-reloadable via SIGHUP (`config_reload.go` copy).
- **Surfaces:** `/status` gains a "📊 Regime transitions (24h)" note; the dashboard gains `GET /api/regime/transitions` (auth + drain-gated, `?limit=` capped at 500). Relationship to #907 `regime_window_divergence` is unchanged — that remains the per-strategy 2-window direction-override mechanism.

## Verification

- `go -C scheduler build .`, `go -C scheduler vet .` clean; full `go -C scheduler test -count=1 ./...` passes.
- 15 new Go tests (pure helpers, no Python subprocess): transition differ, debounce boundaries, flap-back suppression, restart idempotence against the same DB file, reversal classification (incl. composite substates, neutral-longest, k-of-N), signature dedupe + re-arm after confirmed clear, retention prune, disabled-feature no-op, sorted window formatting, config validation.

## Notes

- The dashboard surface is the queryable API endpoint (matching the `/api/regime` #879 precedent — the static UI has no regime-store panel today); a static-UI transitions panel would be a follow-on if wanted.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-03T21:36:22Z) in 52d970d.

### Fixed (Recommended Optional)
1. **Debounce/flap-back counted raw per-cycle populations, not distinct bars** — Confirmed against current code: `processRegimeTransitionAlerts` wrote one `regime_window_history` row per population and `regimeTransitionConfirmed` counted those rows, so a strategy interval shorter than the regime timeframe fired two DMs on a single-bar A→B→A flap instead of zero. Now the processor records at most one row per `(bundle key, window, closed bar)`: it skips the insert + debounce/transition work when `b.BarTime` equals the last stored `bar_time` for that window (`regime_transitions.go` per-window loop). The debounce run and flap-back suppression are now bar-accurate regardless of cycle cadence, and the history table grows ~1 row/bar instead of ~1 row/cycle.

### Design decisions beyond the reviewer's concrete fix
1. **Reversal debounce made bar-accurate too** — the reviewer flagged only transitions, but the in-memory cross-window reversal counters (`regimeReversalPending.count`/`inactive`) had the same cadence bug (confirm/clear within a fraction of a bar). Added `regimeReversalPending.lastBarTime` and gated counter advancement to once per bar (`processRegimeReversal`). The persisted signature already bounded over-alerting to one DM, but this makes the debounce semantics consistent with the stated invariant.
1. **Empty-BarTime fallback** — when a bundle carries no bar time, dedup is impossible, so both paths fall back to the original per-population behavior (guarded by `b.BarTime != ""`, mirroring the `regime_profile_allocation.go` #998 precedent) rather than silently going dark. Documented in the file header, the `bar_time` schema comment, and the `DebounceCycles` doc.
1. **Rejected: dedup only the history insert but keep re-running confirm/DM every population** — harmless (DM is idempotent via `alerted_at`) but wasteful of DB reads; skipping the whole per-window body when the bar is unchanged is cleaner and expresses the invariant directly.

### Tests
Added the three Must-survive regressions: (a) sub-bar single-bar flap → zero DMs and consumed transition rows; (b) interval==timeframe unchanged (existing suite, now with per-cycle BarTime advancement); (c) skipped bars then a sustained change → exactly one net-change DM with no fabricated intermediate. Plus a bar-accurate reversal-debounce test. Full `go test ./...` green.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-03T21:55:04Z) in 3534852.

### Fixed
1. **`regime_transitions.go` (#1224) wasn't indexed in CLAUDE.md/ARCHITECTURE.md** — confirmed via `grep -rn "regime_transitions\|#1224" CLAUDE.md docs/ARCHITECTURE.md` returning nothing pre-fix. Added a "Regime cluster" bullet in `CLAUDE.md` (scope: alerting-only, post-`regimeStoreReady()`/outside-`mu` placement, fail-open policy, the `regime.transitions` config block, the three SQLite tables, `/status` + `GET /api/regime/transitions`) and a matching file-ownership entry in `docs/ARCHITECTURE.md` (`CLAUDE.md:32`, `docs/ARCHITECTURE.md:29`) covering the bar-accurate debounce invariant, the `BarTime==""` per-cycle fallback, `netRegimeTransition` net-change collapse, the persisted `alerted_at` exactly-once marker (incl. boot false-positive suppression), reversal signature dedupe, retention pruning, and the full `regimeBundleKey`+window keying.

Also checked: `docs/backtesting-registry.md`'s upkeep rule scopes to `backtest/` harnesses only — `regime_transitions.go` is a live scheduler-loop alerting subsystem, not a backtest harness, so no row was added there (would have been spurious).

Docs-only change; `go -C scheduler build .` run clean, `git status` confirmed only `CLAUDE.md`/`docs/ARCHITECTURE.md` staged.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-03T22:10:40Z) in db4a482.

### Fixed
1. **`regime.transitions` was rejected by the SIGHUP compat gate before its "always hot-reloadable" copy branch could ever run** — `regimeConfigEqualIgnoringReloadableFields` (`config_reload.go:349-361`) masked `DisplayWindows`/`Timeframe` but not `Transitions`, so any diff there made `validateHotReloadCompatible` (`config_reload.go:18`) return `"regime changed (restart required)"` before line 60-64 was reached. Now `Transitions` is masked alongside its siblings, and the copy branch deep-copies via a new `cloneRegimeTransitionAlertsConfig` helper instead of sharing `next`'s pointer (matching the defensive posture of `clonePortfolioRiskConfig`, even though `next` is a freshly-`LoadConfig`'d struct discarded after the reload call — no aliasing risk existed, but the copy now matches sibling conventions). Doc comment on `regimeConfigEqualIgnoringReloadableFields` now calls out #1224 and states the invariant explicitly: every masked field needs a matching copy branch in `applyHotReloadConfig`.

### Extra validation performed (per dispatcher request)
1. **`validateHotReloadStateCompatible`** (`config_reload.go:446`) — confirmed it has no independent regime-equality or `Transitions` gate; its only `cfg.Regime`/`next.Regime` reads are the derived symbol/timeframe lookup for the state-shifting timeframe guard. `regime.transitions` is alerting-only and was never gated there — no change needed.
2. **Pointer-copy safety** — confirmed `next` in the only caller (`main.go:604-610`, `reloadConfig()`) is freshly loaded via `LoadConfig` per SIGHUP and goes out of scope after `applyHotReloadConfig` returns, so sharing the pointer would have been safe. Switched to a deep copy anyway (`cloneRegimeTransitionAlertsConfig`) to match the defensive posture of sibling struct-pointer fields (e.g. `clonePortfolioRiskConfig`) rather than relying on that caller-lifetime invariant holding forever.

Added `TestApplyHotReloadConfigRegimeTransitions` (mirrors `TestApplyHotReloadConfigDisplayWindows`) covering: nil→nil no-op, nil→enabled acceptance + deep-copy (with an open position, proving it's non-state-shifting), tunable-only changes while already enabled (also with an open position), and a compound change with a genuinely incompatible field (`db_file`) still correctly rejecting. Confirmed all four subtests fail against the pre-fix code with exactly the reported `"regime changed (restart required)"` error, and pass after the fix. Full scheduler suite (`go test -count=1 ./...`) passes.

---

richkuo:
@claude review

Issue #1229, opened 2026-07-05
[C85] Dashboard UI reaches full operator parity with Discord and CLI functionality (tracking)


richkuo:
**Complexity: 85/100** — scope: very large (new API surface + frontend panels across every subsystem, split into phased child issues); risk: high (mutating trade operations — force-close, SL edits, kill-switch/CB visibility — touch money and auto-protective paths); uncertainty: moderate (confirm-flow design for destructive mutations settles during Phases 1–2)

**This is the parent tracking issue.** Phases 1–2 are filed as fully-specified child issues; Phases 3–5 stay speced here until the earlier phases settle the mutation/confirm-flow design, then get their own child issues.

## Problem

The web dashboard covers only a fraction of what an operator can do with go-trader. Everything else lives in Discord commands or the CLI, so operating the system requires three tools.

**What the dashboard has today** (routes registered in `scheduler/server.go:189-202`, loopback-only bind at `server.go:179`): strategy list/overview, per-strategy candles + trade markers, trade history, equity sparklines, status rail, and the Strategy Tuner — the sole mutating action (POST `/api/strategies/{id}/config`, `ui_tuner.go:162`).

**Missing:** ~15 read-only surfaces (regime store/transitions, pause, CB/kill-switch, certifications #1157, diagnostics #1147, leaderboard, cashflow #1100–#1106, drift, dead/closing strategies, correlation) and 12+ mutating operations (pause/unpause, manual trade actions, SL edits, add/remove strategy, paper-to-live, apply-regime-gate, config set) that exist only in Discord/CLI. Full gap inventory with citations lives in the child issues.

## Goal

The dashboard becomes a complete operator console: every read-only surface is visible and every mutating operation available via Discord/CLI is executable from the UI with equivalent or stronger safety guarantees. Discord and CLI remain; the UI reaches parity.

## Security model (decided)

This is a personal application hosted on the end user's own private server — **no user authentication.** Concretely:

- **No `status_token` requirement anywhere in the UI path.** Reads already work token-free when unset (`requireAPIAuth`, `ui_server.go:105`). Mutations currently hard-require a token (`requireMutatingAPIAuth` 403s when unset, `ui_tuner.go:265-270`) — this requirement is **removed** for the dashboard mutation surface: token checks apply only when a token happens to be configured.
- **Keep loopback-only bind** (`server.go:179`) — network exposure control, not user auth.
[...]

PR #1260, opened 2026-07-06
feat(#1257): dashboard trade actions behind confirm-nonce typed confirmation (Phase 4 of #1229)

### LINKED ISSUE #1257: [C80] Dashboard can place, close, and adjust trades with a typed confirmation step (Phase 4 of dashboard parity)
**Complexity: 80/100** — scope: large (confirm-nonce infrastructure + manual-action endpoints + UI dialogs); risk: very high (money and auto-protective paths — manual trades, force-close, SL edits); uncertainty: moderate (confirm-nonce design is new, though the pattern is specified in parent #1229)

Phase 4 of #1229 (dashboard operator parity). Depends on Phase 3 (child of #1229) landing the UI mutation plumbing.

## Problem

Trade-affecting operations — manual open/add/close, force-close, stop-loss update/cancel — exist only in the manual CLI and Discord. Doing them from the dashboard is impossible, so live position management still requires leaving the UI.

## Goal

Expose trade-affecting mutations in the dashboard behind a server-issued confirm-nonce + typed-confirmation flow (mirroring Discord's `AskDM`), reusing the existing manual pipeline with zero bypass of its fail-closed guards.

## Approach

- **Confirm flow:** server issues a short-lived single-use nonce describing the exact action; the UI shows a typed-confirmation dialog; the mutating POST must present the nonce. Guards misclicks, not attackers (per #1229 security model). Same-origin + token-optional auth as Phase 3.
- **Manual trade actions** reuse the manual pipeline (`manual.go`/`manual_sl.go`/`manual_limit.go`): enqueue `PendingManualAction` — never a direct DB UPDATE. SL edits are cancel-then-queue, fail-closed via `manualSLAutoManaged`/`pendingSLActionExists`. Extract shared helpers from the CLI wrappers rather than duplicating logic.
- **Force-close** stays live-HL-perps-only, reduce-only, and refused while an SL edit is queued — identical constraints to `force-close <id>` (#1140).
- No UI endpoint spawns Python directly; side effects stay inside the scheduler's existing execution paths. Kill-switch and circuit-breaker states must gate these actions the same way the CLI does.
- UI: action buttons on position rows opening the confirm dialog; results reported from the queued action's outcome, not optimistically.

## Acceptance criteria

- Confirm-nonce round trip covered by Go tests: expired nonce rejected, reused nonce rejected, nonce bound to the specific action/strategy.
- SL edit via UI produces a queued `PendingManualAction` identical to the CLI path (regression test asserting no direct DB UPDATE).
- Force-close via UI refused while an SL edit is queued.
- Cross-origin POSTs rejected on every new endpoint.
- `docs/ARCHITECTURE.md` UI-server section updated.

---
LLM: Fable 5 | low | Harness: Claude Code

richkuo:
Closes #1257 — Phase 4 of #1229 (dashboard operator parity): trade-affecting mutations from the dashboard behind a server-issued confirm-nonce + typed-confirmation flow.

## What changed

- **Confirm-nonce (`ui_confirm.go`)** — `POST /api/confirm {action, strategy_id, params}` mints a crypto/rand nonce, 60s TTL, in-memory under a dedicated mutex, bound to `action + strategy + canonicalized(params)` (key order on the wire never matters; any value difference does). Single-use is unconditional: burned on lookup even when validation then fails. Constant-time binding compare. Response carries the server-authoritative description + confirm phrase (the strategy id) the dialog requires typing. Guards misclicks/stale dialogs — same-origin + optional token remain the boundary per #1229.
- **Manual-core extraction (`manual_core.go`)** — the six CLI monoliths (`manual-open|add|close`, `force-close`, `manual-update-sl|cancel-sl`) refactored into shared cores (inputs struct → result + typed error), CLI wrappers keep flag parsing/printing and identical exit-code classes. Every fail-closed guard lives once in the cores: kill switch, per-strategy circuit breaker, position ownership, `manualSLAutoManaged`, `pendingSLActionExists` orphan guard, force-close live-HL-perps-only + reduce-only + refused-while-SL-edit-queued (#1140), `PendingManualAction` enqueue only — never a direct positions UPDATE. On-chain side effects ride the existing `RunHyperliquid*` helpers; no new Python spawning path.
- **Endpoints (`ui_trade_actions.go`)** — `POST /api/strategies/{id}/{open|add|close|force-close|update-sl|cancel-sl}`, body `{nonce, params}`, behind the Phase-3 preamble (POST-only, `requireMutatingAPIAuth`, JSON content type, `requireSameOrigin`) plus nonce consumption before anything deeper. Daemon deps snapshot in-memory state under `ss.mu.RLock`, released before subprocess work (6-phase lock pattern). Responses report what was submitted/queued for the scheduler drain — never optimistic. Usage errors → 400, guard refusals/venue failures → 409, nonce failures → 403.
- **UI** — trade-actions panel (manual open/add form), per-position-row buttons (Close / Edit SL / Cancel SL for manual; Force close for HL perps), `<dialog>` typed-confirmation showing the server-issued description; all dynamic values escaped.
- **Docs** — ARCHITECTURE.md UI-server + manual bullets updated.

## Verification

- `go -C scheduler build .` + `go vet` clean; full `go -C scheduler test ./...` green (32s); new + force-close tests pass under `-race`; `gofmt -l` clean; `node --check app.js` clean.
- Tests: nonce lifecycle (expired/reused/rebound action/strategy/params, burn-on-failure), cross-origin 403 on all 7 new endpoints, token enforcement, UI SL edit queues a `PendingManualAction` byte-identical to the CLI core's row with no direct position mutation, force-close 409 while an SL edit is queued (closer never invoked), kill-switch/CB gating, stubbed-fill close happy path. Existing `runForceCloseWithCloser` CLI tests pin wrapper parity.
- Not verified live: no running daemon here — dashboard exercised via tests + `node --check` only; on-chain paths via stubbed seams (same as prior CLI tests).

---


richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review · 2026-07-06) in 5203678.

### Fixed
1. **Dashboard open could double-fire while the first open sat queued** — confirmed: `manualOpenCore` gates kill-switch/CB but nothing blocked a retry between on-chain submit and scheduler drain. Added a handler-level (UI-only, core untouched) guard: open refused when the strategy already holds the symbol or an open/add is still queued; add-retry refused while one is queued (`pendingPositionIncreasingActionExists`, `scheduler/ui_trade_actions.go:38`, mirroring `pendingSLActionExists`). Must-survive cases pinned in `TestUIOpenGuardsDoubleFire`: pending-open retry 409, re-open after drain passes, shared-coin peer position never blocks (view is strategy-scoped).
2. **Direct endpoint tests for open/add/cancel-sl** — added `TestUIOpenGuardsDoubleFire` (open happy path: queued "open" row, no in-memory position before drain), `TestUIAddQueuesAndGuardsPending` (queued "add" row, position unmutated, guarded retry), `TestUICancelSLQueuesAndErrors` (queued removal + no-resting-SL error path, nothing queued).

Full `go -C scheduler test ./...` green; new tests pass under `-race`.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review · 2026-07-06, second round) in 600ddf8.

### Fixed
1. **Close/force-close could double-fire while the first close sat queued** — confirmed (a sized manual close is a non-reduce-only order, and the queued row records requested qty → flip + double-decrement on drain). Guard extended: close/force-close refused while a `close` for the same strategy+symbol is queued, via generalized `pendingManualActionExists(kinds...)` (`scheduler/ui_trade_actions.go:35`). Must-survive pinned in `TestUICloseGuardsDoubleFire`: partial-close re-click blocked, post-drain close passes, a queued `update-sl` does not trip this guard (the core's `pendingSLActionExists` owns that class).
2. **Non-atomic check-then-submit under concurrent opens** — implemented the serialization remedy: new `StatusServer.tradeActionMu` held across the guard check AND the core (which inserts the pending row), making check+insert atomic (`scheduler/ui_trade_actions.go:238`); `/api/confirm` and read paths never take it. Rejected alternative: a DB uniqueness constraint on un-drained `(strategy_id, action, symbol)` — it would not stop two concurrent opens (both fire on-chain before either inserts) and adds a migration for weaker coverage. `TestUITradeActionsConcurrentOpensSerialized` races two opens through the real router and asserts exactly one venue call, one queued row, one 200 + one 409, under `-race`.

Full `go -C scheduler test ./...` green.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review · 2026-07-06, third round) in c3c4321.

### Fixed
1. **Force-close double-fire guard was a production no-op (empty perps `symbol` field)** — confirmed: `forceCloseCore` queues rows under the args-derived `sym` while the guard compared `sc.Symbol` (empty for perps). The guard now keys on `sym` for force-close (`scheduler/ui_trade_actions.go:266` region), matching exactly what the core writes; the perps test fixture drops its non-production `Symbol` field so `TestUICloseGuardsDoubleFire` genuinely exercises the guard (it fails against the old code — the closer stub trips). Manual `close`/`open`/`add` branches keep `sc.Symbol` (populated for `type=manual`), post-drain force-close still passes.

Full `go -C scheduler test ./...` green; UI/confirm tests pass under `-race`.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review · 2026-07-06, fourth round) in 8eca1dc.

### Fixed
1. **Cross-class guard gap: add-while-close-queued (and inverse) orphaned an on-chain order** — confirmed via the drain trace (close applies first, deletes the position, the add row then fails every cycle while the on-chain buy goes untracked). All four position-changing endpoints now share one in-flight class — the guard checks pending `{open, add, close}` per strategy+symbol for `open`/`add`/`close`/`force-close` (`scheduler/ui_trade_actions.go` guard block), keeping the `view.Pos` open check and the strategy-scoped keying (peer positions never block). `TestUICrossClassPendingGuard` pins add-while-close-queued 409, close- and force-close-while-add-queued 409, venue seams never invoked; drained-then-retry legitimacy stays covered by the existing guard tests.

Full `go -C scheduler test ./...` green; UI/confirm tests pass under `-race`.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review · 2026-07-06, fifth round) in 564a7f3.

### Fixed
1. **SL edits not refused while a close/open/add is queued** — confirmed, and fixed in the shared core rather than the handler: `resolveManualSLTargetCore` (behind both `update-sl` and `cancel-sl`, CLI and UI alike — the CLI had the identical hazard) now refuses when a pending `open`/`add`/`close` exists for the same strategy+symbol, fail-closed on a check error, mirroring the close cores' `pendingSLActionExists` refusal (`scheduler/manual_core.go` SL-target resolver). Must-survive pinned in `TestUISLEditGuardedWhileCloseQueued`: update-sl and cancel-sl 409 while a close is queued with the SL seam never invoked, and a legitimate edit passing once the close has drained (position still present ⇒ the hand-managed-SL workflow is never wedged); peer keying unchanged (strategy-scoped).

Full `go -C scheduler test ./...` green; UI/confirm tests pass under `-race`.

---

richkuo:
@claude review

richkuo:
@claude fix-pr

richkuo:
Addressed the Sonnet review (github-actions[bot] · 2026-07-06 22:09) in 803661c. Its one non-blocking finding is a safety-class item (auto-protective mechanism on the money path), so it was validated against the code and implemented rather than deferred.

### Fixed
1. **Position-changing double-fire guard was atomic only within the daemon process — a CLI racing the dashboard, or two concurrent CLIs, could still double-fire.**
Confirmed real: the CLI opens the state DB directly (`manual.go:263`, `OpenStateDB(cfg.DBFile)`), so it and the daemon are two OS processes on one SQLite file; `tradeActionMu` (`ui_trade_actions.go:243`) and `SetMaxOpenConns(1)` (`db.go:447`) serialize only within one process. Each manual-action core now wraps its guard-read → on-chain submit → pending-row insert in a cross-process advisory file lock (`scheduler/manual_action_lock.go`, `acquireManualActionFileLock`, `<canonicalDBPath>.manual-action.lock`), taken in `manualOpenCore`/`manualAddCore`/`manualCloseCore`/`forceCloseCore`/`manualUpdateSLCore`/`manualCancelSLCore` (`manual_core.go`). The SL cores are included so an SL edit racing a close/open/add is serialized too (the round-5 cross-guard, now cross-process). `tradeActionMu` stays as the in-process fast path + the UI-only "already holds the symbol" open pre-check.

### Resolved judgment call (remedy derivation — was implicit in the "Consider …" suggestion)
1. **Both suggested remedies are unsafe in this code path; derived a flock instead.**
The on-chain submit sits BETWEEN the guard read and the insert (`manual_core.go:790` → `821` → `876`). So a `BEGIN IMMEDIATE` txn spanning check+insert would hold a SQLite write lock across the subprocess network round-trip (up to the 30s script timeout), blocking the daemon's own DB writes; and a unique index on the final insert fires only AFTER both orders already hit the chain — worse, the second process's successful on-chain fill would then fail to insert and never be booked (orphaned fill / on-chain divergence). The flock (modeled on the existing `singleton_lock.go`) is OS-released on process death, so a killed holder never strands the lock — the decisive edge over a DB "reservation" row, which would permanently block that strategy+symbol until manually cleared. Bounded ~8s wait then fail-closed (matches the existing guards' fail-closed posture); in-memory DB → no-op (per-process, no race, no writable lock path). *Alternatives rejected:* `BEGIN IMMEDIATE` (write-lock across the network submit); unique-index-on-final-insert (too late + orphans the second fill); DB reservation row (crash-leaks a permanent block). Override if you'd prefer one of these.

### Tests
1. `TestManualActionLockPreventsCrossProcessDoubleFire` — two goroutines on independent fds contend under `flock()` exactly as two OS processes do; the first parks mid-submit holding the lock (pending row not yet written) and the second must block, then get refused by the guard once the row lands. Proven to fail (`bFired=1`, cross-process double-fire) when the lock is disabled and pass with it. `TestAcquireManualActionFileLock` covers the in-memory no-op plus real contention/release. Full scheduler suite green under `-race`; `gofmt -l`/`go vet` clean; `docs/ARCHITECTURE.md` manual-mechanics updated.

Scope note: the pre-existing `manual-limit-open` path (`manual_limit.go`, CLI-only, not exposed on the dashboard and outside this PR's surface) is left untouched.

LLM: Claude Opus 4.8 | xhigh | Harness: Claude Code

richkuo:
@claude review

richkuo:
Addressed the Opus re-review of `803661c` (github-actions[bot] · 2026-07-06 23:20). Verdict was **LGTM** with a single `Create Follow-up Issue` item and no blocking findings or open threads — no PR code change is warranted this round.

### Deferred to follow-up
1. **`manual-limit-open` resting-order path is outside the cross-process lock and the position-action guard.** Confirmed against the code: `runManualLimitOpen` (`manual_limit.go:245-271`) refuses only on a live position (`:263`) or an existing resting limit (`CountPendingLimitOrders`, `:268`) — it never inspects `pending_manual_actions` and never takes `acquireManualActionFileLock`, and it writes a `pending_limit_orders` row (`:310`) that `refuseIfPositionActionQueued` (`manual_core.go`) never reads. So a resting limit and a market/dashboard open on the same strategy+symbol are mutually blind and can both proceed; when the maker later fills against an already-open position the fill-adoption fails closed (`manual_limit.go:246-248`), orphaning the limit fill's on-chain exposure. Pre-existing, CLI-only, and correctly out of this PR's dashboard scope (it turns on the async resting-OID lifecycle, not the synchronous market double-fire the six cores now cover). Filed as #1261 (C58) with both race orderings, the concurrent-limit case, and the acceptance criteria specified.

The PR is approved (LGTM) with the follow-up tracked separately; no re-review triggered (no new commit to review).

LLM: Claude Opus 4.8 | xhigh | Harness: Claude Code

Issue #1205, opened 2026-07-02
[C55] feat(discord): interactive command to apply a regime-gate wiring to an operator-chosen live strategy


richkuo:
**Complexity: 55/100** — scope: medium (new mutating Discord command + interactive strategy-selection flow + safe config-write path + tests); risk: medium (mutates live strategy entry-gating behavior — fail-safe in that it only blocks entries, never forces trades, but a wrong target strategy or malformed window/label pairing would silently stop that strategy from ever entering); uncertainty: medium (repo has no existing "pick one of N strategies interactively" Discord flow to reuse — only fixed-arg commands and AskDM confirm-or-cancel prompts exist today).

## Problem

Wiring a regime gate (e.g. the #1197 `comp_up_clean_p21` breakout gate) onto a live strategy today means an operator manually hand-edits the out-of-tree config JSON (`/var/lib/go-trader[/<instance>]/config.json`, #1056): adding a `regime.windows` entry, setting `regime_gate_window`, and setting `allowed_regimes` on the target strategy. That's error-prone (typo'd window key, wrong classifier/label pairing, applying to a strategy that isn't flat) and has no operator-facing discoverability — nothing in `/status` or Discord surfaces "you could gate this strategy's entries."

## Goal

A Discord slash command that lets an operator apply a regime-gate wiring to a chosen live strategy interactively — asking which strategy to target rather than requiring the operator to already know/type its exact ID — and writes the resulting config change through the existing safe-write path.

## Approach

- Register as a new mutating command in `opsCommandNames` + `slashCommands()` + the dispatch switch (`scheduler/discord_commands.go`), owner-DM-only per existing mutating-command convention.
- Strategy selection: since no existing command offers a multi-choice interactive picker (today's mutating commands take a bare ID argument, or use `AskDM` for a confirm/cancel prompt — `scheduler/discord_mutating_commands.go:666`), this needs either a Discord select-menu/component interaction or a follow-up `AskDM` numbered-list-reply flow. Pick whichever fits the existing `AskDM` plumbing with the least new surface area; list only strategies eligible for the gate (matching type/platform the gate was validated against, e.g. futures/HL perps for #1197-style composite gates — reject or grey out ineligible ones).
[...]

PR #1262, opened 2026-07-07
feat(#1258): dashboard structural mutations — add/remove strategy, paper-to-live, apply-regime-gate behind confirm-nonce flow (Phase 5 of #1229)

### LINKED ISSUE #1258: [C75] Dashboard can add or remove strategies and promote paper to live (Phase 5 of dashboard parity)
**Complexity: 75/100** — scope: large (three structural operations, each with its own validation surface + confirm flow); risk: high (remove-strategy and paper-to-live are destructive/capital-affecting; apply-regime-gate has cross-strategy blast radius); uncertainty: low-moderate (all three have existing Discord implementations to mirror)

Phase 5 of #1229 (dashboard operator parity) — final phase. Depends on Phase 4 (child of #1229) landing the confirm-nonce flow.

## Problem

Structural operations — add/remove strategy, paper-to-live promotion, apply-regime-gate — exist only as owner-DM Discord commands. The dashboard cannot change the strategy roster or promote a paper strategy.

## Goal

Expose the three structural mutations in the dashboard behind the Phase 4 confirm-nonce + typed-confirmation flow, with validation and blast-radius checks identical to (or stronger than) the Discord commands.

## Approach

- All three reuse the Phase 4 confirm flow (these are the Discord `AskDM`-confirmed commands: `/remove-strategy`, `/paper-to-live`, `/apply-regime-gate` #1205). Same-origin + token-optional auth throughout.
- **Add strategy:** UI form → validated via the same config-validation path as `writeValidatedConfigRoot`; hot-reload constraints apply (adding a strategy is a SIGHUP-blocked change — surface the restart requirement honestly rather than pretending it applied).
- **Remove strategy:** refuse while the strategy has an open position unless explicitly confirmed per the Discord command's semantics; config write on `configWriteMu`.
- **Paper-to-live:** mirror `/paper-to-live` validation (capital, keys, probe status).
- **Apply-regime-gate:** re-check blast radius just before the write, not only at confirm time — the `regime.enabled` flip can reactivate another strategy's dormant `allowed_regimes` gate (parent #1229 requirement, matching `discord_regime_gate_command.go`).
- Config mutations only through `writeValidatedConfigRoot`/`applyStrategyConfigPatch` on `configWriteMu`.
- Out of scope (parent decision): `restart`/`backtest` triggers, `add-platform`.

## Acceptance criteria

- Each operation requires the confirm-nonce round trip; cross-origin POSTs rejected (Go tests per guard).
- Apply-regime-gate blast-radius re-check at write time covered by a test where a second strategy's dormant gate would reactivate.
- Interleaved-write test: UI structural mutation vs concurrent Discord `config set` — no clobbering on `configWriteMu`.
- Parent #1229 closable once this lands: all phase checkboxes done.
- `docs/ARCHITECTURE.md` UI-server section updated.

---
LLM: Fable 5 | low | Harness: Claude Code

richkuo:
Closes #1258 — Phase 5 (final) of #1229: the three structural operations plus add-strategy reach the dashboard behind the Phase-4 confirm-nonce + typed-confirmation flow.

## What

- **Endpoints (`scheduler/ui_structural.go`):** `POST /api/config/add-strategy` and `POST /api/strategies/{id}/{remove-strategy,paper-to-live,apply-regime-gate}`, all behind `uiStructuralGuards` (rejectIfDraining + POST-only + `requireMutatingAPIAuth` + JSON content type + `requireSameOrigin`) and a consumed single-use confirm nonce bound to the exact `(action, id, params)`.
- **Confirm flow:** `/api/confirm` now accepts the four structural actions with per-action eligibility checks and server-authoritative descriptions. `add-strategy` is the one action allowed an empty `strategy_id` — the target doesn't exist yet; the generated ID becomes the typed confirm phrase.
- **Zero duplicate mutation logic:** execute reuses the Discord pure helpers (`addStrategyToRoot` / `removeStrategyFromRoot` / `flipStrategyToLive` / `applyRegimeGateToRoot`) through a new shared `ss.mutateConfigRoot` (read → mutate → `writeValidatedConfigRoot` on `configWriteMu`); `DiscordNotifier.mutateConfig` now delegates to it.
- **Honest restart semantics:** all four are restart-required shape changes (blocked by the hot-reload path). The response states that; `params.restart: true` — part of the nonce binding, so it's what the operator confirmed — fires the injectable `ss.restartFn` (default `restartSelf`) after the response, mirroring the Discord apply path.
- **apply-regime-gate safety model (#1205 parity):** flat-only at confirm AND re-checked at execute; the `regime.enabled`-flip blast radius is computed at confirm, shown in the dialog, pinned into the nonce (`confirmNonceEntry.payload`), and recomputed inside the `configWriteMu` critical section just before the write — growth vs the confirmed set refuses; shrinkage passes.
- **remove-strategy:** confirm description warns when the target holds an open position (management stops after restart); removing the only strategy is refused. **paper-to-live:** REAL-FUNDS warning in the dialog; already-live / modeless strategies fail early at confirm.
- **Frontend:** overview "Add strategy" ops-panel (paper-only creation; promotion is the separate confirmed step) + status-rail "Structural" panel (Remove for all strategies, Paper→Live and Apply regime gate for perps/futures), all through the shared typed-confirmation dialog.

## Tests

- Per-endpoint guards: cross-origin 403, GET 405, missing nonce 400 (before any config touch), bogus nonce 403.
- End-to-end add / remove / paper-to-live / apply-regime-gate against the real `LoadConfigForProbe` validator on a temp config.
- Nonce single-use (replay refused, config untouched) and params binding (same nonce with different params refused).
- Flat re-check at execute (position opens between confirm and execute → 409), blast-radius growth refusal (concurrent config edit adds a dormant `allowed_regimes` → 409, gate not written), only-strategy removal refusal, restart-param fires the injected `restartFn` (and not otherwise).
- Interleaved `mutateConfigRoot` writers — both fields persist, config stays valid.
- Full scheduler suite green under `-race`; `gofmt -l` / `go vet` clean. `docs/ARCHITECTURE.md` UI-server section updated.

LLM: Fable 5 | low | Harness: Claude Code


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-07T00:14:48Z) in 51256fc.

### Fixed
1. **paper-to-live gave no open-position guard despite being a real-funds flip.** Implemented the finding — and went stronger than the suggested "warn": paper→live now **refuses** while the target holds an open position, at confirm (`ui_structural.go` `confirmStructuralAction` paper-to-live case) and re-checked at execute (`ui_structural.go` new `executePaperToLive`, mirroring `executeApplyRegimeGate`'s flat-only structure). Rationale: a simulated paper position has no on-chain backing, so carried into a live strategy it becomes a phantom `syncHyperliquidAccountPositions` flags as a gap and the live strategy would mismanage — there is no coherent live semantics for it, so refusing (not warning) is the correct, hazard-eliminating design; `remove-strategy` still *warns* because stopping management of an existing position is a legitimate operation, whereas carrying a fake position into real funds is not. Covered by `TestUIPaperToLiveFlatChecks` (confirm-refused-while-open; flat-at-confirm-then-opens-before-execute; args untouched on refusal). The guard is type-agnostic (`strategyHasOpenPosition` ignores type), so it also satisfies the futures "Must survive" case.

### Deferred to follow-up
1. **Discord /paper-to-live has the same gap.** The reviewer flagged this as inherited/pre-existing (not a regression here). It is a different, already-shipped surface (#1205), the guard cannot live in the pure `flipStrategyToLive` helper (no `ss.state` access), and there is no Discord-mutating-handler test harness today — testing it properly would materially expand this dashboard PR. Filed with a full spec and the flat-only invariant + Must-survive cases: #1264.

Note on the judgment call: I chose refuse over the finding's suggested warn for the real-funds path. If you'd prefer parity-with-`remove-strategy` (warn-and-proceed) instead, say so and I'll switch it — but refuse is the safer default and matches `apply-regime-gate`.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-07T00:35:02Z) in 21bb30e. Both were Recommended Optional and both validated as real.

### Fixed
1. **`remove-strategy` didn't front-load the only-strategy refusal.** Added `isOnlyStrategyOnDisk` (best-effort on-disk read mirroring `removeStrategyFromRoot`'s `len==1` refusal) and wired it into `confirmStructuralAction`'s remove-strategy case (`scheduler/ui_structural.go`), so a single-strategy config refuses at confirm instead of minting a nonce and 409-ing at execute — matching the file's "fail the dialog early" contract. The execute-time check stays authoritative. Tests: `TestUIRemoveStrategyEndToEnd`'s only-strategy case now asserts the confirm-time 400; new `TestUIRemoveStrategyExecuteRefusesOnlyStrategyAfterConcurrentPrune` proves the execute check still fires when the sibling is removed concurrently after a valid confirm (the reviewer's Must-survive case 2).
2. **`docs/ARCHITECTURE.md` #1258 paragraph was stale after 51256fc.** Updated the guardrail paragraph to document paper-to-live's flat-only open-position refusal (`executePaperToLive`, confirm + execute) and the newly front-loaded remove-strategy check, so the documented safety model matches the code.

Verified: `go build`, `go vet`, `gofmt -l` clean; full `go test -race ./...` green.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-07T01:08Z) in 16f007c.

### Not changed (refuted)
1. **Move the flat check inside the `configWriteMu` critical section in `executePaperToLive` / `executeApplyRegimeGate`.** Declined — the suggested remedy is both ineffective and regressive; kept the check outside and documented why (`scheduler/ui_structural.go`, comments on both functions).
   - **It can't close the race.** Position-opens are `mu`-governed (the trading loop's execute phase), and the trading loop never acquires `configWriteMu` (grep: `configWriteMu` appears only in `ui_mutations.go`/`ui_tuner.go`/`discord_mutating_commands.go`/`ui_structural.go`, never in the loop). So holding `configWriteMu` during the check cannot block a fill — moving the check inside only narrows the wall-clock window, it does not eliminate the race.
   - **The same hazard survives via a far larger window.** Both actions are restart-required: after the write the daemon keeps running in the old mode until the async restart + reload — a multi-second `write→restart→reload` window in which a position can open and be carried in. Tightening the micro-to-millisecond `check→write` window can't prevent a hazard that also occurs across the seconds-long window neither placement addresses.
   - **It would be the codebase's sole `configWriteMu → mu` nesting.** Every config-write path releases `mu` before taking `configWriteMu` (`ui_mutations.go:98`, `ui_tuner.go:216`, `discord_mutating_commands.go:729` all compute `strategyHasOpenPosition` first, then lock). Nesting the `mu.RLock` inside would invert that uniform convention and plant a latent AB-BA deadlock for any future `mu → configWriteMu` path — the exact class the convention prevents. Safe today only because no reverse path exists; keeping the check outside keeps it that way.

   Also corrected `executeApplyRegimeGate`'s doc comment, which wrongly implied the flat re-check runs inside the critical section (only the blast-radius growth check does).

If you'd still prefer the tighter window despite the above, the only correct way to actually close the carried-position hazard is a reload/startup-time guard (refuse to run a newly-live/newly-gated strategy that boots with an open position) — genuinely separate, broader scope than this dashboard PR, and already backstopped by the fail-closed HL reconcile gap alert.

---

richkuo:
@claude sonnet review

PR #1263, opened 2026-07-07
fix(#1261): guard manual limit opens against market races

### LINKED ISSUE #1261: [C58] A resting limit order and a market/dashboard open on the same strategy can both fill, doubling the position
**Complexity: 58/100** — scope: medium (the `manual_limit.go` placement guard + a `pending_manual_actions`↔`pending_limit_orders` cross-check, and extending the #1260 cross-process lock to the limit path); risk: high (position/fill accounting — a doubled position and orphaned on-chain exposure); uncertainty: medium (the correct guard has to span the asynchronous resting-OID lifecycle, a different model than the synchronous pending-row guard #1260 uses).

## Problem

#1260 added a cross-process advisory file lock (`acquireManualActionFileLock`) and the `refuseIfPositionActionQueued` guard so the **synchronous market** trade-action cores (manual open/add/close, force-close, SL edits) can't double-fire when a CLI races the dashboard. The **resting limit-open path is covered by neither mechanism**, so it and a market open are mutually blind:

- `runManualLimitOpen` (`scheduler/manual_limit.go:245-271`) is a check-then-place: it loads state and refuses only when a **live position** already exists (`ss.Positions[sc.Symbol]`, `manual_limit.go:263`) or a **resting limit** already exists (`CountPendingLimitOrders`, `manual_limit.go:268`). It does **not** inspect `pending_manual_actions` (queued market opens), and it does **not** take `acquireManualActionFileLock`. It persists a `pending_limit_orders` row (`InsertPendingLimitOrder`, `manual_limit.go:310`), not an `open`/`add`/`close` row.
- The market-open guard `refuseIfPositionActionQueued` (`scheduler/manual_core.go`) reads only `pending_manual_actions` (`open`/`add`/`close`), never `pending_limit_orders`.

### The race

A `manual-open` (or dashboard open) submits its market order on-chain and queues a `pending_manual_actions{open}` row, but the scheduler has not yet drained it into a live `Position`. During that window an operator runs `manual-limit-open` on the same strategy+symbol: its placement guard sees no live position and no resting limit, so it places a resting maker order and inserts a `pending_limit_orders` row. Now there is one un-drained market open AND one resting limit for the same strategy+symbol. When the scheduler adopts the market fill it creates the position; when the resting maker later fills, the fill-adoption path meets an already-open position and fails closed — the placement guard's own comment (`manual_limit.go:245-248`) documents that the fill path "fails closed rather than adopt a foreign position" — leaving the limit fill's on-chain exposure **orphaned** (a real HL position the daemon won't book). The inverse ordering (limit placed first, market open second) is symmetric: the market open's guard never sees the resting limit and the loadState position check sees nothing until the limit fills.

Two concurrent `manual-limit-open` invocations are also unguarded across processes: `CountPendingLimitOrders` (`manual_limit.go:268`) is a read-then-decide with no lock, so both can observe 0 and both place.

This is pre-existing (the limit-open path predates #1260), CLI-only (not exposed on the dashboard), and was explicitly scoped out of #1260 — a resting order fills asynchronously, so it isn't the synchronous market double-fire the six cores now cover. But because it touches position/fill accounting it should be closed.

## Goal

At most one un-drained position-*establishing* action per (strategy, symbol) across every process and caller — whether that action is a synchronous market open/add or an asynchronous resting limit order. A market open must see a pending resting limit and vice-versa, atomically across a CLI-vs-dashboard or CLI-vs-CLI race, and it must survive a process crash without stranding the guard.

## Approach (the design decision)

1. **Extend the #1260 lock + make the guard bidirectional (preferred).** Have `runManualLimitOpen` acquire `acquireManualActionFileLock` around its placement-guard read → on-chain place → `InsertPendingLimitOrder`; add a `pending_manual_actions` check to the limit path (refuse a resting-order placement while a market open/add/close is queued); and extend `refuseIfPositionActionQueued` (used by the market cores) to also refuse when `CountPendingLimitOrders > 0`. This reuses the existing crash-safe lock family and the pending-row guard model with the least new surface.
2. **Model the resting limit as a position-establishing `pending_manual_actions` row** so the existing guard sees it directly, retiring the separate `pending_limit_orders` visibility gap. Larger change to the resting-OID lifecycle; only pursue if option 1's cross-check proves awkward.

## Acceptance criteria

- A `manual-limit-open` racing a `manual-open`/dashboard open on the same strategy+symbol: at most one proceeds; the other is refused with a clear operator message. **Both orderings** covered.
- Two concurrent `manual-limit-open` invocations on the same strategy+symbol: at most one places.
- A legitimate `manual-limit-open` still succeeds once no market action is queued and no position/resting order exists (no false refusal).
- Regression tests over the CLI seam (mirroring `TestManualActionLockPreventsCrossProcessDoubleFire`) proving the cross-path refusal in both directions, each shown to fail with the new guard/lock removed.

---
LLM: Claude Opus 4.8 | xhigh | Harness: Claude Code

richkuo:
Closes #1261

## What changed

- Made `refuseIfPositionActionQueued` also fail closed when a resting manual limit order exists for the same strategy and symbol.
- Wrapped `runManualLimitOpen` in the existing manual-action advisory lock from guard reads through `pending_limit_orders` insert.
- Added CLI-seam regressions for queued market-vs-limit, queued limit-vs-market, and concurrent limit-open races.
- Documented the #1261 manual limit-order guard mechanics in `docs/ARCHITECTURE.md`.

## Verification

- Red check: the new targeted tests failed against the unfixed guard behavior, then passed after the fix.
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler test -run 'TestManualLimitOpenRefusesQueuedMarketAction|TestManualOpenCoreRefusesQueuedLimitOrder|TestManualLimitOpenLockPreventsCrossProcessDoubleFire|TestManualCoresGuardPositionDoubleFire|TestManualActionLockPreventsCrossProcessDoubleFire' .`
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler fmt ./...`
- `env GOCACHE=/tmp/go-build-cache /opt/homebrew/bin/go -C scheduler vet ./...`


richkuo:
@claude review

richkuo:
@claude review

richkuo:
@claude fix-pr

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-07T01:13:04Z) in 2a55ba6.

### Fixed
1. **Fully-drained limit order residual-leak race (Needs Fixing).** Validated against code: `reconcilePendingLimitOrders` grows the position in-memory (`applyLimitFillProgress`) and deletes the terminal `pending_limit_orders` row (`manual_limit.go:706`) before the end-of-cycle `SaveState` (`main.go:2696`) flushes the grown position — so a cross-process CLI `manual-close` (`newCLIManualCoreDeps.loadState` reads `state.db` fresh, `manual_core.go:161`) can observe the row gone while the DB position is still stale, find nothing to reconcile, and size a shared-coin close (`closeFullPosition=false`) to the stale qty. A CLI-only fix can't recover the true size once the row is gone and the DB is stale, so the race is closed at the writer: the daemon now persists the position to `state.db` **before** deleting the terminal row, gated on an actual fill (`o.FilledSize > 0`), establishing the invariant "terminal row absent ⇒ `state.db` reflects the adopted fill" (`manual_limit.go`, terminal branch). On a save failure it keeps the row and retries next cycle rather than deleting against a stale DB. The existing CLI-side reconcile stays for the row-still-present case (authoritative exchange read).

1. **`--qty` bounds check ran before reconciliation.** The bounds check + intent determination now run **after** `clearRestingLimitRemainderForPositionAction` reconciles `pos.Quantity` (`manual_core.go`). An explicit `--qty` matching the true adopted size is accepted instead of refused against the stale snapshot; a `--qty` between the stale and true sizes is a valid partial; and the "exceeds open position" error reports the reconciled size. Dry-run stays advisory against the current snapshot (it must not cancel/reconcile, which is an on-chain side effect). One deliberate behavior refinement: an explicit `--qty` is never scaled up to the reconciled full size — only an omitted (or within-lot-of-full) `--qty` flattens the reconciled position — so an explicit partial never removes more than the operator asked for. Omit `--qty` for a full close.

### Tests
1. `TestReconcilePendingLimitOrdersFullFillFlushesPositionBeforeRowDelete` — daemon flush-before-delete on a terminal full fill (a fresh cross-process reload sees the true size).
1. `TestManualCloseAcceptsExplicitQtyMatchingReconciledSize` / `...PartialQtyBetweenStaleAndReconciledSize` / `...RejectsQtyExceedingReconciledSize` — the three must-survive `--qty` cases against the reconciled size.

Build, `go vet`, `gofmt`, and the full `scheduler` test suite pass.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-07T01:13:04Z re-review) in 0afba98.

### Fixed
1. **`manualCloseCore` sized against the pre-lock snapshot when the row was deleted between `loadState` and `clearResting` (Needs Fixing).** Confirmed against code: `pos` was read at `manual_core.go:933` before the lock (`:971`) and before `clearRestingLimitRemainderForPositionAction` (`:983`); the daemon adopts+flushes+deletes a terminal limit row without the manual-action lock, so in that window `clearResting` returns `(0,0,nil)`, the reconcile branch didn't fire, and the close sized `pos.Quantity` from the stale snapshot — leaking a residual on a shared-coin sized close. The flush-before-delete invariant is now leveraged on the reader side: when `clearedQty > 0` the exchange-authoritative reconcile is kept (row present, `state.db` not yet authoritative); when `clearedQty == 0` the position is re-read from `state.db` under the lock — guaranteed fresh by "terminal row absent ⇒ `state.db` reflects the adopted fill," and no new resting row can appear while the lock is held (limit-open takes it too) — with nil + ownership re-validated and `closeSide` recomputed from the resolved position (`manual_core.go`). The `--qty` bound and sizing now run against the resolved size, so must-survive case 3 (explicit `--qty` = true size, row gone) is accepted, not refused.

### Tests
1. `TestManualCloseRereadsFreshPositionWhenRowDeletedBeforeClearResting` — injected stale 0.4 pre-lock snapshot, row deleted + position flushed to 0.7 before `clearResting`, shared coin: full close sizes the fresh 0.7 (`closeFullPosition=false`), not 0.4.
1. `TestManualCloseExplicitQtyValidatedAgainstRereadWhenRowGone` — `--qty 0.7` accepted against the re-read (must-survive case 3).

Build, `go vet`, `gofmt`, and the full `scheduler` test suite pass.

---

richkuo:
@claude review

PR #1265, opened 2026-07-07
fix(#1264): refuse Discord /paper-to-live while strategy holds open position

### LINKED ISSUE #1264: [C40] Discord /paper-to-live can promote a strategy to live while it holds an open paper position
**Complexity: 40/100** — scope: small (one Discord handler + a testable seam); risk: medium (real-funds paper→live command, but the change is a refusal guard with low blast radius); uncertainty: no test harness exists for Discord mutating handlers yet.

## Problem

The Discord `/paper-to-live` command (`scheduler/discord_mutating_commands.go` `handlePaperToLive`, ~L825) flips a strategy's `--mode=paper` arg to `--mode=live` after a DM confirm, with **no open-position check**. Its pure helper `flipStrategyToLive` (~L286) operates only on the config-map `root` and has no access to `ss.state`, so the guard cannot live in the helper — it must live in the handler.

The dashboard equivalent shipped in #1258 (`scheduler/ui_structural.go`) now **refuses** paper→live while the target strategy holds an open position, at both confirm time and execute time (mirroring `apply-regime-gate`'s flat-only rule). The Discord surface still allows it.

Hazard (why flat-only): a simulated paper position has no on-chain backing. Carried into a live strategy it becomes a phantom that `syncHyperliquidAccountPositions` flags as a reconcile gap, and the now-live strategy would place reduce-only orders against a position that does not exist on-chain. This is a real-funds / data-integrity path, so the invariant should hold on **every** surface that can perform the flip, not just the dashboard.

This is pre-existing behavior in the Discord command (#1205), not a regression introduced by #1258 — the #1262 reviewer explicitly scoped it out of that PR. It is filed separately because (1) it is a different, already-shipped surface than #1258's dashboard, and (2) doing it properly needs a Discord-mutating-handler test harness that does not exist today (only the pure `flipStrategyToLive` is unit-tested; there is no discordgo session mock anywhere in the suite), which would materially expand the dashboard PR's diff.

## Goal

Enforce the same flat-only invariant on Discord `/paper-to-live` so a simulated paper position can never be carried into a real-funds live strategy from any surface.

**Invariant:** paper→live may only run while the target strategy is flat, on every surface (dashboard and Discord).

1. In `handlePaperToLive`, call `d.ss.strategyHasOpenPosition(id)` and refuse via `followupText` when the strategy holds an open position. Place the check **right before the `d.mutateConfig` write** so it catches a position that opens during the (blocking) DM-confirm wait; optionally also a fast pre-check before `confirmDestructive` so the operator is not prompted for an action that will be refused.
2. Add a minimal test harness (or extract a testable precheck seam callable without a live discordgo session) and cover: (a) an open perps position → refused, config unchanged; (b) flat → flips to live; (c) a position that opens between the confirm and the write → refused.
3. **Must survive:** (i) a perps strategy with an open long paper position; (ii) a strategy flat when the confirm is issued that opens a position before the write lands; (iii) a `futures` paper strategy holding an open position (`strategyHasOpenPosition` is type-agnostic, so one guard must cover both).
4. Audit sibling mutating commands for consistency: confirm `/apply-regime-gate` already refuses while open, and that `/remove-strategy` intentionally *warns* rather than refuses (removal is a legitimate stop-managing operation, so a warning — not a refusal — is correct there).

---

richkuo:
## Summary
- Add `paperToLiveBlockedReason` as the shared flat-only refusal helper used by dashboard confirm, execute, and Discord.
- Discord `/paper-to-live` now pre-checks before the DM confirm and delegates the write to `executePaperToLive` (same path as the dashboard).
- Add `TestPaperToLiveFlatChecks` covering pre-confirm refusal, flat success, and execute-time refusal without a discordgo session.

Closes #1264

## Test plan
- [x] `go test ./scheduler -run 'TestPaperToLiveFlatChecks|TestUIPaperToLiveFlatChecks'`

---
Created with LLM: Composer 2.5 | high | Harness: Cursor

### DIFF
--- a/scheduler/discord_mutating_commands.go
+++ b/scheduler/discord_mutating_commands.go
@@ -824,8 +824,7 @@ func (d *DiscordNotifier) handleAddPlatform(s *discordgo.Session, i *discordgo.I
 // it switches a strategy to placing real orders with real funds.
 func (d *DiscordNotifier) handlePaperToLive(s *discordgo.Session, i *discordgo.InteractionCreate, opts []*discordgo.ApplicationCommandInteractionDataOption) {
 	deferAck(s, i)
-	path, err := d.configOpsReady()
-	if err != nil {
+	if _, err := d.configOpsReady(); err != nil {
 		followupText(s, i, err.Error())
 		return
 	}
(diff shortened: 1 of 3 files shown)


richkuo:
@claude review

PR #1296, opened 2026-07-09
test(#1279): registry-wide forward-peek (truncation-invariance) sweep over all open strategies

### LINKED ISSUE #1279: [C40] Strategies that peek at future bars are not caught automatically — add a registry-wide forward-peek (truncation-invariance) test
**Complexity: 40/100** — scope: medium (one test module sweeping ~49 registered open strategies, per-strategy fixtures/kwargs); risk: medium (test-only, no runtime path, but it guards backtest-integrity decisions — a vacuous harness would give false confidence); uncertainty: medium (per-strategy warmup lengths and special-input needs discovered during implementation)

## Problem

Strategy signal purity is entirely caller-trusted, and nothing sweeps the registry for forward-peeking strategies automatically.

Verified in code:

- The backtester's `shift(1)` enforces **fill timing only, not signal purity**. `backtest/backtester.py:1374-1384` shifts `signal` / `_open_action` / `_close_fraction` / `_profile_label` by one row, and the module docstring (`backtest/backtester.py:41`) states the signal `shift(1)` "is the only look-ahead guard on the signal path".
- `backtest/tests/test_backtester_lookahead.py:230` (`test_forward_peek_signal_documents_caller_responsibility`) explicitly demonstrates the gap: a `close.shift(-1)` peeking signal sails through the engine and inflates returns more than 20 points past buy-and-hold. Its docstring calls this a "DOCUMENTED LIMITATION (#730): the engine enforces fill timing, not signal purity."
- Per-strategy truncation-invariance tests exist for exactly **three** strategies: `test_anchored_vwap_no_lookahead` (line 371, #1016), `test_anchored_vwap_channel_no_lookahead` (line 450, #1169), `test_anchored_vwap_reversion_no_lookahead` (line 533, #1170).
- `shared_strategies/open/registry.py` registers roughly 49 open strategies (`register(...)` calls; `STRATEGY_REGISTRY` + `PLATFORM_ORDER` at line 1376). The other ~46 have no automated look-ahead check.

Consequence: a new strategy — or a refactor of an existing one — that reads future bars inside its function (rolling windows with `center=True`, `shift(-1)`, whole-series normalization, pivot confirmation without a confirmation lag, etc.) passes CI, backtests with inflated edge, and can be promoted toward live on contaminated evidence. The three anchored-VWAP tests were each added reactively per strategy; that coverage model does not scale.

## Goal

Every registered open strategy is automatically checked for truncation-invariance (no forward peek) by the pytest suite, so a peeking strategy fails CI the day it is added — with a sensitivity guard proving the harness itself is not vacuous.

## Approach

Add a registry-wide parameterized test module (alongside the existing look-ahead tests or under `shared_strategies/`):

1. **Enumerate the registry**: iterate `STRATEGY_REGISTRY` from `shared_strategies/open/registry.py` (via `importlib.util.spec_from_file_location`, matching existing test-import convention), including hidden and `backtest_only` entries — research strategies feed promotion decisions too.
2. **Synthetic fixture**: generate seeded, non-trivial OHLCV (trend + noise + swings, positive volume) with a `DatetimeIndex` — required because `amd_ifvg` reads `index.hour` and `vwap_reversion` buckets by `index.date` (existing repo rule). Fixture must be long enough to clear every strategy's warmup and rich enough that most strategies emit nonzero signals.
3. **Truncation-invariance assertion**: compute `signal` on `df.iloc[:N]` and on `df.iloc[:N+k]`, plus a variant of the longer frame whose future bars (> N) are perturbed; assert signals at bars `<= N - warmup_margin` are identical across all three. Default params from the registry entry.
4. **Sensitivity check (non-vacuity)**: a deliberately forward-peeking fixture strategy (`s[:-1] = s[1:]`, mirroring the pattern at `test_backtester_lookahead.py:390`) must FAIL the same harness assertion — guarding against a fixture/cut choice that makes the sweep vacuous. Additionally report which strategies produced all-zero signals on the fixture, so per-strategy vacuity is visible rather than silent.
5. **Special-input strategies**: handle strategies needing extra columns/kwargs with per-strategy fixture hooks — e.g. `delta_neutral_funding` (funding-rate params, `registry.py:1010`), position-aware entries (framework kwargs must be named explicitly per repo rule), HTF-consuming strategies. Any strategy placed on a skip-list must carry an inline justification comment, and the test asserts the skip-list is a closed allowlist — an unexplained or stale skip is a failure.
6. **Suite wiring**: module runs in the FULL pytest suite (repo rule: registry/sys.path tests must not run isolated — open-vs-close `registry.py` import-order conflicts are masked otherwise).
7. **Follow-through on findings**: if the sweep surfaces an existing peeking strategy, do not weaken the harness — mark that strategy xfail with a linked follow-up issue per offender, so the sweep lands green while offenders are tracked explicitly.

## Acceptance criteria

- A parameterized test sweeps every strategy in `STRATEGY_REGISTRY` (including hidden and `backtest_only`); adding a new `register(...)` call automatically adds it to the sweep with no test edit.
- For each swept strategy: signals at bars up to `N - warmup_margin` are identical between the truncated frame, the extended frame, and the future-perturbed frame.
- A deliberately forward-peeking fixture strategy fails the harness (sensitivity check), and the harness reports strategies producing all-zero signals on the fixture.
- Skip-list entries (if any) each carry an inline justification and are asserted against a closed allowlist.
- Existing tests (`test_backtester_lookahead.py`, full pytest suite) still pass; the new module passes inside the full-suite run.
- No changes to live/runtime code paths — test-only.

---

richkuo:
Closes #1279

## What

Adds `shared_strategies/open/test_registry_forward_peek.py` — a registry-wide forward-peek (truncation-invariance) sweep. Every strategy in `shared_strategies/open/registry.py` (including hidden and `backtest_only` entries, one case per distinct platform-variant param set) is checked automatically: adding a new `register(...)` call adds it to the sweep with no test edit (`test_sweep_covers_every_registered_strategy` pins this).

## How it works

- **Invariant:** at every checked cut, signals before the cut must be identical across (a) the truncated frame, (b) the full frame, and (c) a frame whose bars past the cut are replaced with a differently-seeded, rebased tail. Cuts sit at a fixed prefix boundary AND at the strategy's own signal transitions — transition cuts are what expose single-bar peeks (a fixed cut can land where the signal is locally constant and let a `shift(-1)` through).
- **Fixture:** seeded 960-bar hourly OHLCV with trend, swings, volatility-regime segments, and sparse shocks; carries `close_b` (pairs_spread) and `funding_rate` (delta_neutral_funding / funding_skew series path) so special-input strategies run their real per-bar paths.
- **Per-strategy fixture hooks** (closed allowlist, input data only — params stay registry defaults): purpose-built frames for `range_scalper` (tight-bandwidth consolidation with engineered band-cross slides) and `sweep_squeeze_combo` (sweep wicks aligned onto stoch-RSI trigger bars), so their coverage is real instead of allowlisted away.
- **Non-vacuity guards:** a deliberately forward-peeking wrapper must FAIL the same checker the sweep uses; all-zero-signal strategies are asserted against a closed, justified allowlist (only `hold`, which is signal=0 by contract) — stale entries fail in both directions; skip-list and fixture-hook dicts are validated as closed allowlists (skip-list is currently empty).

## Sweep result

No offender found: all 49 registered strategies are truncation-invariant on the sweep, so no xfail/follow-up issues were needed. (During development the sweep also confirmed `liquidity_sweeps`' centered swing detection is consumed with the proper confirmation lag.)

## Verification

- Red→green: planted `result["signal"].shift(-1)` in `sma_crossover` → sweep fails at transition cuts (`cut=822: truncation changed signals at bars [821]`); reverted → green.
- Module: 59 passed in ~4s.
- Full suite (`shared_strategies/ shared_tools/ platforms/ backtest/`): 2453 passed, 1 skipped — the module runs inside the full-suite import-order context per repo rule.
- Test-only change; no runtime/live code paths touched.

---
Created with LLM: Fable 5 | medium | Harness: Claude Code

### DIFF
--- a/shared_strategies/open/test_registry_forward_peek.py
+++ b/shared_strategies/open/test_registry_forward_peek.py
@@ -353,9 +353,6 @@ def _sweep_cases():
     return cases
 
 
-_ZERO_SIGNAL_OBSERVED = {}
-
-
 @pytest.mark.parametrize("name,params", _sweep_cases())
 def test_registry_strategy_is_truncation_invariant(name, params):
     if name in SKIP_STRATEGIES:
@@ -372,7 +369,6 @@ def test_registry_strategy_is_truncation_invariant(name, params):
     # above prove nothing for this strategy, so it must be on the closed
     # allowlist (with a justification comment) — not silently green.
     nonzero = np.nan_to_num(full) != 0
-    _ZERO_SIGNAL_OBSERVED[name] = _ZERO_SIGNAL_OBSERVED.get(name, False) or bool(nonzero.any())
     if not nonzero.any():
         assert name in EXPECTED_ALL_ZERO_SIGNAL, (
             f"{name} produced all-zero signals on the sweep fixture — the "
@@ -385,25 +381,24 @@ def test_registry_strategy_is_truncation_invariant(name, params):
 def test_zero_signal_allowlist_is_exact():
     """Stale-entry guard: every EXPECTED_ALL_ZERO_SIGNAL entry must (a) name a
     registered strategy and (b) still be all-zero on the fixture for every
-    swept param set. Runs after the sweep in file order; pytest -p no:randomly
-    is not assumed, so it recomputes when the sweep didn't populate the cache.
+    swept param set. Always recomputed here (one strategy eval per allowlisted
+    name — cheap) so the verdict never depends on which sweep cases happened
+    to run in this session (test selection, ordering plugins, xdist workers).
     """
     unknown = set(EXPECTED_ALL_ZERO_SIGNAL) - set(_REGISTRY.STRATEGIES)
     assert not unknown, f"EXPECTED_ALL_ZERO_SIGNAL names unregistered strategies: {sorted(unknown)}"
     for name in sorted(EXPECTED_ALL_ZERO_SIGNAL):
         if name in SKIP_STRATEGIES:
             continue
-        emitted = _ZERO_SIGNAL_OBSERVED.get(name)
-        if emitted is None:
-            entry = _REGISTRY.STRATEGIES[name]
-            emitted = False
-            for platform in entry["platforms"]:
-                merged = {
-                    **entry["default_params"],
-                    **entry["variants"].get(platform, {}).get("default_params", {}),
-                }
-                sig = _signal(entry["fn"], merged, _df_for(name)[1])
-                emitted = emitted or bool((np.nan_to_num(sig) != 0).any())
+        entry = _REGISTRY.STRATEGIES[name]
+        emitted = False
+        for platform in entry["platforms"]:
+            merged = {
+                **entry["default_params"],
+                **entry["variants"].get(platform, {}).get("default_params", {}),
+            }
+            sig = _signal(entry["fn"], merged, _df_for(name)[1])
+            emitted = emitted or bool((np.nan_to_num(sig) != 0).any())
         assert not emitted, (
             f"{name} now emits signals on the sweep fixture — remove it from "
             f"EXPECTED_ALL_ZERO_SIGNAL (stale vacuity entry)."
--- a/shared_strategies/open/test_registry_forward_peek.py
+++ b/shared_strategies/open/test_registry_forward_peek.py
@@ -1,0 +1,464 @@
+"""Registry-wide forward-peek (truncation-invariance) sweep (#1279).
+
+The backtester's ``shift(1)`` enforces *fill timing only* — signal purity is
+caller-trusted (``backtest/tests/test_backtester_lookahead.py`` documents the
+gap, #730). Per-strategy no-lookahead tests existed for only three strategies
+(#1016/#1169/#1170); this module sweeps EVERY registered open strategy —
+including hidden and ``backtest_only`` entries, since research strategies feed
+promotion decisions too — so a forward-peeking strategy fails CI the day it is
+registered, with no test edit.
+
+Invariant per strategy (with registry-default params, per platform variant):
+at every checked cut, signals on bars before the cut must be byte-identical
+whether the frame is
+
+* truncated at the cut (future bars absent),
+* the full extended frame (future bars present), or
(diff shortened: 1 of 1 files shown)


richkuo:
LGTM

### Recommended Optional

1. **Stale-entry guard can silently trust a partially populated cache — recompute allowlisted names unconditionally.**
   `test_zero_signal_allowlist_is_exact` (shared_strategies/open/test_registry_forward_peek.py:396) only recomputes when `_ZERO_SIGNAL_OBSERVED.get(name)` is `None`. Under `-k` selection or random test ordering, a multi-variant strategy may have only some sweep cases recorded, so a listed strategy that fires on an unobserved variant false-passes the stale check (the guard weakens, never the sweep itself). Since recompute cost is one strategy eval per allowlisted name (currently just `hold` on 960 bars), drop the cache read in this test and always recompute; `_ZERO_SIGNAL_OBSERVED` can then be deleted entirely.
   **Invariant:** the stale-entry verdict for an allowlisted strategy reflects all its swept param sets, regardless of which sweep cases ran in this session.
   **Must survive:** (1) running only this test via `-k zero_signal`; (2) random/partial test ordering where the allowlist test runs mid-sweep; (3) a future allowlist entry whose spot variant is all-zero but futures variant fires.

---
Validated with LLM: Fable 5 | medium | Harness: Claude Code

richkuo:
Addressed review feedback (richkuo · 2026-07-09T23:28:17Z) in 8c4fa1b.

### Fixed
1. **Stale-entry guard trusted a partially populated cache** — `test_zero_signal_allowlist_is_exact` now recomputes the all-zero verdict unconditionally for every allowlisted name across all platform-variant param sets; `_ZERO_SIGNAL_OBSERVED` deleted (`shared_strategies/open/test_registry_forward_peek.py:381`). Verified: module 59 passed; isolated `-k zero_signal` run passes.

---

richkuo:
@claude sonnet review

richkuo:
@claude review

PR #1300, opened 2026-07-10
Add configurable regime_gate_on_failure entry-gate failure policy

### LINKED ISSUE #1278: [C50] Regime entry gate fails open on regime-store failure — add configurable regime_gate_on_failure with fail-closed for entries
**Complexity: 50/100** — scope: medium (config field + single gate funnel + 6 dispatch sites + backtest parity + status/alerting surfaces); risk: high (entry gating is an auto-protective money path — a wrong default or a fail-closed bug that also blocks closes has direct trading impact); uncertainty: low (behavior fully traced in code; open question is only the default policy)

## Problem

The `allowed_regimes` entry gate fails open when the regime store cannot produce a label, so a transient regime-subprocess outage admits exactly the entries the filter exists to block.

Verified chain:

- `scheduler/regime_store.go:467-468` — a failed regime bundle signature leaves NO store entry; consumers see an empty `RegimePayload` and "fail open (issue #879 failure policy b)". The same applies when the phase budget seals the store (`regime_store.go:463-465`).
- `scheduler/regime_multi_window.go:562-565` — `applyRegimeGate` resolves the gate label from the payload; an empty payload yields `""`.
- `scheduler/regime.go:20-23` — `regimeAllowsEntry` returns `true` when `current == ""` ("regime not available from check script"), i.e. the gate admits the entry. `regimeBlocksOpen` (`regime.go:46-51`) only ever blocks when flat, so open-position management is unaffected either way.
- All 6 regime-gated dispatch sites funnel through `applyRegimeGate`: `scheduler/main.go:1808` (OKX spot), `1845` (Robinhood), `1880` (spot), `1951` (OKX perps), `1996` (HL perps), `2289` (TopStep) — one helper change covers every site.
- Alerting lag: regime-subprocess failures thread through the existing `scriptFailureTracker` via `regimeBundleAlertConfig` (`regime_store.go:446-452`), with the primary operator alert at `scriptFailureAlertThreshold=3` consecutive failures. So a `check_regime.py` outage produces up to ~3 cycles of silently ungated entries before anyone is told — and the entries themselves are never flagged.

Fail-open is the right policy for display and management paths (#879), but for an entry gate the conservative option is fail-closed: skip NEW opens while the regime is unknown; existing positions keep managing (closes, trailing SL, ratchet, protection sync all pass through `regimeBlocksOpen`'s `posQty > 0` short-circuit unchanged).

Also found while verifying: the backtester in fact fails open on empty regime labels too — `shared_tools/regime.py:78-94` (`regime_label_allows_entry`: empty `current` allows entry, "matching the Go `regimeAllowsEntry` contract for parity") and the `backtest/backtester.py:126` fallback. But two backtester comments (`backtest/backtester.py` ~1563-1567 and ~1911-1916) incorrectly claim an empty label "fails the `in allowed_regimes` check and blocks the entry". The comments are stale and contradict the code; they should be reconciled as part of this change.

## Goal

Make the regime entry gate's failure policy configurable — `regime_gate_on_failure: "open" | "closed"` — with fail-closed available for entries, without any silent behavior change and without ever affecting close/management paths.

## Approach

1. **Config surface.** Per-strategy `regime_gate_on_failure` on `StrategyConfig` (`config.go`), values `"open"` (default, preserves current #879 behavior) and `"closed"`; optionally a global default under the `regime` config block that per-strategy values override. Reject unknown values at load. Hot-reloadable always (it only changes flat-strategy open gating, never state-shifting fields), mirroring the #1150 pause pattern.
2. **Gate semantics.** Thread the policy into `applyRegimeGate`/`regimeBlocksOpen` (`regime_multi_window.go:562`, `regime.go:46`). Fail-closed fires ONLY when all of: `len(allowed_regimes) > 0` (a gate is actually configured), the resolved gate label is empty (store failure / sealed budget / missing window), and `posQty == 0`. Strategies without a gate are untouched; closes and open-position management are untouched by construction.
3. **Dispatch sites.** No per-site changes needed — all 6 `main.go` sites already call `applyRegimeGate`. Distinguish the block reason in the existing log line (`main.go:1809` et al.): `regime gate: open blocked (regime unknown, fail-closed)` vs the current not-in-allowed-set block.
4. **Operator visibility.** When a fail-closed gate is actively suppressing opens: mark it in the Phase 6 status log and `/status` (alongside the #1189 display label — e.g. `regime=? (gate closed)`), and note it in the existing regime-bundle failure alert text so the scriptFailureTracker DM says whether affected strategies were failing open or closed during the outage. No new alert channel needed.
5. **Backtest parity.** Thread the policy through `--config` in `backtest/run_backtest.py` → `Backtester` so `regime_label_allows_entry`/`_regime_allows_entry` honor it for empty/NaN regime rows (warmup bar 0 after the #730 shift, mid-series gaps). Default `"open"` keeps existing baselines byte-identical. Fix the two stale `backtester.py` comments that claim empty labels already block.
6. **Default policy.** Keep `"open"` as the effective default for existing configs (no silent change). Have `go-trader init`/the wizard emit `"closed"` for newly generated configs that set `allowed_regimes`, and update the CLAUDE.md #879 note to say the entry gate is policy-configurable. If review concludes fail-closed should be the universal default instead, that flip is a one-line change behind this same surface — but it must be an explicit, documented decision, not a side effect.
7. **Tests.** Go: `regimeBlocksOpen`/`applyRegimeGate` table tests covering policy x empty/known label x flat/open x gate configured/absent; hot-reload acceptance. Python: backtester gate test for empty-regime bars under both policies; parity test asserting Go and Python agree on the empty-label decision per policy.

## Acceptance criteria

- [ ] `regime_gate_on_failure: "closed"` on a strategy with `allowed_regimes` blocks fresh opens (all 6 dispatch sites) whenever the gate label resolves empty — regime-store subprocess failure, sealed phase budget, or missing window — and the log line names the fail-closed reason.
- [ ] Fail-closed never blocks: closes, any action while `posQty > 0`, or strategies with no `allowed_regimes` configured.
- [ ] Default (field omitted or `"open"`) reproduces current behavior exactly; existing configs load and behave unchanged.
- [ ] Unknown values rejected at config load; the field is SIGHUP hot-reloadable including while a position is open.
- [ ] `/status` and the Phase 6 status log surface an actively fail-closed gate; the regime-bundle failure DM states the gate policy in effect for affected strategies.
- [ ] Backtester honors the policy for empty-regime bars via `--config`; default keeps existing backtest results identical; the two stale `backtester.py` comments claiming empty labels block are corrected.
- [ ] Go table tests + Python backtest/parity tests cover policy x label-availability x position-state combinations.

---

richkuo:
Closes #1278

## Summary
- Per-strategy `regime_gate_on_failure` (`"open"` default | `"closed"`) plus a global `regime.gate_on_failure` fallback; resolved via `resolveRegimeGateOnFailure`, never read directly.
- Fail-closed fires only when a gate is configured, the resolved gate label is empty (regime-store failure, sealed phase budget, or missing window), AND the strategy is flat — closes, posQty>0 management (trailing SL/ratchet/protection sync), and ungated strategies are untouched by construction.
- Single funnel: `applyRegimeGate` covers all 6 regime-gated dispatch sites; the block log line now distinguishes `regime unknown, fail-closed` from a not-in-allowed-set block.
- Load validation: unknown values rejected on both surfaces; `closed` + `allowed_regimes` + `regime.enabled=false` is a load error (deterministic permanent entry block). Hot-reloadable always, including while a position is open.
- Operator visibility: Phase 6 status line shows `regime=? (gate closed)` (computed from the live regime store, not the synced state labels, which go stale during exactly the outage); `/status` and the dashboard overview serialize `regime_gate_fail_closed`; the regime-bundle failure DM names each gated strategy's policy ("fail-closed (opens held)" vs "fail-open (entries ungated)").
- Backtest parity: `Backtester(regime_gate_on_failure=...)` resolves empty/NaN regime rows identically; `--config` threads strategy>global>default resolution through `load_strategy_config`; `simulate_strategy.py` mirrors it. Default `"open"` keeps existing baselines byte-identical. Fixed the two stale `backtester.py` comments that claimed empty labels already block.
- `go-trader init` emits `"closed"` for wizard-gated strategies (composite-ranging defaults); existing configs keep `"open"`. CLAUDE.md + docs/ARCHITECTURE.md updated.

## Testing
- Full Go suite passes (new: policy x label x position x gate matrix, resolution order, load validation, hot-reload-while-open, store end-to-end per-strategy policy split, outage-note formatting, status-marker helper, init default).
- Full pytest suite passes (2410 passed, 1 skipped; new: backtester gate under both policies incl. warmup/mid-series gaps and closes-never-gated, `load_strategy_config` threading, Go-mirrored empty-label parity matrix).

---
LLM: Claude Fable 5 | high | Harness: anthropics/claude-code-action@v1


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-10T05:21:34Z) in 56a1e92.

### Fixed
1. **Backtest gate policy validates only the resolved value, not each surface independently** — validated the raw global `regime.gate_on_failure` independently before the existing `or`-chain resolution, so a garbage global value is now rejected even when a valid per-strategy `regime_gate_on_failure` override would otherwise short-circuit past it (`backtest/run_backtest.py:890-901`).

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-10T08:39:52Z) in 638ed67.

### Fixed
1. **`regime_gate_on_failure` resolution short-circuits validation across the three Python surfaces** — adopted `normalize_regime_gate_on_failure` (`shared_tools/regime.py:88`) as the single source of truth. `simulate_strategy.py` previously resolved the policy with an `or`-chain (old `shared_scripts/simulate_strategy.py:200-204`) so a garbage global `regime.gate_on_failure` silently passed whenever a valid per-strategy override existed — the same class 56a1e92 fixed in `run_backtest.py`. New pure helper `_resolve_gate_on_failure` (`shared_scripts/simulate_strategy.py:43`) validates both surfaces independently (global always normalized; per-strategy set-value wins), used at `shared_scripts/simulate_strategy.py:200`. `run_backtest.py` refactored from its hand-rolled checks to the same helper, preserving the config-path/strategy-id error context via re-raise (`backtest/run_backtest.py:886-909`). Resolution semantics stay byte-identical (per-strategy set wins, else global, else `open`).

### Note on backtester.py
`backtester.py`'s constructor check (`backtest/backtester.py:863-869`) was intentionally left inline. `regime` is lazy-imported there (`_load_regime`, `backtest/backtester.py:106`) so the module keeps its documented standalone/fallback path (`backtest/backtester.py:124` "Falls back to plain membership if the helper import fails"); a module-scope hard import to reach the helper would break that. Validation is unchanged and never weakened — the inline check is retained by design.

### Tests
New `shared_scripts/test_simulate_strategy_gate_on_failure.py` pins the tuner path, including the garbage-global + valid-per-strategy-override regression. Existing `backtest/tests/test_backtester_gate_on_failure.py` (already covering the run_backtest refactor case) and `shared_tools/test_regime.py` still pass — 70 passed.

---

richkuo:
@claude sonnet review

Issue #1150, opened 2026-06-30
[C60] feat: per-strategy pause/resume via paused config flag (manage-only gating)


richkuo:
**Complexity: 60/100** — scope: medium (config field + main-loop gating + control surfaces + status display); risk: high (gating sits on the live-exec dispatch and must never suppress SL/TP management); uncertainty: low (follows the existing #1046/#1048 patterns)

### Problem

There's no clean way to temporarily stop a strategy from opening new positions:

1. **Removing it from config** — SIGHUP hot reload rejects strategy add/remove (`validateHotReloadCompatible`), so a restart is required; state rows persist in the DB but open positions lose their managing strategy.
2. **`circuit_breaker: false`** — only disables the circuit-breaker firing arms (#1048); signal evaluation and new entries continue.
3. **Editing `capital` to 0** — not a pause switch. `shouldSkipZeroCapital` (scheduler/main.go) is the #100 balance-fetch guard: it only fires when `capital_pct > 0` resolved to $0, so a plain `capital: 0` doesn't trigger it — and when it does fire it skips the strategy's entire dispatch *before* the check script runs, leaving open positions unmanaged.

Operators need a pause that:
- Stops the strategy from opening new positions (entry/add/flip suppressed)
- Keeps existing positions fully managed (SL/TP, trailing stops, ratchet, protection sync, `close_strategy`)
- Preserves all state (cash, positions, trade history, risk state)
- Is reversible instantly without a restart

### Design

Add `paused *bool` to `StrategyConfig` in **config.json**:

```json
{
  "id": "momentum-btc",
  "paused": true
}
```

Nil/false/omitted → normal operation (backwards compatible).

**Semantics — manage-only gating, NOT a `dueStrategies` skip.** Trailing-SL updates, paper SL/TP simulation, the TP ratchet, and `runHyperliquidProtectionSync` all run inside the per-strategy dispatch of the main loop — a bare `if sc.Paused { continue }` before `dueStrategies` would leave paper positions with no SL/TP at all and live positions with stale software management. Instead, mirror the #1046 latched-circuit-breaker manage-only path (`circuitBreakerPermitsManagement`): the paused strategy stays in `dueStrategies` and runs its cycle, but entry/add/flip actions are suppressed. **Unlike CB manage-only (where Signal==0 suppresses closes too), pause must keep `close_strategy` evaluation live** so positions ride their natural exits.

**Control surfaces (no new CLI subcommand, no state.db column):**
[...]

Issue #1269, opened 2026-07-09
[C65] Daily losses never stop trading — add a hard daily loss limit that blocks new entries for the rest of the day


richkuo:
**Complexity: 65/100** — scope: medium-high (config + migration, risk gate, 6 dispatch sites, manual CLI, notifier, /status, hot-reload); risk: high (auto-protective money path — a wrong gate blocks all trading or fails to halt losses); uncertainty: medium (portfolio-wide aggregation basis needs a design call)

## Problem

Daily realized PnL is tracked per strategy but never gates trading — nothing stops the system from continuing to open new positions after an arbitrarily bad day.

Verified in code (line refs as of `d36d2fe`):

- `RiskState.DailyPnL` / `DailyPnLDate` exist at `scheduler/risk.go:633-634` and roll over at the UTC day boundary via `rolloverDailyPnL` (`scheduler/risk.go:1132-1137`), called from both `CheckRisk` (`risk.go:1416`) and `RecordTradeResult` (`risk.go:1580`).
- The doc comment on `RecordTradeResult` (`risk.go:1577-1578`) even says it updates risk state "for daily limits" — but no daily limit exists anywhere.
- `CheckRisk` contains exactly three halting arms: the latched circuit-breaker check (`risk.go:1419-1425`), the peak/margin drawdown circuit breaker (`risk.go:1491-1505`), and the 5-consecutive-loss circuit breaker (`risk.go:1509-1521`). `DailyPnL` is read by none of them; its only consumer outside persistence (`db.go:1253`/`db.go:1676`) and tests is a display estimate in `scheduler/portfolio_warning.go:136`.
- No `daily_max_loss` (or similar) config field exists in `scheduler/config.go` — searched `daily_max_loss` / `DailyMaxLoss` / `MaxDailyLoss` / `daily_loss` across `scheduler/`, zero hits.

So the only portfolio-protective mechanisms today are the peak-drawdown kill switch and per-strategy circuit breakers. A day of many small, non-consecutive losses spread across strategies trips neither: consecutive-loss counters reset on every win, and peak-relative drawdown can stay under threshold while daily bleed accumulates. A hard daily max-loss that halts new entries is a standard risk control this system lacks.

## Goal

[...]

PR #1301, opened 2026-07-10
Add same-direction exposure cap that blocks correlated new opens

### LINKED ISSUE #1270: [C55] Portfolio risk treats correlated crypto positions as diversified — add a same-direction exposure cap that blocks new opens
**Complexity: 55/100** — scope: medium (risk config, CheckRisk/dispatch gate, exposure aggregation, hot-reload, operator surfaces); risk: high (new blocking arm on the trade-dispatch path — a bug could wrongly block all entries or silently gate nothing); uncertainty: low-medium (bucketing design chosen; correlation-aware extension path needs spec, not research)

## Problem

Portfolio risk treats correlated crypto exposure as if it were diversified. The only portfolio-level exposure control is a **gross** notional cap:

- `scheduler/risk.go:582-605` — `PortfolioNotional` sums `quantity * price` (and option strike/value) across all strategies with no netting and no direction: a long and an equal short add together, and three same-direction positions look identical to three offsetting ones.
- `scheduler/risk.go:570-574` — the `MaxNotionalUSD` cap blocks new trades when gross notional exceeds the cap, nothing else. `PortfolioRiskConfig` (`scheduler/config.go:61-65`) has only `max_drawdown_pct` / `max_notional_usd` / `warn_threshold_pct`.
- `scheduler/main.go:1788-1793` — the per-strategy dispatch skip (`notionalBlocked`) is the sole exposure gate before a new open.

So a book of long BTC + long ETH + long SOL perps — near-perfectly correlated in a crypto selloff — passes every check as long as the gross sum stays under `max_notional_usd`. The realistic worst case is a leveraged, all-long crypto book that draws down as a single position while every per-strategy check stays green until the portfolio kill switch fires.

What exists today is **advisory only**:

- `scheduler/correlation.go:34-173` — `ComputeCorrelation` already computes signed per-asset delta (`NetDeltaUSD`/`GrossDeltaUSD`, options delta-weighted), per-asset concentration %, and same-direction strategy-count warnings. But its output is a snapshot + warning strings sent to Discord (`scheduler/main.go:1411-1428`) and surfaced via the `/correlation` command (`scheduler/discord_commands.go:298`); it never gates an entry. Its concentration check is also **per-asset vs portfolio gross** — BTC, ETH, and SOL can each sit under the 60% threshold while the book is 100% same-direction crypto.

Cross-reference (not duplicated here): #1159 covers **per-strategy** correlated hedge legs — opening an offsetting position for one strategy's thesis. This issue is the **portfolio-level blocking cap** on aggregate same-direction exposure; the two are complementary (a #1159 hedge leg would actually *reduce* the same-direction bucket measured here).

## Goal

New opens are blocked when aggregate same-direction exposure in a correlated bucket exceeds a configured cap, with an operator-visible skip reason. Blocking-only — the cap never force-closes, reduces, or mutates positions (same contract as the existing notional cap at `risk.go:570`). Position-decreasing actions (closes, trailing SL/ratchet/protection sync, reduce-only exits) always pass through.

## Approach

Start with **direction + asset-class bucketing**, not full covariance/VaR:

1. **Config** — extend `PortfolioRiskConfig` (or a sibling `exposure` block) with:
   - `max_same_direction_notional_usd` (0 = disabled): cap on the sum of same-direction signed exposure within a bucket. Phase 1 defines a single `crypto` bucket covering all crypto spot/perps (options delta included, mirroring `ComputeCorrelation`'s delta-weighting).
   - optional `max_asset_concentration_pct`: per-asset |net delta| as % of a configured reference (portfolio value, not gross — the existing gross-relative % in `correlation.go:128` self-normalizes and can't catch "everything is one asset").
2. **Measurement** — reuse/extract the signed-delta aggregation from `ComputeCorrelation` (`correlation.go:47-118`) rather than writing a second exposure model; compute long-bucket and short-bucket sums from per-asset `NetDeltaUSD`. Keep the pure computation in a lock-free helper testable without subprocesses (repo testing convention).
3. **Enforcement** — evaluate alongside the notional cap in the portfolio-risk phase and thread a per-direction blocked flag to the dispatch sites, gating only position-increasing signals for the offending direction (a short entry is still allowed when the long bucket is capped, and vice versa). Follow the existing `notionalBlocked` skip shape at `main.go:1788` including the `cbManageOnly` manage-only carve-out; use a distinct, explicit skip reason (e.g. `same-direction crypto exposure $X exceeds cap $Y — new long opens blocked`).
4. **Hot reload** — the new fields are threshold-only and blocking-only, so make them SIGHUP hot-reloadable. Note the deliberate divergence from `max_notional_usd`, which is currently restart-required (`scheduler/config_reload.go:415-417`); do not silently change that field's behavior in this issue.
5. **Operator surfaces** — surface the cap and current bucket utilization in `/status` and the startup summary; DM once on first block per direction (throttled), not every cycle.
6. **Extension path (spec only, not built here)** — the bucket definition becomes configurable: named buckets with asset membership and an optional pairwise correlation weight (measured or static), so the same-direction sum generalizes to a correlation-weighted exposure without touching the enforcement plumbing. Full covariance/VaR stays out of scope until bucketing proves insufficient.

## Acceptance criteria

- A book of long BTC + long ETH + long SOL perps whose combined signed exposure exceeds `max_same_direction_notional_usd` blocks the next long open with an explicit skip reason; a short entry on the same cycle is not blocked.
- Netting is honored: long $10k BTC + short $10k ETH contributes $10k long / $10k short — neither bucket double-counts, and neither blocks under a $15k cap.
- The cap never force-closes or reduces a position; closes, trailing SL/ratchet, and protection sync run normally while blocked (manage-only semantics preserved).
- Disabled by default (`0` = off); existing configs behave identically after upgrade.
- Threshold changes apply via SIGHUP without restart; existing `max_notional_usd` reload behavior unchanged.
- Options exposure is delta-weighted consistently with `ComputeCorrelation`; corrupt/zero-price positions fail safe (excluded from the sum, logged) rather than blocking everything or nothing.
- Go unit tests cover the pure bucket-aggregation helper and the gate decision (long-capped/short-allowed, netting, disabled, manage-only carve-out) without spawning Python.

---

richkuo:
## Summary

Closes #1270

Adds a blocking, direction-aware same-direction exposure cap:

- `portfolio_risk.max_same_direction_notional_usd` (0 = disabled): caps aggregate same-direction signed exposure in the phase-1 crypto bucket (spot/perps/manual positions + delta-weighted options; per-asset net deltas bucketed by sign, so netting is honored and a long/short pair never double-counts).
- `portfolio_risk.max_asset_concentration_pct` (0 = disabled): blocks new opens in an asset's net direction when its |net delta| exceeds the configured percent of portfolio value (not gross).

Measurement reuses one exposure model: `computeAssetDeltas` is extracted from `ComputeCorrelation` and shared by the advisory /correlation snapshot and the blocking gate — adds manual-type positions, an AvgCost fallback when no live price resolves, and fail-safe exclusion (recorded + logged) of corrupt/unpriceable legs.

Enforcement composes the #1150 `pausedBlocksSignal` classifier with the signal's sign at the 5 crypto dispatch sites, the options open-action filter (coarse delta direction), and the manual open/add/limit-open guards: a short entry passes while only the long bucket is capped, a long-to-short flip holds only under a short cap, all position-reducing actions (registry closes, pure-close exits, trailing SL/ratchet, protection sync) pass untouched, nothing is ever force-closed, and the TopStep futures site is deliberately ungated (CME futures are outside the crypto bucket).

Both thresholds hot-reload via SIGHUP (deliberate divergence from `max_notional_usd`, which stays restart-required). Operator surfaces: `[config]` startup line, per-cycle `[WARN]` while blocking, `/status` bucket-utilization note, and an owner DM edge-triggered on the transition into blocked per direction/per asset (re-arms on clear); a concentration arm that cannot evaluate is loudly inert.

Tests: pure-Go coverage (no Python subprocess) of the aggregation (all-long book, cross- and same-asset netting, AvgCost fallback, manual inclusion, fail-safe exclusions), the gate decision (direction-aware blocks, flips, pure-close pass-through, manage-only carve-out), the options filter, alert edge-triggering, config validation bounds, and hot-reload compatibility. Full `go test` suite passes.

---
LLM: Claude Fable 5 | high | Harness: anthropics/claude-code-action@v1


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-10T05:22Z) in a41e491.

1. **Concentration arm silently unenforced on manual entries** — implemented the reviewer's first suggestion: the manual path now derives a portfolio-value basis instead of passing 0. `manualExposureCapStatus` (`scheduler/exposure_cap.go:125`) sums `displayStrategyValue` at the same AvgCost fallback the bucket sums already use on this feed-less path — the identical basis this PR's `/status` note uses — so `max_asset_concentration_pct` now evaluates and blocks on manual open (`manual_core.go:530`), manual-add (`manual_core.go:851`), and limit-open (`manual.go:139`) via the shared `exposureCapManualEntryBlock` (bucket arm first, then per-asset concentration, entry direction only — same order as the dispatch gate). When the basis is still unusable (zero-value book), `exposureCapPVBasisMissWarning` is surfaced on all three paths instead of being dropped. Basis caveat is documented in code + ARCHITECTURE.md: the standalone CLI virtual-sums shared-wallet members (can overstate the basis, never the bucket sums); the daemon-embedded dashboard path picks up reconciled values.

All three "must survive" cases are covered by new tests in `exposure_cap_test.go` (concentration-only config through open/add/limit-open, net-direction-only blocking, PVBasisMiss loudly-inert). Full `go test` suite passes.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-10T08:48Z) in 4de9dec.

1. **No end-to-end regression test through the manual guard call sites** — added `TestManualOpenCoreRefusesExposureCap` and `TestManualAddCoreRefusesConcentrationOnly` (`scheduler/exposure_cap_test.go`), mirroring the #1269 daily-loss pair: blocking view injected via `manualCoreDeps.loadState`, asserting both the refusal text and that `execute` is never invoked. All three "must survive" cases covered, including the inverse-direction over-block checks (short open under a long-capped bucket, and a short position's add under a long-over-concentrated asset, both reach `execute`). Full `go test` suite passes.

---

richkuo:
@claude sonnet review

Issue #1085, opened 2026-06-20
[C85] Evidence-gated, default-off migration for regime directional selection (#1076 follow-up)


richkuo:
**Complexity: 85/100** — scope: medium-large (config-load gate, a certified-edge artifact, the directional resolver, hot-reload, shared-coin orphan handling, a from-flat migration, tests/docs); coupling: high (dual-language SSoT — the statistical test in Python vs the gate in Go — plus live↔backtest parity and shared-coin coordination on a live-direction path); risk: high (changes live HL-perps long/short direction behavior; shared-coin positions can be stranded if decertified unsafely); uncertainty: medium (certification provenance/lifetime is open).

## Context

#1076 validated the premise behind `regime_directional_policy` (#779) and regime directional entry-gating — regime label → forward DIRECTION — and found it **empirically false across the tested universe** (0/2121 per-state forward-return tests survive global FDR/Bonferroni; the regime-timing book never beats its block-shuffled-label null, 0/60 after FDR). Evidence + harnesses: PR #1084 (`backtest/research/regime_1076_*`, `README_1076_directional_premise.md`).

PR #1084 shipped the safe, non-breaking guard: a `[WARN]` at config load (`regimeDirectionalPolicyWarnings`, `scheduler/config.go`). This issue is the **principled end-state** the warning points toward.

## Proposal

Make the directional-selection surface **default-off / fail to base direction**, and activate it for a given strategy only where a **per-(asset, timeframe, classifier) validation gate** certifies real, multiplicity-honest directional edge — mirroring the #1073/#1078 `gate_verdict` "abstain unless the incumbent is trustworthy" philosophy. None of the tested universe currently qualifies, so the default is: no regime-based side selection.

## Safety requirements (the hard part — why this is its own issue)

1. **From-flat migration only.** Disabling/changing the policy on a strategy with an open position must not strand it. SIGHUP already blocks the change while open (`config_reload.go`); a config-rewrite + restart relies on the #822 orphan auto-close (`perpsRegimeDirectionOrphanConflict` → `runRegimeDirectionOrphanCloses`, `scheduler/hyperliquid_balance.go`), which fires **only for sole-owner coins**. A shared-coin live short would be left for manual close — the migration must detect and surface this, not silently flip behavior.
[...]

PR #887, opened 2026-06-04
fix(ratchet+atr): tighten default final-tier trail to 0.8×ATR; round ATR to whole numbers

richkuo:
## Summary

- Raises the tightest rung in all default ratchet ladders (scalar, clean, choppy, ranging) from `0.5×ATR` to `0.8×ATR` — prevents the final trailing stop from sitting so close that normal volatility triggers an early exit before the position has fully run. Go and Python mirrors kept in sync.
- Rounds all computed ATR values to whole numbers at the three computation sites (`standard_atr`, `breakout_strategy`, `atr_breakout_strategy`) so stop/TP distances don't carry sub-dollar precision noise.

## Test plan

- [ ] Verify `defaultTrailingRatchetTiers` and `ratchetTierGroupDefaults` all reflect `0.8` as final-tier `TrailingMultAfter`
- [ ] Verify `DEFAULT_RATCHET_TIERS` and `DEFAULT_RATCHET_TIERS_BY_GROUP` Python mirrors match
- [ ] Run `uv run --no-sync python -m pytest shared_strategies/ shared_tools/` and confirm no regressions
- [ ] Smoke-test a ratchet strategy config with `use_defaults:true` and confirm `trailing_mult_after` on the last tier is `0.8`
- [ ] Confirm `latest_atr` returns integer-valued floats for a sample asset

---
Created with LLM: Sonnet 4.6 | high | Harness: commit-push-pr

PR #1324, opened 2026-07-10
feat(#1277): configurable Wilder-RMA ATR behind a gated atr_method migration (v17)

### LINKED ISSUE #1277: [C65] ATR is a simple average with whole-number rounding above 100 — switch to the standard Wilder formula behind a gated migration
**Complexity: 65/100** — scope: medium (consolidation already landed via #1281/#1321; remainder is the method param, config surface + v17 migration, check-script/backtester threading, regime decision, baseline re-runs); risk: high (ATR feeds every live stop-loss distance, trailing stop, tiered TP band, and regime classification — changing the math moves live stops and invalidates documented backtest baselines); uncertainty: moderate (magnitude of baseline/verdict deltas and regime-certification impact unmeasured until re-runs)

## Problem

The codebase computes ATR as a **simple rolling mean of True Range**, not the standard Wilder smoothing (RMA, `ewm(alpha=1/period, adjust=False)`), and quantizes it to whole numbers once it reaches 100.

Current state (post-#1281/PR #1321, which consolidated all indicator math):

- The math lives solely in `shared_strategies/open/indicators_core.py` — `atr_from_true_range` is the single choke point (`indicators_core.py:87` `tr.rolling(window=period).mean()`), with `atr_sma`/`atr_sma_series` wrappers. `shared_tools/atr.py:standard_atr` is a thin re-export. Zero inline TR-rolling-mean copies remain; the module's docstring explicitly reserves the smoothing-method change for this issue (`indicators_core.py:20`).
- The ≥100 integer rounding is the documented #887 convention (`round_atr_large`, `indicators_core.py:66-72`): BTC-scale ATRs round to whole numbers (up to ~0.5% error near the 100 boundary); sub-100 values pass through. The convention split is now explicit via the `round_large` flag — `False` at the supertrend / squeeze_momentum / order_blocks / session_breakout / sweep_squeeze_combo / chart_patterns sites, default `True` everywhere else — so different strategies deliberately disagree on ATR for the same bars.
- `standard_atr` (and its `ensure_atr_indicator`/`latest_atr` wrappers) is consumed by all five live check scripts (`check_hyperliquid.py:62,200,228,1206`; `check_okx.py:24`, `check_robinhood.py:27`, `check_topstep.py:25`, `check_strategy.py:25`) for `EntryATR` stamping and live `market_ctx["atr"]`, by `backtest/backtester.py:1818`, by `shared_scripts/simulate_strategy.py:20`, and by the **regime classifier** (`shared_tools/regime.py:205,424,458` — `atr_pct` feeds composite regime thresholds).
- No Go-side ATR computation exists (`scheduler/*.go` has no True Range/ATR calculator — Go consumes Python-stamped `EntryATR` and `market_ctx` values), so no Go mirror is required.

Why it matters: Wilder's RMA is the published ATR definition. Reference parameterizations of supertrend, Keltner/squeeze, and ATR breakout assume it; a simple mean responds faster and is noisier, so the current bands and stops do not match published behavior for the same period settings. The integer rounding coarsens live stop geometry on BTC-scale prices.

## Goal

Standardize ATR on Wilder smoothing and remove the ≥100 integer rounding — behind an explicit, version/config-gated migration that never silently moves a live stop, with an explicit decision on which ATR sites cut over, and re-established backtest baselines under the new math.

## Approach

1. **Consolidation is done** — #1281/PR #1321 already reduced every inline copy to `indicators_core.py` with frozen numerics. This issue starts from that substrate.
2. **Add the method.** Add a `method` parameter (`"simple"` default, `"wilder"` = `tr.ewm(alpha=1/period, adjust=False).mean()`) to `atr_from_true_range` — the single choke point — threaded through `atr_sma`/`atr_sma_series` and `shared_tools/atr.py:standard_atr`. Byte-identical output at the default, snapshot-asserted.
3. **Decide the cutover scope explicitly.** The config-gated switch governs the `standard_atr` path (EntryATR stamping, live `market_ctx["atr"]`, backtester ATR injection — the stop/TP geometry surface). Strategy-internal ATR calls inside cores (supertrend/Keltner/breakout band math, including the `round_large=False` sites) are hardcoded conventions, not config-driven — decide per-site whether each adopts Wilder (their reference parameterizations assume it) or stays frozen, and document the roster. "Every site in lockstep" is not assumed.
4. **Gated cutover.** Introduce the switch as an explicit config surface (global with per-strategy override, e.g. `atr_method`), migrated at a `config_version` bump (v16 → v17; `config_migration.go:13`). Treat it as state-shifting: blocked by `validateHotReloadStateCompatible` while a position is open (an in-flight `EntryATR`/frozen geometry must not be re-based). Fresh opens after cutover use Wilder. Thread to Python via the standard per-strategy-flag pattern (StrategyConfig field → `run*Check` CLI flag → check-script parse → init/wizard), appending to `probeArgv` if the flag is runtime-required.
5. **Rounding removal rides the same gate.** `wilder` mode never rounds; `simple` mode keeps the #887 rounding frozen for reproducibility of old baselines.
6. **Regime classifier decision.** `regime.py` composite thresholds and the #1085 directional certifications were calibrated on simple-mean `atr_pct`. Either pin `regime.py` to `method="simple"` explicitly (decoupling stop math from regime labels), or re-run the bounded-window/regime-promotion validation before letting Wilder flow into classification. The implementation must pick one and document it.
7. **Baseline re-runs.** Thread the method through `run_backtest.py`/`backtester.py` and re-run documented baselines under Wilder: `docs/research/1181-htf-filter-baseline-revalidation.md`, and flag which M1/M2/M6 study verdicts (`docs/backtesting-registry.md` rows) depended on ATR-based exits (`tiered_tp_atr*`, `atr_stop`, trailing ATR) and need re-checking before any future promotion decisions cite them.
8. **Tests.** Regression tests pinning both maths: simple-mean output unchanged (including the #887 rounding), Wilder output matching a hand-computed RMA fixture. The no-new-inline-copies guard exists from #1281. Existing constraints stay honored: `stampEntryATRIfOpened` rejects ATR > 50% of AvgCost; Go tests keep ATR < 50% of entry price.
9. **Docs.** Update the CLAUDE.md ATR note and ARCHITECTURE.md (method surface, cutover scope roster, regime decision) in the same PR.

## Acceptance criteria

- `method` supported at the `indicators_core` choke point (`simple`/`wilder`); `standard_atr` threads it.
- With the gate off (default at merge), all live check scripts, the backtester, and `--list-json` strategy discovery produce output identical to today (snapshot-diffed).
- Cutover to Wilder is an explicit versioned config change (v17), blocked while positions are open, and removes the ≥100 integer rounding on the Wilder path.
- The cutover-scope roster is explicit and documented: which sites follow `atr_method` (standard_atr path) and the per-site verdict for strategy-internal ATR conventions.
- Regime classification method is explicitly pinned and documented (simple kept, or Wilder with re-validated certifications).
- Re-run results for the documented baselines are committed, with a list of prior study verdicts that must be re-checked under Wilder.
- Tests cover both maths; CLAUDE.md/ARCHITECTURE.md updated.

---

richkuo:
Closes #1277

## What

ATR smoothing is now configurable behind an explicit, versioned gate. `atr_method: "wilder"` (global or per-strategy, config v17) switches the **standard-ATR surface** — entry-ATR stamping when the open strategy emits no `atr` column, live `market_ctx["atr"]`, manual `--fetch-atr`, backtester injection, tuner simulate — from the frozen simple rolling mean to the published Wilder RMA (`tr.ewm(alpha=1/period, adjust=False).mean()`), which never applies the ≥100 integer rounding (#887). Default (`"simple"`) is byte-identical to today.

## Key decisions (per the validated issue)

- **Single choke point:** `method` lands on `indicators_core.atr_from_true_range` and threads through `atr_sma`/`atr_sma_series`/`standard_atr`. Unknown values fail loud on both sides (Go `validateConfig`, Python `normalize_atr_method`, argparse `choices`).
- **Explicit cutover roster** (no "every site in lockstep"): regime classification is **pinned `method="simple"`** at all three `regime.py` sites (composite thresholds + #1085 certifications were calibrated on it); strategy-internal `indicators_core` call sites keep their frozen per-site conventions; a strategy-emitted `atr` column always wins over injection. Roster documented in `docs/research/1277-wilder-atr-cutover.md`.
- **Never silently moves a live stop:** hot-reload of the *resolved* method (so a global flip through inheritance is caught too) is blocked while the strategy holds an open position; applies when flat. Options excluded (no ATR surface; the per-strategy field is rejected on options at load). Startup summary + `inspect` tag non-default resolution `atr=wilder`.
- **Deploy contract:** Go appends `--atr-method` on all 5 signal-check dispatches + HL `--fetch-atr`; `probeArgv`/`probeCompositeArgv`/`fetchATRProbeArgv` carry the flag so an asymmetric deploy fails at startup. `executeProbeArgv` deliberately doesn't (the execute argv never carries it).
- **v17 is stamp-only/additive** — no on-disk rewrite; operators get a one-time migration DM describing the new surface.
- **Backtest parity:** `Backtester(atr_method=…)`; `run_backtest.py --config` reads the config's value (both surfaces validated independently); CLI `--atr-method` drives config-less single/compare/multi runs, is rejected alongside `--config` and in optimize mode (optimizer engines run on the default).

## Baseline impact

- Gate off ⇒ nothing moves: `--list-json` snapshot diffs byte-identical vs `main`; a default single-mode backtest diffs byte-identical against explicit `--atr-method simple`.
- Committed a measured wilder-vs-simple delta (`mean_reversion_pro` BTC/USDT 1h, 13,337 candles: PF 0.792→0.756, 49→46 trades, same entries — exit geometry only) plus the list of documented study verdicts that must be re-checked before being cited under wilder (`docs/research/1277-wilder-atr-cutover.md`). `1181-htf-filter` baselines verified unaffected (open-signal-as-close runs never inject ATR).

## Verification

- `go -C scheduler test ./...` — full suite green (two version-pin tests updated for v17: `TestCurrentConfigVersion_IsSeventeen`; the agent-info snapshot test now forces v15 so it still exercises a synchronous write path).
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 2680 passed, 1 skipped.
- New tests: hand-computed RMA recursion fixture, wilder-never-rounds + `round_large`-inert guards, explicit-simple byte-identity, regime pin (on a frame where the two methods provably diverge), `atr.py` threading, Backtester validation + injected-series divergence, `load_strategy_config` resolution/rejection; Go: precedence, validation (incl. options reject), hot-reload block matrix (per-strategy, removal, global-through-inheritance, masked-by-explicit, options exclusion), restart-shape mask, apply-when-flat change lines, probe argvs, v17 stamp, `init --json` emission, summary tag.
- All five check scripts + `--fetch-atr` probed OK with the new argv shape; `go-trader init --json '{"atrMethod":"wilder"}'` emits the field, invalid values fail the init.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11T01:32:50Z) in d8d0664.

### Fixed

1. **Restart-time atr_method drift can silently re-base a live position's stop/TP geometry, and the v17 notice overstates the guard** — the SIGHUP hot-reload guard (`validateHotReloadStateCompatible`, `scheduler/config_reload.go:603`) only fires on the SIGHUP path; a config edit + process restart bypasses it (a fresh process has no "old" resolved value to diff against). Implemented both remedies suggested rather than only the cheaper one:
   - `Position.ATRMethodAtOpen` freezes the resolved method at open (`scheduler/atr_method.go` `stampATRMethodAtOpenIfOpened`, wired into all 5 execute dispatch sites in `scheduler/main.go`; never re-stamped on a scale-in add, mirroring the `RiskAnchorPrice`/`EntryATR` freeze-at-entry convention). Persisted via a new `atr_method_at_open` column (`scheduler/db.go`).
   - `checkATRMethodDriftAtStartup` (`scheduler/atr_method.go`) runs once per boot, compares each open position's frozen stamp against the live-resolved method, and DMs the owner on a mismatch — mirrors the `ValidatePerpsDirectionConfig` startup-warning shape (`scheduler/main.go:270-276`, `:490-497`).
   - Reworded `v17ATRMethodNotice` (`scheduler/config_migration.go:150`) so the one-time migration DM no longer implies an absolute block; it now states the restart caveat and that a startup check DMs the owner if it happens.

New tests: `TestStampATRMethodAtOpenIfOpenedFreshOpen/SkipsAdds/NoOp`, `TestCheckATRMethodDriftAtStartup` (7 subtests covering drift detection, global-flip propagation, no-drift, pre-#1277/flat/options skips, nil-safety), `TestSaveLoadState_ATRMethodAtOpenRoundTrip`. `go build`, `go vet`, and the full scheduler test suite pass.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11T02:05:13Z) in db87f05.

### Fixed
1. **Manual-opened positions never get `ATRMethodAtOpen` stamped** — Confirmed: both `applyManualAction`'s open case (`scheduler/manual.go`) and `applyLimitFillProgress`'s first-fill case (`scheduler/manual_limit.go`) built `Position` literals with an empty stamp, permanently hiding manual positions from `checkATRMethodDriftAtStartup`. Implemented the review's suggested shape, following the `PendingManualAction.EntryATR` precedent: the market/record path resolves `atr_method` at queue time in `manualOpenCore` (next to the EntryATR fetch) and carries it on the row via a new `atr_method` column (`scheduler/db.go`, idempotent migration), so the drain stamps the method the EntryATR was actually computed under even when a config edit + restart lands between queue and drain — a drain-time re-resolve would mask exactly that window. Pre-upgrade rows (empty carried value) fall back to drain-time resolution instead of staying unstamped. The limit path resolves at fill-adoption time — the same moment `entryATR` is fetched in `reconcilePendingLimitOrders` — and `applyLimitFillProgress` freezes it at first fill only (partial fills never re-stamp, matching the EntryATR/RiskAnchorPrice freeze-at-entry convention). No empty-stamp imputation anywhere, so an explicitly-configured non-default method never yields a false-positive warning. Regression tests cover all three "must survive" cases plus the DB round-trip (`scheduler/atr_method_test.go`). Full Go suite passes.

---
Created with LLM: Fable 5 | medium | Harness: Claude Code

richkuo:
@claude review

PR #1336, opened 2026-07-11
feat(#1159): per-strategy correlated hedge legs (phase 1, live HL perps) [Opus 4.8, high, fableplan]

richkuo:
Implements the Fable 5 plan posted on the issue (`## Implementation plan (Fable 5)`). Phase 1 of per-strategy correlated hedge legs: an opt-in `hedge` block opens an inverse leg on a correlated coin (e.g. long ETH → short BTC) whose lifecycle is strictly coupled to the primary. Hyperliquid perps only.

## Design
Hedge management is a **per-cycle state-derived reconciler** (`runHedgeSync` → `hedgeTargetDecision`), not per-event mirror hooks. It converges the hedge leg to the primary by diffing primary quantity against a persisted watermark (`Position.HedgePrimaryQtyBasis`) — mark-price drift never re-trades, only a real primary qty/side change does. Every primary event (fresh open, scale-in add, partial/full close, external close, ratchet) produces the matching hedge action within the same or next cycle. Events the dispatch never sees (kill switch, per-strategy CB) get explicit extensions.

**Fail-closed:** a fresh-open cycle whose hedge open fails unwinds the primary reduce-only (cancelling its just-armed SL/TP OIDs) + CRITICAL owner DM — never runs unhedged. Later add/manage failures alert + retry (the state-derived design self-heals).

## What changed
1. `config.go` — `HedgeConfig` + accessors + `validateHedgeConfigs` collision matrix (hedge coin ≠ own primary, any configured strategy coin, or another hedger's coin; `direction=both` rejected); nested unknown-key guard.
1. `hedge.go` (new) — decision core, open/add/reduce/close booking, fail-closed primary unwind, orchestrator with the 6-phase lock discipline.
1. Persistence — `Position.HedgeFor`/`HedgePrimaryQtyBasis` (DB migration + save/load), startup consistency check.
1. Accounting — hedge legs book `trade_type="hedge"` (excluded from `#T`/W-L; ledger sums still attribute to the owner) and route PnL through `RecordHedgeTradeResult` (daily-PnL only, **never the loss streak**); diagnostics skipped.
1. Integrations — marks, wallet uPnL/funding attribution, fill resolver, reconcile (own pass; #822 orphan-close guarded off; foreign-coin non-adoption WARN), kill-switch on-chain roster, CB pending close.
1. Hot-reload blocked while a hedge leg is open; backtester loud-rejects enabled hedge configs; operator surfaces (inspect `[config]` line, Discord `/status`).

## Deviations from the plan
[...]

PR #1333, opened 2026-07-11
feat(#1159): per-strategy correlated hedge legs (phase 1, live HL perps) [Fable 5, high]

richkuo:
Closes #1159 (phase 1: live Hyperliquid perps only).

> Note: this is an independent implementation, opened deliberately alongside PR 1331 and PR 1332 per operator instruction — pick one, close the others.

## What

A hedge-enabled strategy carries one auto-managed inverse leg on a different coin, strictly coupled to the primary position's lifecycle. New per-strategy `hedge` config block (`enabled`/`symbol`/`side:"inverse"`/`ratio`/`platform`/`type`/`margin_mode`/`leverage`); default behavior unchanged for everyone else.

## Design: one convergence engine, not per-path mirroring

`hedge_sync.go` runs once per cycle **after** the dispatch loop and converges each hedge leg with its primary from virtual state. This single mechanism covers every lifecycle source — fresh open, scale-in add, partial close, full close, **on-chain SL fills booked by reconcile**, and CB force-closes — instead of hooking each dispatch path separately (SL-fill closes never pass through dispatch, so per-path mirroring would leave naked hedges).

The ongoing target is **quantity-anchored** via a persisted covered watermark (`Position.HedgeCoveredPrimaryQty`): marks are consulted only to size opens/adds (`primary_qty_delta × primary_mark × ratio ÷ hedge_mark`), reductions are proportional to covered quantity — so mark moves never churn orders (regression-tested).

**Fail-closed (constraint 4):** any hedge open/add that cannot be placed or confirmed — order error, missing mark, unverifiable clearinghouse state, or a foreign on-chain position on the hedge coin — immediately closes the *uncovered* primary quantity reduce-only (sized partial, never a full-coin close, so shared-primary peers survive) and DMs the operator. Reduce/close failures alert and retry next cycle (over-hedged is the safe direction).

## Coupling with every wallet mechanism

- **Kill switch** — hedge coins join the traded-coin roster (config-derived + state-claimed via `HLHedgeStateCoins` for legs whose hedge block was removed across a restart); hedge fills book to the owner before generic cleanup.
- **Circuit breaker** — pending closes capture the hedge leg at fire time (before `forceCloseAllPositions` wipes virtual state); the stuck-CB drain reconstruction re-derives it from config + on-chain size.
[...]

PR #1332, opened 2026-07-11
feat(#1159): add correlated hedge legs [GPT-5.6 Sol, high]

richkuo:
## Summary

- add opt-in inverse Hyperliquid perps hedge legs with static primary-notional sizing and explicit hedge margin settings
- couple hedge open, scale, reduction, full close, circuit-breaker, kill-switch, force-close convergence, and restart reconciliation to the owning primary position
- fail closed by unwinding the primary when hedge placement cannot be confirmed, and preserve exact-fill fee/PnL attribution without guessed hedge reconciliation
- persist and expose hedge ownership metadata, reject coin collisions and unsupported configurations, and reject hedge-enabled backtests until parity is implemented

## Verification

- `go -C scheduler test ./...`
- `go -C scheduler test -race ./...`
- `go -C scheduler build .`
- `uv run --extra test pytest shared_strategies/ shared_tools/ platforms/ backtest/ -q` — 2727 passed, 1 skipped

## Follow-ups

- allow hedge coins to overlap peer strategies by extending shared-coin ownership, margin, circuit-breaker, kill-switch, and reconcile attribution
- add hedge PnL, fee, and slippage modeling to the backtester with live-parity guarantees

Closes #1159

LLM: GPT-5.6-Sol | high | Harness: Codex

PR #1331, opened 2026-07-11
feat(#1159): phase-1 per-strategy correlated hedge legs [Grok 4.5, high]

richkuo:
## Summary
- Add opt-in per-strategy `hedge` config for live Hyperliquid perps: inverse side, static notional×ratio sizing, own margin/leverage, peer-collision rejection (own coin / strategy coin / hedge-vs-hedge).
- Scheduler mirrors primary lifecycle (open / scale-in / partial / full close) with fail-closed primary unwind when hedge open/sizing fails; no hedge SL/TP; kill-switch, CB drain, reconcile, and protection paths treat hedge legs via persisted `IsHedge` metadata.
- Hot-reload blocks hedge changes while any position (or hedge leg) is open; backtester loudly rejects enabled hedges until parity lands; inspect/status/Discord surfaces tag hedge legs.

Closes #1159

## Test plan
- [x] `go -C scheduler test -count=1 -run 'TestHedge|TestValidateHedge|TestApplyHedge|TestSyncStrategy|TestCollectPerpsMarkSymbolsIncludesHedge|TestForceCloseHyperliquidLiveIncludes' .`
- [x] `go -C scheduler test -count=1 .`
- [x] `uv run --no-sync python -m pytest backtest/tests/test_backtester_scale_in.py -k hedge -q`
- [ ] Manual: enable hedge on a paper-ineligible live HL strategy config and confirm load validation / collision rejects
- [ ] Manual: after merge, smoke a sole-owner hedge open→close on a small live size and confirm fail-closed unwind path via injected hedge failure in tests

## Follow-ups (unfiled)
- Allow hedge coins that overlap peer strategies (extend coin-membership / margin / CB / KS / reconcile).
- Backtest hedge PnL/fee/slippage modeling with parity.

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

PR #1337, opened 2026-07-11
feat(#1159): per-strategy correlated hedge legs (phase 1, HL perps) [Opus 4.8, xhigh, fableplan]

richkuo:
## Summary

Phase-1 implementation of per-strategy correlated hedge legs: an opt-in `hedge` config block lets a Hyperliquid perps strategy automatically open/manage an inverse hedge leg on a second coin, strictly coupled to the primary position's lifecycle.

Closes #1159

**Note:** this issue already has several other independent implementation attempts open (#1331, #1332, #1333, #1336). This PR is an intentional, user-approved 4th independent attempt, planned by a Fable 5 subagent (plan posted at https://github.com/richkuo/go-trader/issues/1159#issuecomment-4948757648) and built by Claude Code following that plan.

## What changed

- **Config** (`config.go`, `config_migration.go`): `HedgeConfig` struct + accessors (`HedgeEnabled`, `hedgeCoin`, `hedgeRatio`, `hedgeExchangeLeverage`, `hedgeMarginMode`), full validation including phase-1 collision rejection (`hyperliquidHedgeCollisionErrors`: hedge symbol vs own coin, vs any peer strategy's coin, vs another strategy's hedge symbol). `CurrentConfigVersion` bumped 17→18, stamp-only/additive (mirrors #1277's v17).
- **Data model** (`portfolio.go`, `db.go`): `Position.IsHedge`/`HedgeForSymbol`/`HedgeForPositionID` (hedge leg) and `Position.HedgeSymbol` (primary leg) are the sole restart/reconcile pairing key — never coin→config inference. `Trade.IsHedge` excludes hedge legs from `#T`/W-L (`LifetimeTradeStats*`) while keeping them in `tradeLedgerDeltaSQL` sums (still books to the owning strategy).
- **Execution engine** (`hedge.go`, `hedge_dispatch.go`): pure sizing/side helpers (`hedgeOpenQty`, `hedgeReduceQty`, `hedgeSideForPrimary`), HL order wrappers, and state appliers. `mirrorHedgeAfterPrimaryFill` (wired into the main.go HL perps dispatch, right after the primary's fill is applied) is **delta-based**: it diffs the primary's pre-cycle snapshot against its post-apply state and mirrors whatever happened — open, scale-in add, partial close, full close, or flip — onto the hedge, so no per-branch special-casing was needed in the already-dense dispatch block. Hedge-open failure fails closed (constraint 4): a sized (never `sz=None`) reduce-only unwind of the just-opened primary; if the unwind itself fails, the primary open stays booked (the fill was real) and the sweep retries.
- **Coherence sweep** (`hedge_sweep.go`, wired after reconcile/before dispatch): the single repair engine for every asynchronous desync path — reconcile-booked external changes, kill-switch/CB residue, manual force-close, restart. Detects P-without-H (fail-closes the primary), H-without-P (closes the orphaned hedge), and an over-sized hedge on a both-open pair (reduces toward the configured ratio using live marks; never auto-upsizes outside a primary entry event).
- **Kill-switch / circuit-breaker / reconcile / shared-wallet** (`hyperliquid_balance.go`, `risk.go`, `shared_wallet_reconcile.go`): hedge coin added to `tradedCoins` (kill-switch), CB pending (including the case where the *primary* is shared with a peer but the hedge coin is still sole-owned — stripped rather than dropping the whole pending job), the reconcile sole-owner pass, virtual-quantity snapshots, and shared-wallet funding attribution.
- **Hot-reload** (`config_reload.go`): the `hedge` block is state-shifting — blocked by `validateHotReloadStateCompatible` while *either* leg is open (the hedge Position lives in the same `Positions` map `strategyHasOpenPositions` already scans), freely reloadable while flat. Reload-time collision re-validation added alongside the existing peer-strategy re-check.
- **Manual surfaces** (`manual_core.go`, `manual.go`): `force-close` notes that the hedge leg will be handled by the scheduler's next sweep cycle; `validatePendingManualActionStrategy` fail-closes any pending manual action that targets a hedge coin (the daemon owns it exclusively).
- **Operator surfaces** (`inspect_cmd.go`): `inspect` (text + JSON) shows the hedge config block and tags hedge/hedged positions in the position list.
- **Backtest** (`run_backtest.py`): `--config` loudly rejects `hedge.enabled=true` (HL-live-only in phase 1; parity modeling deferred).

## Deviations from the Fable 5 plan (per the code, not the prose)

- **`checkHedgeStateDriftAtStartup` is a warning-only owner-DM, not a fail-loud process exit.** The plan described this as "refuse startup, mirrors `ValidatePerpsDirectionConfig`" — but tracing the actual code, `ValidatePerpsDirectionConfig` and `checkATRMethodDriftAtStartup` (the two closest precedents) are both `[]string` warning collectors, not fail-loud gates; nothing in this codebase refuses to boot on a startup-drift finding. I followed the real, established pattern instead — a hard refuse-to-boot would add an outage risk of its own, and the coherence sweep is the actual repair path regardless.
- **Operator surfaces scoped to `inspect`** (text + JSON) rather than also touching `/status` HTTP, Discord `/status`, and the UI static assets. Hedge positions are not hidden — they remain visible in every existing generic position listing via `IsHedge`/`HedgeForSymbol` — but dedicated chip/suffix rendering in those surfaces is not yet wired. Flagging as follow-up scope rather than silently dropping it.
- **`buildCachedHyperliquidReconcileFillResolver`'s pre-lock userFills prefetch was not extended to hedge coins.** Reconcile still reuses `reconcileHyperliquidPositionsForStrategy` correctly for the hedge coin; without the prefetch a hedge-coin external-close fill lookup falls back to a per-OID on-demand call (correct, just not batched) instead of a cache hit. Efficiency gap only, not correctness.
- **Flip handling** (primary `direction=both` reversing side) reuses the same close+open primitives as the partial-close and fresh-open paths but has no dedicated unit test beyond the shared building blocks (`hedgeReduceQty`/`hedgeOpenQty`/collision validation are all directly tested).

## Test plan

- `scheduler/hedge_test.go` (33 tests): pure sizing/side helpers, hedge-coin normalization + accessor defaults, config-validation collision rejection (own-coin, peer-coin, hedge-vs-hedge), ratio bounds, manual-type rejection, `BTC/USDC:USDC`-form normalization, hot-reload state-compat (blocked while either leg open, allowed while flat), startup-drift warning generation (all three arms + the consistent-pair no-warning case), and coherence-sweep job detection (P-without-H, H-without-P, both-absent, oversized-hedge-reduces, consistent-pair-no-job).
- `backtest/tests/test_backtester_hedge_reject.py`: `--config` with `hedge.enabled=true` raises; `enabled=false` and no-block both load clean.
- `go build`, `go vet`, `gofmt -l` clean; `go test ./...` — full suite green (no regressions), including the version-pin test updated for v18.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 2739 passed, 1 pre-existing skip.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11 21:59) in 28d81c9.

### Fixed
1. **Live primary closing in-cycle mirrored onto the hedge as PAPER, stranding the on-chain hedge leg** — confirmed: `execResult` stays `nil` on a `Signal==0` live cycle where `hyperliquidArmFixedATRStopLossLive`'s immediate-fill branch closes the primary via `recordPerpsStopLossClose` (`main.go:2362-2384`), so `mirrorHedgeAfterPrimaryFill`'s old `live := execResult != nil` silently took the no-order paper path. Fixed by deriving `live` from `hyperliquidIsLive(sc.Args)` (`hedge_dispatch.go:63`).
2. **Flip re-opens the hedge on the inverse side even when the preceding hedge close failed, corrupting the Position via a same-position opposite-side blend** — confirmed: `applyHedgeOpenFill`'s blend branch summed quantities regardless of `Side` (`hedge.go:132-140` pre-fix). `mirrorHedgeReduce` now returns a success bool and the flip's reopen is skipped when the close failed (`hedge_dispatch.go:78-91`); `applyHedgeOpenFill` also independently refuses to blend into an opposite-side Position and alerts CRITICAL when a real fill is refused a home in state (`hedge.go:113-116`, `hedge_dispatch.go:167-190`).

1. **Crash-between-legs on-chain hedge orphan (on-chain hedge order confirms, state write never commits)** — implemented recovery rather than leaving it an accepted gap: the coherence sweep now checks the cycle's on-chain snapshot for the hedge coin before fail-closing an "unhedged" primary (`hedge_sweep.go` `onChainCoinQty` + `runHedgeSweepClosePrimary`). Collision rejection guarantees the hedge coin is exclusively this strategy's, so a positive on-chain observation there is a real fact, not coin→config inference — constraint 5 is preserved. When a real position is found, it's recovered and closed via `applyHyperliquidCircuitCloseFill` (the same defensive no-persisted-position booking primitive the CB/kill-switch paths already use) and the primary is left untouched; falls through to the original fail-closed primary unwind only when no on-chain hedge position exists. Alternatives rejected: (a) accept the window as documented phase-1 risk — rejected, this is a money-adjacent auto-protective gap the codebase's own "detect the specific gap, don't guess, but do act on positively-observed on-chain state" convention (e.g. `hl_reconcile_gap_alerts.go`) already has a template for; (b) durable pre-order intent stamping as the recovery key — considered but not needed, since collision rejection alone gives unambiguous ownership without it. Override if you'd prefer accepting the smaller residual risk instead.

Tests: 7 new regression tests (`hedge_test.go`) cover the state-level halves of both Needs Fixing items and the new sweep recovery detector — `applyHedgeOpenFill` opposite-side refusal + same-side blend + fresh-open, `mirrorHedgeReduce`'s paper/no-op success contract, and `onChainCoinQty`. The live/paper mode fix itself (a one-line change reusing the already-tested `hyperliquidIsLive` helper) and the live-order halves of the flip fix and orphan recovery aren't independently unit-tested — they call `RunHyperliquidExecute`/`RunHyperliquidClose` (real subprocess spawns), and this codebase's existing pattern for testing that layer is dependency-injected closers (e.g. `HyperliquidLiveCloser` in the kill-switch path), which `hedge_dispatch.go`/`hedge_sweep.go` don't yet have. Flagging rather than fabricating coverage that doesn't exist.

`go build`/`go vet`/`gofmt -l` clean; `go test ./...` full suite green.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11 22:36) in e81f2e8.

### Fixed
1. **Mark-driven oversized-hedge reduce ratchets the hedge smaller on routine relative price movement, never re-grows it** — confirmed: `targetHedgeQty` was computed from `prices[primaryCoin]`/`prices[hedgeCoin]` (`hedge_sweep.go:102-103` pre-fix), so a >2% relative move with no quantity change on either leg fired a permanent reduce. Fixed by computing the target from `primary.AvgCost`/`hedge.AvgCost` instead — a trade-derived value immune to live mark noise, reproducing the exact original sizing at rest and only diverging on an actual out-of-band quantity change (`hedge_sweep.go` both-open branch). `prices` param dropped from `snapshotHedgeCoherenceJobs` (now unused).

### Resolved judgment calls (implemented the optional improvement)
1. **One-cycle HL mark-fetch gap unwinds a freshly-opened primary** — confirmed: `mirrorHedgeOpen` sized from `prices[hedgeCoin(sc)]` with no fallback, unlike every other mark consumer (which degrades to `AvgCost` + WARN). Implemented `resolveHedgeMid` (`hedge.go`): a one-shot inline re-fetch of just the hedge coin's mid before falling through to the fail-closed unwind, wired into both `mirrorHedgeOpen` and `mirrorHedgeAdd`. There's no "next cycle" retry available for a fresh open (the position is gone once the unwind fires), so an inline retry within the same event is the correct mechanism, not a deferred one.

Tests: 4 new (`TestSnapshotHedgeCoherenceJobs_PriceMovementAloneNoJob` pins the exact regression scenario from the finding; `TestResolveHedgeMid_FastPathNoRetry`; the oversized/consistent-pair tests updated to the AvgCost basis). `go build`/`vet`/`gofmt -l` clean; `go test ./...` full suite green.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11 22:55) in 2ec1b74.

### Fixed
1. **Scale-in-add unwind books a FULL primary close in state while only reversing the add on-chain, stranding the pre-existing primary untracked and eventually unhedged** — confirmed: `unwindPrimaryAfterHedgeOpenFailure`'s on-chain unwind is correctly sized to `primaryQty` (`hedge_dispatch.go:225` post-fix numbering), but the old state booking unconditionally called `bookPerpsCloseWithFillFee` (full close, deletes the position) regardless of whether `primaryQty` represented the whole position or (via `mirrorHedgeAdd`) just the add delta on top of a pre-existing, already-blended primary quantity. Confirmed the blend timing too: `executeHyperliquidScaleInDeferredOpen` (`main.go:2460`) runs before the mirror, so state already shows `preQty+addQty` when the unwind books. Net effect on failure: on-chain ends at `preQty` (add reversed) but state was deleted entirely, and the surviving hedge Position (now H-without-P) would get closed by the next sweep cycle, unhedging the still-open, now-untracked `preQty` primary. Fixed with `unwindShouldBookPartial` (`hedge_dispatch.go`): routes to `bookPerpsPartialCloseWithFillFee` when the unwound quantity is less than the current state quantity (the add case), reserving the full-close path for when they match (fresh-open/flip, where the unwound quantity IS the whole position).

Tests: `TestUnwindShouldBookPartial` pins all three branch cases (add-partial, fresh-open-full, and the defensive over-unwind case); `TestUnwindAfterScaleInAddFailure_BooksPartialNotFull` is an end-to-end state-level regression proving the pre-existing primary+hedge pair survives an add-failure unwind at the correct residual quantity, using the exact booking primitives and branch selection `unwindPrimaryAfterHedgeOpenFailure` now uses. `go build`/`vet`/`gofmt -l` clean; `go test ./...` full suite green.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-11 23:27) in 29243bf.

### Fixed
1. **Hedge-leg closes feed `RecordTradeResult`, defeating the consecutive-loss circuit breaker** — confirmed: `bookPerpsCloseWithFillFee`/`bookPerpsPartialCloseWithFillFee` (`portfolio.go:410,498` pre-fix), `forceCloseAllPositions` (`risk.go:1291`), and `applyHyperliquidCircuitCloseFill` (`hyperliquid_balance.go:2762`) all called `RecordTradeResult` unconditionally, with no `pos.IsHedge` guard, on the single `RiskState` shared by the primary and its hedge. Added `RecordHedgeTradeResult` (`risk.go`): credits `DailyPnL` as before (real money, belongs in the portfolio-wide daily-loss basis) but never touches `ConsecutiveLosses`. Routed every hedge-leg close through it at all four sites.
2. **CB force-close and kill-switch/CB-drain primitives build hedge-leg close Trades without `IsHedge`** — confirmed: `forceCloseAllPositions` and `applyHyperliquidCircuitCloseFill` didn't stamp `IsHedge` on the Trade structs they construct, unlike `bookPerps{,Partial}CloseWithFillFee`, so a hedge closed via CB/kill-switch was counted in `LifetimeTradeStats` `#T`/W-L. Both now set `IsHedge: pos.IsHedge`. `applyHyperliquidCircuitCloseFill` also gained an explicit `isHedge` parameter for its no-virtual-position branch (no `Position` to derive hedge-ness from there — needed by the round-3 crash-orphan recovery path, which calls it with no persisted position by construction); the has-position branch ORs the caller's value with `pos.IsHedge`, so under-reporting callers still get the correct answer. Every pre-existing call site (5 production, 6 test) updated to pass `false`, preserving exact prior behavior — confirmed by the full existing test suite passing unchanged.

Tests: 9 new — `RecordHedgeTradeResult` direct, the two exact "must survive" ordering scenarios from the finding (losing-primary-then-winning-hedge, winning-primary-then-losing-hedge) at the `bookPerpsCloseWithFillFee` level, the partial-close path, `forceCloseAllPositions` tagging + streak exclusion, and `applyHyperliquidCircuitCloseFill`'s `isHedge` param on both branches (no-position/crash-recovery, has-position/OR-derivation, and a non-hedge sanity check that the streak still moves normally). `go build`/`vet`/`gofmt -l` clean; `go test ./...` full suite green.

---

richkuo:
@claude review

richkuo:

PR #1360, opened 2026-07-15
feat(#1159): per-strategy correlated hedge legs (phase 1, HL perps) [Opus 4.8, high]

richkuo:
Closes #1159

Opt-in per-strategy **correlated hedge leg** (phase 1, Hyperliquid perps, live + paper). When a strategy opens (e.g. long ETH), the scheduler simultaneously opens an inverse position on a different, collision-validated coin (e.g. short BTC), reducing market beta while preserving the strategy thesis.

## Design — one state-derived reconciler

`runHedgeSync` (`scheduler/hedge.go`) runs once per cycle at the perps dispatch tail (trade path + `Signal==0` manage path) and converges the hedge to a target computed by the pure `hedgeTargetDecision` from the primary position vs. a persisted quantity watermark (`Position.HedgePrimaryQtyBasis`). **Qty-event mirroring, not price mirroring** — mark drift never re-trades the hedge. Because it is state-derived, every primary event (fresh open, scale-in add, partial/full close, reconcile-booked SL/TP fill, kill-switch/CB force-close, external close, crash-between-legs restart) produces the matching hedge action automatically, without touching each close path.

## Safety invariants

1. **Fail-closed open:** a confirmed primary fill + failed hedge open on a fresh-open cycle immediately reduce-only unwinds the primary fill (cancelling its SL/TP OIDs) + CRITICAL DM; later drift cycles alert + retry. Fill-confirmed state mutation only — no fill means no state change, so the sync itself is the retry.
2. **Sole ownership by construction:** `validateHedgeConfigs` rejects a hedge coin equal to any strategy's configured coin, its own coin, or another hedger's hedge coin, so every shared-coin mechanism keyed on `hyperliquidConfiguredCoin` stays untouched.
3. **Attribution:** hedge PnL/fees book into the owning strategy's ledger (`tradeLedgerDeltaSQL` ignores `trade_type`) but are tagged `trade_type="hedge"` (excluded from lifetime `#T`/W-L) and route through `RecordHedgeTradeResult`, never `RecordTradeResult` — a hedge loss must not extend the circuit-breaker loss streak. Diagnostics skipped for hedge legs.

## Coverage

Config + validation (collision matrix, inverse-only, `direction:"both"` rejected, nested unknown-key guard) · persisted `Position` fields + DB migration/round-trip · marks (`collectPerpsMarkSymbols`) · reconcile hedge-coin pass (external-close booking, foreign non-adoption, orphan-close skip) · fill resolver · shared-wallet attribution · kill switch (roster gated on a held leg + fill booking, no double-book) · CB pending (sole-owned hedge enqueued even when the primary coin is shared) · manual force-close note · hot-reload blocked while open + restart-bypass startup check (`validateHedgeStateConsistency`) · operator surfaces (inspect/startup summary) · backtester loud reject.

## Verification

- `go test ./...` — pass (full scheduler suite)
- backtest pytest suite — 1368 passed, 1 skipped
- `gofmt` clean; `py_compile` clean
- New tests: `hedge_test.go` (decision core, sizing, apply, DB round-trip, `RecordHedgeTradeResult` streak isolation), `hedge_config_test.go` (collision/vocabulary/direction-both/nested-key), `backtest/tests/test_backtester_hedge_reject.py`

## Known follow-ups (documented, out of scope)

- Stuck-CB reconstruction does not rebuild the hedge symbol from persisted metadata (rare double-fault: pending never set **and** CB latched).
- Backtester PnL/fee/slippage parity for hedge legs (phase 1 rejects loudly).

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-15T01:49:30Z) in 84f3c75.

### Fixed
1. **Live partial hedge fills book the requested size, not the actual filled size** — `applyHedgeFillLocked` now takes `filledQty` and threads `execFill.TotalSz` (live) / requested size (paper) into `applyHedgeOpenLocked`/`applyHedgeAddLocked`/`applyHedgeReduceLocked`; each books the actual filled qty and derives the `HedgePrimaryQtyBasis` watermark via `hedgeAppliedBasis(basisBefore, target, action.RequestedQty, filledQty)` (fill-fraction interpolation), matching the primary path (`scheduler/hedge.go:334,426,453,514,570`). Reduce now interpolates the watermark instead of jumping to `PrimaryQtyTarget`, and books only the filled reduce size so virtual qty stays ≥ on-chain remaining. `closeFull` stays a full virtual flatten by construction (full `market_close`/`AlreadyFlat`). New regression tests `TestApplyHedgeOpen_PartialFillBooksActualSize` (60% fill → qty 0.06, basis 1.2) and `TestApplyHedgeReduce_PartialFillInterpolatesBasis` (0.02 of 0.05 filled → remaining 0.08, basis 1.6).

### Resolved judgment calls / optional improvements
1. **A transient missing hedge mark on a fresh-open cycle unwinds the just-opened primary** — a `FailClosed` decision (couldn't SIZE the hedge because `prices[hedgeCoin]` was absent) is now treated as a retryable data gap: it defers and retries next cycle, and never unwinds the primary, in both live and paper (`scheduler/hedge.go:282`). The genuine fail-closed unwind is now reached only after an order was actually attempted and rejected (the `!ok` branch after `runHedgeOrder`) — distinguishing a retryable mark gap from a real order failure, per the invariant. Tradeoff: a persistent (non-transient) missing mark leaves the primary hedged-pending for that window with an escalating WARN rather than churning open+close fees / phantom paper round-trips on every hiccup; the mark comes from the reliable per-cycle `collectPerpsMarkSymbols` fetch, so a persistent gap is an ops signal, not a normal state. Override if you'd prefer a strict immediate unwind on any sizing failure.
2. **Kill-switch hedge fill skipped when the primary coin produced no on-chain fill** (safety carve-out) — `applyHyperliquidKillSwitchCloseFill` no longer early-returns on an absent/empty primary fill; it books the primary fill if present and the held hedge leg's fill independently (`scheduler/hyperliquid_balance.go:2204`). A primary-externally-flat kill switch now attributes the hedge's real on-chain fill at fill price/fee (sole-owner full share) instead of losing it to the mark-priced `forceCloseAllPositions` cleanup. The #954 dup-OID guard still prevents double-booking against that pass.

`go test ./...` passes; `gofmt` clean.

---

richkuo:
@claude review

richkuo:

PR #1361, opened 2026-07-15
feat(#1159): per-strategy correlated hedge legs (phase 1, HL perps) [Opus 4.8, xhigh]

richkuo:
Closes #1159

## Summary

Adds an opt-in per-strategy `hedge` config block (phase 1, Hyperliquid perps only) that auto-manages a correlated inverse position on a **distinct** coin, mirroring the primary strategy's full lifecycle to reduce market beta while preserving the thesis. Example: a long-ETH strategy simultaneously runs a short-BTC hedge.

## Architecture

Hedge management is a **per-cycle, state-derived reconciler** (`runHedgeSync`) — the single mirror choke point, modeled on `runHyperliquidProtectionSync`'s lock discipline — not scattered per-event hooks. A pure decision core (`hedgeTargetDecision`) computes the hedge target from the *current* primary position vs. a persisted quantity watermark (`Position.HedgePrimaryQtyBasis`), so every primary event — fresh open, scale-in add, partial/full close, an on-chain SL/TP fill detected by reconcile, kill-switch, circuit-breaker — automatically produces the matching hedge action within the same or next cycle.

**Key invariants**
- **Qty-event mirroring, not price mirroring** — mark drift never re-trades the hedge; only primary quantity/side changes do.
- **Fill-confirmed state mutation only** — hedge virtual state mutates solely from a confirmed fill (mirrors `runHyperliquidExecuteOrder`'s `ok=false → no state update`).
- **Fail-closed open** — a confirmed *fresh-open* cycle whose hedge open fails immediately unwinds the primary fill reduce-only (cancelling its just-armed SL/TP OIDs) + CRITICAL owner DM; other-cycle failures alert and the state-derived sync retries. Never runs unhedged silently.
- **Sole ownership by construction** — collision validation guarantees a hedge coin is never any strategy's configured coin, its own primary coin, or another hedger's coin, so every shared-coin mechanism stays untouched.

## What changed

- **Config + validation** — `HedgeConfig` + accessors; `hyperliquidHedgeConfigErrors` enforces the collision matrix, vocabulary, and `direction="both"` reject; nested unknown-key guard (a typo'd leaf key like `ration` is rejected, not silently defaulted).
- **State/persistence** — two persisted `Position` fields (`HedgeFor`, `HedgePrimaryQtyBasis`); `trade_type="hedge"` (excluded from `#T`/W-L and `trade_diagnostics`); hedge legs book via `RecordHedgeTradeResult` — **never** the loss streak (a hedge loses by construction when the primary wins).
- **Execution/booking** (`hedge.go`) — decision core, open/add/reduce/close orchestration, fill-proportional watermark advance, dust deferral, primary unwind on fail-closed.
- **Valuation/wallet/reconcile** — marks include hedge coins; shared-wallet attribution, the reconcile fee resolver, and a dedicated hedge-coin reconcile pass cover the hedge leg (no phantom orphan-drift, real-fee external closes).
- **Kill switch / circuit breaker / manual force-close** — flatten the *held* hedge coin without double-booking PnL/fees; the shared-coin CB drain guard now strips only the primary symbol so a sole-owned hedge still drains when the primary coin is shared.
- **Hot reload** — blocked while a hedge leg is open (`hedgeConfigEqual`), free while flat; `strategyRestartShape` masks `Hedge`; reload re-runs the collision matrix.
- **Operator surfaces** — inspect text + `--json`, `/status`, Discord `hedgeStatusNote`; startup consistency check for config-edit-plus-restart drift.
- **Backtester** — loud-rejects an *enabled* hedge block (single-instrument model would drop the hedge's PnL/fees); `enabled: false` passes through.
- **Docs** — CLAUDE.md guardrail + full ARCHITECTURE.md mechanics section.

## Verification

- `go build ./scheduler` ✓, `go test ./...` ✓ (full suite), `gofmt` clean, `go vet` clean.
- `uv run --no-sync python -m pytest -n auto` ✓ (full suite), `py_compile backtest/run_backtest.py` ✓.
- End-to-end config smoke: a hedge-vs-configured-coin collision is rejected at load; a valid hedge renders in `inspect` text/JSON.
- New tests cover acceptance-6 cases: decision core (sizing, inverse side, proportional reduce, add deltas, dust deferral, fail-closed price), collision matrix + vocabulary + nested-key guard, booking/persistence/risk-routing, diagnostics skip, hot-reload block-while-open + allowed-flat, DB round-trip, startup consistency, and the backtest reject.

## Notes / follow-ups (out of scope, per the issue)

- Backtest PnL/fee/slippage parity for hedge legs (deferred).
- Allowing hedge coins that overlap peer strategies (would extend the shared-coin coin-membership derivation).

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-15T01:49:30Z) in 900a2a4.

### Fixed

1. **CB flattens a shared-primary strategy's hedge, then the sync re-opens it (churn + unhedged window)** — Confirmed. The hedge is now enqueued for CB close **only alongside a closeable (sole-owned) primary**; a shared primary enqueues nothing, so the hedge and primary always share the same fate and the latched-CB manage-only `runHedgeSync` never re-opens a just-flattened hedge. Applied at all three sites the review named: `setHyperliquidCircuitBreakerPending` restores the shared-primary early-return (`scheduler/risk.go:953`); the stuck-CB reconstruction gains the same shared-primary gate (`scheduler/hyperliquid_balance.go:2350`); and the drain's shared-coin guard reverts to clearing the whole pending (`scheduler/hyperliquid_balance.go:2447`) — with the enqueue rule fixed, a pending never legitimately pairs a hedge with a shared primary, and a peer added mid-drain correctly leaves both legs open.
   - Verifies all three Must-survive cases via new tests: shared-primary hedger enqueues nothing; sole-owned-primary hedger enqueues both legs; the stuck-CB reconstruction path uses the same decision (`scheduler/hedge_cb_test.go`).

2. **Startup drift check ignored ratio/side/margin/leverage from a config-edit-plus-restart** (Recommended Optional) — Confirmed. Froze the resolved hedge sizing config at open in two new persisted `Position` fields, `HedgeRatioAtOpen` + `HedgeMarginModeAtOpen` (`Leverage` was already stamped; `side` is fixed to `inverse` in phase 1 so it's implicitly frozen), following the `ATRMethodAtOpen` (#1277) precedent. `validateHedgeStateConsistency` now compares all three and warns on any drift, noting the open leg keeps its original sizing while the change applies to the next add (ratio) / re-open (margin/leverage) (`scheduler/hedge.go:645`). A dimension whose stamp is missing (pre-column leg) is skipped to avoid false positives. Covered by drift-detection + no-false-positive + DB round-trip tests.

Verification: `go build ./scheduler` ✓, `go test ./...` ✓ (full suite), `gofmt`/`go vet` clean.

---

richkuo:
@claude review

richkuo:

PR #1377, opened 2026-07-16
feat(#1159): per-strategy correlated hedge legs (phase 1, HL perps) [Sonnet 5, high, fableplan]

richkuo:
## Summary

Implements per-strategy correlated hedge legs (#1159 phase 1: Hyperliquid perps only) per the [Fable 5 plan posted on the issue](https://github.com/richkuo/go-trader/issues/1159#issuecomment-4991828694), built independently in its own worktree via `fableplan-work-on-issue` (multiple other PRs already exist for this issue from other agent runs — this one was built without looking at them, per the requester's explicit instruction).

- **Config** (`scheduler/hedge.go`): `HedgeConfig` (enabled/symbol/side/ratio/platform/type/margin_mode/leverage), per-strategy validation, and a whole-config collision pass (`hyperliquidHedgeStrategyErrors`) rejecting a hedge coin that equals the strategy's own primary coin, any peer's primary coin, or another hedge-enabled strategy's hedge coin.
- **Data model**: `Position.IsHedge`/`HedgeForPositionID`/`HedgeRatioQty`, `Trade.IsHedge`, threaded through the SQLite schema/migrations and `SaveState`/`LoadState`.
- **Dispatch mirror** (`main.go`): one hook after the primary HL perps open/scale-in/close booking diffs the pre/post primary position to mirror opens, adds, partial/full closes, and flips from ACTUAL fills. Constraint 4 (fail closed): a fresh-open hedge failure unwinds the primary reduce-only and alerts; a scale-in-add hedge failure unwinds only the added delta.
- **Coherence-sync backstop** (`runHedgeCoherenceSync`, every cycle): re-derives each hedge's target from the current primary qty and reduce-only corrects any over-hedge — self-healing drift from CB drains, kill-switch closes, manual force-close, on-chain TP/SL fills, or a restart mid-sequence. Never increases a hedge or guesses a fill; under-hedge is alert-only. Also serves as the hedge's startup reconcile (ownership read only from persisted state, never inferred from the configured symbol).
- **Kill switch**: hedge coins are added to the live close pass's traded-coin set so a hedge position is flattened on-chain in the SAME emergency pass as the primary, with its real fill booked.
- **Hot-reload**: the hedge block is blocked from changing while any position on the strategy is open, freely reloadable while flat; reload re-runs the collision check.
- **Backtester**: loud reject on `hedge.enabled=true` — no parity model for hedge PnL/fees this release.
- **Marks**: hedge coins are included in the mark-price fetch set so PortfolioValue/drawdown and the dashboard price the hedge leg correctly.

### Deliberate scope reduction from the plan (named per instructions)

Circuit-breaker drain and manual force-close do **not** get dedicated inline hedge hooks (the plan's steps 6/8). Both already reduce the primary same-cycle; the coherence-sync backstop closes/reduces the resulting over-hedged leg on the **next** cycle instead of the same one. This trades a bounded one-cycle unwind lag for far fewer touch points in already-complex, money-critical existing CB/manual-close code — a safety-first tradeoff given the scope of a single session. Also deferred as follow-ups: dedicated `inspect`/status/Discord hedge display beyond the automatic JSON field exposure, shared-wallet funding attribution for the hedge coin, and `classifyPositionTradeType` hedge close-reason labels — cosmetic/operator-surface gaps, not correctness gaps.

## Test plan

- [x] `go build` + `go vet` + `gofmt` clean
- [x] `go test ./...` — full suite passes, no regressions
- [x] New `scheduler/hedge_test.go`: side mapping, notional sizing (fail-closed on non-positive inputs), coherence-sync target/tolerance math (never grows a hedge), full validation rule set, all three collision rules (+ deterministic ordering), open/add booking (blended avg cost, ratio re-stamp), partial/full close booking (PnL math, `IsHedge` stamp), kill-switch coin set + fill booking (+ duplicate-OID rejection)
- [x] New hot-reload test (`config_reload_test.go`): hedge block changes rejected while open, accepted while flat, no-op when unchanged
- [x] New DB round-trip test (`db_test.go`): hedge Position/Trade fields survive `SaveState`→`LoadState`
- [x] `uv run python -m py_compile backtest/run_backtest.py`
- [x] New `backtest/tests/test_backtester_hedge_reject.py`: loader rejects `hedge.enabled=true`, accepts `enabled=false`/no block
- [x] `uv run python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2853 passed, 1 pre-existing skip, no regressions
- [x] Binary smoke test: `--once` against `scheduler/config.example.json`, clean full cycle, no crashes

Closes #1159

---
Created with LLM: Sonnet 5 | high | Harness: fableplan-work-on-issue


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T13:12:02Z) in ea7bcef.

### Fixed

1. **Fresh-open hedge-failure unwind could flatten a peer strategy's share of a shared primary coin.** `failHedgeOpenClosed` now always unwinds by the exact filled delta (`&deltaQty`), never a nil-sized `market_close` — `hedge.go:619-630`. Booking unified onto `bookPerpsPartialCloseWithFillFee`, which still fully removes a sole-owned position when the delta equals the whole quantity, so non-shared-coin behavior is unchanged.
2. **Scale-in add hedge was sized off the blended `AvgCost` instead of the add's own fill price.** `syncHedgeAfterPrimaryFill` now threads `execResult.Execution.Fill.AvgPx` through as `primaryFillPx` for all three sizing call sites (open/add/flip) instead of `Position.AvgCost` — `hedge.go:515-573`, `main.go:2459-2466`. Fresh-open/flip were only coincidentally correct before (nothing to blend with); adds now size on the true incremental notional regardless of price drift since entry.

### Fixed (Recommended Optional)

1. **Under-hedge coherence-sync alert had no debounce.** Now routed through the existing `LiveExecFailureThrottle` (1st/10th/hourly cadence, same mechanism live-exec failures already use), plus an explicit "recovered" notice when the condition clears — `hedge.go:781-822`, new `Had()` method on the throttle in `failure_alerts.go`.
2. **`Trade.IsHedge` was persisted but never consumed, inflating #T/W-L.** `LifetimeTradeStatsAll` and `LifetimeTradeStatsForStrategy` now exclude `is_hedge=1` rows from both open-count and win/loss queries — `db.go`. PnL/cash sums are untouched (hedge PnL still books to the strategy). New regression test: `TestLifetimeTradeStatsAll_ExcludesHedgeLegs`.

Verification: `go build`, `go vet`, `gofmt`, `go test ./...` (full suite, no regressions, new tests pass).

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T13:33:00Z) in 12b1ea4.

### Fixed

1. **A hedge flip (or flat-then-reverse re-open) could blend an opposite-side fill into a residual hedge position, corrupting Side/Quantity/AvgCost and inverting downstream PnL.** `applyHedgeOpenToState` now detects a fill whose side differs from an existing same-symbol hedge position and nets it like an exchange would — closes up to the residual's quantity first (booking real realized PnL via `bookHedgeCloseFill`, fee pro-rated), then opens any excess on the new side — `hedge.go:313-343`. Enforced inside the booking function itself, so it holds for every caller, not just the flip path. New tests cover full-net-with-excess, partial-net-no-flip, and same-side-add (no regression).

### Fixed (Recommended Optional)

1. **A hedge-enabled paper strategy silently traded unhedged with no warning.** `validateHedgeConfig` now rejects `hedge.enabled=true` on a non-live HL perps strategy — `hedge.go:139-148` — since `syncHedgeAfterPrimaryFill`/`runHedgeCoherenceSync` both early-return on `!hyperliquidIsLive`, making a paper hedge permanently inert. Rejecting at load also means a SIGHUP reload can't land a config back into that silently-inert state (`LoadConfig` re-validates every reload).

Verification: `go build`, `go vet`, `gofmt`, `go test ./...` (full suite, no regressions, new tests pass).

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T13:50:01Z) in 22d5d96.

### Fixed

1. **The hedge open fee was recorded on the trade but never debited from cash, overstating cash and breaking on-chain drift reconciliation.** `applyHedgeOpenToState` now does `s.Cash -= fillFee` on every open/add, including the flip-net excess-open leg (`hedge.go:376-383`), matching how every primary perps open debits its fee (`portfolio.go:1352`,`:1554`). New tests cover fresh open, scale-in add, and the flip-net excess-open case (fee correctly split between the netted-close portion and the fresh-open portion).

2. **Hedge close PnL fed into the loss-streak circuit breaker via `RecordTradeResult`, masking it on any hedged strategy.** `bookHedgeCloseFill` now calls a new `recordHedgeDailyPnLOnly` (`hedge.go:427-440`) that still updates `RiskState.DailyPnL` (real cash/daily-loss-limit impact preserved) but never touches `ConsecutiveLosses` — the same "hedge is not alpha" reasoning that already excludes `IsHedge` legs from lifetime W/L stats. New test asserts a winning hedge close (i.e. a losing primary) no longer resets `ConsecutiveLosses`, and a losing hedge close still updates `DailyPnL`/cash correctly.

3. **A flip whose old-hedge close under-filled/failed could leave a same-direction leg mislabeled as a hedge, undetected by the side-blind coherence sync.** `runHedgeCoherenceSync` now checks side inversion first via a new pure helper `hedgeLegSideMismatched(hedgeSide, expectedSide)` (`hedge.go:290-298`, wired at `hedge.go:817-820`) — a mismatched leg is flattened immediately (`hedge_side_mismatch_flatten`) with a CRITICAL alert, rather than being trimmed toward a bogus quantity target. New unit tests cover the pure helper directly (inverse/mismatched/no-resolvable-primary cases).

### Fixed (Recommended Optional)

1. **The hedge subsystem had no `docs/ARCHITECTURE.md` mechanics/ownership entry.** Added a full `hedge.go` entry covering mirroring, the coherence-sync-backstop-not-inline-hooks design (explicitly recorded so it isn't "fixed" by duplicating the backstop later), the reduce-only never-grow invariant, the new side invariant, the netting invariant, constraint-4 fail-closed unwind, all three collision rules, and money accounting.

Verification: `go build`, `go vet`, `gofmt`, `go test ./...` (full suite, no regressions, new tests pass).

---

richkuo:
@claude review

richkuo:

Issue #496, opened 2026-04-30
Support position-aware close evaluators (avg_cost, qty, entry_atr)


richkuo:
## Problem

Close strategies in the open/close composition system (#480/#483) can currently only see OHLCV data — they're invoked via `apply_strategy(name, df, params)` in `shared_tools/strategy_composition.py` and return a `close_fraction` derived from candle features alone.

This rules out an entire useful class of exit strategies that need **position context** to make decisions:

- Tiered take-profit (close 50% at +1.5×ATR, remainder at +3×ATR — needs `avg_cost`, `entry_atr`, `current_quantity / initial_quantity` to track tier progress)
- Trailing stops based on highest mark since entry
- Time-in-trade exits ("close after N hours regardless of signal")
- Profit-locking exits ("close if unrealized pnl > X% of margin")
- R-multiple exits ("close at +2R from initial stop distance")

All of these need at minimum: `avg_cost`, `current_quantity`, `initial_quantity`, `entry_atr`, `side`, `mark_price`. Today only `position_side` is threaded through (`scheduler/strategy_composition.go:51` `appendOpenCloseArgs`).

## Proposed change

Extend the open/close composition plumbing so close evaluators can optionally receive position context. The OHLCV `fn(df, **params)` interface stays unchanged for existing strategies — position fields are passed as additional kwargs that position-unaware strategies ignore.

### 1. Go: thread position data through `appendOpenCloseArgs`

`scheduler/strategy_composition.go` — extend signature to accept a position-context struct:

```go
type PositionCtx struct {
    Side            string
    AvgCost         float64
    Quantity        float64  // current
    InitialQuantity float64
    EntryATR        float64
}

func appendOpenCloseArgs(args []string, sc StrategyConfig, pos PositionCtx) []string
```

When `usesOpenCloseConfig(sc)` and the relevant fields are non-zero, append:

- `--position-avg-cost=<f>`
- `--position-qty=<f>`
- `--position-initial-qty=<f>`
- `--position-entry-atr=<f>`

(plus existing `--position-side`).

### 2. Go: capture position context at Phase 1 RLock

`scheduler/main.go` — Phase 1 already captures `hlPosSide`, `hlPosQty`, `hlAvgCost` for HL (similarly for OKX/RH/TS). Add `initialQty` and `entryATR` to the snapshot, then thread them into the existing `runHyperliquidCheck` / `runOKXCheck` / `runRobinhoodCheck` / `runTopStepCheck` / spot equivalents.

[...]

PR #1379, opened 2026-07-16
feat(#1159): add per-strategy correlated hedge legs [Sonnet 5, xhigh, fableplan]

richkuo:
## Summary

Implements phase 1 of #1159: an opt-in per-strategy correlated hedge leg for Hyperliquid perps strategies. When a hedge-enabled strategy opens/scales/closes its primary position, the scheduler mirrors an inverse position on a different, sole-owned coin, sized on notional exposure x a configurable ratio. The hedge carries no independent SL/TP and no check script — it is strictly slaved to the primary's lifecycle.

Built independently in a fresh worktree from a Fable 5 plan (posted at https://github.com/richkuo/go-trader/issues/1159#issuecomment-4992214573), per explicit instruction to work this issue independently of the many other open PRs attempting it.

## What changed

- **`scheduler/config.go`**: `HedgeConfig` schema, `hyperliquidHedgeConfigErrors` — rejects a hedge coin equal to the strategy's own primary coin, any other configured strategy's coin, or another strategy's hedge coin. Hedge coins are always sole-owned by construction, so every existing shared-coin mechanism (peer detection, margin compatibility, CB drain, kill-switch share, reconcile owner mapping) stays correct without needing to know hedge legs exist.
- **`scheduler/hedge.go` (new)**: the hedge engine — pre-open gate (mirrors `PerpsOrderSkipReason`'s skip-before-submit discipline), open/scale-in/close mirrors, fail-closed unwind on a hedge-open failure (primary fill confirmed but hedge can't open → reduce-only close the primary immediately, CRITICAL alert — never runs unhedged silently), a per-cycle reduce-only coherence safety net, and the reconcile lane for externally-closed/reduced hedge legs.
- **`scheduler/main.go`**: wires the gate/mirrors/coherence into the existing HL perps dispatch.
- **`scheduler/hyperliquid_balance.go` / `hyperliquid_fills.go` / `risk.go`**: circuit-breaker drain, kill-switch `tradedCoins`/fill-share, and mark-price collection all extended to cover the hedge coin.
- **`scheduler/manual_core.go`**: rejects an operator force-close or SL edit addressed directly at a hedge symbol — "hedge legs are scheduler-managed; act on the primary."
- **`scheduler/config_reload.go`**: hot-reload of the `hedge` block is blocked while the primary or the hedge is open (same treatment as scale-in / stop-owner fields).
- **`scheduler/config_migration.go`**: config_version bumped to 18 (purely additive, notice-only migration).
- **`scheduler/correlation.go`**: hedge exposure now buckets under its own asset instead of being silently skipped, so it shows up in correlation warnings and the exposure cap.
- **`scheduler/db.go`**: `is_hedge`/`hedge_for`/`hedge_symbol` persistence — restart/reconcile recovers ownership from these columns alone, never from coin-to-symbol inference.
- **`backtest/run_backtest.py`**: loud reject for `hedge.enabled=true` configs (no PnL/fee/slippage model yet).
- **`scheduler/inspect_cmd.go`**: hedge config + live status surfaced in `inspect` (text and JSON).

## Deviation from the posted plan

The Fable 5 plan's coherence-sync design compared `Quantity/InitialQuantity` coverage fractions between the primary and hedge. I found this breaks after `unwindPrimaryAddAfterHedgeAddFailure` reduces a botched scale-in add back off: `InitialQuantity` is intentionally monotonic (#496, shared with the TP-tier fraction-of-initial system), so a coverage ratio would permanently read the primary as "under 100%" even though it's economically back to its pre-add size — spuriously shrinking an already-correctly-sized hedge. I replaced it with a notional-ratio target (`hedgeExpectedQty`/`primaryQtyForHedgeQty`: primary notional x ratio / hedge mark), which has no such history-dependent drift and is closer to the issue's own "hedge sizing should be based on notional exposure" language. The fail-closed intent (reduce-only, never add exposure, CRITICAL-alert the under-hedged direction) is unchanged.

## Scope boundaries (named, not silent gaps)

- Operator-surface work is JSON/inspect-only; no dashboard HTML or Discord trade-alert tagging beyond what the generic `Position` JSON serialization already exposes.
- A foreign on-chain position appearing on a hedge coin outside an open attempt has no dedicated gap-alert channel beyond the pre-open gate (refuses new opens) and the reconciler (never adopts it).

## Test plan

- [x] `go build .` / `go vet ./...` / `gofmt -l` clean from `scheduler/`
- [x] `go test ./...` — full existing suite passes unchanged, plus new `scheduler/hedge_test.go` (sizing/gate/validation/apply/reconcile-lane math, incl. the coverage-fraction-vs-notional-ratio edge case)
- [x] `uv run --no-sync python -m pytest backtest/` — 1451 passed, 1 pre-existing skip, incl. new `backtest/tests/test_backtest_hedge_reject.py`
- [x] Binary smoke test: a hedge-enabled config loads/validates correctly through the real `go-trader` binary; a colliding hedge config (`hedge.symbol` == primary coin) is rejected with the expected error

Closes #1159

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] / Opus 4.8 · 2026-07-16T13:59:11Z) in cf2cfbe.

### Fixed

1. **Coherence net ratchets a winning primary on ordinary mark drift, not a real desync.** `scheduler/hedge.go:582` (as reviewed) recomputed `hedgeExpectedQty` from live primary/hedge marks every cycle; since the two coins never move identically, price drift alone crossed the 2% tolerance and reduce-only trimmed a winning primary. Replaced with `Position.HedgeQtyRatio` (`scheduler/portfolio.go`) — a fixed hedge-qty-per-primary-qty basis captured once at each confirmed open/add event (`applyHedgeOpen`/`applyHedgeScaleIn`, `scheduler/hedge.go:122,189`) and compared purely by quantity thereafter via the new pure `hedgeCoherenceDecision` (`scheduler/config.go:797`). A `ratio<=0` bootstraps from current quantities instead of misreading an absent basis as "expect zero hedge" (covers brand-new positions and rows persisted before this column). Persisted via a new `positions.hedge_qty_ratio` column (`scheduler/db.go`) so a restart can't reset it to zero and misfire.
2. **Fail-closed unwind retry never fires when the reduce-only close also fails.** `unwindPrimaryAfterHedgeOpenFailure` (`scheduler/hedge.go:349`) never stamped the primary's `HedgeSymbol`, so `syncHedgeCoherence`'s `primary.HedgeSymbol != ""`-gated retry branch could never pick it up — a primary left open after both a failed hedge open and a failed unwind close silently degraded to a permanently-unhedged position. The function now takes the intended hedge coin and stamps it onto the primary on the close-failure path (`scheduler/hedge.go:379`), so coherence retries the reduce-only close every cycle until it succeeds.

Both re-validated against current code before changing anything (traced the exact call sites and conditions cited in the review). New/updated tests: `TestHedgeCoherenceDecisionIgnoresMarkDrift`, `TestHedgeCoherenceDecisionBootstrap`, `TestHedgeCoherenceDecisionOverAndUnderHedged`, plus `HedgeQtyRatio` stamping assertions in `TestApplyHedgeOpen`/`TestApplyHedgeScaleIn` (`scheduler/hedge_test.go`). Full `go build`/`go vet`/`go test ./...` (gofmt clean) passes.

---

richkuo:
@claude review

richkuo:
Addressed round-2 review feedback (github-actions[bot] / Opus 4.8 · 2026-07-16T14:31:20Z) in 0b21f82.

### Fixed

1. **A refused hedge pre-open gate booked a phantom live primary position instead of skipping the open entirely.** `scheduler/main.go:2418-2432` (as reviewed) set `hedgeGateBlocked=true` but never `liveExecFailed`, so the booking guard further down still ran `executeHyperliquidResultDeferredOpen` with `execResult==nil` — read as the modeled/paper path (`fillQty==0`), booking a virtual position + fictional trade + modeled fee with nothing on-chain behind it. `hedgeGateBlocked=true` now also sets `liveExecFailed=true` so the booking path is skipped entirely when nothing was actually submitted. `syncHedgeCoherence` still runs unconditionally afterward — unaffected, nothing to reconcile since nothing opened.
2. **After an external hedge close/liquidation the reconciler cleared `HedgeSymbol`, permanently disabling coherence's "never run unhedged" primary close.** Both the external-close path (`scheduler/hedge.go`, was ~line 700) and the on-chain direction-flip path (was ~line 718) booked the lost hedge and then zeroed the primary's `HedgeSymbol`, directly contradicting the function's own docstring intent and `syncHedgeCoherence`'s `HedgeSymbol != ""` gate on its "hedge gone" reduce-only close. Stopped clearing `HedgeSymbol` in both paths — only the hedge row itself is removed/booked-closed; `HedgeSymbol` now only clears once coherence's own reduce-only close actually removes the primary. Kept the fix inside `syncHedgeCoherence`'s existing Phase-3 (no-lock) reduce-only+CRITICAL machinery rather than having the Lock-held reconciler drive a live close itself, which would violate the repo's subprocess-never-under-`mu.Lock()` invariant.

Both re-validated against current code before changing anything. Updated `TestReconcileHedgeLegsForStrategy`'s external-close and direction-flip subtests, which previously asserted `HedgeSymbol` was cleared — that assertion was itself encoding the bug. Full `go build`/`go vet`/`go test ./...` (gofmt clean) passes.

---

richkuo:
@claude review

richkuo:
Addressed round-3 review feedback (github-actions[bot] / Opus 4.8 · 2026-07-16T17:37:45Z) in dbc9975.

### Fixed

1. **The hedge scale-in-add unwind booked the primary reduce before the primary add, inverting the ledger order.** `unwindPrimaryAddAfterHedgeAddFailure` ran immediately on a hedge-add failure and reduced the primary against its PRE-add virtual quantity (`scheduler/main.go`, was line 2531) — before `executeHyperliquidScaleInDeferredOpen` booked the add. When the add size >= the pre-add quantity, `bookPerpsPartialCloseWithFillFee` clamped to a full close and deleted the primary, leaving virtual state flat while the chain still held the base position; even when it didn't clamp, PnL was realized against the wrong (pre-add) cost basis and fed the CB loss-streak tracker. Reordered: the add now always books first, and the unwind reduce runs immediately after that booking's `mu.Unlock()` (`scheduler/main.go`), so it's always valued against the post-add quantity.
2. **The coherence net's "hedging disabled while a hedge leg is open" freeze/CRITICAL backstop was unreachable.** `syncHedgeCoherence`'s own `!strategyHedgeEnabled(sc)` freeze branch was correct, but its sole caller in `main.go` gated the *call itself* on `strategyHedgeEnabled(sc)`, so the branch could never fire. `syncHedgeCoherence` is now called unconditionally; a new `hedgeCoinForStrategyOrOrphan` (`scheduler/hedge.go`) falls back to scanning state for an orphaned `IsHedge` row when config resolution comes up empty, used by both coherence's coin lookup and `reconcileHedgeLegsForStrategy`'s gate (which had the same problem).

### Resolved judgment calls

1. **Scope of finding 2's fix — kill-switch `tradedCoins` / CB drain left untouched.** The review named `forceCloseHyperliquidLive`'s `tradedCoins` set (`hyperliquid_balance.go:1970`) and the CB drain (`risk.go:966`) as "likewise gated," alongside coherence and the reconciler. I fixed coherence + the reconciler (which fully satisfy the review's stated invariant and all three must-survive cases) but deliberately left the kill-switch/CB-drain hedge-coin inclusion on the existing `strategyHedgeEnabled` gate. Reasoning: re-enabling hedging instantly restores all four paths at once since they share the identical gate — the coherence CRITICAL alert is the intended signal to do that or intervene manually; extending into `forceCloseHyperliquidLive` would need a new parameter threaded through `KillSwitchCloseInputs` and touching ~10 existing tests in one of the highest-blast-radius functions in the codebase, for a narrow edge case the review didn't test directly. Alternative considered and rejected: making that change anyway under time pressure — rejected as higher regression risk in emergency-safety code than the residual gap it closes. Override if you'd like this pursued as a follow-up.

Both findings re-validated against current code before changing anything. New tests: `TestHedgeCoinForStrategyOrOrphan` (enabled/disabled/orphan/wrong-owner/nil cases), a new `TestReconcileHedgeLegsForStrategy` subtest proving an orphaned hedge still reconciles while disabled. Finding 1's fix is a pure statement-reorder in the main.go HL-perps dispatch — no existing unit-test scaffold for that function (same as round 1/2's main.go-only fixes), verified via full build+test pass. `go build`/`go vet`/`go test ./...` (gofmt clean) all pass.

---

richkuo:
@claude review

richkuo:
Addressed round-4 review feedback (github-actions[bot] / Opus 4.8 · 2026-07-16T18:16:31Z) in 2c2bbd0.

### Fixed

1. **A hedge-enabled strategy whose primary coin is shared with another strategy never reconciled its hedge leg.** `reconcileHyperliquidAccountPositions`'s due-strategy loop (`scheduler/hyperliquid_balance.go`, was line 854) `continue`d past the `reconcileHedgeLegsForStrategy` call whenever the strategy's *primary* coin was shared — but `hyperliquidHedgeConfigErrors` only guarantees the *hedge* coin is sole-owned; the primary is free to be shared with a peer. The shared-coin lane below never touches hedge coins either, so an externally-closed/liquidated hedge went stale in virtual state forever and `syncHedgeCoherence` (virtual-quantity comparison only) never noticed the primary had gone unhedged. Restructured so only the primary-coin reconcile stays behind the shared-coin skip; `reconcileHedgeLegsForStrategy` now runs unconditionally per due strategy (no-op when hedging isn't enabled and no orphaned row exists).

Re-validated against current code before changing anything. New test `TestReconcileHedgeLegReconciledDespiteSharedPrimaryCoin` (`scheduler/hyperliquid_balance_test.go`) — proven red on the unfixed code (reverted the fix locally, confirmed the test failed, restored it) before committing, green after. `go build`/`go vet`/`go test ./...` (gofmt clean) all pass.

---

richkuo:
@claude review

richkuo:

PR #1380, opened 2026-07-16
feat(#1159): per-strategy correlated hedge legs (phase 1, HL perps live) [C90, Sonnet 5, xhigh]

richkuo:
## Summary

Independent implementation of phase 1 from #1159 (per-strategy correlated hedge legs), built fresh in its own worktree/session per the requester's instruction — not derived from or dependent on any of the other open PRs targeting this issue.

- **Config**: `HedgeConfig` block (`enabled`, `symbol`, `side`, `ratio`, `platform`, `type`, `margin_mode`, `leverage`) + accessors. Validation rejects: hedge coin == any configured strategy's coin, hedge coin == the strategy's own primary coin, two hedge-enabled strategies sharing a coin, `direction="both"` + hedge, and **paper mode** (phase 1 is live HL perps only — no exchange to verify hedge economics against, so this PR scopes tighter than the original plan's "both paper and live" design note).
- **State**: `Position.HedgeFor`/`HedgePrimaryQtyBasis` (persisted, migrated), excluded from the `#T` open count, W/L grading, `trade_diagnostics`, and the `#822` regime/direction orphan-close check.
- **Decision core** (`hedge.go`, pure + fully unit tested): `hedgeTargetDecision` diffs the *current* primary qty against a persisted watermark — a qty-event mirror, immune to mark-price drift. Handles open/add/reduce/close-full, dust-defers small reduces without losing the delta, and fails closed (no action) on an unusable price rather than guessing a notional.
- **Execution**: live open/add/reduce/close via `RunHyperliquidExecute`/`RunHyperliquidClose`. **Fail-closed on hedge-open failure tied to a fresh primary open**: sized (never full `market_close`) reduce-only unwind of the primary + CRITICAL owner DM; failures on a later add/reduce alert and retry next cycle instead. `RecordHedgeTradeResult` keeps hedge PnL out of `ConsecutiveLosses` (a hedge loses by construction whenever the primary wins) while still counting toward `DailyPnL`.
- **Main-loop wiring**: `runHedgeSync` runs at the manage-only (`Signal==0`) and post-trade (`trades>0`) dispatch points — mutually exclusive per cycle, so a fresh open never runs the sync twice against stale-then-fresh state.
- **Marks / shared-wallet / reconcile**: hedge coin included in the mark-fetch set, `buildSharedWalletBooks`'s `virtualQty` map, and the reconcile fee-resolver candidate set; `reconcileHyperliquidAccountPositions` reconciles the hedge coin per strategy independently of whether the strategy's own primary coin is shared with peers.
- **Kill switch / circuit breaker / manual force-close**: all extended to close the hedge leg too, gated on the leg being **currently held** (never on config alone) — a foreign position sitting on a declared-but-flat hedge coin is never touched by the kill switch.
- **Hot reload**: `hedge` block is a stop-owner-field shape — blocked while any position (primary or hedge) is open, freely reloadable while flat.
- **Backtester**: `load_strategy_config` rejects any `hedge.enabled=true` config loudly (HL-live-only, no parity in phase 1); `enabled:false` passes through unchanged.
- **Operator surfaces**: `[config]` startup summary line (`hedge=BTC×1(inverse,cross,3x)`), inspect `--json` `hedge` block, Discord `/status` `🛡️ hedged:` note.

## Scope decisions vs. the issue's design discussion

- **Live-only in phase 1** (paper rejected at load) — a deliberate narrowing from the plan sketched in the issue comments, to keep this increment's live-money paths fully verified rather than diluting effort across a second (paper) modeling surface with no exchange to check it against.
- Dashboard/UI badge rendering for hedge positions is not included — the `hedge_for`/`hedge_primary_qty_basis` fields already flow through the `/status` JSON via the `Position` struct (no extra plumbing needed), but the frontend JS doesn't yet render a distinct badge for it.

## Test plan

- [x] `go build` / `go vet` / `go test -race ./...` — all clean (scheduler package)
- [x] `gofmt -l` — clean
- [x] `uv run --no-sync python -m py_compile` on touched files — clean
- [x] Full pytest suite (`shared_strategies/ shared_tools/ platforms/ backtest/`) — 2852 passed, 1 skipped (pre-existing, unrelated)
- [x] New `hedge_test.go`: exhaustive decision-core coverage (inverse mapping, ratio sizing, add deltas, proportional reduce, dust deferral, wrong-side defense, no-basis-watermark, skip-reason mirror) + config validation/collision tests + a DB round-trip test for the new `Position` fields + a lifetime-stats exclusion test
- [x] New backtester reject tests (`enabled:true` rejected, `enabled:false` passes)
- [x] Live-binary smoke test: paper+hedge config rejected with a clear message; live+hedge config passes validation and renders the `[config]` summary line correctly; the unmodified example config still loads and runs a full cycle with no regressions

## Unfiled follow-ups (out of scope for this PR)

- Paper-mode hedge support
- Allowing hedge coins to overlap peer strategies (currently rejected — phase 1 constraint)
- Hedge PnL/fee/slippage backtest parity modeling

Closes #1159

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T14:16Z) in f7846df.

### Fixed
1. **CB/kill-switch margin drawdown counts hedge leg's unrealized loss** — confirmed at `scheduler/risk.go:1382` (`perpsMarginDrawdownInputs` swept every `Multiplier>0` position with no `HedgeFor` exclusion). Added `|| pos.HedgeFor != ""` to the skip condition, excluding the hedge leg from both the loss numerator and margin denominator, mirroring the existing `ConsecutiveLosses` exclusion. Verified the fix doesn't dilute a *real* primary-side drawdown (new test `TestPerpsMarginDrawdownInputs_PrimaryLossStillCounted`) and that a stray hedge alone can't trip the CB (`TestPerpsMarginDrawdownInputs_StrayHedgeWithPrimaryFlat`).

### Resolved judgment calls (was Recommended Optional)
1. **Partial-fill hedge open/add silently under-hedges** — confirmed at `scheduler/hedge.go:345` (basis stamped to the full intended primary qty regardless of actual fill size). Implemented a `hedgeBasisAfterFill` helper that prorates the basis watermark by `fillQty/action.Qty`, so a short fill leaves the unconverged remainder as a live delta for the next cycle to retry. Extended the same fix to `applyHedgeReduceOrCloseFill`'s reduce path (`hedge.go:417`) — the review only named open/add, but a partial-fill reduce has the identical failure mode in the opposite direction (stale excess hedge left undetected), so both are fixed for consistency. New tests cover full-fill, partial-fill (open/add/reduce), zero-requested-qty, and overfill-clamp cases.

Verification: `go build`, `go vet`, `go test -race ./...` (scheduler package) — all clean. `gofmt -l` clean.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T14:32Z) in 0836fa1.

### Fixed
1. **Partial manual force-close double-reduces the hedge leg** — confirmed at `scheduler/manual_core.go:1438` (mirror queues a proportional hedge reduce) and `scheduler/manual.go:596` (the hedge leg's own drain reduced `pos.Quantity` but never advanced `HedgePrimaryQtyBasis`, so `runHedgeSync` recomputed a stale delta and reduced it again). Fixed by setting `pos.HedgePrimaryQtyBasis` to the primary's current qty (already updated, since the primary's own action inserts and drains first) whenever a hedge leg survives a partial close. Verified all three "must survive" cases: the mirror-succeeds case now converges to the correct proportional size with no second reduce (new test `TestApplyManualAction_PartialHedgeMirrorAdvancesBasisWatermark`, red-verified against the pre-fix code); full force-close still deletes the position (no basis-touch needed); the mirror-fails backstop path is untouched since no hedge `PendingManualAction` is ever queued on that path.
2. **Kill switch strands a held hedge leg on-chain when the primary is flat** — confirmed at `scheduler/hyperliquid_balance.go:2091` (`snapshotHyperliquidVirtualQuantities`'s hedge-capture block was nested under a `continue` gated on the primary coin, so a hedge leg dangling from a prior failed `closeFull` was never captured, and `forceCloseHyperliquidLive`'s held-hedge gate then never added it to the on-chain flatten roster). Moved the hedge capture out from under the primary's guard so it runs independent of the primary's held state. Verified with new tests: stranded-hedge-gets-closed-on-chain (red-verified) and a companion foreign-position-on-hedge-coin-not-touched case confirming the `HedgeFor != ""` ownership check still excludes non-owned positions.

Verification: `go build`, `go vet`, `go test -race ./...` (scheduler package) — all clean. `gofmt -l` clean. Both new test cases were confirmed to fail against the pre-fix code (via a temporary stash of each fix) before the fix was restored, then passed after.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot]/Claude Opus 4.8 · 2026-07-16T17:54Z) in 34833a1.

### Fixed
1. **Per-strategy CB strands the hedge leg on-chain when the primary is flat but the hedge is still held** — confirmed at `scheduler/risk.go:961` (`setHyperliquidCircuitBreakerPending` returned early before ever reaching the hedge-append block whenever the primary coin wasn't in state) and the Phase-2 stuck-CB recovery block (`scheduler/hyperliquid_balance.go:2354`, same early-`continue` shape). Both restructured to build the pending-close symbol list conditionally per leg instead of returning/continuing on the primary. Verified all four "must survive" cases via new tests: `TestCheckRisk_HedgeStrandedWithFlatPrimary_StillQueuesOnChainClose` and `TestRunPendingHyperliquidCircuitCloses_RecoversStrandedHedgeWithFlatPrimary` (both red-verified against the pre-fix code), plus `TestCheckRisk_ForeignPositionOnHedgeCoinNotQueued` confirming a foreign position is still never queued.

### Resolved judgment calls (was Recommended Optional)
1. **Protective-path hedge capture keyed on `HedgeEnabled(sc)` instead of persisted state** — confirmed the finding: every safety/reconcile site (`snapshotHyperliquidVirtualQuantities`, `setHyperliquidCircuitBreakerPending`, reconcile, shared-wallet, fee-lookup) derived the hedge coin from live config, so a leg orphaned by `hedge.enabled` being flipped off (or the block removed) via a config edit + cold restart — the one path that bypasses the hot-reload guard — became invisible to all of them. Implemented `hedgeCoinForProtection(sc, ss, primarySym)` (`scheduler/hedge.go`): prefers the live config's coin when enabled (covers a fill not yet booked to state), falls back to scanning `ss.Positions` for the persisted `HedgeFor` tag otherwise. Rewired all five call sites (`hyperliquid_balance.go:858,1989,2111`, `hyperliquid_fills.go:507`, `shared_wallet_reconcile.go:543`) to use it. Deliberately did **not** apply this to the ongoing mirror (`hedgeTargetDecision`/`runHedgeSync`'s own gate) or the operator-initiated manual `force-close` mirror — disabling the feature should still stop new opens/adds/reduces; only the safety nets needed to see a disabled-but-held leg. Alternatives rejected: threading a config-independent coin string through every call site's params (more invasive, same result); leaving it as a documentation-only gap (fails the safety carve-out). Covered by new tests in `hedge_test.go` (`TestHedgeCoinForProtection_*`) and the round-3 CB/kill-switch integration tests above. Override if you'd prefer the disabled-but-held case stay a documented gap instead.

Verification: `go build`, `go vet`, `go test -race ./...` (scheduler package) — all clean. `gofmt -l` clean. Both Needs Fixing regression tests were confirmed to fail against the pre-fix code (via a temporary stash of each fix) before the fix was restored, then passed after.

---

richkuo:
@claude review

richkuo:

Issue #1341, opened 2026-07-12
[C95] One-click apply of a tuning suggestion to the live config, refused when live values have drifted


richkuo:
**Complexity: 95/100** — Capability 3 (Risk 4 — writes live strategy parameters); Volume 20 — Fable 5, xhigh

Builds on merged issues #1339 (persistent tuning-run API) and #1340 (`/tuning` page). Part 3 of 3.

**Prerequisite #1386 is merged** (PR #1387): `SCHEMA_VERSION = 2` with the raw `promotion_baseline` is live in `backtest/tune_live.py`. This issue is unblocked.

## Problem

The merged `/tuning` page displays ranked survivors but has no promotion action. Operators must retype a winning parameter set into the inline tuner or edit the config by hand. That creates avoidable transcription risk on a money-adjacent surface.

The tuning pipeline is deliberately suggest-only. Promotion must remain an explicit human action, and a suggestion must never be promoted after the live raw inputs used to produce its evidence have changed.

The merged implementation also establishes constraints that the original issue did not account for:

- Suggestion keys such as `baseline` and `cand_1` repeat across strategies, so a run id plus suggestion key is not a unique identity.
- Schema-v1 artifacts contain raw explicit open parameters and resolved close information, but not the complete raw baseline required for a reliable drift check.
- Survivor patches currently contain only `open_strategy` name/parameters. They do not contain close-stack fields.
- The existing inline-tuner helper merges open-strategy parameters, while a survivor patch represents the complete promoted open-strategy value. Reusing that helper verbatim would change the artifact contract.
- A validated config write and an audit-artifact write are separate filesystem operations, so crash recovery must be designed explicitly.

## Goal

Let an operator promote one ranked survivor from `/tuning` with one click. The server must apply the exact server-stored `open_strategy` patch only when the current raw baseline still matches the run baseline, reject legacy or ambiguous requests, and preserve an idempotent, recoverable audit trail. The client never submits parameter values, and the system never self-promotes.

## Invariants

- Promotion identity is the triple `(run_id, strategy_id, suggestion_key)`.
- Only a completed run's ranked survivor carrying a server-stored `patch.open_strategy` is promotable.
- This issue promotes open-strategy values only. Close-stack promotion is out of scope.
- Drift comparison is raw-to-raw in Go. Python remains responsible for producing the tuning evidence; Go already performs live close-default resolution and must not attempt to reproduce the backtester for this endpoint.
[...]

PR #1387, opened 2026-07-20
feat(#1386): persist schema-v2 raw promotion baseline in tune_live artifacts [C37, Cursor Grok 4.5, high]

### LINKED ISSUE #1386: [C37] Save the raw baseline settings in tuning results so a suggestion can later be applied safely
**Complexity: 37/100** — Capability 1 (Uncertainty 2 — raw-capture placement vs shared loader return contract); Volume 12 — Cursor Grok 4.5, high

Split out of #1341 (one-click apply of a tuning suggestion). Part 1 of 2: #1341 remains the promotion endpoint, journal, and UI, and is blocked on this issue. Landing this first means tuning runs start producing promotable artifacts immediately, shrinking the population of legacy runs that will need a re-run once #1341 ships.

## Problem

`backtest/tune_live.py` artifacts (schema v1) record evidence fields used for display and search, not the raw config blocks #1341 needs for a promotion drift check:

- `baseline_params` — the strategy's **explicit** `open_strategy.params` from the config (unmerged with registry defaults; UI already treats this as raw-unmerged).
- `close_strategies` — after `inject_user_defaults=True`, so user-defaults close injection can be present.
- `stop_owner` — strategy-level stop fields from the load, which are usually raw `sc` values but can include user-defaults-injected regime trail/SL when that path runs.

They do **not** persist the raw `open_strategy` object or the raw root `user_defaults` object with presence bits.

#1341's promotion flow needs a raw-to-raw drift check in Go: "has the live config changed since this evidence was produced?" That check cannot be built by re-resolving evidence fields — Go would have to reproduce Python's resolution logic, which #1341 explicitly forbids. The artifact must therefore carry the raw baseline itself.

## Goal

Bump the tune_live artifact to schema v2. Every per-strategy result persists a `promotion_baseline` object holding the raw values read from the same config load used to resolve that strategy:

```json
{
  "open_strategy": <raw JSON verbatim, or null>,
  "open_strategy_present": <bool>,
  "user_defaults": <raw JSON verbatim, or null>,
  "user_defaults_present": <bool>
}
```

## Invariants

- Raw means verbatim from the config JSON — `"open_strategy" in sc` / `"user_defaults" in cfg` (value `null` when absent) — captured **before** `_effective_user_close_defaults` transforms `user_defaults`, before absent `open_strategy` collapses to `{}`, and before any close/stop injection. Resolved or injected values must never leak into `promotion_baseline`.
- Presence bits are explicit: an absent block is `null` + `false`, distinguishable from an empty object.
- The baseline is taken from the same `load_strategy_config` load that builds the resolved evidence for that strategy — no separate reload for the baseline (no window for the two to diverge).
- Existing evidence/display fields (`baseline_params`, `close_strategies`, `stop_owner`) stay unchanged in meaning and shape.
- No v2 baseline is synthesized for already-persisted v1 runs; they stay as-is and #1341 will classify them `legacy_artifact`.

## Approach

- Bump `SCHEMA_VERSION` from 1 to 2 in `backtest/tune_live.py` (stamped in artifact, progress JSON, and summary header as today).
- Extend `load_strategy_config` (`backtest/run_backtest.py`) to snapshot raw `open_strategy` and raw root `user_defaults` with presence bits **before** transform/collapse, and return them nested under a non-Backtester key (e.g. `promotion_baseline`) so allowlisted/`**kwargs` consumers of the loader stay safe. Do not treat the transformed `_effective_user_close_defaults` view as the raw baseline.
- In `tune_strategy`, copy that `promotion_baseline` onto each per-strategy result next to `baseline_params`.
- Update the `tune_live.py` row in `docs/backtesting-registry.md` for schema v2 in the same PR (registry rule).

No Go changes: `readTuningResults` (`scheduler/ui_tuning.go`) only presence-checks `schema_version`, so v2 artifacts render on `/tuning` unchanged. Go-side version parsing and eligibility land with #1341.

## Acceptance criteria

- Artifact and progress JSON report `schema_version: 2`; the existing artifact-structure test is extended rather than replaced.
- `promotion_baseline` is byte-faithful to the raw config file: a config that uses `user_defaults` close injection still has `promotion_baseline.open_strategy` / `promotion_baseline.user_defaults` matching the file verbatim, with no injected close data inside the baseline.
- Absent `open_strategy` and absent `user_defaults` each produce `null` + presence bit `false` (empty `{}` is present + empty object).
- When `open_strategy` is present with explicit params, `promotion_baseline.open_strategy.params` equals the recorded `baseline_params` (same explicit dict — `baseline_params` is not registry-merged). A separate assertion may show `baseline_params != effective_params` when registry defaults fill omitted keys; that is search-universe behavior, not a promotion_baseline divergence.
- When the config uses args-form open naming (empty/`open_strategy.name` filled from `args[0]` at load), the raw baseline still stores the file's `open_strategy` block verbatim (including empty name), not the resolved name.
- `backtest/tests/test_tune_live.py` passes; full pytest suite passes.

## Touch set

- `backtest/tune_live.py`, `backtest/run_backtest.py`
- `backtest/tests/test_tune_live.py`
- `docs/backtesting-registry.md`

## Out of scope

- Everything promotion-side: apply endpoint, drift comparison, journal, eligibility, UI (#1341).
- Migrating or re-validating persisted v1 artifacts.

---
Updated with LLM: Cursor Grok 4.5 | high | Harness: Cursor

richkuo:
## Plain simple English

Tuning results now save a copy of the exact config settings that were in effect when the run started. Later, when someone wants to apply a suggestion, the system can check whether those settings changed — without guessing or re-running old logic.

## Summary

- Bump `tune_live` `SCHEMA_VERSION` to 2; each successful per-strategy result carries `promotion_baseline`.
- Add opt-in `include_promotion_baseline` to `load_strategy_config` (Finding 1: unconditional key would break `Backtester(**kwargs)` spread callers).
- Baseline captures raw `open_strategy` / `user_defaults` / `user_close_defaults` with presence bits from the same config read, before any close/stop injection (Finding 2: include deprecated legacy block).
- Update `docs/backtesting-registry.md` tune_live row for schema v2.

Closes #1386.

## Verification

- `uv run --no-sync python -m py_compile backtest/run_backtest.py backtest/tune_live.py`
- `uv run --no-sync python -m pytest backtest/tests/test_tune_live.py` — 77 passed
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2868 passed, 1 skipped
- `go test -C scheduler ./...` — ok

## Note for #1341

Baseline includes deprecated `user_close_defaults` (presence + value). Presence = key present in the JSON object (`null`+`true` ≠ absent `null`+`false`).


richkuo:
@claude review

Issue #1382, opened 2026-07-18
[C37] Cap or prune accumulated tuning run artifacts


richkuo:
**Complexity: 37/100** — Capability 1 (Risk 2 — wrong prune can erase tuning audit history; Uncertainty 2 — config/hot-reload wiring); Volume 12 — Opus-class, high

## Problem

The #1339 persistent tuning-run API never prunes run directories or in-memory records. Each enqueue creates a new `tuning_runs/<id>/` directory; nothing deletes them. `loadPersistedRuns` reads every valid `run.json` at startup (O(n) metadata — not `results.json`/`progress`), and `list()` copies+sorts all in-memory records on every `GET /api/tuning/runs`. On-disk footprint (especially large `results.json` artifacts) grows without bound across the daemon's lifetime.

Tracked out of https://github.com/richkuo/go-trader/pull/1381 (review follow-up) rather than expanding that diff.

## Goal

Add an explicit count-based retention cap for tuning runs so operators can bound disk and list/restart metadata without losing in-flight work or deleting history until they opt in.

## Approach

1. **Config:** top-level `tuning.max_retained_runs` on `Config` (omitempty int). `0` / omitted = keep-all (prune off) — the safe default so operators are not surprised by auto-delete. Positive `N` retains the newest `N` *terminal* runs.
2. **Prune when:** after `loadPersistedRuns` at manager construction, and again after each terminal `storeRecord` (completed/failed/interrupted/rejected). No main-loop periodic timer.
3. **Never delete** `queued` or `running` runs.
4. **How:** choose oldest terminal runs beyond the cap by `CreatedAt` (tie-break `ID`); `RemoveAll` the whole run directory, then drop the id from `m.runs`. Per-dir fail-open on `RemoveAll` errors (log + continue) so one stuck directory cannot block the rest. Corrupt/id-mismatched/invalid-status dirs that `loadPersistedRuns` already skips stay skippable (do not invent deletes for them).
5. **Touch sites:** `Config` field + loadConfig validate/default; pass retention into `newTuningRunManager` (or read from server cfg); SIGHUP-adoptable via `applyHotReloadConfig` + clone mask; `config.example.json`; ARCHITECTURE/`ui_tuning` note.
6. **Non-goals:** no age-based retention in this issue; serial lane / auth / results schema unchanged.

## Acceptance criteria

1. `tuning.max_retained_runs` documented; default `0` = keep-all / prune off; positive `N` enables count-based prune of oldest terminal runs.
[...]

Issue #1304, opened 2026-07-10
[C20] Speed up CI Python tests by running them in parallel with pytest-xdist


richkuo:
**Complexity: 20/100** — scope: small (pyproject test extra + one CI step); risk: low (CI-only); uncertainty: some — hidden shared state between tests could surface under parallel workers.

## Problem
The CI `python` job takes ~12.5 minutes; the `uv run pytest` step alone is 739s (run 29081931971). Five regime test files account for ~703s of that, with `backtest/tests/test_regime_bounded_window.py` alone at 404s. All 2,517 tests run sequentially on one core.

## Goal
Cut the pytest step's wall-clock substantially (target: well under 7 minutes) without removing or weakening any tests.

## Approach
1. Add `pytest-xdist` to the `test` extra in `pyproject.toml`.
2. Change the CI `Run tests` step in `.github/workflows/ci.yml` to `uv run pytest -n auto` (and the `shared_scripts/` step if it benefits — it's only 1.6s today, so likely leave it).
3. Verify all 2,517 tests still pass under parallel workers; fix any test-isolation issues that surface (shared temp files/DBs, module-level state).
4. Compare before/after step timings on the PR's CI run and record them in the PR body.

## Acceptance criteria
- CI pytest step time drops materially (expected under ~7 minutes; bounded by the slowest single test).
- No tests skipped, removed, or flaky under `-n auto`.

---
Created with LLM: Fable 5 | high | Harness: Claude Code

PR #1388, opened 2026-07-20
feat(#1341): one-click apply of tuning suggestion with drift refusal [C95, Cursor Grok 4.5, high]

### LINKED ISSUE #1341: [C95] One-click apply of a tuning suggestion to the live config, refused when live values have drifted

richkuo:
## Plain simple English

Operators can now click Apply on a winning tuning suggestion to put those settings live. The system blocks the change if live settings moved since the run, keeps an audit trail, and never promotes anything on its own.

## Summary

- Add `POST /api/tuning/apply` for identity triple `(run_id, strategy_id, suggestion_key)` with mutating auth + same-origin guards
- Exact `open_strategy` replacement via `mutateConfigRoot` (not tuner merge), refused on schema-v1 / incomplete baseline / non-survivor / raw baseline drift
- Crash-recoverable `tuning_runs/promotions.json` journal (`pending` → `applied` / `manual_review`) outside per-run dirs so #1382 prune cannot erase audit state
- GET run detail overlays transient `apply_eligibility` / `applied_at`; `/tuning` Apply button is gated by server eligibility and refreshes after every attempt

Closes #1341

## Verification

- `go -C scheduler build .`
- `go -C scheduler test -count=1 -timeout 300s .` (pass)
- Coverage includes auth/origin matrix, per-strategy identity, legacy/non-survivor refusal, three-block + presence drift, crash recovery, prune-surviving journal, `configWriteMu` serialization, eligibility overlay, and UI wiring/node eligibility mapping


richkuo:
@claude fable effort:high review

richkuo:
Addressed review feedback (claude[bot] · 2026-07-20T05:48Z) in 2c9a9ad.

### Fixed
1. **Apply never signaled SIGHUP** — `commitTuningPromotion` and the pending-finalize path now call `triggerConfigReload()` and return its message; idempotent `already_applied` retries and refusals do not (`scheduler/ui_tuning_apply.go`).
2. **Pruned applied retry returned `manual_review`** — journal `applied` is checked before run-map/dir lookup so prune-surviving promotions stay idempotent (`scheduler/ui_tuning_apply.go`).
3. **Open-position open-as-close confirm** — confirm + inline advisory warn when `has_open_position` and no `close_strategy` (`scheduler/static/ui/app.js`).

### Not changed (refuted)
1. **CI Failure — python** — failure is `.github/scripts/test_patch_claude_comment.py::test_run_id_selects_own_comment_despite_newer_same_author` under xdist (`patched == ''`); this PR touches no Python/workflow files and Go passed. Pre-existing isolation defect per #1304.

### Deferred to follow-up
1. **Hermetic patch_claude_comment tests under `pytest -n auto`** — filed https://github.com/richkuo/go-trader/issues/1389.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-20T06:15Z) in $(git rev-parse --short HEAD).

### Fixed
1. **Eligibility overlay re-read promotions.json per ranked row** — `overlayTuningApplyEligibility` now loads the journal once via `promotionIndex` and looks each row up in memory (`scheduler/ui_tuning_apply.go`); regression asserts one load per detail GET (`TestTuningEligibilityOverlayLoadsJournalOnce`).

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-20T06:36Z) in 7597b1d2.

### Fixed
1. **Pending stuck on strategy remove/dup mid-apply** — `commitTuningPromotion` / `recoverPendingTuningPromotion` now settle `errTuningStrategyMissing`/`errTuningStrategyDup` to journal `manual_review` (response `manual_review`); transient errors still leave `pending` for retry (`scheduler/ui_tuning_apply.go`).
2. **Eligibility overlay rescanned live strategies per ranked row** — one `extractLivePromotionBaseline` per strategy id, reused across that strategy's ranked rows (`scheduler/ui_tuning_apply.go`).

---

richkuo:
@claude review

PR #1390, opened 2026-07-20
fix(#1389): eliminate pipefail race in patch_claude_comment fake gh stub [C27, Cursor Grok 4.5, high]

### LINKED ISSUE #1389: [C27] Fix random CI failure in patch_claude_comment tests: pipefail race in the fake gh stub
**Complexity: 27/100** — Capability 1 (Uncertainty 2 — flake mechanism hypothesized from code, must be reproduced/confirmed); Volume 2

## Problem

CI on unrelated PRs (observed on #1388, run 29718904917) intermittently fails the Python job on:

`.github/scripts/test_patch_claude_comment.py::test_run_id_selects_own_comment_despite_newer_same_author`

under `pytest -n auto` (worker gw1): the returned patch log was empty, so `assert "repos/.../comments/202" in patched` fails against `''`. The script exited 0 (a non-zero exit would have tripped the rc assertion at `test_patch_claude_comment.py:111` first), meaning it took the clean-noop "No comment found" path (`patch_claude_comment.sh:73-83`). The same PR surface touches no Python or workflow files; a rerun passed. Per #1304, treat intermittent parallel-run failures as test-isolation/determinism defects, not flakes.

## Root cause (validated)

Cross-worker contamination is ruled out: `tmp_path` is per-test and per-xdist-worker (the failure log shows a `popen-gw1` basetemp), the env dict is a private copy, and the stub `gh`, comments fixture, and patch log all live under `tmp_path` (`test_patch_claude_comment.py:71-115`). The shared `cwd=REPO_ROOT` is read-only.

Prime suspect is inside the fake `gh` stub itself (`test_patch_claude_comment.py:55-68`): under `set -euo pipefail`, the branch condition `printf '%s\n' "$@" | grep -q -- '--paginate'` races — `grep -q` exits at first match, `printf` can take EPIPE under scheduler load, and `pipefail` then reports the matched pipeline as failed. The condition falls through to the `cat "${GH_STUB_SINGLE:-/dev/null}"` fallback, the outer script sees empty comment JSON, and lands on the exit-0 noop path with an empty patch log — exactly the observed failure. Load-sensitive (hence surfacing under `-n auto`), not shared state.

## Approach

1. Reproduce/confirm the stub race (e.g. stress loop under load), then eliminate the pipeline: replace both `printf … | grep -q` branch conditions in `FAKE_GH` with `case " $* " in` string matching — no pipe, no EPIPE, deterministic.
2. Strengthen diagnostics in `run_patch_script`: also capture stdout so an unexpected miss fails with the script's actual output and stderr, not a blank-string assertion. (The non-zero-exit path already fails loudly at `test_patch_claude_comment.py:111`.)
3. Distinguish expected noop from unexpected miss: the noop tests (`test_run_id_without_match_is_a_clean_noop`, `test_run_id_match_is_not_a_prefix_match`, `test_no_matching_comment_is_a_clean_noop`, `test_on_miss_post_without_status_note_is_a_noop`) legitimately assert an empty patch log — have them additionally assert the "No … comment found" stdout message so a mis-routed stub can't masquerade as a valid noop.
4. Confirm the suite is green under repeated `pytest -n auto` runs.

## Acceptance criteria

- `test_run_id_selects_own_comment_despite_newer_same_author` passes reliably under repeated `pytest -n auto` runs.
- `FAKE_GH` branch dispatch contains no `printf | grep` pipelines.
- An unexpected selection miss fails with the script's stdout/stderr in the assertion message, never a bare `''` comparison.
- Intentional-noop tests assert the noop message, not just an empty log.

---

richkuo:
## Plain simple English

CI sometimes failed a workflow helper test for no real code reason. The fake GitHub CLI used in that test could mis-read its arguments under load. This change makes that stub deterministic and makes real misses easier to diagnose.

## Summary

- Replace `FAKE_GH`'s `printf | grep -q` branch dispatch with `case " $* "` matching so `set -o pipefail` + early `grep -q` close can no longer EPIPE a matched `--paginate`/`--method` call into the empty single-comment GET fallback (#1389).
- Return stdout/stderr from `run_patch_script` via `PatchScriptResult`; patch-expecting asserts include them on failure instead of a bare `''` comparison.
- Intentional-noop tests now also require the script's `No … comment found — nothing to update.` stdout message.
- Add `test_fake_gh_stub_has_no_printf_grep_pipelines` as an acceptance guard.

Closes #1389

## Verification

- Red→green: `test_fake_gh_stub_has_no_printf_grep_pipelines` failed against the old `printf|grep` stub, then passed after the `case` fix.
- `uv run --no-sync python -m pytest .github/scripts/test_patch_claude_comment.py -q` → 14 passed
- Repeated `pytest -n auto` ×5 on the same module → 14 passed each run

### DIFF
--- a/.github/scripts/test_patch_claude_comment.py
+++ b/.github/scripts/test_patch_claude_comment.py
@@ -11,10 +11,13 @@ on PATH; the script's comment recomposition (compose_claude_comment.py) runs
 for real.
 """
 
+from __future__ import annotations
+
 import json
 import os
 import stat
 import subprocess
+from dataclasses import dataclass
 from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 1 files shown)


richkuo:
@claude review

PR #1398, opened 2026-07-20
fix(#1395): bound LoadState's trade query in SQL instead of Go truncation [C8, Sonnet 5, high]

### LINKED ISSUE #1395: [C8] LoadState loads a strategy's entire trade history from SQLite before trimming it in Go
**Complexity: 8/100** — Capability 0 (Risk 1 — perf-only, no money/data-integrity path; Uncertainty 0 — fix pattern is standard); Volume 8 (Scope 1, Coupling 1, Verification 2) — Sonnet-class, high

## Problem
`StateDB.LoadState()` (`scheduler/db.go:1795`) queries `SELECT ... FROM trades WHERE strategy_id = ? ORDER BY timestamp ASC` with no SQL `LIMIT`, appends every row into `allTrades` (`scheduler/db.go:1799-1820`), and only then truncates to `maxTradeHistory` in Go (`scheduler/db.go:1826-1828`). The `trades` table has separate single-column indexes on `strategy_id` (`idx_trades_strategy`, `db.go:192`) and `timestamp DESC` (`idx_trades_timestamp`, `db.go:194`), but no composite `(strategy_id, timestamp)` index, so SQLite can't satisfy this query's filter+order together from one index.

## Impact
Startup/restart load time and memory for `LoadState()` scale with a strategy's *total lifetime* trade count, not `maxTradeHistory` — the opposite of the bound the in-memory truncation implies. This gets worse as strategies accumulate history over time.

## Acceptance criteria
- Bound the query in SQL, e.g. `ORDER BY timestamp DESC, rowid DESC LIMIT ?` in a subquery, then re-order ascending for in-memory chronology (mirrors the pattern already used at `db.go:1025`).
- Add a composite index `(strategy_id, timestamp DESC)` (or `(strategy_id, timestamp DESC, rowid DESC)` to match ordering) if one doesn't already cover this access path.
- Add a regression/benchmark test with a strategy having more than `maxTradeHistory` trade rows, asserting both correctness (same trades survive truncation) and that the query no longer scans the full history.

## Context
Restated from a withdrawn upstream-audit issue (originally filed against a private derivative and later withdrawn by its author); this specific claim is grounded entirely in current `richkuo/go-trader` Go code and was independently verified.

---

richkuo:
## Summary
- `StateDB.LoadState()` queried a strategy's entire lifetime trade history from SQLite with no `LIMIT`, then truncated to `maxTradeHistory` (1000) in Go — so restart time/memory scaled with total trade count instead of the intended bound.
- Bounded the query in SQL: `ORDER BY timestamp DESC, rowid DESC LIMIT ?`, mirroring the existing pattern in `RecentTradesForStrategy` (`db.go:1025`), then reversed to chronological order in memory.
- Added composite index `idx_trades_strategy_timestamp(strategy_id, timestamp DESC, rowid DESC)` via `migrateSchema` so SQLite can satisfy the filter + order from one index instead of falling back to a single-column index scan plus a temp B-tree sort.

## Verification
- New regression test `TestLoadState_TradeHistoryBoundedInSQL`: seeds 1200 trades for one strategy, asserts only the newest 1000 survive and are in ascending chronological order, and asserts via `EXPLAIN QUERY PLAN` that the composite index is used (no `SCAN TABLE trades`).
- Confirmed red→green: ran the new test against the unfixed query first — it failed (plan showed `SEARCH ... USING INDEX idx_trades_strategy` + `USE TEMP B-TREE FOR ORDER BY`, no LIMIT) — then restored the fix and it passed.
- `go build .`, `gofmt -l`, and the full `go test ./...` suite pass.

Closes #1395

## Plain simple English
When the app restarts, it used to pull a strategy's entire trade history out of the database just to throw away everything except the most recent 1000 trades. This change tells the database to only fetch the last 1000 in the first place, so restarts stay fast no matter how much trading history a strategy has built up.

---

### DIFF
--- a/scheduler/db.go
+++ b/scheduler/db.go
@@ -192,9 +192,9 @@ CREATE TABLE IF NOT EXISTS trades (
 CREATE INDEX IF NOT EXISTS idx_trades_strategy ON trades(strategy_id);
 CREATE INDEX IF NOT EXISTS idx_trades_symbol ON trades(symbol);
 CREATE INDEX IF NOT EXISTS idx_trades_timestamp ON trades(timestamp DESC);
--- idx_trades_close (#455) and idx_trades_strategy_position (#471) are created
--- in migrateSchema, not here, so legacy DBs add columns before indexes
--- reference them.
+-- idx_trades_close (#455), idx_trades_strategy_position (#471), and
+-- idx_trades_strategy_timestamp (#1395) are created in migrateSchema, not
+-- here, so legacy DBs add columns before indexes reference them.
 
 CREATE TABLE IF NOT EXISTS portfolio_risk (
     id INTEGER PRIMARY KEY CHECK (id = 1),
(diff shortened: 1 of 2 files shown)


richkuo:
@claude review

PR #1399, opened 2026-07-20
fix(#1396): synchronize kill-switch reset prompt flag [C33, Cursor Grok 4.5, high]

### LINKED ISSUE #1396: [C33] Kill-switch reset prompt flag is read and written from two goroutines with no lock
**Complexity: 33/100** — Capability 1 (Risk 2 — affects kill-switch reset DM prompting, contained/reversible blast radius; Uncertainty 0 — fix pattern is standard); Volume 8 (Scope 1, Coupling 1, Verification 2) — Opus-class, high

## Problem
`resetGoroutineRunning` (`scheduler/main.go:755`) is a plain `bool`, checked and set to `true` in the main loop (`scheduler/main.go:1580-1581`), then spawned into a `go func()` that clears it back to `false` via `defer` (`scheduler/main.go:1585`) with no mutex or `sync/atomic` guard. The main loop and the spawned goroutine read/write the same variable across goroutines with no synchronization — a genuine data race, confirmed under `go test -race` / `go build -race`.

## Impact
Under concurrent access this is undefined behavior per the Go memory model. In practice it can suppress a legitimate kill-switch reset DM prompt (main loop reads a stale `false` or `true`) or spawn a duplicate one, depending on scheduling — an unsynchronized read/write on the flag that gates that prompt.

## Acceptance criteria
- Replace `resetGoroutineRunning` with a `sync/atomic.Bool`, or guard both the main-loop read/write and the goroutine's clear with the same mutex.
- Add a race-detector-covered test (or a targeted unit test if a full `-race` integration test isn't feasible) that exercises concurrent set/clear of the flag.

## Context
Restated from a withdrawn upstream-audit issue (originally filed against a private derivative and later withdrawn by its author); this specific claim is grounded entirely in current `richkuo/go-trader` Go code and was independently verified.

---

richkuo:
## Plain simple English

The kill-switch reset chat prompt used a shared on/off flag from two places at once with no lock. That could skip a real reset ask or start a duplicate one. It now uses a safe atomic handoff so only one prompt runs at a time.

## Summary

- Replace the plain `resetGoroutineRunning bool` with `atomic.Bool`.
- Claim the single-flight slot via `CompareAndSwap` (`tryClaimKillSwitchResetPrompt`) and clear it on goroutine exit (`releaseKillSwitchResetPrompt`), matching the existing `shutdownDraining` pattern.
- Add single-owner + concurrent claim/release tests; verified under `go test -race`.

Closes #1396

## Verification

- Red proof: unsynchronized plain-bool pattern fails under `go test -race` (DATA RACE).
- Green: `go -C scheduler test -race -count=1 -run 'TestTryClaimKillSwitchResetPrompt_' .` passes.
- `go -C scheduler test -count=1 -run 'KillSwitchReset' .` passes.
- `go -C scheduler build .` succeeds.

---

### DIFF
--- a/scheduler/kill_switch_reset_dm.go
+++ b/scheduler/kill_switch_reset_dm.go
@@ -3,8 +3,23 @@ package main
 import (
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 )
+
+// tryClaimKillSwitchResetPrompt atomically claims the single-flight slot for
+// the kill-switch reset AskOwnerDM goroutine (#1396). Returns true iff this
+// caller owns the prompt; the owner MUST call releaseKillSwitchResetPrompt
+// when the goroutine exits (typically via defer).
+func tryClaimKillSwitchResetPrompt(running *atomic.Bool) bool {
+	return running.CompareAndSwap(false, true)
+}
+
+// releaseKillSwitchResetPrompt clears the single-flight slot so a later cycle
+// can prompt again after the AskOwnerDM goroutine finishes.
+func releaseKillSwitchResetPrompt(running *atomic.Bool) {
+	running.Store(false)
+}
 
 // DefaultKillSwitchResetDMTimeout is the AskOwnerDM wait for the portfolio
 // kill-switch reset prompt when kill_switch_reset_dm_timeout is omitted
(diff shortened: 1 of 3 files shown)


richkuo:
@claude review

PR #1404, opened 2026-07-24
feat(#1159): add per-strategy correlated hedge legs (phase 1, HL perps) [C90, Opus 5, xhigh]

richkuo:
## Plain simple English

A strategy can now automatically open a second, opposite trade on a related coin — long Ethereum while short Bitcoin, say. The two move together, so the pair is much less exposed to the market's direction. The hedge shrinks and closes with the main trade; if it fails to open, that trade is undone.

## Summary

Adds an opt-in per-strategy `hedge` block (Hyperliquid perps only, live and paper) that runs an auto-managed correlated hedge leg on a second coin, opened opposite the primary. Off by default; nothing changes for any existing strategy.

**Core design — one state-derived reconciler, not per-event hooks.** `hedgeTargetDecision` is a pure function computing the hedge target from the *current* primary position versus a persisted quantity watermark (`Position.HedgePrimaryQtyBasis`); `runHedgeSync` converges the leg to it once per HL dispatch cycle. A primary position changes size through at least eight paths (fresh open, scale-in add, evaluator partial, evaluator full close, on-chain SL fill, on-chain TP fill, ratchet close, external close). Hooking each one would mean eight places to keep correct forever, and any path added later would silently skip the hedge. Deriving from state means all of them mirror automatically, and a restart mid-sequence resumes from the watermark with no recovery logic of its own.

Sizing keys on primary **quantity**, not price — `hedge_qty = primary_qty_delta × primary_px × ratio / hedge_px` — so mark drift never re-trades the leg. A reduce takes the same *fraction* off the held hedge quantity, making it immune to price movement between open and reduce.

### Safety properties

| Property | How |
|---|---|
| Fail-closed sizing | Unusable marks or an unknown primary side return `Blocked` — never size off a stale basis. A primary-flat close still executes without marks, so a feed outage can't strand a naked leg. |
| Live-exec guard | Virtual state mutates only on a confirmed fill, and books the size the exchange **actually filled**. A short fill books a partial and advances the watermark proportionally; a short-filled *full close* books a partial close rather than marking the leg flat while exposure remains on-chain. |
| Fail-closed open (constraint 4) | A hedge failure on a cycle that *added* primary exposure closes exactly that increment reduce-only + CRITICAL owner DM. A fresh open unwinds fully (cancelling its resting protection OIDs); a failed scale-in add unwinds only the add — the pre-add position is still correctly hedged, and its protection is preserved. A failure on any later cycle alerts and retries instead of liquidating an aged position. A failed unwind needs no latch: the reconciler self-heals next cycle. |
| Ownership | Only the persisted `Position.HedgeFor` stamp. Never coin→config inference; an unstamped position on the hedge coin is never adopted, and reconcile surfaces it instead of staying silent. |
| Collision rejection | `validateHedgeConfigs` rejects a hedge coin equal to the strategy's own primary, **any** configured strategy's coin (perps + manual, live + paper), or another hedger's hedge coin — plus `direction:"both"`, non-HL-perps owners, and out-of-vocabulary/bounds fields. This is what lets every shared-coin mechanism keyed on `hyperliquidConfiguredCoin` stay untouched. |
| Risk accounting | `RecordHedgeTradeResult` books hedge PnL into `DailyPnL` but **never** `ConsecutiveLosses`. A hedge loses by construction when the primary wins, so counting it would both mis-fire the loss-streak circuit breaker *and* spuriously **reset** a genuine losing streak — disarming the breaker exactly when it's needed. All perps close sites route through `recordPositionTradeResult`. |

### Coverage extensions

Hedge coins are invisible to `hyperliquidConfiguredCoin`, so each place one must be seen is extended explicitly:

- **Marks** (`collectPerpsMarkSymbols`) — else the leg values at `AvgCost`, its unrealized loss reads as exactly zero forever, and the circuit breaker never sees it.
- **Shared-wallet books** — else the coin reads as an orphan and drift alerts fire every cycle; funding events on the hedge coin would land nowhere instead of on the owner's ledger.
- **Reconcile fill resolver** + a dedicated `reconcileHyperliquidHedgeLeg` (reuses the existing resolver for drift/external-close/non-adoption; passes `pendingOrphanCloses=nil` so the #822 check never auto-closes a correctly-inverse leg).
- **Kill switch** — roster, virtual-quantity snapshot, and fill booking. Gated on a *held* leg, never config alone, so a foreign position on a declared-but-flat hedge coin is not liquidated.
- **Circuit breaker** — pending closes and stuck-CB reconstruction. Flattening only the primary would convert a market-neutral pair into naked **inverse** exposure at the moment the breaker decided to stop trading.
- **Manual `force-close`** — closes the proportional hedge share in the same command; the drain re-anchors the basis so the next sync doesn't reduce twice.

`trade_type="hedge"` is excluded from lifetime #T / W-L on **both** the open-count and the close-side round-trip aggregation, while `tradeLedgerDeltaSQL`/`tradeNetPnLSQL` ignore `trade_type` so PnL and fees still book to the owner's ledger. Hot reload is blocked while either leg is open; a config edit + restart bypasses that, so `validateHedgeStateConsistency` warns at startup and leaves an orphaned leg **frozen** rather than guessing a close. The backtester rejects an enabled block loudly — a single-instrument simulation would drop the hedge's PnL and fees and report the naked primary as the strategy.

## Verification

- `gofmt`, `go vet ./...`, `go build` — clean.
- `go test ./...` — full suite green; `-race` clean on the hedge tests.
- `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2873 passed, 1 skipped.
- **End-to-end paper run** against live Hyperliquid marks (`--once` × 4): primary opens 0.5 ETH → hedge opens 0.01448 BTC short (notional-matched); primary 0.5 → 0.25 → hedge reduces 50%; primary closed → hedge flattened. No residual position; all three hedge legs booked with `trade_type=hedge`.

Two defects were found by that end-to-end run *after* the unit tests were already green, and both now have regression tests proven red on the unfixed code:

1. Hedge **close** legs were booked as `trade_type="perps"`, so the stats exclusion only covered opens — every hedge round trip would still have counted as a win or loss.
2. `applyHedgeFill` booked the *requested* quantity rather than the exchange's reported fill.

88 new test functions (plus 35 sub-cases) cover the decision core (inverse mapping, ratio sizing, proportional reduce, add deltas, dust deferral, wrong-side/corrupt legs, unusable-mark fail-closed), the skip-reason mirror, fill booking and partial fills, both unwind scopes, failed-primary-open, startup reconcile and non-adoption, kill-switch/CB coupling, ledger attribution, persistence round-trip, and the hot-reload block.

## Notes

Phase-1 scope per the issue: HL perps only, no independent hedge SL/TP or close evaluator, backtest parity deferred. The issue's two named follow-ups — allowing hedge coins that overlap peer strategies, and backtest hedge PnL/fee/slippage modelling — remain **unfiled**.

Closes #1159

---


richkuo:
@claude review

richkuo:
Addressed review feedback (`github-actions[bot]` · 2026-07-25T03:23:03Z) in 708ab483.

Both findings validated against current code and confirmed. CI was green at review time (`go`, `python` both `pass`), so no CI findings.

### Fixed

1. **Partially-filled hedge reduce advanced the watermark to the full target, stranding an oversized hedge** — confirmed exactly as described, at both sites: the `hedgeActionReduce` branch and the short-filled residual-full-close branch each stamped `HedgePrimaryQtyBasis = action.NewBasis` after `bookPerpsPartialCloseWithFillFee` had booked only `filledQty`. New `hedgeReducedBasis(oldBasis, target, filled, requested)` (`scheduler/hedge.go:734`) interpolates between the PRE-reduce basis and the target by the fill ratio — the same proportional rule the add (`hedge.go:788`) and partial-open (`hedge.go:794`) paths already used — and is applied at `hedge.go:858` and `hedge.go:883`. The pre-reduce basis is captured *before* booking, since the booker mutates or (on a full drain) deletes the position.

   Your suggested formula is what I implemented; verified it lands exactly on the target for a full fill and leaves precisely the residual surplus otherwise. All three **Must survive** cases are regression-tested in `TestPartialHedgeReduceLeavesDeltaForTheNextCycle` (`scheduler/hedge_test.go`): (a) a 50% fill with no further primary event still trims the surplus next cycle (0.3 → 0.2); (b) the residual-below-minimum full close when it itself partially fills; (c) two consecutive partial reduces converging on the true remaining size rather than compounding. `TestHedgeReducedBasisInterpolatesByFillRatio` pins the helper's boundaries (full fill, over-fill clamp, zero fill, unanchored basis, zero request).

2. **Collision detection normalized its two sides differently, letting a ccxt-suffixed symbol slip a colliding hedge coin past validation** — confirmed. New `hedgeCollisionCoin` (`scheduler/config.go:1702`) runs the primary through `normalizeHedgeCoin` before comparison, applied at both keying sites (`config.go:1736`, `config.go:1811`).

   Scoped deliberately to the hedge validator: `hyperliquidConfiguredCoin` itself is left untouched because it is the key the entire shared-coin machinery (peer margin checks, kill-switch roster, CB drain, reconcile ownership) is built on, and widening its semantics has a blast radius this PR can't verify. Normalizing both sides *here* can only ADD collision detections, never remove one, so it's a strict tightening with no reachable regression. All three **Must survive** cases plus a distinct-coins negative control are covered in `TestValidateHedgeConfigsNormalizesBothSidesOfCollisionChecks` (`scheduler/hedge_config_test.go`).

### Verification

Both regression tests were run against the **unfixed** code first and fail there; the subtests that should still pass on old code (full-fill reduce, primary-flat self-heal, distinct coins, bare-primary-vs-suffixed-hedge) do pass, so the tests discriminate rather than failing broadly.

`gofmt` · `go vet ./...` · `go test ./...` (full suite) · `go test -race` on the hedge tests · `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — all green. `docs/ARCHITECTURE.md` updated so the partial-fill/proportional-basis rule is documented alongside the rest of the hedge mechanics.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (`github-actions[bot]` · 2026-07-25T04:02:15Z) in ce2ce4e3.

All four findings validated against current code and confirmed — nothing refuted. CI was green at review time (`go`, `python` both `pass`), so no CI findings.

### Fixed

1. **Stuck-CB recovery reconstructed the hedge from virtual state the fire cycle had already deleted** — confirmed end to end. On a CB fire whose HL fetch failed, `setHyperliquidCircuitBreakerPending` bails on the empty snapshot (`risk.go:947`) while `shouldForceCloseAllPositionsOnCircuitBreaker` still returns true for a sole owner (`risk.go:996`), so `forceCloseAllPositions` deletes **both** virtual legs — making `heldHedgeCoin` structurally incapable of naming the hedge on the recovery cycle.

   The hedge is now reconstructed from **config + the on-chain snapshot, exactly as the primary is** (`scheduler/hyperliquid_balance.go:2392`). New `hedgeIsInverseOfPrimaryOnChain` (`scheduler/hedge.go:762`) is the discriminator that survives the virtual delete: an auto-managed hedge is *always* inverse to its primary, so a same-side position on the sole-owned hedge coin provably was not opened by this mechanism and is refused with a CRITICAL log instead of liquidated. It cannot prove the inverse case *is* ours — no evidence surviving the delete can — but it rules out the whole class of same-side foreign positions, and both outcomes are logged. Safety posture otherwise matches the primary reconstruction, which already trusts the same sole-ownership guarantee (stronger here: `validateHedgeConfigs` rejects every collision).

   Covers all three **Must survive** cases in `TestCircuitBreakerFireWithFailedFetchThenRecoveryClosesBothLegs`, `TestLossStreakCircuitBreakerWithFailedFetchLeavesNoPending`, and `TestHedgeIsInverseOfPrimaryOnChain` — the first asserts the precondition that virtual state *cannot* name the hedge there, so the test fails if anyone reverts to reading it.

2. **Short-filled manual full hedge close was booked as full, deleting the leg while exposure remained** — confirmed. `IsFullClose` keyed off `fullClose` (the operator's primary-close intent), so a short fill still took the drain's `closedFull` branch and erased the `HedgeFor` stamp. Now keyed off the **actual drain** (`scheduler/manual_core.go:1564`), mirroring the primary path's `actualFullClose`.

   `TestShortFilledCoupledHedgeCloseStaysTracked` drives `forceCloseCoupledHedgeLeg` itself through a stub closer and asserts the flag it *queued* — asserting the formula in the test would not have caught this, since the defect was in that expression.

3. **Short-filled manual partial hedge reduce re-anchored to the primary quantity, stranding a permanent over-hedge** — confirmed; this is the previous round's defect re-introduced on the manual path, and (as the review notes) invisible to reconcile because virtual and on-chain both moved by the same filled quantity. New `hedgeBasisAfterPartialReduce(oldBasis, preQty, remainingQty)` (`scheduler/hedge.go:796`) re-anchors by the fraction of the leg that actually remains, applied at `scheduler/manual.go:617`.

   I used the held-quantity derivation you suggested rather than reusing `hedgeReducedBasis` directly, because the drain knows the before/after sizes but not the originally requested order size. `TestHeldQuantityAndFillRatioBasisRulesAgree` pins that the two derivations produce identical answers, so they can't drift apart. All three **Must survive** cases are in `TestManualDrainReAnchorsHedgeBasisProportionally`.

### Resolved judgment calls (was Recommended Optional)

4. **The reconcile comment promised re-convergence the delta math cannot deliver** — confirmed: `hedgeTargetDecision` keys only on `PrimaryQty - HedgeBasis` and never reads the hedge's own size, so an externally reduced leg with an unchanged primary read as in-sync forever.

   You offered "re-grow, or correct the comment." **I implemented re-grow**, because leaving it was incoherent with behavior already shipped: a **full** external close deletes the leg and `runHedgeSync` re-opens it from scratch next cycle. Closing 100% of the hedge got it rebuilt while closing 99% did not — so "don't re-grow" was never actually the design, just a gap.

   A downward resync now shrinks the basis by the same fraction the leg lost (`scheduler/hedge.go:1206`), letting the existing delta math add the shortfall back. Routing it through the **basis** rather than comparing held size to a target recomputed at the current mark is what preserves the qty-event invariant — the basis moves only when the exchange actually takes size, never on price, so this cannot degrade the hedge into a continuous rebalancer paying taker fees on noise. That directly answers your **must survive (b)**: it fights a deliberate operator hedge close no more than the already-accepted full-close path does, and the alert tells the operator to disable the hedge block if that was intentional.

   Deliberately **one-directional**: an upward resync leaves the basis untouched, because surplus this scheduler never opened is not its to trade away. Both alert texts now state what actually happens instead of overpromising.

   *Alternatives rejected:* correct the comment only — leaves the pair silently under-hedged and keeps the 99%/100% incoherence. Add a persisted `HedgeTargetQty` field and compare held against it — strictly more state and a migration for no behavioral gain over the basis route. Compare held against a mark-recomputed target — the one option that actually breaks the qty-event invariant.

   Covered by `TestExternallyReducedHedgeIsRegrown`, `TestExternallyINCREASEDHedgeIsLeftAlone`, and `TestPrimaryReduceAfterExternalHedgeReductionSizesOffTheShrunkBasis`.

### Verification

Every fix was reverted individually and its test confirmed **failing** on the unfixed code before restoring. `gofmt` · `go vet ./...` · `go test ./...` (full suite) · `go test -race` on the hedge/manual/CB tests · `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — all green. `docs/ARCHITECTURE.md` updated for the config-derived CB reconstruction, the actual-drain full-close rule, and the re-grow semantics.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (`github-actions[bot]` · 2026-07-25T04:54:12Z) in ffd47cc6.

Confirmed the finding against the code. CI was green at review time (`go`, `python` both `pass`), so no CI findings.

1. **Stuck-CB recovery left the hedge naked when the primary had gone flat on-chain during the same outage** — confirmed exactly as traced. `computeHyperliquidCircuitCloseQty` on the primary returns `!ok` with no on-chain position, and the loop `continue`d before ever reaching the hedge block, so the compound case enqueued nothing.

   **I implemented the alternative you named — enqueue the hedge on primary-reconstruction-equivalent trust — rather than the recommended alerts-only posture**, and folded your alerting improvements in on top rather than instead.

   The reconstruction now evaluates both legs independently (`scheduler/hyperliquid_balance.go:2415`). With the primary **live**, the inverse discriminator still applies and a same-side position is refused with a CRITICAL log **plus an owner DM** (previously log-only). With the primary **flat**, the discriminator cannot run and the hedge is closed on sole-ownership trust.

   Why not alerts-only: the trust being extended here is not *new*. The primary reconstruction already closes an on-chain position on a configured coin under a latched CB without proving ownership beyond sole-ownership — and the hedge coin's guarantee is strictly **stronger**, since `validateHedgeConfigs` rejects a hedge coin that is any strategy's primary or any other hedger's hedge, whereas a primary coin need only be unshared among live peers. Weighing realistic worst cases rather than averages: if the leg is ours — overwhelmingly the likely reading, since it is our declared coin, our breaker just fired, and our virtual row was deleted *by that fire* — alerts-only leaves naked inverse leveraged exposure running unbounded on a strategy the breaker just halted, the precise harm the CB exists to stop. If instead it is an operator's manual trade on a coin they declared machine-managed, the cost is one bounded, reversible, loudly-announced close. Unbounded loss outweighs bounded reversible loss, so the auto-protective mechanism should complete its job.

   Your alerting recommendations are implemented regardless: a dedicated CRITICAL owner DM names this specific case and tells the operator what to do if the leg was in fact theirs ("re-open it and remove `<coin>` from this strategy's hedge block"), and the `docs/ARCHITECTURE.md` safety claim now states the compound-outage semantics explicitly instead of implying blanket coverage.

   *Alternatives rejected:* alerts-only (your recommendation) — leaves the naked leg the CB was meant to flatten, and the generic reconcile alert never closes anything. Gate the close on some stronger ownership proof — none exists once the virtual row is deleted and the primary is flat; inventing one would be guessing. Re-derive ownership from `ClosedPositions`/the trade ledger — those are in-memory buffers subject to flush timing, so a restart between the fire and the recovery would silently reopen the same gap.

   Two related tightenings while I was in the path: a peer-shared primary is now skipped explicitly (`hyperliquid_balance.go:2369`) — that strategy routes to the operator-required path by design and must never receive a pending close — and owner DMs raised inside the locked reconstruction are buffered and drained after the unlock (#880).

   Regression coverage: `TestStuckCBClosesOrphanedHedgeWhenPrimaryWentFlatDuringTheOutage` (confirmed **failing** on the unfixed code), plus `TestStuckCBClosesBothLegsWhenPrimaryIsStillLive`, `TestStuckCBRefusesSameSideHedgeCoinPositionButStillClosesPrimary`, and `TestStuckCBWithNothingOnChainIsANoOp` for the neighbouring branches.

### Verification

`gofmt` · `go vet ./...` · `go test ./...` (full suite) · `go test -race` on the hedge/CB tests · `pytest shared_strategies/ shared_tools/ platforms/ backtest/` — all green.

Thanks for the explicit no-action list — I re-checked the `perpsPositionTradeType(nil)` branch and the coupled-force-close drain ordering and agree with both readings.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (`github-actions[bot]` · 2026-07-25T05:17:49Z, LGTM) in 62ba4c63.

### Fixed
1. **Hedge open path missing the minimum-notional floor (Recommended Optional)** — confirmed: the open branch (`scheduler/hedge.go:235`) checked only `hedgeQtyEpsilon` while add (`hedge.go:270`) and reduce (`hedge.go:315`) both apply `hedgeMinOrderNotionalUSD`. A sub-$10 open is now deferred (`hedgeActionNone`). The primary runs un-hedged; no unwind — a deferral is a sizing decision, and the decision recomputes each cycle so a scale-in above the floor fires the open. Regression tests: `TestHedgeTargetDecisionDefersOpenBelowMinNotional`, `TestHedgeTargetDecisionDeferredOpenFiresAfterGrowth`.

### Merge conflict resolution
1. Merged `origin/main` in 09167985. `CLAUDE.md`: kept main's condensed structure and re-added the `hedge.go` bullet. `discord_commands.go`: kept `hedgeStatusNote` plus main's new `dailyLossStatusNote` signature. `main.go`: kept both the #1159 `hedgeFreshExposureQty` tracking and main's #1416 ratchet-tighten pass-through to `scaleInResizeTrailingSLNow`.

Verification: `go build` + full `go test ./...` pass after the fix and the merge.

---

richkuo:
@claude sonnet review

PR #1406, opened 2026-07-25
feat(#1159): add per-strategy correlated hedge legs (phase 1, HL perps) [C90, Opus 5, high]

richkuo:
## Plain simple English

A strategy can now automatically open an offsetting trade on a related coin — for example, going long ETH also opens a short BTC — so it keeps its trading idea while carrying less overall market risk. The offsetting trade is fully managed: it opens, grows, shrinks, and closes together with the main one. If it ever cannot be opened, the main trade is closed instead of running exposed.

## Summary

Opt-in `hedge` block on a Hyperliquid perps strategy opens an auto-managed **inverse** position on a second coin, strictly coupled to the primary. Default behavior is unchanged for every strategy that does not opt in.

```json
"hedge": { "enabled": true, "symbol": "BTC/USDC:USDC", "ratio": 1.0, "margin_mode": "cross", "leverage": 3 }
```

### Core design

Hedge management is a **state-derived per-cycle reconciler**, not scattered per-event mirror hooks. `hedgeTargetDecision` (pure, no locks/IO/clock) computes the target from `(current primary position, current hedge position, persisted qty watermark)`; `runHedgeSync` converges the exchange to it. Every primary lifecycle event therefore mirrors automatically — fresh open, scale-in add, evaluator partial/full close, on-chain SL/TP fill booked by reconcile, ratchet close, kill switch, circuit breaker, manual force-close, external close, and a crash between the two legs. **One place decides a hedge order; one place books a hedge fill.**

It runs twice per cycle, idempotently: inline at the tail of the HL perps dispatch (so a fresh open is hedged in the same cycle and a failure unwinds immediately, not after a full interval of naked exposure), and via a post-loop sweep over every hedge-relevant strategy the dispatch did not reach.

### Safety invariants

1. **Qty-event mirroring, never price mirroring.** The target keys on `Position.HedgePrimaryQtyBasis`, so mark drift can never re-trade the hedge — only a real change in primary size or side does.
2. **Fill-confirmed state mutation only.** A failed hedge order mutates nothing, mirroring `runHyperliquidExecuteOrder`'s `ok2=false` contract, so *"state disagrees with target" IS the retry mechanism* — no pending table, no latch, restart-safe. A partial fill advances the watermark only proportionally, so the remainder is retried rather than assumed hedged.
3. **Fail closed (constraint 4).** A hedge that cannot be opened *or grown* — rejected, unconfirmed, or unsizeable for want of a mark — closes the **unhedged slice** of the primary reduce-only plus a CRITICAL owner DM. Scoping to the unhedged slice makes one rule cover both shapes exactly: a failed `open` has basis 0 so the whole primary unwinds; a failed `add` unwinds only the add leg and the originally-hedged position rides on. A failed reduce/close is *not* escalated — over-hedged is risk-reducing — and the reduce-only retry is bounded. A consecutive-failure counter holds new entries after 3 failures so a bad hedge venue cannot turn every signal into a fee-burning open→unwind loop.
4. **Sole ownership by construction.** Validation rejects a hedge coin equal to the strategy's own coin, ANY configured strategy's coin (perps *and* manual, live *and* paper), or another strategy's hedge coin — which is what lets every shared-coin mechanism keyed on `hyperliquidConfiguredCoin` stay correct untouched. Hedge closes are nonetheless **always sized**, never `market_close(sz=None)`, so a position an operator opened by hand is never liquidated.
5. **Ownership from persisted metadata only (constraint 5).** `Position.HedgeFor` is written once at fill time and read back from SQLite; hedge legs are never re-derived from a coin→config lookup, so a foreign position on a declared hedge coin is never adopted (alerted instead).

### Accounting (requirement 6)

Hedge trades use `trade_type="hedge"`, **excluded** from lifetime `#T` and W/L — one thesis is not two round-trips, and the hedge close loses by construction when the primary wins — but **deliberately included** in `tradeLedgerDeltaSQL`, which is what books hedge PnL and fees to the owning strategy. All perps close-booking sites route through `recordTradeResultForPosition` → `RecordHedgeTradeResult`, which updates `DailyPnL` but never `ConsecutiveLosses` (counting it would double-count one thesis and mis-fire the loss-streak circuit breaker). Trade diagnostics skip hedge closes.

### Extensions to existing machinery

Marks (`collectPerpsMarkSymbols`); reconcile (hedge-coin pass over all strategies, plus a #822 orphan-close exemption — an inverse hedge always "conflicts" with the strategy's direction); kill switch (held-hedge-coin roster + fill booking applied first and unconditionally); circuit breaker (hedge enqueued independently, **including the previously-early-returning shared-primary case**, where `forceCloseAllPositions` would otherwise zero the leg virtually and strand real exposure; drain now strips only the shared primary instead of clearing the whole pending); shared-wallet funding attribution; hot reload (blocked while any position incl. a residual hedge leg is open, reload-time collision re-check); startup drift check (warn-and-continue, since refusing to boot would leave live positions unmanaged); inspect / `/status` / Discord surfaces; manual force-close note + queued-action refusal on hedge coins; loud backtester reject.

## Verification

- `go build` clean, `gofmt` clean, `go test ./...` — all green.
- Full pytest suite (`shared_strategies/ shared_tools/ platforms/ backtest/`) — 2872 passed, 1 skipped.
- ~70 new Go tests across `hedge_test.go` (decision core, sizing, side mapping, skip-reason mirror, dust deferral, fail-closed blocking, failure tracker, surfaces), `hedge_config_test.go` (full collision matrix, vocabulary/bounds, nested unknown-key guard, hot-reload block/allow/add/remove, reload-introduced collision), and `hedge_sync_test.go` (paper end-to-end open/add/reduce/close/stale-unwind, live path with injected executors, fail-closed unwind shapes, partial fills, accounting, marks, reconcile, kill switch, CB, persistence round-trip, lifetime-stats exclusion, sweep).
- Four regression tests were **confirmed red** against the unfixed code before the fix: the CB drain shared-primary strip, the #822 orphan exemption, the kill-switch hedge roster, and the direction-check exemption.
- End-to-end smoke via the built binary: a hedge-enabled config loads and renders in `inspect`; a same-coin hedge and a nested typo (`ration`) both reject at load with actionable messages.
- Go tests spawn no Python — the live order path is exercised through injected seams.

## Not covered (phase 1, by design)

Hedge coins overlapping peer strategies, and backtester PnL/fee/slippage parity — both tracked as follow-ups in the issue, both currently rejected loudly rather than silently mishandled.

Closes #1159

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T04:20:46Z) in e52fff33.

### Fixed

1. **Failed side-mismatch hedge close misclassified as a benign over-hedge (Needs Fixing)** — confirmed exactly as described. Escalation no longer keys on the action *kind*; `hedgeAction` gained `FailureLeavesPrimaryExposed` + `HedgedPrimaryQtyOnFailure` (`scheduler/hedge.go:390-420`), set per-branch, and `hedgeHandleOrderFailure` routes on that field (`scheduler/hedge_sync.go:588-600`). The side-mismatch close sets exposed=true with hedged=**0** (`scheduler/hedge.go:452-465`) — the stale basis still reads as the pre-flip quantity while covering none of the new side, so the whole primary unwinds. `unwindUnhedgedPrimary` now takes the watermark from the decision rather than re-deriving it from the snapshot (`scheduler/hedge_sync.go:660-673`), which is the part that was structurally wrong for a flip. The failure also now counts toward the entry hold, which the close path previously skipped.

2. **Entry hold promised a recovery path it blocked (Recommended Optional 1)** — confirmed: `clear`-on-success is unreachable while the hold suppresses the very opens that would clear it, so a flat strategy stayed frozen until restart. The hold is now time-bounded (`hedgeEntryHoldCooldown` = 1h, `scheduler/hedge.go:76-87`); `hedgeFailureTracker` expires lapsed holds lazily with an injectable clock (`scheduler/hedge.go:745-800`), granting a fresh retry budget and re-arming the DM for the next episode. A successful hedge still clears it immediately — hedge sync is not gated by the hold, so an open primary keeps getting attempts. The DM now states that path (`hedgeEntryHoldMessage`, `scheduler/hedge.go:824-831`).

3. **Flip left the primary unhedged for up to a full interval (Recommended Optional 2)** — confirmed. A close that is half of a replace sets `ReopenAfterClose` (`scheduler/hedge.go:398-411`), and `runHedgeSync` loops to perform the reopen in the same cycle (`scheduler/hedge_sync.go:60-72`). Applied to hedge-symbol changes too, which had the identical two-cycle gap. Deliberately **narrow**: an ordinary open/add/reduce does not loop — my first cut looped on any successful action and re-issued orders on partial fills, tripling order count on a thin book to chase a gap the next cycle closes for free. `hedgeSyncMaxStepsPerCycle` is the hard backstop.

### Verification of the reviewer's "Must survive" cases

All added as regression tests, and the four that exercise the blocking finding were **confirmed red** against the pre-fix code before the fix landed:

- (1a/1b) both flip directions with a repeatedly-failing hedge close → whole primary fail-closed — `TestHedgeFlipCloseFailureUnwindsPrimary`
- (1c) flip immediately after a scale-in (residual leg larger than the pre-flip hedge) → whole primary unwound, stale basis not credited — `TestHedgeFlipAfterScaleInUnwindsWholePrimaryNotJustTheDelta`
- (1d) genuine proportional reduce failure on a still-inverse hedge → **not** unwound, not counted — `TestHedgeInverseReduceFailureStillDoesNotUnwind`; plus `TestHedgeCloseFailureWithFlatPrimaryDoesNotEscalate` for the other non-escalating case
- (2a/2b/2c) hold expires and re-arms, success during hold clears immediately, DM wording — `TestHedgeEntryHoldExpiresAndReArms`, `TestHedgeSuccessDuringHoldClearsItImmediately`, `TestHedgeEntryHoldMessageStatesTheRealRecoveryPath`
- (3a/3b/3c) same-cycle replace, reopen failure still fail-closes, rapid double-flip — `TestHedgeFlipReplacesHedgeInOneCycle`, `TestHedgeFlipReopenFailureStillFailsClosed`, `TestHedgeRapidDoubleFlipConvergesEachCycle`; plus `TestHedgePartialFillDoesNotReorderWithinTheCycle` guarding the loop's narrowness

`go build`, `go test ./...`, and the full pytest suite (2872 passed, 1 skipped) are green. Both CI checks were already passing before this push, so there were no CI findings. `CLAUDE.md` and `docs/ARCHITECTURE.md` updated to match the new escalation rule, the bounded hold, and the same-cycle replace.

---

richkuo:
@claude review

richkuo:

PR #1407, opened 2026-07-25
feat(#1159): add per-strategy correlated hedge legs (phase 1, HL perps) [C90, Opus 5, medium]

richkuo:
## Plain simple English

A strategy can now automatically open a matching offsetting trade on a related coin. If it buys ETH, it can at the same time sell an equivalent amount of BTC, cutting how much the position swings with the overall market. The offsetting trade grows, shrinks, and closes with the main one. If it can't be placed, the main trade is closed immediately rather than left exposed.

## Summary

Adds an opt-in per-strategy `hedge` block for Hyperliquid perps (paper and live). A long ETH primary opens a short BTC hedge sized on notional exposure; the hedge then tracks every primary quantity event for the life of the position.

**Design: a state-derived reconciler, not per-event mirror hooks.** `hedgeTargetDecision` is a pure function that diffs the *current* primary quantity against a persisted watermark (`Position.HedgePrimaryQtyBasis`). `runHedgeSync` converges the hedge leg to that target on every HL perps dispatch cycle, including the `Signal==0` manage path. Fresh open, scale-in add, evaluator partial/full close, on-chain SL/TP fill detected by reconcile, ratchet close, orphan close, and external close are therefore all mirrored with no per-path hook. Only the two paths that bypass the dispatch loop — the portfolio kill switch and the per-strategy circuit-breaker drain — get explicit extensions. Keying on quantity (not price) is what stops mark drift from re-trading the hedge.

**Safety.**

- **Fail-closed open** (constraint 4): primary filled + hedge open failed on the *same* cycle immediately closes the primary reduce-only — always **sized**, never a bare `market_close`, because the primary coin may have shared-coin peers — cancelling its just-armed protection OIDs, with a CRITICAL owner DM + all-channels alert. A failed unwind latches nothing: the state-derived sync retries the hedge or the unwind next cycle, restart-safe with zero new persisted flags.
- **Later-cycle failures alert and retry only.** A transient RPC error must never liquidate an already-hedged position.
- **Fill-confirmed state mutation only**, mirroring `runHyperliquidExecuteOrder`'s `ok2=false` → no state update; plus a `hedgeOrderSkipReason` re-read immediately before every spawn.
- **Collision validation is load-bearing** (constraint 2): a hedge coin may not equal the strategy's own coin, any configured strategy's coin (perps *and* manual, live *and* paper), or another strategy's hedge coin. That exclusivity is exactly what keeps every `hyperliquidConfiguredCoin`-derived shared-coin mechanism correct while blind to hedge coins. `direction: "both"` + hedge is rejected (a mid-flight flip would change the hedge side). `side` accepts only `"inverse"`.
- **No hedge protection**: the hedge symbol never enters the protection sync, the trailing walker, `post_tp_sl`, the regime store, the LLM analysis queue, or any check script.

**Accounting.** Hedge rows carry `trade_type='hedge'`, excluded from the W/L round-trip stats but deliberately **not** from ledger sums — `tradeLedgerDeltaSQL`/`tradeNetPnLSQL` ignore `trade_type`, which is what books hedge PnL and fees into the owning strategy's ledger (requirement 6). `RecordHedgeTradeResult` updates `DailyPnL` (a hedge loss is real money) but never `ConsecutiveLosses` — a hedge loses by construction when the primary wins, so counting it would double-count one thesis and mis-fire the CB loss-streak arm. `captureTradeDiagnostics` is skipped for hedge legs.

**Cross-subsystem extensions.** Perps mark symbols, shared-wallet uPnL + funding attribution, the reconcile fill resolver's coin universe, the reconciler itself (sole-owned, never adopts a foreign position on a declared hedge coin), the kill switch (`heldHedgeCoins`, built from *held* legs so a declared-but-flat coin is never liquidated), the CB pending close and its stuck-CB recovery, and `ValidatePerpsDirectionConfig` / the `#822` orphan check (an inverse leg is opposite the configured direction by design) all learn about hedge legs.

**Hot reload + restart.** `sc.Hedge` is hot-reloadable when flat and blocked by `validateHotReloadStateCompatible` while any position is open — including a residual hedge leg with the primary already flat, since the hedge lives in the same `Positions` map. Reload re-runs `validateHedgeConfigs` so a collision can't be introduced. A config edit + process restart bypasses all of that, so `validateHedgeStateConsistency` warns loudly at startup and leaves the leg **frozen** — a config warning must never close live exposure.

**Backtest** rejects an *enabled* hedge block loudly (a disabled block still backtests): the single-instrument simulator would silently drop the hedge's PnL/fees/funding and report the primary's standalone edge as the hedged strategy's.

**Operator surfaces.** Startup summary `hedge=BTC×1.00(inverse,cross,3x)`, `inspect` text + `--json`, `/status` + dashboard `HedgeStatus`, Discord `/status` note, and an explicit note on manual `force-close`.

**Schema.** Two additive `positions` columns (`hedge_for`, `hedge_primary_qty_basis`) via idempotent migrations. No `config_version` bump — the block is purely additive optional.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean.
- `go test ./...` — full scheduler suite passes.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 2872 passed.
- End-to-end binary smoke: a valid hedged config loads and renders the hedge in `inspect` text + `--json`; a same-coin hedge is rejected with the collision message; a typo'd `"ration"` key is rejected by the nested unknown-key guard.
- New tests (`scheduler/hedge_test.go`, `backtest/test_hedge_reject.py`) cover every acceptance criterion: inverse-side mapping, ratio sizing, add/reduce/close quantity events, the steady-state no-op under moved marks, dust deferral without advancing the watermark, fail-closed blocking on unusable inputs, wrong-side unwind, the skip-reason mirror, owning-strategy ledger attribution, the loss-streak exemption (with an ordinary-perps control), failed hedge open → primary unwound, failed *primary* open → no hedge order, later-cycle failure → no unwind, partial close → proportional reduce + idempotence, paper virtual-only mode, the full config collision matrix, the nested unknown-key guard, hot-reload block while open (with the flat inverse), startup consistency warnings, direction-validator exemption, kill-switch held-hedge closing (with a foreign-coin control), CB pending including the hedge leg (with a flat-hedge control), mark symbols, shared-wallet books, and the operator surfaces.

Closes #1159

---
Created with LLM: Opus 5 | medium | Harness: work-on-issue


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T05:51:16Z) in 6d745264.

### Fixed

1. **Hedge-add failure during a scale-in liquidated the whole already-hedged primary** — confirmed exactly as described, on both escalation paths. Fixed on both sides so neither alone is load-bearing:
   - `scheduler/main.go:2623` now sets `PrimaryOpenedFromFlat` (renamed from `FreshOpen`) only for a genuine flat→open: `openTrade != nil && scaleInAddQty == 0`. Both a fresh open and a scale-in add produce an `openTrade`, so `scaleInAddQty` is what distinguishes them.
   - `scheduler/hedge.go:562` introduces `hedgeFailureMayUnwindPrimary(openedFromFlat, snap)` and gates **both** the error path (`hedge.go:646`) and the `Blocked` path (`hedge.go:596`) on it. It requires all three of: the primary went flat→open this cycle, the primary is held, and **no hedge leg is currently held**.

   Two deliberate design points, since the review offered `action.Kind == hedgeActionOpen` as an alternative:
   - **Why the hedge-flat clause rather than `Kind`:** the `Blocked` branch carries `Kind == hedgeActionNone` (the decision could not be computed at all), so a `Kind` check there would have silently disabled the fail-closed unwind on a genuine fresh open — regressing must-survive case (3). Hedge-flat is the property that actually encodes "this exposure is uncovered", and it holds on both branches.
   - **Why keep it as defense in depth rather than fixing only main.go:** the hedge-flat clause makes an already-hedged position *structurally* un-unwindable by a hedge error, even if a future caller mis-computes the flag. It also fixes a second case the review didn't name: a long-running primary whose hedge was closed **externally** produces `hedgeActionOpen` with the hedge flat, so under the old `Kind`-based logic a failed re-open would have liquidated it — contradicting the documented "re-open next cycle + DM" recovery.

   Tests (`scheduler/hedge_test.go:415-560`) cover all three must-survive cases plus the extra one: scale-in add failure on the error path *and* the `Blocked` path, each run twice — once with the flag correctly `false` and once with it wrongly `true` — asserting the primary, the hedge leg, and the unadvanced watermark are all untouched; a genuine flat→open still unwinds on both paths; an externally-closed-hedge re-open failure does not unwind; plus `hedgeFailureMayUnwindPrimary`'s own truth table. The two cases that reproduce the production scale-in path were verified **failing** against the unfixed code before the fix was restored.

2. **`hedgeCoinsForStrategies` was dead code with a misleading comment** — confirmed: the definition (`hedge.go:734`) was its only occurrence in the tree, and both named consumers call `hedgeCoin(sc)` directly (`risk.go:76`, `hyperliquid_fills.go:312`). Removed rather than wired up: both consumers need per-strategy iteration anyway (the fill resolver needs `sc` for its state lookup), so a coin-set builder doesn't fit either loop and `hedgeCoin` already is the single source of truth. Hedge-coin visibility to those subsystems is unchanged — `TestCollectPerpsMarkSymbols_IncludesHedgeCoins` still passes.

Docs updated to match the corrected semantics: `CLAUDE.md` and the fail-closed matrix in `docs/ARCHITECTURE.md` now state the unwind gate precisely and call out the scale-in and external-close cases explicitly.

### Verification

`go build ./...`, `go vet ./...`, `gofmt -l` clean; `go test ./...` green; `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 2872 passed.

---
Created with LLM: Opus 5 | medium | Harness: Claude Code

richkuo:
@claude review

richkuo:

PR #1418, opened 2026-08-05
fix(#1416): tighten resting SL same cycle after ratchet scale-out [C85, Grok 4.5, high]

### LINKED ISSUE #1416: [C85] Bug: trailing TP ratchet tightens the multiplier but the resting stop-loss keeps the old wider trigger for the rest of the cycle
**Complexity: 85/100** — scope: small (HL execute path in `main.go` + tests); risk: high (live on-chain stop-loss protection — money/auto-protective path); uncertainty: low (mechanism traced end-to-end) · fableplan: no

### Summary

`applyTrailingTPRatchetToPosition` stamps a tighter `pos.PostTPTrailingATRMult` and emits a "trail tightened to X×ATR" log + alert, but on a cycle where the close evaluator fires a partial close, the on-chain stop-loss trigger is **not** recomputed against the new multiplier in the same cycle. The next cycle's trailing walker fixes it (`effectiveTrailingStopPct` reads the new mult, and the tighter trigger passes the favorable/min-move check in `computeTrailingStopUpdateInternal`), but until then the position rests at the OLD wider stop distance — the ratchet promise ("tightened to X×ATR") is broken for up to one full cycle interval.

### Scope

**HL perps only, live and paper.** Not other platforms and not HL manual:

- The validator rejects `trailing_tp_ratchet*` close evaluators on anything but HL perps/manual (`scheduler/trailing_tp_ratchet.go:320`), and `effectiveTrailingStopPct` returns 0 off-Hyperliquid — OKX/spot cannot be affected.
- The HL manual manage path runs ratchet → re-read position → trailing walker back-to-back with no signal gate (`scheduler/main.go:2839`–`2855`), so manual self-heals in the same cycle.

### Concrete reproduction

`hl-vwap-eth-60` (live, HL perps, ETH, 1h). `vwap_reversion` opening + `trailing_tp_ratchet_regime` closing (default `ranging_volatile` ladder).

Position state immediately after open (rowid 666, 2026-08-05 00:03:19Z):
- `avg_cost = 1868.80`, `entry_atr = 10.498943`, `stop_loss_atr_mult = 2.5`
- Fixed SL trigger (entry-time): `1868.80 × (1 − 2.5·10.498943/1868.80/100) = 1842.55` ✓

**15:04:25** — first trailing SL update (mark 1878.75, first favorable print):
- `pct = 2.5 × 10.4989 / 1868.80 × 100 = 1.4044%`
- `new trigger = 1878.75 × (1 − 0.014044) = 1852.40` ✓
- OID 510296624630 canceled, OID 510751808470 placed @ $1852.40.

**16:04:33** — cycle ran, no trailing update (mark 1875.75 < hwm 1878.75; trailing walker no-ops correctly).

**17:04:42** — ratchet fires (mark 1895.95, atrProfit = (1895.95−1868.80)/10.499 = 2.585 ≥ 2.0):
- Tier 1 cleared (close 80% of 0.535 = 0.428 @ 1895.70, PnL $11.51)
- `pos.PostTPTrailingATRMult = 0.75` (was 2.5)
- `pos.SLAdjustedTiersProcessed = 2`
- Log: `trailing_tp_ratchet: ETH tier 1 cleared — trail tightened to 0.75×ATR (from 2.5×ATR)`
- BUT the resting SL on HL is **still OID 510751808470 @ $1852.40** — it was never touched.

Intended trigger at this point (the value the alert said the SL was tightened to):
- `pct = 0.75 × 10.4989 / 1868.80 × 100 = 0.4213%`
- `intended trigger = 1895.70 × (1 − 0.004213) = 1887.71`

Actual: **$1852.40** vs intended **$1887.71** — a **$35.31 (~1.89%)** wider-than-promised downside window on the residual 0.107 ETH.

Eventually the position was signal-closed at 18:04:52 @ $1916.20 (close_reason=`signal`, PnL $4.98) before any Signal==0 cycle could re-run the walker — so the under-protection never self-healed for this position.

### Root cause

Two interacting facts:

1. **`applyTrailingTPRatchetToPosition` only mutates `PostTPTrailingATRMult` and `SLAdjustedTiersProcessed`** (`scheduler/trailing_tp_ratchet.go:634`–`642`). It never touches `StopLossHighWaterPx` or `StopLossTriggerPx`. The "intended" trigger is computed only for the alert payload (`buildRatchetTriggerAlert` → `IntendedSLTriggerPx`), not for the on-chain order.

2. **Cycle ordering + signal gating.** The trailing walker blocks run *before* execution and are gated on `result.Signal == 0` (`scheduler/main.go:2471` live, `scheduler/main.go:2444` paper). When the close evaluator emits a partial close (`close_fraction > 0`, non-zero Signal), those blocks are skipped, execution runs, and the ratchet fires *after* the fill inside `executeHyperliquidResultDeferredOpen` (`scheduler/main.go:4153`). Nothing after that point recomputes or replaces the resting SL in the same cycle.

On a Signal==0 cycle the defect does not occur: the ratchet runs first (`scheduler/main.go:2423`), the snapshot is re-read, and the walker immediately re-derives the trigger from the new mult.

### Suggested fix

**After execute, when the ratchet returned a tighten alert, run the trailing walker once more** — mirroring the existing manual sequence at `scheduler/main.go:2839`–`2855`:

- Re-snapshot the position under RLock (post-partial-close residual qty, new `PostTPTrailingATRMult`).
- Live: call `runHyperliquidTrailingStopUpdate` + `applyTrailingStopUpdateResult` **outside `mu`** (subprocess contract — no Python calls under the state lock). Sizing via `hlSLEffectiveQty` as usual; the walker's existing defer-and-retry alerting covers a failed replace.
- Paper: recompute the synthetic trigger via the same `computeTrailingStopUpdateInternal` path.

The walker already handles all the hard parts: min-move threshold, cancel+replace, immediate-fill booking, cancel-without-rest cleanup.

**Rejected alternative** — recomputing `StopLossTriggerPx` inside `applyTrailingTPRatchetToPosition`: that function runs under `mu.Lock` at all three call sites, and the cancel+replace is a Python subprocess, which must never run under `mu`. Stamping the intended trigger onto the position before the on-chain order actually moves also desyncs the virtual trigger from the resting OID and can make the walker's favorable/min-move comparison no-op the real replacement.

### Workaround

For operators: the next Signal==0 cycle naturally tightens the trigger as the trailing walker sees the new mult. The under-protection window is bounded by the strategy `interval` — an hour for 1h strategies, worse for 4h/1d (and unbounded if every subsequent cycle emits a close signal, as in the repro).

If you can't wait, cancel the resting SL and let the next cycle re-place it at the tighter trigger (or place a new SL manually at `intended_trigger = hwm × (1 − newMult × entry_atr / avg_cost)`).

### Evidence

- DB rows: `positions` (strategy_id=`hl-vwap-eth-60`), `trades` rowid 666 (open), 684 (17:04 partial close), `closed_positions` 1 row for 2026-08-05T18:04:52Z (signal close).
- Logs: `journalctl -u go-trader-live --since "2026-08-05 15:00"` shows the 15:04:25 trailing update, the 17:04:42 partial close + ratchet, and the 18:04:52 signal close.
- Code: `scheduler/trailing_tp_ratchet.go:597` (`applyTrailingTPRatchetToPosition`), `scheduler/main.go:2423` / `2839` / `4153` (ratchet call sites), `scheduler/main.go:2471` / `2444` (trailing walker `Signal == 0` gates, live/paper), `scheduler/hyperliquid_trailing_stop.go:675` (`runHyperliquidTrailingStopUpdate`).

---
Created with LLM: MiniMax-M3 | unknown | Harness: unknown

richkuo:
## Plain simple English

When a take-profit step sells part of a position and tightens the trailing stop, the exchange stop now moves to that tighter level on the same cycle. Before, the old wider stop could sit for a full cycle (or longer).

## Summary

- After execute stamps a ratchet tighten alert, call `runTrailingStopUpdateAfterRatchetTighten` (live cancel+replace + paper virtual trigger) **outside** the `execResult != nil` gate so paper is covered.
- Reuses existing walker primitives (`runHyperliquidTrailingStopUpdate` / `runHyperliquidTrailingStopPaper`); subprocess stays outside `mu`.
- Closes the #1416 under-protection window on scale-out tiers (`close_fraction > 0`); trail-only tiers already self-healed on the Signal==0 manage path.

Closes #1416

## Verification

- Red→green: stubbed helper → live/paper tests failed; restored helper → passed.
- `go test -C scheduler -count=1 -run 'TestRunTrailingStopUpdateAfterRatchetTighten|TestArmTrailingStopAtOpenNow|TestApplyTrailingTPRatchet|TestEffectiveTrailingStopPct_HonorsPostTP' .`
- `go build -C scheduler -o /dev/null .`

---
LLM: Grok 4.5 | high | Harness: Cursor

### DIFF
--- a/scheduler/hyperliquid_open_trailing.go
+++ b/scheduler/hyperliquid_open_trailing.go
@@ -87,7 +87,7 @@ func armTrailingStopAtOpenNow(
 
 	// currentTrigger==0 / currentOID==0: the primitive computes the initial
 	// trigger and rests a fresh SL (no forceResize needed).
-	newHighWater, slUpdate, updateConfirmed := runHyperliquidTrailingStopUpdate(sc, symbol, side, slEffectiveQty, &posSnap, mark, 0, 0, 0, false, notifier, logger)
+	newHighWater, slUpdate, updateConfirmed := runHyperliquidTrailingStopUpdate(sc, symbol, side, slEffectiveQty, &posSnap, mark, 0, 0, 0, trailingReplacePolicy{}, notifier, logger)
 	mu.Lock()
 	defer mu.Unlock()
 	if immediateFill, fillPx := applyTrailingStopUpdateResult(stratState, symbol, side, 0, newHighWater, updateConfirmed, slUpdate, logger); immediateFill {
--- a/scheduler/hyperliquid_stop_loss_test.go
+++ b/scheduler/hyperliquid_stop_loss_test.go
@@ -258,7 +258,7 @@ func TestRunHyperliquidTrailingStopUpdate_CancelThenPlaceArgs(t *testing.T) {
 	logger := silentStrategyLogger("hl-test")
 	defer logger.Close()
 
-	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, false, nil, logger)
+	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, trailingReplacePolicy{}, nil, logger)
 	if !ok {
 		t.Fatalf("runHyperliquidTrailingStopUpdate returned ok=false")
 	}
@@ -306,12 +306,12 @@ func TestRunHyperliquidTrailingStopUpdate_RatchetFallbackNormalizeWidensOnce(t *
 				return &HyperliquidStopLossUpdateResult{StopLossOID: 222, StopLossTriggerPx: triggerPx}, "", nil
 			}
 
-			_, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", tc.side, 0.5, &Position{AvgCost: 100}, 100, 100, tc.currentTrigger, 111, false, nil, logger)
+			_, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", tc.side, 0.5, &Position{AvgCost: 100}, 100, 100, tc.currentTrigger, 111, trailingReplacePolicy{}, nil, logger)
 			if !ok || result != nil || called {
 				t.Fatalf("without marker expected no replace: ok=%v result=%+v called=%v", ok, result, called)
 			}
 
-			_, result, ok = runHyperliquidTrailingStopUpdate(sc, "ETH", tc.side, 0.5, &Position{AvgCost: 100, RatchetFallbackNormalizePending: true}, 100, 100, tc.currentTrigger, 111, false, nil, logger)
+			_, result, ok = runHyperliquidTrailingStopUpdate(sc, "ETH", tc.side, 0.5, &Position{AvgCost: 100, RatchetFallbackNormalizePending: true}, 100, 100, tc.currentTrigger, 111, trailingReplacePolicy{}, nil, logger)
 			if !ok || result == nil || !called {
 				t.Fatalf("with marker expected widen replace: ok=%v result=%+v called=%v", ok, result, called)
 			}
@@ -343,7 +343,7 @@ func TestRunHyperliquidTrailingStopUpdate_DefersOnCancelFailure(t *testing.T) {
 		ownerID:  "owner",
 	})
 
-	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, false, notifier, logger)
+	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, trailingReplacePolicy{}, notifier, logger)
 	if ok || result == nil {
 		t.Fatalf("runHyperliquidTrailingStopUpdate = (%+v, %v), want deferred result", result, ok)
 	}
@@ -374,7 +374,7 @@ func TestRunHyperliquidTrailingStopUpdate_DefersOnOpenOrderCheckFailure(t *testi
 	logger := silentStrategyLogger("hl-test")
 	defer logger.Close()
 
-	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, false, nil, logger)
+	newHighWater, result, ok := runHyperliquidTrailingStopUpdate(sc, "ETH", "long", 0.5, &Position{AvgCost: 100}, 110, 100, 97, 111, trailingReplacePolicy{}, nil, logger)
 	if ok || result == nil {
 		t.Fatalf("runHyperliquidTrailingStopUpdate = (%+v, %v), want deferred result", result, ok)
 	}
@@ -1849,7 +1849,7 @@ func TestRunHyperliquidTrailingStopPaper(t *testing.T) {
 	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			gotHW, gotTrig, gotBreach, gotPx := runHyperliquidTrailingStopPaper(c.sc, c.side, c.pos, c.mark, c.highWater, c.currentTrigger)
+			gotHW, gotTrig, gotBreach, gotPx := runHyperliquidTrailingStopPaper(c.sc, c.side, c.pos, c.mark, c.highWater, c.currentTrigger, trailingReplacePolicy{})
 			if floatDiff(gotHW, c.want.newHighWater) > 1e-9 ||
 				floatDiff(gotTrig, c.want.newTrigger) > 1e-9 ||
 				gotBreach != c.want.breach ||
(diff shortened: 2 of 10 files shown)


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-07T17:41:48Z) in 9b987f71.

### Fixed

1. **Same-cycle walker inherited the 0.5% min-move debounce** — confirmed at `scheduler/hyperliquid_trailing_stop.go:534-538`: the debounce measures the trigger shift, but a ratchet tier shrinks the trail *distance*, so a tier step of `Δmult × EntryATR / anchor` lands under the default on any coin whose ATR is a small fraction of price. Added `trailingReplacePolicy` (`scheduler/hyperliquid_trailing_stop.go:483-508`) carrying the existing #873 `forceResize` plus a new `ratchetTightened` bypass; the bypass drops only the debounce and leaves the favorable-direction gate intact, so a forced tighten can never widen the stop on either side (`scheduler/hyperliquid_trailing_stop.go:562-566`).

2. **Widened the fix beyond the new helper** — the same defect exists at the two other ratchet-then-walker dispatches, which the reviewer's invariant also covers. All three now pass the bypass, each gated on a non-nil tighten alert: the #1416 execute helper (`scheduler/trailing_tp_ratchet_sl.go:76,88`), the perps `Signal==0` manage path for live and paper (`scheduler/main.go:2429,2461,2492`), and the manual manage path (`scheduler/main.go:2864,2884`). A tier alerts at most once (`SLAdjustedTiersProcessed` advances past it), so no cancel+replace churn is possible.

3. **Test coverage gaps** — `scheduler/trailing_tp_ratchet_sl_test.go:243-448` adds: a table case at the shipped 0.5% default with the issue's real geometry (`anchor 1868.80`, `EntryATR 10.4989`, `2.5 → 2.0`, ~0.28% shift) for **both** long and short, asserting the tighten lands and the trigger moves toward the mark; a case asserting a watermark-only cycle with a zero policy stays fully debounced; a case asserting the bypass never widens on either side; a two-tier-in-one-cycle case asserting exactly one replace at the tightest multiplier. `TestRunTrailingStopUpdateAfterRatchetTighten_PaperUpdatesVirtualTrigger` now stubs `runHyperliquidUpdateStopLossFunc` and asserts it is never called (`scheduler/trailing_tp_ratchet_sl_test.go:77-120`).

Regression check: with the bypass branch disabled, both new default-min-move cases fail with `cancel+replace calls = 0, want 1`.

Verification: `gofmt` clean, `go vet ./...`, `go test ./...`, and `go test -race ./...` all pass.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-07T18:37:15Z, `LGTM` + 1 optional) in 4e3dd599.

### Fixed

1. **Scale-in add branch never applied the ratchet** — confirmed: `executeHyperliquidScaleInDeferredOpen` books the add with no ratchet call, and `scheduler/main.go:2624-2625` left `ratchetAlert` nil, so the same-cycle helper was skipped. `config.go:1967` requires a resizable or trailing stop for live scale-in, so `allow_scale_in` + `trailing_tp_ratchet` is reachable. The add branch now applies `applyTrailingTPRatchetToPosition` after the add lands and returns the alert (`scheduler/scale_in.go:417-424,444`), mirroring `executeHyperliquidResultDeferredOpen`. Tier geometry is unaffected by the blend, since #873 pins the ladder to `RiskAnchorPrice`.

2. **One replace, post-add size, new multiplier** — `scaleInResizeTrailingSLNow` takes the tighten signal and combines `forceResize` with the min-move bypass in a single walker pass (`scheduler/scale_in.go:40,71`; wired at `scheduler/main.go:2678`). Exactly one owner moves the stop per execute cycle: a live add routes through `scaleInResizeTrailingSLNow`, which alone knows the grown on-chain size, while every other case keeps the #1416 helper (`scheduler/main.go:2639-2652`). Running both would first rest an under-sized stop against the stale pre-add snapshot and then replace it again.

3. **Closed the tail of the same gap, beyond the finding** — a forced resize previously always re-armed at the *existing* trigger (`scheduler/hyperliquid_trailing_stop.go:731-750`). That meant a tighten stamped by an earlier cycle but not placed — for example one deferred by the #621 capped-qty guard at `scheduler/scale_in.go:64-70` — was re-frozen at the old wide trigger on the very pass that rewrote the order. The forced path now adopts a strictly favorable candidate instead. The debounce exists to prevent order churn, and a replace we are already making has no churn to prevent; the direction gate still blocks any widening.

Tests (`scheduler/trailing_tp_ratchet_scale_in_test.go`): add plus tier-clear in one cycle on both sides — exactly one replace, post-add quantity, trigger moves toward the mark and never away; add with no tier clear keeps the #873 behavior with no trigger move; a forced resize adopts a pending unplaced tighten; the add branch alerts and stamps only when a tier newly clears.

Regression check: with item 3 reverted, `ForcedResizeAdoptsPendingTighten` fails; with both item 2 and item 3 reverted, the long and short add-plus-tighten cases fail at the old wide trigger.

Verification: `gofmt` clean, `go vet ./...`, `go test ./...`, and `go test -race ./...` all pass.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-07T19:07:39Z, `LGTM` + 1 optional) in 1083639a.

### Not changed (refuted as a behavior change) — but confirmed and pinned

1. **Scale-in add does not seed `StopLossHighWaterPx`** — confirmed deliberate, and it is the safer of the two paths. `executeHyperliquidResultDeferredOpen` re-seeds from the fill price on every trade-executing cycle (`scheduler/main.go:4190-4195`), which it can afford because the comment there is right: `StopLossTriggerPx` is the durable ratchet and `computeTrailingStopUpdateInternal` only ever replaces on a FAVORABLE candidate (`scheduler/hyperliquid_trailing_stop.go:553-559`), so a reset high-water can never loosen a resting stop. An add is different in kind: the high-water belongs to the trail already in progress, and the add's fill price carries no information about it. Stamping it would push the stored high-water backward whenever the add fills below the best price seen, then hold the trail frozen until price climbed back through it. The only zero case is already covered — the walker seeds an unset high-water from `riskAnchorPrice()`, the frozen first entry rather than the blended average (`scheduler/hyperliquid_trailing_stop.go:719-728`, #873).

No behavior change. The asymmetry is now recorded in code (`scheduler/scale_in.go:456-465`) so a later edit does not "restore symmetry", and `TestExecuteHyperliquidScaleInDeferredOpen_LeavesHighWaterUntouched` pins both halves: an add that fills below an established high-water leaves it untouched, and a zero high-water resolves to the frozen anchor rather than the blended average.

Verification: `gofmt` clean, `go vet ./...`, `go test ./...`, and `go test -race ./...` all pass.

---

richkuo:
@claude sonnet review

Issue #1412, opened 2026-07-28
[C99] Fuse the Hurst exponent into the composite regime classifier so labels reflect persistence


richkuo:
**Complexity: 99/100** — Capability 3 (Risk 4 — changes which regime label a window maps to, feeding live entry gates, `*_atr_regime` close blocks, and #1085 certifications; Uncertainty 3 — Stage 0 evidence and re-calibration outcomes unknown); Volume 24 (Scope 4; Coupling 4 — Python classifier + Go threshold structs + backtester + cert/fixture surfaces; Verification 4) — Fable 5, high · fableplan: yes

**Chain:** depends on #1409 (Hurst estimator/metric), #1410 (calibration study — its report file is the Stage 0 substrate), #1411 (standalone `hurst_gate`, which stays functional and independent).

## Stage 0 — Evidence gate (STOP if it fails)

Before any implementation: extend `backtest/research/hurst_gate_calibration.md` (from #1410) with a **joint-bucket analysis** — ADX-level × Hurst-level (at minimum high/low ADX × H `<0.45` / `0.45–0.55` / `>0.55` / NaN) per strategy family, reporting win rate, net PnL, and max drawdown per joint bucket. Commit it as a new section of that report.

The question: do joint buckets separate beyond what either metric alone predicts — specifically, is high-ADX + low-Hurst (strong move, anti-persistent tape) materially worse for momentum-style entries than high-ADX alone? If not: write `NO JOINT SEPARATION` in the report section, commit only that, **STOP**, and comment on this issue that full fusion is not justified — the standalone `hurst_gate` (#1411) remains the correct amount of Hurst. Proceed to Stage 1 only on demonstrated separation.

## Problem (Stage 1)

The composite classifier (`map_composite_label`, `shared_tools/regime.py:260`) splits clean vs choppy trends from efficiency + ADX only (`clean = efficiency >= eff_th and high_adx`, regime.py:286). A strong-ADX but anti-persistent window still labels `trending_*_clean`, so every label consumer (entry gates, `*_atr_regime` close blocks, directional policies) treats persistence-hostile tape as clean trend. #1411's gate can block entries on H, but labels themselves stay persistence-blind.

## Goal

[...]

Issue #1411, opened 2026-07-28
[C97] Hurst entry gate and optional persistence-scaled position sizing, shipped default-off


richkuo:
**Complexity: 97/100** — Capability 3 (Risk 3 — live entry gating + position-sizing money path; Coupling 3 — config/dispatch/state/sizing/diagnostics/backtester; Uncertainty 2 — precedents all identified); Volume 22 (Scope 4; Coupling 3; Verification 4 — per-site wiring tests, hysteresis sequences, parity) — Fable 5, high · fableplan: no

## Dependency gate — RESOLVED, with an INCONCLUSIVE calibration verdict

`backtest/research/hurst_gate_calibration.md` now exists on main (PR #1420, closing #1410). The transitive dependency #1409 (Hurst estimator + metric surfacing) also merged (PR #1419). No blockers remain.

**The report's Recommendation is `INCONCLUSIVE`.** 0 of the 30 tested configurations passed Benjamini-Hochberg at alpha=0.05; 16 met every economic condition but failed only on significance. The report closes with, verbatim:

> Do not build a Hurst entry gate on this evidence. Re-run this study before revisiting the question.

**Owner reaffirmation (2026-08-08).** The implementer gate in the original body pre-authorized building under an INCONCLUSIVE verdict, and that pre-authorization predates the report. The owner has reviewed the report's do-not-build sentence and **reaffirms building this feature**, under the approved milestone run plan, with these binding conditions:

1. The feature ships **default-off**, per-strategy opt-in only. It never auto-enables. An operator who never writes `hurst_gate` sees byte-identical behavior.
2. **No recommended thresholds ship anywhere.** `config.example.json` gets no `hurst_gate` values. Document the key shapes in `docs/ARCHITECTURE.md` only. The study supports no numbers, so the code offers none.
3. The PR description must state the INCONCLUSIVE verdict and quote the do-not-build sentence prominently, near the top.

The reasoning: the mechanism is the deliverable, and the evidence for *turning it on* is separate from the evidence for *being able to*. A default-off gate with no shipped thresholds adds no live risk, and it lets a re-run of the study (which the report asks for) be acted on without a build cycle. This reaffirmation does not authorize enabling the gate on any strategy.

## Problem

[...]

Issue #1410, opened 2026-07-28
[C41] Calibration study for a future Hurst-based entry gate (report only, no live changes)


richkuo:
**Complexity: 41/100** — Capability 1 (Risk 1 — offline research only; Uncertainty 2 — study design choices: family bucketing, cross-window entry dedup, multiplicity control; max(Risk, Uncertainty) = 2 maps to Capability 1, and Coupling 2 gives no bump); Volume 16 (Scope 3 — study script + report + tests + registry row; Coupling 2 — #1409 estimator + eval-windows reuse; Verification 3 — helper tests, noise/multiplicity gates, re-runnability) — Opus-class, xhigh · fableplan: no

#1409 landed on the default branch through merged PR #1419 (commit `bfcf4cc9`). Branch from the default branch.

## Problem

There is no evidence base for whether a Hurst-based entry gate would help. #1409 adds the estimator and surfaces the metric for observability, but nothing measures how entry outcomes vary with H at entry time, so any future gate threshold would be guessed. Prior repo research (#983/#984 close sweeps, #1054 noise gate) shows that plausible-sounding gates frequently fail validation — a calibration study must come first.

## Goal

A committed, re-runnable calibration study and report at `backtest/research/hurst_gate_calibration.md` (exact path — the follow-up gate-implementation work reads its thresholds from this file), ending in an unhedged **Recommendation** section: either a concrete per-strategy-family configuration, or the single word **INCONCLUSIVE** with justification. No scheduler, config, gating, or sizing changes of any kind.

## Approach

Add a study script under `backtest/research/` following the naming and style of the existing research one-shots (`regime_1076_directional_premise.py`, `regime_1152_exit_retune.py`) and the dependency-light statistics style of `backtest/regime_stats.py` (numpy-only helpers). Reuse the M1 eval-window definitions (`backtest/eval_windows.py:79` `WINDOWS`, protocol `is`/`oos` + held-out `2023`/`2024`/`2025H1`) rather than inventing new windows.

**Measurement**
- Rolling Hurst imported from the #1409 SSoT estimator — the `hurst_exponent` symbol in `shared_strategies/open/indicators_core.py`. No reimplementation of the DFA estimator.
- The estimator carries an advisory-only invariant. This study is report-only and must not feed gating, sizing, or config surfaces.
[...]

PR #1419, opened 2026-08-07
feat(#1409): add Hurst exponent metric to regime payloads for observability [C41, Sonnet 5, high]

### LINKED ISSUE #1409: [C41] Add a Hurst exponent metric to regime payloads for observability only
**Complexity: 41/100** — Capability 1 (Risk 1 — advisory-only metric, label invariance enforced by tests); Volume 16 (Scope 3 — four dirs touched; Coupling 2 — Python↔Go payload contract + #879 injection path; Verification 3 — synthetic-series tolerance tests + byte-identical label fixtures) — Sonnet-class, high · fableplan: no

## Problem

The regime system classifies market state from an efficiency/ADX metric tuple (`_composite_efficiency_metrics`, `shared_tools/regime.py:220` — `return_eff`, `range_eff`, `efficiency`, plus `adx`), but carries no persistence/mean-reversion measure. A Hurst exponent would let operators and offline research distinguish trending-persistent from mean-reverting conditions, but today there is no Hurst implementation anywhere in the repo (verified: zero `hurst` hits in `*.py`/`*.go`).

## Goal

A Hurst exponent (DFA method) rides along in regime payload metrics — visible in `check_regime.py` output, the Go regime store, and offline backtest feature matrices — with **zero behavior change**: no label mapping, gating, sizing, or config surface changes. Observability only, so the metric can be studied before anyone proposes acting on it.

## Approach

**1. Estimator SSoT — `shared_strategies/open/indicators_core.py`**
Add a DFA (detrended fluctuation analysis) Hurst estimator alongside `wilder_rsi`/`atr_*`, matching that module's conventions. DFA over classic R/S: R/S is too noisy at the window lengths this system uses. Requirements:
- Input: pandas close-price Series; operate on log returns internally.
- Signature takes window/min-points; returns **NaN** (never a raised error, never 0.5) below the documented minimum (~100 points), so "unknown" is distinguishable from "random walk".
- Deterministic, no randomness, vectorized where practical.
- Unit tests in `shared_strategies/open/test_indicators_core.py`: persistent (trending) series → H meaningfully > 0.5; strongly mean-reverting series → H < 0.5; Gaussian random walk → H ≈ 0.5 within tolerance; insufficient data → NaN; constant-price series edge case.

**2. Regime metrics surfacing — `shared_tools/regime.py` (advisory only, critical constraint)**
Metrics dicts are built at two sites: `latest_regime_composite` (`regime.py:326`, `CLASSIFIER_COMPOSITE`) and `latest_regime` (`regime.py:498`, ADX classifier — keys `adx`/`plus_di`/`minus_di`/`atr_pct`, built independently of `_composite_efficiency_metrics`). `regime.py:437` is `compute_regime_composite`'s per-bar label loop; it emits labels only and builds no metrics dict, so it is not a hurst emission site. `composite_feature_matrix` (`regime.py:445`, using `_composite_efficiency_metrics` at `:472`) is the offline/enriched-feature path, covered separately under item 4.

**This issue's scope is the composite metrics dict only: add a `hurst` key at `regime.py:326`.** The ADX-classifier payload (`latest_regime`, `regime.py:498`) does **not** get a `hurst` key under this issue — its metrics dict is a distinct, independently-built tuple and adding hurst there is out of scope unless a follow-up issue amends it explicitly.

**Do NOT change `map_composite_label` or any thresholds** — composite thresholds and directional certifications (#1085) were calibrated on the existing metric tuple; hurst must not alter which label any input produces. The #879 injection path (`--regime-payload-json`, handled by `regime_from_injected_payload` at `regime.py:593`) already passes unknown/absent metric keys through untouched, so it accepts payloads without a `hurst` key today with no code change — the acceptance test here is a regression guard confirming that behavior stays, not new tolerance work. Extend `shared_tools/test_regime.py` to assert (a) `hurst` appears in the composite payload metrics when data is sufficient and (b) labels for all existing fixtures are byte-identical to before.

**NaN-at-the-JSON-boundary constraint:** when the estimator returns NaN (insufficient data), the composite metrics dict must **omit** the `hurst` key entirely rather than emit a NaN value. `json.dumps` writes bare `NaN` for a Python float NaN, and Go's `json.Unmarshal` into `map[string]float64` (`scheduler/regime_multi_window.go:21`) rejects that token, which would fail the whole `RegimePayload` parse and break live regime output for that strategy on that cycle. Absence, not a sentinel value, is the correct "unknown" signal on the wire.

**Input for the estimate:** compute hurst over the full fetched OHLCV frame passed into `latest_regime_composite` (`check_regime.py` default `--ohlcv-limit 200`), not the `period`-length window slice used for `_composite_efficiency_metrics` (typically 14-50 bars). Window-length slices sit below the ~100-point DFA minimum and would return NaN on nearly every live cycle, making the metric permanently absent in practice.

**3. Operator visibility**
- `shared_scripts/check_regime.py`: hurst flows out inside each window's `metrics` dict (the script already emits `{"regime": {"<window>": {"regime","score","classifier","metrics"}}}`).
- Go side: `RegimeSnapshot.Metrics` is already a generic `map[string]float64` (`scheduler/regime_multi_window.go:21`), so hurst is carried with **no Go schema change**. Go consumers that read specific metric keys today (`regime_divergence.go:369`, `regime_transitions.go:409`) only read `return_eff` and need no edits. Leave Go untouched; state so in the PR.

**4. Backtest availability**
Expose hurst in `backtest/regime_enriched_features.py` by importing the estimator from the SSoT (`indicators_core.py`) — no reimplementation — as an enriched-only column, so it can be studied in `backtest/regime_stats.py`-style offline analysis before any gating proposal.

HMM fit/decode (`backtest/regime_hmm.py`) is an offline-only path; live `check_regime.py` builds no feature columns today (`LIVE_WIRING_DELTA`, `backtest/regime_enriched_features.py:54-61` — live wiring is unbuilt, tracked separately as #1074), so there is no live HMM contract to preserve here. The real constraint is `CANONICAL_COLUMNS` ordering at `regime_enriched_features.py:44-51`: the naming/fit path consumes those columns by position, so hurst must **never** be inserted into or reorder that block — append it after the canonical block, following the existing `ENRICHED_EXTRA_COLUMNS` pattern.

## Out of scope

- No new regime labels; no changes to label mapping or thresholds.
- No entry gating, position sizing, or config surface for acting on Hurst (separate follow-up issues).
- No changes to ATR method resolution or the #1277-pinned `regime.py` `method="simple"` path.
- No `hurst` key added to the ADX-classifier (`latest_regime`, `regime.py:498`) metrics dict — composite path only.

## Acceptance criteria

- All new and existing tests pass: `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/`; `go test ./...` only if Go files are touched.
- `test_regime.py` proves label outputs unchanged for all existing fixtures.
- Injected `--regime-payload-json` payloads without `hurst` keep working (regression guard).
- Below-minimum-data cases omit the `hurst` key from the emitted metrics dict rather than emitting NaN into JSON.
- Hurst is computed over the full fetched frame, not the period-length window slice.
- PR description states: estimator method (DFA) and why R/S was rejected, minimum-data/NaN-omission behavior, exactly which payloads/displays now carry hurst (composite only, not ADX classifier), and confirmation no gating behavior changed.

## Execution

- Depends on: none
- Runs after: none
- Build model: Sonnet 5
- Build effort: high
- Validate effort: medium
- fableplan first: No
- PR review: Opus 5 (first review)

---
Updated with LLM: Sonnet 5 | high | Harness: milestone-pipeline

richkuo:
## Plain simple English

This change adds a new market-condition number, the Hurst exponent, to the regime data the bot already tracks. The number shows whether a market is trending or bouncing back and forth. It is for humans and research only. It does not change any trading decision, label, or config.

## Summary

- New estimator: `shared_strategies/open/indicators_core.hurst_exponent` — detrended fluctuation analysis (DFA) over log returns. Chosen over classic R/S (rescaled range), which is materially noisier at the series lengths this system observes in practice (a few hundred bars). Deterministic, never raises, returns NaN (never a fabricated 0.5) below the ~100-point DFA floor, on non-positive prices, or on a degenerate (e.g. constant-price) series.
- `shared_tools/regime.py`'s composite classifier (`latest_regime_composite`, `regime.py:326`) adds a `hurst` key to its metrics dict, computed over the **full fetched OHLCV frame** (`check_regime.py` default `--ohlcv-limit 200`) — not the `period`-length composite window (14-50 bars live), which sits below the DFA floor and would be NaN almost every cycle.
- **NaN-at-the-boundary:** when the estimate is NaN, the `hurst` key is **omitted** from the metrics dict, never serialized as NaN. `json.dumps` writes a bare `NaN` token for a Python NaN float; Go's `json.Unmarshal` into `map[string]float64` (`scheduler/regime_multi_window.go:21`) rejects that token, which would break the entire `RegimePayload` parse for that strategy on that cycle. A dedicated test asserts on the raw serialized JSON text (not a round-tripped dict — Python's `json` module is itself NaN-tolerant and would mask a regression).
- `backtest/regime_enriched_features.py` exposes `hurst` as an enriched-only column, imported from the SSoT (never reimplemented), appended after `CANONICAL_COLUMNS` following the existing `ENRICHED_EXTRA_COLUMNS` pattern. It uses its own trailing `hurst_window` (default 100, independent of `period`) — the composite `period` alone is below the DFA floor.
- **This issue's scope is the composite metrics dict only** (`latest_regime_composite`, `regime.py:326`). The ADX-classifier payload (`latest_regime`, `regime.py:498`) does **not** get a `hurst` key — it is a separate, independently-built metrics dict; adding hurst there is explicitly out of scope.
- **No Go changes.** `RegimeSnapshot.Metrics` (`scheduler/regime_multi_window.go:21`) is already a generic `map[string]float64`, so `hurst` is carried with no schema change. The two Go readers of `Metrics` (`regime_divergence.go:369`, `regime_transitions.go:409`) only read `return_eff` and need no edits.

## Zero behavior change confirmed

- `map_composite_label` and all composite thresholds are byte-for-byte untouched.
- A dedicated test (`test_latest_regime_composite_metrics_and_label_unchanged_by_hurst`) independently recomputes the label, score, and every pre-existing metric via the same helper functions `latest_regime_composite` calls, and asserts the function's output is byte-identical aside from the new `hurst` key.
- The `#879` injection path (`regime_from_injected_payload`, `regime.py:593`) already passes unknown/absent metric keys through untouched — a payload snapshot without `hurst` keeps working. This is covered as a regression guard, not new tolerance work.
- `_DEFAULT_METRICS` stays hurst-free (asserted by test) — absence is the correct "unknown" signal; a fabricated default like `0.0` would masquerade as strong mean-reversion.

## Verification

- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — all pass (including the 6 pre-existing files touched or newly-added tests).
- `uv run --no-sync python -m pytest shared_scripts/test_regime_wiring.py` — explicit invoke (not in `testpaths`), all pass, including new JSON-boundary tests proving no bare `NaN` token reaches the payload.
- `go -C scheduler build .` — builds clean (no Go files touched by this PR).
- Manual sanity: DFA estimator gives H ≈ 0.5 on a Gaussian random walk, H > 0.5 on a persistent AR(1) series, H < 0.5 on a mean-reverting AR(1) series, across multiple seeds.

## Merge order

Head of the serial chain #1409 → #1410 → #1411 → #1412 (milestone run `wf_fbd2b98e-ad4`). No dependency for this PR; it must land first.

Closes #1409

---
Created with LLM: Sonnet 5 | high | Harness: milestone-pipeline

### DIFF
--- a/backtest/regime_enriched_features.py
+++ b/backtest/regime_enriched_features.py
@@ -15,6 +15,12 @@ canonical features (kept FIRST, in their canonical order) plus signals the hand-
                           higher-timeframe bar that has already CLOSED at or before the base bar's
                           open (strictly causal; conservatively staler than the canonical row,
                           never leaking).
+  * ``hurst``           — #1409 trailing DFA Hurst exponent (``indicators_core.hurst_exponent``,
+                          the SSoT — imported, never reimplemented). Uses its OWN trailing
+                          ``hurst_window`` (default 100 bars, independent of ``period``), because
+                          DFA needs ~100 points and the composite window (``period``, 14-50 bars
+                          live) is below that floor. Warmup (< ``hurst_window`` observations) or a
+                          degenerate segment -> NaN, same warmup convention as the other extras.
 
 Offline / research only. Live ``check_regime.py`` still builds only the canonical four-column
 matrix, so an enriched model is NOT drop-in for the live classifier — that delta is recorded for
@@ -32,7 +38,8 @@ import sys
 
 _THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 for _p in (_THIS_DIR, os.path.abspath(os.path.join(_THIS_DIR, "..")),
-           os.path.abspath(os.path.join(_THIS_DIR, "..", "shared_tools"))):
+           os.path.abspath(os.path.join(_THIS_DIR, "..", "shared_tools")),
+           os.path.abspath(os.path.join(_THIS_DIR, "..", "shared_strategies", "open"))):
     if _p not in sys.path:
         sys.path.insert(0, _p)
 
@@ -40,23 +47,28 @@ import numpy as np
 import pandas as pd
 
 from regime import composite_feature_matrix, _DEFAULT_COMPOSITE_THRESHOLDS
+from indicators_core import hurst_exponent, HURST_DFA_MIN_POINTS
 
 # Canonical four, in the order composite_feature_matrix emits them. The naming path
 # (map_composite_label) consumes these by position; they MUST stay first and in this order.
 CANONICAL_COLUMNS = ["return_eff", "range_eff", "efficiency", "adx"]
 # Extra signals the hand-rule does not use. Appended AFTER the canonical block.
-ENRICHED_EXTRA_COLUMNS = ["funding_rate", "volume_z", "htf_range_eff"]
+ENRICHED_EXTRA_COLUMNS = ["funding_rate", "volume_z", "htf_range_eff", "hurst"]
 ENRICHED_COLUMNS = CANONICAL_COLUMNS + ENRICHED_EXTRA_COLUMNS
 # Position of (return_eff, range_eff, efficiency, adx) inside any canonical-first matrix.
 CANONICAL_INDICES = (0, 1, 2, 3)
+# #1409: hurst's own trailing window, independent of the composite `period`. Matches
+# indicators_core.HURST_DFA_MIN_POINTS (the DFA floor) so the column isn't NaN by construction.
+HURST_WINDOW_DEFAULT = HURST_DFA_MIN_POINTS
 
 # Recorded for #1074. An enriched model decodes from this column set on-cycle; live
 # check_regime.py builds only CANONICAL_COLUMNS, so live wiring would need to feed the same
-# enriched matrix (funding fetch + volume z-score + HTF resample) every cycle. Not solved here.
+# enriched matrix (funding fetch + volume z-score + HTF resample + hurst) every cycle. Not solved
+# here.
 LIVE_WIRING_DELTA = (
     "Enriched models decode from ENRICHED_COLUMNS; live check_regime.py builds only "
-    "CANONICAL_COLUMNS. Live wiring (#1074) must feed funding_rate + volume_z + htf_range_eff "
-    "on-cycle in this exact order, or forward_filter_labels reads garbage."
+    "CANONICAL_COLUMNS. Live wiring (#1074) must feed funding_rate + volume_z + htf_range_eff + "
+    "hurst on-cycle in this exact order, or forward_filter_labels reads garbage."
 )
 
 
@@ -121,18 +133,43 @@ def _htf_range_eff_column(df: pd.DataFrame, period: int, thresholds: dict,
     return merged["htf_range_eff"].to_numpy(dtype=float)
 
 
+def _hurst_column(df: pd.DataFrame, window: int) -> np.ndarray:
+    """Trailing per-bar Hurst exponent (DFA) over the last `window` closes, inclusive of the
+    current bar — same causal window convention as the other enriched columns.
+
+    `window` is independent of the composite `period`: DFA needs ~`indicators_core.
+    HURST_DFA_MIN_POINTS` points to produce a non-NaN estimate, well above the composite's
+    14-50 bar window, so this column intentionally does NOT reuse `period`/`vol_window`.
+    Estimator comes from ``indicators_core.hurst_exponent`` (the repo SSoT) — never
+    reimplemented here. `hurst_exponent` operates on LOG RETURNS internally, so a `window`-price
+    segment yields only `window - 1` returns — one short of `min_points=window` and NaN every
+    time. The segment therefore spans `window + 1` closes (`window` log returns), warming up one
+    bar later than the level-only extras above. Warmup or a degenerate segment -> NaN.
+    """
+    if window < 2:
+        raise ValueError("hurst_window must be >= 2")
+    closes = df["close"].astype(float).to_numpy()
+    n = len(closes)
+    out = np.full(n, np.nan, dtype=float)
+    for i in range(window, n):
+        seg = pd.Series(closes[i - window : i + 1])
+        out[i] = hurst_exponent(seg, min_points=window)
+    return out
+
+
 def enriched_feature_matrix(df: pd.DataFrame, period: int, thresholds: dict | None = None, *,
                             funding: pd.DataFrame | None = None, vol_window: int | None = None,
-                            htf_multiple: int = 4,
+                            htf_multiple: int = 4, hurst_window: int | None = None,
                             columns: list[str] | None = None) -> pd.DataFrame:
     """Per-bar enriched feature matrix (canonical four FIRST, then the enabled extras).
 
     All joins are causal — canonical window ends at the bar, funding/HTF use backward merges,
-    volume z-score uses a trailing window. Warmup and atr<=0 bars are NaN (dropped downstream by
-    the model's NaN mask). `funding` is the DataFrame(timestamp, rate) from
+    volume z-score and hurst use a trailing window. Warmup and atr<=0 bars are NaN (dropped
+    downstream by the model's NaN mask). `funding` is the DataFrame(timestamp, rate) from
     ``funding_fetcher.load_cached_funding``; when None/empty the funding column is all-NaN (the
     model mask then drops every row, so a funding-bearing subset is effectively unavailable — the
-    bake-off harness detects and reports that rather than fitting on nothing).
+    bake-off harness detects and reports that rather than fitting on nothing). `hurst_window`
+    defaults to `HURST_WINDOW_DEFAULT` (the DFA floor, NOT `period` — see `_hurst_column`).
 
     `columns` selects a subset for ablations; the result preserves ENRICHED_COLUMNS order so the
     canonical block stays first (canonical_indices == (0, 1, 2, 3)). Unknown names raise.
@@ -149,6 +186,7 @@ def enriched_feature_matrix(df: pd.DataFrame, period: int, thresholds: dict | No
     columns = [c for c in ENRICHED_COLUMNS if c in columns]
 
     vol_window = int(vol_window) if vol_window else int(period)
+    hurst_window = int(hurst_window) if hurst_window else HURST_WINDOW_DEFAULT
     out = pd.DataFrame(index=df.index)
     canon = composite_feature_matrix(df, period, thresholds)
     for col in CANONICAL_COLUMNS:
@@ -160,6 +198,8 @@ def enriched_feature_matrix(df: pd.DataFrame, period: int, thresholds: dict | No
         out["volume_z"] = _volume_z_column(df, vol_window)
     if "htf_range_eff" in columns:
         out["htf_range_eff"] = _htf_range_eff_column(df, period, thresholds, htf_multiple)
+    if "hurst" in columns:
+        out["hurst"] = _hurst_column(df, hurst_window)
     return out[columns]
 
 
--- a/backtest/tests/test_regime_enriched_features.py
+++ b/backtest/tests/test_regime_enriched_features.py
@@ -155,6 +155,51 @@ def test_volume_z_is_trailing_and_warmup_nan():
     assert vz[i] == pytest.approx(expected, rel=1e-9)
 
 
+def test_hurst_column_is_trailing_and_warmup_nan():
+    df = _synthetic_ohlcv(n=300)
+    mat = ref.enriched_feature_matrix(df, period=48, columns=ref.CANONICAL_COLUMNS + ["hurst"],
+                                      hurst_window=100)
+    h = mat["hurst"].to_numpy()
+    assert np.isnan(h[:100]).all()   # fewer than `hurst_window` observations -> NaN
+    assert np.isfinite(h[100])       # first full trailing window (101 closes) is defined
+    # recompute by hand from the SSoT estimator over the same trailing window
+    from indicators_core import hurst_exponent
+    close = df["close"].to_numpy()
+    i = 200
+    expected = hurst_exponent(pd.Series(close[i - 100: i + 1]), min_points=100)
+    assert h[i] == pytest.approx(expected, rel=1e-12)
+
+
+def test_hurst_column_uses_own_window_independent_of_period():
+    """#1409: hurst must NOT inherit `period` (14-50 bars live) as its window -- that would put it
+    below the ~100-point DFA floor and make it NaN almost every live cycle. Two calls with the
+    SAME hurst_window but different `period` must produce the identical hurst column."""
+    df = _synthetic_ohlcv(n=300)
+    mat_a = ref.enriched_feature_matrix(df, period=20, columns=ref.CANONICAL_COLUMNS + ["hurst"],
+                                        hurst_window=100)
+    mat_b = ref.enriched_feature_matrix(df, period=48, columns=ref.CANONICAL_COLUMNS + ["hurst"],
+                                        hurst_window=100)
+    pd.testing.assert_series_equal(mat_a["hurst"], mat_b["hurst"], check_exact=True)
+
+
+def test_hurst_column_is_causal_future_bars_do_not_change_past_rows():
+    df = _synthetic_ohlcv(n=300)
+    base = ref.enriched_feature_matrix(df, period=48, columns=ref.CANONICAL_COLUMNS + ["hurst"])
+    t = 200
+    mutated = df.copy()
+    mutated.iloc[t + 1:, mutated.columns.get_loc("close")] *= 3.0
+    after = ref.enriched_feature_matrix(mutated, period=48, columns=ref.CANONICAL_COLUMNS + ["hurst"])
+    a = base["hurst"].iloc[: t + 1].to_numpy()
+    b = after["hurst"].iloc[: t + 1].to_numpy()
+    assert np.array_equal(np.nan_to_num(a, nan=-7.0), np.nan_to_num(b, nan=-7.0))
+
+
+def test_hurst_extra_column_appended_after_canonical_block():
+    assert ref.ENRICHED_COLUMNS[-1] == "hurst"
+    assert ref.ENRICHED_COLUMNS[:4] == ref.CANONICAL_COLUMNS
+    assert "hurst" in ref.ENRICHED_EXTRA_COLUMNS
+
+
 # ---------------------------------------------------------------- fit / decode contract
 
 def test_fit_unsupervised_enriched_schema_and_decode():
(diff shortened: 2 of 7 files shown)


richkuo:
LGTM

### Recommended Optional

1. **The Hurst estimate reads systematically above 0.5 on memoryless data at both frame sizes this metric is actually computed at.**
Measured by running the estimator at this head commit over 300 independent Gaussian random walks (true H = 0.5, no memory). At n=200 closes — the live path, since `check_regime.py --ohlcv-limit` defaults to 200 (`shared_scripts/check_regime.py:198`) and `latest_regime_composite` feeds the full frame (`shared_tools/regime.py:355`) — the estimator returns mean H = 0.539, sd = 0.061. At n=101 closes — exactly the per-bar segment the enriched column uses, because `HURST_WINDOW_DEFAULT = HURST_DFA_MIN_POINTS = 100` (`backtest/regime_enriched_features.py:62`) and `_hurst_column` slices `window + 1` closes (`backtest/regime_enriched_features.py:154-156`) — mean H = 0.558, sd = 0.087. The driver is `_HURST_DFA_MIN_SCALE = 4` (`shared_strategies/open/indicators_core.py:267`): holding everything else fixed and raising it to 8 moves the random-walk mean to 0.505 at n=200 and 0.514 at n=101, while persistent and mean-reverting AR(1) series still separate cleanly (0.750 vs 0.398 at n=2000). This matches the primary literature on the method — *On the Validity of Detrended Fluctuation Analysis at Short Scales*, Entropy 24(1):61, states "we have observed a systematic deviation of α1 with respect to the correct and unique scaling exponent, which is in many cases largely overestimated by α1", and that the overestimation "is independent of the time series length", so a longer fetch does not remove it. The consequence is interpretive, and it lands precisely on the purpose the metric was added for: the docstring at `shared_strategies/open/indicators_core.py:316-324` calls 100 points "the floor below which the DFA scaling-exponent regression becomes unstable" and states "H > 0.5 marks a persistent/trending series" with no null-distribution caveat, so an operator or a later gating proposal reading `hurst: 0.60` off a 200-bar payload would take it as evidence of persistence when a memoryless series produces that value routinely. Nothing reads the value today (verified: no `hurst` consumer exists in `scheduler/` or in any Python gating path), so this blocks nothing. Raise `_HURST_DFA_MIN_SCALE`, and/or record the measured random-walk null band next to the H > 0.5 sentence so the caveat travels with the number.
**Invariant:** an advisory metric published for human interpretation must center on its own null value under the null hypothesis, at the sample size it is actually computed at — or carry its null distribution where a reader of the value will see it.
**Must survive:** (1) 300 memoryless random walks at n=200 give a mean H within 0.02 of 0.50; (2) the same assertion at n=101, the enriched column's exact per-bar segment and the noisiest operating point; (3) a persistent AR(1) series still returns H clearly above 0.5 and a mean-reverting one clearly below, so removing the bias does not remove the signal.
**Plain simple English:** The new market number is set too high. On data with no memory, it usually reads 0.54 instead of 0.50. Almost half of the readings look like a trend when no trend exists. Change the smallest measurement block from 4 bars to 8 bars, or write the correct expected range in the notes.

2. **The per-bar Hurst column recomputes the full analysis in a Python loop, at roughly a hundred times the necessary cost.**
`_hurst_column` (`backtest/regime_enriched_features.py:154-156`) calls `hurst_exponent` once per bar, and each call runs `np.polyfit` once per segment per scale — on the order of two hundred separate least-squares fits per bar at `window=100`. Measured at this head commit: 12.9 s for a 3000-bar frame, 4.3 ms per bar. `SUBSETS["all_enriched"]` (`backtest/research/regime_1095_enriched_vol_model.py:85`) is `list(ENRICHED_COLUMNS)`, which now carries `hurst`, and `run_bakeoff` builds that subset's matrix for every entry in `all_windows` — the five spans in `backtest/eval_windows.py:79-85`, roughly 30,000 bars at 1h — so a full sweep gains about two minutes purely in this column. The detrend is a fixed-design regression: the design matrix depends only on `scale`, so its pseudo-inverse can be computed once per scale and applied to the whole reshaped segment block as one matrix product, replacing every per-segment `polyfit` call. Worth folding in while the code is new, since the enriched bake-off is the harness this column exists to serve.
**Invariant:** a fixed-design per-segment linear detrend is one matrix product over all segments, not one `polyfit` call per segment; adding a research column must not change the harness's cost class.
**Must survive:** (1) `_hurst_column` output stays equal to the current per-bar values to within a relative tolerance of 1e-9 on a 300-bar frame; (2) the NaN warmup boundary stays at index `window` exactly, so `np.isnan(h[:window]).all()` and `np.isfinite(h[window])` both still hold; (3) degenerate segments — constant price, non-positive price, embedded non-finite values — still return NaN and still never raise.
**Plain simple English:** Making the new column for research is slow. It fits a straight line one small block at a time, many thousands of times. A long research run takes about two extra minutes. Do the same fit for all blocks in one step instead. The resulting numbers must stay the same.

---
Validated with LLM: Opus 5 | high | Harness: milestone-pipeline

richkuo:
Addressed review feedback (richkuo · 2026-08-07T23:07:51Z, Opus 5) in 98b64133.

### Fixed

1. **Hurst estimate reads systematically above 0.5 on memoryless data** — confirmed by re-measuring: 300-trial fuzz gave mean H ~= 0.5405 at n=200 / ~= 0.5532 at n=101 with `_HURST_DFA_MIN_SCALE=4` (matching the review's own 0.539 / 0.558). Raised `_HURST_DFA_MIN_SCALE` 4 -> 8 (`shared_strategies/open/indicators_core.py:203`), which drops the measured mean to ~0.51 / ~0.52 while AR(1) persistent/mean-reverting series still separate cleanly (regression-tested). Also added the measured null-distribution caveat to `hurst_exponent`'s docstring (`shared_strategies/open/indicators_core.py:270-278`) so the residual bias travels with the number, since raising the floor further trades away real signal at the sample sizes this system actually runs at (`test_hurst_persistent_series_above_half`/`test_hurst_mean_reverting_series_below_half` still pass at scale=8). New regression tests: `test_hurst_random_walk_mean_near_half_at_live_frame_size`, `test_hurst_random_walk_mean_near_half_at_enriched_column_frame_size` (`shared_strategies/open/test_indicators_core.py`).

2. **Per-bar Hurst column recomputes DFA via ~200 separate `np.polyfit` calls** — confirmed: `_hurst_dfa_fluctuation` ran one `np.polyfit` per segment per scale. Vectorized it into one batched pseudo-inverse product per scale, since every segment at a fixed scale shares the same design matrix (`shared_strategies/open/indicators_core.py:216-245`). This is the SSoT (`indicators_core.hurst_exponent`) itself, so `_hurst_column`'s per-bar loop in `backtest/regime_enriched_features.py` benefits without any change or reimplementation there. Verified numerically equivalent to the old per-segment-`polyfit` path via a 50-trial fuzz test against the original naive implementation, matching to <1e-9 relative tolerance (`test_hurst_dfa_fluctuation_vectorization_matches_naive_per_segment_polyfit`). `backtest/tests/test_regime_enriched_features.py` (23 tests, including the `hurst`-column `pytest.approx(..., rel=1e-12)` cross-check against the SSoT) all still pass.

### Verification

- `uv run --no-sync python -m pytest shared_strategies/open/test_indicators_core.py -k hurst` — 13 passed (10 pre-existing + 3 new).
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2901 passed, 0 failed.
- `uv run --no-sync python -m pytest shared_scripts/test_regime_wiring.py` — 21 passed (explicit invoke, not in `testpaths`).
- `go -C scheduler build .` — builds clean (no Go files touched).
- Zero-behavior-change invariant confirmed untouched: `map_composite_label`, composite thresholds, and #1085 directional certifications were not modified; `test_latest_regime_composite_metrics_and_label_unchanged_by_hurst` and the byte-identical-label fixture tests in `shared_tools/test_regime.py` still pass.

No `Needs Fixing` or `Requires Human Review` items were raised, no inline diff threads were open, and CI (`go`, `python`) was already green on this review — nothing to refute or defer.

---
Created with LLM: Sonnet 5 | high | Harness: milestone-pipeline

richkuo:
LGTM

### Recommended Optional

1. **The new null-distribution caveat understates the random-walk spread at n=101, the enriched column's exact frame size.**
`shared_strategies/open/indicators_core.py:271-278` (added in this fix commit) states the memoryless-data standard deviation is "sd ~0.06-0.09" at both n=200 and n=101. Independently re-measuring at this head commit with the same AR1-random-walk generator the PR's own tests use (`phi=0.0`, 1000 independent trials, well above the 200-trial sample the tests use): n=201 gives mean=0.508, sd=0.078 (inside the claimed band), but n=101 gives mean=0.513, sd=0.116 — about 30% above the claimed upper bound. At n=101 the empirical 95th/99th percentile is 0.72/0.79, so a memoryless series regularly reads well above the "0.50-0.55" band the same docstring calls consistent with no memory. The mean-bias fix itself (raising `_HURST_DFA_MIN_SCALE` to 8) is confirmed correct and reproduces the disposition's own numbers; only the sd figure in the added caveat is off. Nothing consumes `hurst` today (confirmed: no reader in `scheduler/` or any Python gating path), so this blocks nothing, but the caveat exists specifically to stop a future reader from over-interpreting the number, and an understated spread partly defeats that purpose at the exact sample size the enriched backtest column runs at.
**Invariant:** a stated null-distribution caveat on an advisory metric must bound the actual empirical spread at the sample size it describes, not just the sample size that happens to look best.
**Must survive:** (1) 1000+ independent memoryless (phi=0) trials at n=101 give an empirical sd within the caveat's stated range; (2) the same check at n=200; (3) the caveat's implied "not-persistence" band (~0.50-0.55) covers at least the 90th percentile of the memoryless distribution at n=101, not just the mean.
**Plain simple English:** The code comment says how much a "no trend" reading can vary, but the real variation is bigger than the comment says at the shortest data length. A reading of 0.72 can happen by chance, well above what the comment calls safe. Update the comment with the correct range.

---

richkuo:
Addressed review feedback (richkuo · 2026-08-07T23:42:02Z, Sonnet 5) in aea81737.

### Fixed

1. **Null-distribution caveat understates the sd at n=101, the enriched column's exact frame size.** — Independently re-measured on this head commit with 3000 memoryless (phi=0) trials: n=201 gives mean=0.507, sd=0.081 (matches the old "sd ~0.06-0.09" claim); n=101 gives mean=0.513, sd=0.117, about 30% above the old claimed upper bound, with 95th/99th percentile ~0.72/0.82. Updated the docstring caveat with accurate per-frame-size sd figures and made explicit that a single reading up to and above 0.7 at n=101 is not reliable evidence of persistence on its own (`shared_strategies/open/indicators_core.py:271-282`). Added three regression tests pinning the corrected sd and percentile figures at both frame sizes so a future estimator change that silently widens or narrows the null spread gets caught (`shared_strategies/open/test_indicators_core.py:663-711`).

No other findings in this review — it carried only this one Recommended Optional item.

**No behavior change:** `hurst_exponent`'s estimator logic, `_HURST_DFA_MIN_SCALE`, and all thresholds are untouched — this is a docstring-accuracy and test-coverage fix only. `regime.py`/`map_composite_label` were not touched; `shared_tools/test_regime.py` (57 tests) and `backtest/tests/test_regime_enriched_features.py` (23 tests, the actual `hurst`-column consumer) both pass unchanged.

### Verification

- `uv run --no-sync python -m pytest shared_strategies/open/test_indicators_core.py -k hurst` — 16 passed (13 pre-existing + 3 new).
- `uv run --no-sync python -m pytest shared_tools/test_regime.py` — 57 passed.
- `uv run --no-sync python -m pytest backtest/tests/test_regime_enriched_features.py` — 23 passed.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2904 passed, 0 failed.
- `go -C scheduler build` and `go -C scheduler test ./...` — both pass (no Go files touched; consistent with the issue's no-Go-schema-change requirement).

Per this run's instructions, no `@claude` re-review is triggered — only non-blocking Optional feedback was addressed and this run stops here for a fresh in-session subagent re-review.

---
Created with LLM: Sonnet 5 | high | Harness: milestone-pipeline

richkuo:
LGTM

### Recommended Optional

1. **A DFA docstring still says the min segment scale is 4 bars, contradicting the constant that was raised to 8 in this same fix cycle.**
`_hurst_dfa_scales` (`shared_strategies/open/indicators_core.py:208`) reads `"""Log-spaced integer segment sizes for DFA, from 4 bars to n/4 bars."""`, but `_HURST_DFA_MIN_SCALE` two lines above it (`indicators_core.py:203`) was raised from 4 to 8 in the first fix commit specifically to remove the short-scale upward bias on memoryless data, and the comment directly above the constant documents that change. The function body reads `_HURST_DFA_MIN_SCALE`, not a literal 4, so behavior is correct — only the docstring text is stale, giving a reader two contradictory statements about the estimator's actual scale floor a few lines apart in the exact area two review rounds already focused on.
**Invariant:** a docstring stating a numeric parameter must match the constant the function body actually reads, especially in a region a prior review cycle already revised.
**Must survive:** (1) grep for the literal "4 bars" phrase finds no docstring claim that contradicts `_HURST_DFA_MIN_SCALE`'s current value; (2) a future change to `_HURST_DFA_MIN_SCALE` doesn't require hunting separate prose copies of the number to stay accurate.
**Plain simple English:** A code comment says segments start at 4 bars. The real setting, changed earlier in this same fix, is 8 bars. The comment is out of date and could confuse someone reading the code later. Update the comment to say 8.

---

richkuo:
Addressed review feedback (richkuo · 2026-08-07T23:58:09Z, Sonnet 5) in 54cb073.

### Fixed

1. **Stale DFA scale-floor docstring** — confirmed: `_hurst_dfa_scales` (`shared_strategies/open/indicators_core.py:208`) said segments run "from 4 bars", but `_HURST_DFA_MIN_SCALE` (line 203) was raised to 8 in this fix cycle's first commit. The function body already read the constant, not a literal 4, so only the docstring was wrong. Rather than hardcoding "8" (which would go stale again on the next tuning pass), the docstring now names the constant itself, satisfying both of the finding's own must-survive checks: no prose claim can contradict the current value, and a future change to `_HURST_DFA_MIN_SCALE` needs no separate prose hunt.

No `Needs Fixing` or `Requires Human Review` items were raised. CI (`go`) was green at review time; `python` was still in progress and is unaffected by this docstring-only change.

### Verification

- `uv run --no-sync python -m pytest shared_strategies/open/test_indicators_core.py -k hurst` — 16 passed.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2904 passed, 0 failed.
- `uv run --no-sync python -m pytest shared_scripts/test_regime_wiring.py` — 21 passed (explicit invoke, not in `testpaths`).
- `go -C scheduler build .` and `go -C scheduler test ./...` — both pass (no Go files touched).
- Zero-behavior-change invariant confirmed untouched: this is a docstring-only edit, no logic, thresholds, `map_composite_label`, or #1085 directional certifications touched.

No merge conflicts with `main` — the branch already contains latest `main` and `git status` showed only this one file dirty.

Per this run's instructions, no `@claude` re-review is triggered — only a non-blocking Optional item was addressed, and this run stops here for a fresh in-session subagent re-review.

---
Created with LLM: Sonnet 5 | high | Harness: milestone-pipeline

richkuo:
LGTM

---

PR #1420, opened 2026-08-08
feat(#1410): add Hurst entry-gate calibration study, report only [C41, Opus 5, xhigh]

### LINKED ISSUE #1410: [C41] Calibration study for a future Hurst-based entry gate (report only, no live changes)

richkuo:
## Plain simple English

We wanted to know if a market "trendiness" number could tell our bots when to skip a trade. This adds an offline study that measures it and writes a report. The answer is no. The setting looks good only because it makes the bots trade less, not better. Nothing about the live system changed.

## Summary

Adds a re-runnable research one-shot plus its committed report at the exact contract path `backtest/research/hurst_gate_calibration.md`, which the follow-up gate issue reads thresholds from.

**Verdict: `INCONCLUSIVE`.** 0 of 30 swept configurations reached Benjamini-Hochberg significance at alpha=0.05. 16 met every economic condition and failed only on significance — their drawdown and chop-loss reductions come from trading less, not from trading better. The report says so and tells the follow-up work not to build the gate on this evidence.

Report-only. Zero scheduler, config, gating, sizing, or live-path changes. The `#1409` advisory-only invariant stamped beside `hurst_exponent` holds: the gate and size arms are offline simulations whose whole purpose is to decide whether such a gate should ever exist.

### Files

| File | What |
|------|------|
| `backtest/research/hurst_1410_gate_calibration.py` | the study |
| `backtest/research/hurst_1410_gate_calibration.json` | aggregate output of the committed full run |
| `backtest/research/hurst_gate_calibration.md` | the rendered report (contract path) |
| `backtest/tests/test_hurst_1410_gate_calibration.py` | 96 unit tests over the pure helpers |
| `docs/backtesting-registry.md` | Research one-shot row, same PR per the registry SSoT rule |

`indicators_core.py`, `eval_windows.py`, `regime_stats.py`, `backtester.py` are import-only — untouched.

### Reuse, never reimplemented

- `hurst_exponent` from `shared_strategies/open/indicators_core.py` (the #1409 SSoT). NaN stays NaN: its own bucket, never coerced to 0.5, never a size opinion, and it holds the gate state rather than arming or disarming it.
- `benjamini_hochberg` from `backtest/regime_stats.py` — the same helper `auto_suggest.py` imports. Raw and corrected results appear side by side.
- `eval_windows` `WINDOWS` (is / oos / 2023 / 2024 / 2025H1) verbatim, plus `DATASETS`, `DEFAULT_CAPITAL`, `run_leg`, `trade_samples_from_results`, and the exclusive-end window slice.
- The two platform axes stay uncoupled: data source `PLATFORM=binanceus`, fee model `FEE_PLATFORM=hyperliquid`, both imported.

### Method

1. **Part A** buckets ungated trades by H at entry (`<0.45`, `0.45-0.50`, `0.50-0.55`, `>=0.55`, `NaN`) per family per Hurst window, after pooling and deduplicating on `(strategy, symbol, timeframe, entry_date)` with windows iterated in chronological start order, first wins.
2. **Part B** is real `Backtester` re-runs with entry signals zeroed while a hysteresis gate is disarmed; `-1` closes are never masked. Backtester kwargs are identical to `run_leg`.
3. **Part C** re-compounds the same ungated trade sequence with `m = clamp(1 + gain x e, 0, 1.5)`, NaN giving exactly 1.0. Its drawdown is trade-granular; the report says so beside the table so it is never compared to a Part B bar-level drawdown.
4. One-sided seeded permutation tests (10000 permutations, seed 1410), all 30 hypotheses corrected together.

### Correctness guards

- **Harness identity.** Every leg asserts its ungated arm reproduces `eval_windows.run_leg` exactly — 150 of 150 verified in the committed run. Without it the gated arms could silently be scored on a different harness than the baseline they are compared against.
- **Look-ahead.** Rolling H at bar `i` uses closes `[i-W+1, i]`; the decision series is that shifted one bar, so a bar-N signal reads H through N-1 (one bar more conservative than the live regime gate). A trade is stamped at its fill bar with its signal bar's decision value. A unit test mutates bar N's close and proves the bar-N decision does not move, with a paired assertion that the unshifted series *did* move.
- **Warmup padding.** H is computed on a frame padded before the window, then sliced with the exclusive-end rule before signals are computed, so the ungated arm stays byte-identical to `run_leg`. `end=None` skips the end slice.
- **Volume floors** (20 suppressed / 30 kept pooled) mean a "trade nothing" configuration can never win.
- **BH denominator** covers every tested hypothesis, including configs with no testable contrast, so dropping them can never make the correction more permissive.
- **Recommendation is mechanical.** `decide_recommendation` is a pure function of the pre-registered rules, and `--render-only` re-renders the report from the committed JSON — verified byte-identical.

### Notes for the reviewer

1. `atr_band_revert` emits only `+1` opens on the spot path, so ungated it opens once and never exits. It is paired with `tiered_tp_atr` plus `stop_loss_atr_mult=2.0` so it produces real round trips. Both arms of every comparison use that same stack, so the contrast is unaffected. This is recorded in the report's family-mapping section, as the plan required.
2. The `NaN` bucket is empty in every table. That is a property of the padded-frame harness, not of the estimator. The report states this explicitly so a future implementer does not conclude NaN never happens live, where fetch depth is bounded.
3. Config ids use `/` as their separator, because a `|` inside a Markdown table cell splits the row even in a code span. A test asserts no id contains a pipe and that every rendered table has a consistent cell count.

### Dependency merge order

No blockers. `#1409` already landed on `main` through merged PR #1419 (commit `bfcf4cc9`), which carries the estimator. This branch is cut from that commit.

### Verification

- `uv run --no-sync python backtest/research/hurst_1410_gate_calibration.py --jobs 4 --write-report` — full run; the committed JSON and report come from it. 150 legs, 1350 gated arms, 150/150 harness-identity checks passed. Re-running reproduces every p-value (fixed seed).
- `uv run --no-sync python backtest/research/hurst_1410_gate_calibration.py --render-only` — report byte-identical to the full run's.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3000 passed.
- `uv run --no-sync python -m py_compile backtest/research/hurst_1410_gate_calibration.py`.
- `go -C scheduler build ./...` and `go -C scheduler test ./...` — green (no Go changed).

Closes #1410

---
Created with LLM: Opus 5 | high | Harness: milestone-pipeline


richkuo:
LGTM

### Recommended Optional

1. **The report asserts the empty `NaN` bucket is a harness property, but nothing in the harness guarantees the padding that claim rests on**

   `backtest/research/hurst_gate_calibration.md:48` (rendered from `hurst_1410_gate_calibration.py:1197-1200`) states: "Every leg computes rolling H on a frame padded well before the window start, so H is defined on every scored bar." That holds only when the cached OHLCV series reaches at least `max(HURST_WINDOWS)+2` bars before the earliest window start. `first_needed` is `min(WINDOWS[w][0])`, which resolves to `2023-01-01` (`:1423`), and the `2023` window itself starts on `2023-01-01`. I measured the cache in this checkout: all six datasets begin exactly at `2023-01-01`, so the available lead is zero. Running the head commit's estimator against that cache, the first 129 / 257 / 513 bars of the `2023` window carry NaN stamps at W=128/256/512. Real trades land in that span. On `BTC/USDT 1h` at W512 alone I counted 21 `atr_band_revert`, 4 `mean_reversion` and 1 `momentum` entries, so a re-run there produces a populated `NaN` bucket and different Part A tables, with no warning emitted. The committed run's own lead depth is neither asserted in code nor recorded in `run_summary`, so a reader cannot tell which case produced the shipped tables. Consider asserting the per-dataset lead, or recording it in the JSON and narrowing the sentence to name its condition. `test_report_explains_why_an_empty_nan_bucket_is_expected` (`backtest/tests/test_hurst_1410_gate_calibration.py:818`) pins the present wording, so it moves with any fix. The `INCONCLUSIVE` verdict is unaffected either way.

   **Invariant:** A committed report may only assert a property that the harness enforces or records.
   **Must survive:** (1) a cache whose first bar equals the earliest window start, where the run must warn or fail instead of silently emitting NaN-bucket trades; (2) a cache carrying at least `max(HURST_WINDOWS)+2` bars of lead, where the `NaN` bucket stays empty and no warning fires; (3) `--windows oos` alone on a cache that does have lead, which must not raise a false alarm.
   **Plain simple English:** The report says the "unknown" group is empty because each test always has enough earlier price history. The price data in this repository has no earlier history, so that group would not be empty. The study does not test this condition or record it. Add a check, or state clearly when the sentence is true.

2. **The rolling-Hurst npz cache can be reused across runs that need different coverage**

   `hurst_1410_gate_calibration.py:1442` accepts a cached array whenever `len(cached[key]) == len(frame)`, and the key at `:1440` is `symbol|timeframe|hw`. The array's NaN coverage depends on `first_needed`, which is derived from `--windows` (`:1423`), and that value appears in neither the key nor the guard. Two runs over one dataset therefore yield same-length arrays with very different coverage. I measured both paths on a 3000-bar series: the late-`first_needed` array carries 2496 NaNs and the early one carries 127, and the length guard passes in both directions. So `--out-dir X --windows oos` followed by `--out-dir X` over all windows silently reuses the near-empty series. Because a NaN reading holds the gate state (`hysteresis_mask`, `:238`), every gated arm then reproduces its ungated arm and every trade falls into the `NaN` bucket. That output looks like a clean "no effect" negative result, and the stale cache leaves no other trace. Consider folding `first_needed` into the cache key, or storing it beside the arrays and recomputing on mismatch. The committed run passed no `--out-dir`, so the shipped artifacts are unaffected.

   **Invariant:** A cache hit must be indistinguishable from a fresh recompute.
   **Must survive:** (1) `--out-dir X --windows oos` then `--out-dir X` over all windows, where the second run recomputes or refuses; (2) the same selection run twice, where the cache still hits; (3) a dataset that gains bars between runs, which must recompute.
   **Plain simple English:** The study can save Hurst values into a folder to save time. It only checks that the saved list has the correct length. If you run one date range and then a wider one, it reuses the wrong saved values and reports "no effect" although it measured almost nothing. Put the date range into the saved name.

---
Validated with LLM: Opus 5 | high | Harness: milestone-pipeline

richkuo:
Addressed review feedback (Opus 5 review · 2026-08-08T01:56Z) in c1f145db. Both items were `Recommended Optional`; both are now implemented. No blocking findings were raised and CI was green on the reviewed commit.

### Fixed

1. **Rolling-Hurst npz cache could be reused across runs needing different coverage** — CONFIRMED and fixed. The guard compared array length only (`hurst_1410_gate_calibration.py:1442` on the reviewed commit), while an array's NaN coverage comes from `first_needed`, which is derived from `--windows`. Each cached array now carries an int64 metadata row — `cache_meta` (`:251`) storing `first_needed`, first bar, last bar, length — and `cache_entry_is_usable` (`:262`) admits a hit only when the bar series matches exactly AND the cached `first_needed` is no later than this run's. A deeper cached array is a superset of the requested coverage, so it stays a legitimate hit; a shallower one recomputes. Verified end-to-end, not only in unit tests: `--out-dir X --windows oos` wrote a 13142-NaN array, and the following `--out-dir X --windows 2023,oos` recomputed to 6567 NaNs (the old guard would have reused it); an identical repeat run hit the cache and reproduced the recompute's `buckets`, `configs` and `decision` byte-for-byte.

### Corrected scope (partial)

1. **The report asserted a padding property the harness did not enforce or record** — the INVARIANT is confirmed and fixed; the finding's supporting MEASUREMENT is refuted.

   Refuted part: the claim that "all six datasets begin exactly at `2023-01-01`, so the available lead is zero" does not hold in this checkout. `load_cached_data(..., exchange_id="binanceus")` returns BTC/USDT and ETH/USDT from `2020-01-01` and SOL/USDT from `2020-09-18`. Measured lead before `2023-01-01` is 26270 / 6571 / 26270 / 6571 / 20017 / 5006 bars against a 514-bar requirement, so the committed run's empty `NaN` bucket was genuine, and the counted `atr_band_revert` / `mean_reversion` / `momentum` entries in a NaN span do not exist. Regenerating the artifacts from a full run reproduced every bucket count, every config, every p-value and the `INCONCLUSIVE` verdict bit-for-bit — only the new warm-up block and the wall time moved.

   Fixed part: the invariant "a committed report may only assert a property the harness enforces or records" is right, and the study did neither. The run now measures the per-dataset lead (`warmup_lead_bars`, `:217`), audits it against `required_lead_bars` = `W + 2` (`:211`, and the audit at `:223`), stores the result in `run_summary.warmup`, prints `warm-up OK: min lead 5006 bars before 2023-01-01 (need 514)` or a shortfall WARNING, and renders the `NaN`-bucket paragraph from that measurement (`render_nan_bucket_note`, `:1164`). The three must-survive cases are unit-tested: zero lead is recorded insufficient, ample lead raises no false alarm, and a `--windows oos`-only run is judged against the Hurst windows it actually selected. `--render-only` over an older JSON with no audit now says the property is not attested rather than inheriting the claim. `test_report_explains_why_an_empty_nan_bucket_is_expected` was replaced by three tests pinning the sufficient / short / unattested branches.

### Unchanged invariants

The study stays report-only: no scheduler, config, gating, sizing or live-path change, `hurst_exponent` still imported from `shared_strategies/open/indicators_core.py` (#1409 advisory-only), NaN still its own bucket and never 0.5, `benjamini_hochberg` still from `regime_stats.py`, `eval_windows.WINDOWS` / `DATASETS` / `PLATFORM` / `FEE_PLATFORM` untouched, and `hurst_gate_calibration.md` still ends with `## Recommendation` containing exactly `INCONCLUSIVE`.

### Verification

Full pytest suite under `-n auto`: 3017 passed. `go build` + `go test ./...` in `scheduler/`: green. `py_compile` on the study script: clean. Artifacts regenerated from an actual full run (150 legs, 1350 gated arms).

---
Created with LLM: Opus 5 | high | Harness: milestone-pipeline

richkuo:
LGTM

---

Issue #1344, opened 2026-07-13
[withdrawn: mistakenly filed against upstream] #1344


Xorit:
Withdrawn by Xorit on 2026-07-20.

This issue was created accidentally by automation auditing a private derivative. Xorit does not contribute to or operate against richkuo/go-trader. The original content has been removed, and no action is requested from upstream.

richkuo:
## Validation update (verified against current `main`)

**Complexity: 62/100** — scope: medium (6 dispatch sites in `scheduler/main.go` + tests); risk: high (live auto-protective stop/close path); uncertainty: low (fix pattern already established by #1150/#1269/#1270)

The reported behavior is **confirmed**, with corrected citations and two nuances.

### Corrected citations

1. Notional-cap branch: `scheduler/risk.go:592-595` (the cited 531-535 is the kill-switch margin branch).
2. Whole-strategy skip: `scheduler/main.go:1938-1943` — the `continue` skips the entire check-script dispatch (`main.go:1948+`), including close/reduce signal evaluation.
3. Skipped maintenance: perps trailing-SL/TP-ratchet at `main.go:2251-2320`, manual at `main.go:2649-2685`.

### Nuances

1. **Paper positions:** the trailing SL is enforced entirely in-process, so over-cap it stops being enforced at all.
2. **Live HL positions:** the resting on-chain SL trigger remains armed (not fully naked), but trailing/ratchet tightening and protection sync stop running.
3. **Exception:** a latched circuit breaker with an open HL perps position takes the `cbManageOnly` path (`main.go:1938`), which bypasses the notional skip — a CB-latched strategy keeps SL/TP management running while healthy strategies do not.

### Recommended approach

Replace the strategy-level `continue` with a fourth per-signal hold in the existing entry-hold stack — the same `pausedBlocksSignal(...)` pattern used by paused (#1150), daily-loss (#1269), and exposure-cap (#1270) — at all 6 dispatch sites (`main.go:1971, 2021, 2069, 2171, 2229, 2545`). Position-reducing actions and the Signal==0 manage-only paths then pass by construction. Per-strategy hold log should follow the #1269/#1270 per-site convention (the cycle-level WARN at `main.go:1280` already exists).

---
Created with LLM: Fable 5 | high | Harness: Claude Code

Issue #1268, opened 2026-07-09
[C70] Position sizing ignores stop distance — equal-notional trades carry unequal dollar risk; add opt-in risk-per-trade sizing


richkuo:
**Complexity: 70/100** — scope: large (sizing helpers, live+paper perps sizers, hot-reload validation, backtester parity, init wizard); risk: high (money path — every perps open's dollar exposure); uncertainty: moderate (stop-owner resolution at sizing time, flip/scale-in interaction)

## Problem

Position sizing is notional-based, not risk-normalized. Every perps open sizes from cash and a leverage multiplier, never from the distance to the stop — so two strategies with equal notional but different ATR stops carry different dollar risk per trade.

Verified sizing paths:

- `scheduler/config.go:766-784` — `PerpsOpenNotional`: `notional = cash × sizing_leverage`, or margin-space `min(margin_per_trade_usd, cash) × exchange_leverage` when `margin_per_trade_usd` is set. These are the only two formulas.
- `scheduler/config.go:567-568` — the two sizing fields (`sizing_leverage`, `margin_per_trade_usd`); `scheduler/config.go:789-791` — `ComputePerpsOpenNotional` wrapper.
- `scheduler/portfolio.go:907,949` — live sizer `perpsLiveOrderSize` budgets via `PerpsOpenNotional`; paper sizers at `scheduler/portfolio.go:1333,1529` do the same; scale-in default at `scheduler/main.go:2159`.
- Meanwhile the stop side is fully resolved elsewhere and never consulted for size: `EffectiveStopLossPct` at `scheduler/config.go:831` plus the seven mutually-exclusive stop owners (ATR-mult owners defer to `Position.EntryATR`; system default `DefaultStopLossATRMult = 1.0` at `scheduler/config.go:803`).
- `backtest/backtester.py:2082,2138,2342` — backtester sizes `cash × entry_fraction`, also with no stop-distance input.

Concrete effect: strategy A with a 1.0×ATR stop and strategy B with a 3.0×ATR stop on the same coin, same `sizing_leverage`, risk 1× vs 3× the dollars per trade. The industry-baseline alternative (fixed-fractional: risk a constant fraction of equity per trade, quantity derived from stop distance) is not expressible in the current config. Prior sizing work (#497 `sizing_leverage` decoupling, #518 margin-space sizing — both closed) stayed in notional/margin space and did not address risk normalization.

## Goal

[...]

PR #1421, opened 2026-08-08
feat(#1411): add default-off Hurst entry gate and persistence-scaled sizing [C97, Fable 5, high]

### LINKED ISSUE #1411: [C97] Hurst entry gate and optional persistence-scaled position sizing, shipped default-off

richkuo:
## Plain simple English

This adds an optional filter that can hold new trades, or make them smaller, when the market shows little persistence. It is **off by default** and does nothing until an operator turns it on for one strategy. The research study behind it found **no proof that it helps**, so this ships the machinery with **no suggested settings at all**. Existing behaviour does not change.

---

## Calibration verdict — read this first

The study that this gate depends on, `backtest/research/hurst_gate_calibration.md` (#1410, merged in PR #1420), returned:

> **Recommendation: INCONCLUSIVE**
>
> No configuration of the 30 tested passed the pre-registered acceptance rule. 0 reached Benjamini-Hochberg significance at alpha=0.05; 16 met every economic condition but failed only on significance. A Hurst entry gate is not supported by this evidence.
>
> **"Do not build a Hurst entry gate on this evidence. Re-run this study before revisiting the question."**

Issue #1411 carries the owner's explicit reaffirmation to build the mechanism anyway, default-off, under three binding conditions. All three are met here:

1. **Default-off, per-strategy opt-in only.** Nothing auto-enables. A config without `hurst_gate` behaves byte-identically to pre-#1411.
2. **No recommended thresholds ship anywhere.** `scheduler/config.example.json` gets no `hurst_gate` block; no default bound exists anywhere in the code. A test (`TestConfigExampleShipsNoHurstThresholds`) pins that promise.
3. **The verdict is stated prominently** — here, and in `docs/ARCHITECTURE.md`, and in the header comment of `scheduler/hurst_gate.go`.

This PR does not authorize enabling the gate on any strategy.

Closes #1411.

## Summary

A per-strategy `hurst_gate` block with two modes, layered as a **standalone gate on top of** the `allowed_regimes` label gate:

- **`mode: "gate"`** — a hysteresis state machine (armed/disarmed) over the composite regime window's `metrics["hurst"]`. While disarmed, position-increasing signals are held.
- **`mode: "size"`** — a stateless multiplier `m = clamp(|H − 0.5| / 0.15, size_floor, 1.0)` applied to the computed open size. `m ≤ 1.0` always — the gate can only shrink an open, never grow one.

### 1. Dispatch sites (all six, each paired with `pausedBlocksSignal`)

`advanceHurstGate` is wired immediately after every `applyRegimeGate` call in `scheduler/main.go`: OKX spot, Robinhood spot, default/BinanceUS spot, OKX perps, HL perps, TopStep futures. It runs **unconditionally and signal-independently**, so the hysteresis latch advances on every cycle including `Signal==0` manage cycles.

Every hold is classified through `pausedBlocksSignal`, exactly as the #1150 pause / #1269 daily-loss / #1344 notional-cap arms do. Held: fresh opens, scale-in adds, bidirectional flips. Always passing: closes, `closeFraction>0` registry actions, pure-close directional exits, trailing SL, the TP ratchet, protection sync, paper SL/TP simulation, hedge sync (#1159), and every kill-switch path.

Three source-scan tests in `hurst_gate_wiring_test.go` pin this: the arm count must equal the `applyRegimeGate` count, each arm must sit inside its own dispatch group and read the same `posQty` as its label gate, and no use of the hold flag may appear without `pausedBlocksSignal`. A seventh dispatch site added later without a Hurst arm fails the build.

### 2. `on_failure` mirrors #1278

`resolveHurstGateOnFailure(sc, rc)` is the single accessor — per-strategy `hurst_gate.on_failure` > global `regime.hurst_gate_on_failure` > `"open"`. The raw fields are never read at a decision site. Vocabulary rejection is styled after the `regime.gate_on_failure` validation at `scheduler/config.go:2682-2701`.

Fail-closed is **flat-only**, the exact `regimeBlocksOpen` shape (`scheduler/regime.go:102-110`): an unknown H under `on_failure="closed"` holds a fresh open and never touches management of an open position. Both directions are tested, including the inverse case.

A **known** disarmed state carries no flat-only restriction — `pausedBlocksSignal` at the site already scopes it to adds and flips, so closes still pass.

### 3. The load-bearing composite-classifier check

`metrics["hurst"]` is emitted **only** by `latest_regime_composite` (`shared_tools/regime.py`); the default `adx` path never emits it. `validateHurstGateConfigs` therefore **rejects at config load** any `hurst_gate` whose resolved window is adx-classified, missing, in single-window mode, or with regime detection disabled.

Without that check, H would be permanently absent and `on_failure` would silently govern every cycle — under `closed`, a permanent entry block indistinguishable from "the feature is off". Verified end-to-end: the daemon refuses to start on an adx-window config with a named error.

### 4. Sizing composition order

The multiplier is applied where size is **computed**, before any order is placed, and nothing rescales after the exposure-cap check or after execution:

- **HL/OKX perps (live + paper), flips, shared-wallet pool budget** — `PerpsSizing.EntrySizeMult`, applied inside `PerpsOpenNotionalSized`, the single point where an open's notional is computed.
- **Spot and futures** — an `openSizeMult` parameter on new additive `Execute{Spot,Futures}SignalWithFillFeeSized*` variants. **The existing public signatures are unchanged**, so no existing executor test needed rewriting.
- **Live order sizers** — the RH/OKX spot buy notional and the TopStep contract count.
- **Scale-in adds (#873)** — applied to the decided add quantity (so an explicit `scale_in.add_notional_usd` is scaled too). The frozen `RiskAnchorPrice` geometry is untouched: the multiplier scales quantity only. The caps and ATR spacing are evaluated on the unscaled intent, so shrinking can only leave a guardrail satisfied.

**With `risk_per_trade_pct` (#1268):** effective risk = `pct × m`, and the `cash × exchange_leverage` cap inside `PerpsRiskBasedNotional` still applies afterward. Because `m ≤ 1`, a scaled open can never breach a cap the unscaled one satisfied. The mode stays mutually exclusive with `sizing_leverage` / `margin_per_trade_usd` / `scale_in`, so the multiplier composes with exactly one sizing mode at a time and replaces none of them.

**Futures floor-to-zero:** the contract count is scaled *before* the whole-contract floor, so a shrunk size that floors to 0 **refuses the open** rather than rounding up to 1.

Every out-of-range multiplier (0, negative, NaN, >1.0) resolves to **1.0** at read time — a mis-threaded value can never grow an open.

### 5. Hysteresis persistence

`StrategyState.HurstGate` persists to a new `strategies.hurst_gate_state` TEXT column as JSON, following the `RiskState` save/load shape. It is keyed by a SHA-256 of `(window_key, min, max, disarm_min, disarm_max)`; **a threshold edit discards the latch** rather than reinterpreting a disarmed state under a new band. `mode` / `on_failure` / `size_floor` are excluded from the key on purpose — they change the decision taken from a state, never the state series itself. An absent or NaN H never transitions state (the #1410 NaN policy: unknown is never 0.5).

### 6. Hot-reload policy (while open: **allowed**)

`hurst_gate` and `regime.hurst_gate_on_failure` are **not** added to `validateHotReloadCompatible`'s blocked list. SIGHUP adopts them always, including while a position is open — the #1278 precedent. The justification: gate mode only holds position-increasing signals from the next cycle, and size mode only scales an order at the moment it is placed. Neither rewrites any persisted position or order state, so no protection can be stranded. A threshold edit resets the hysteresis via the key hash, and that reset is safe by design (unknown → `on_failure` governs until the next valid reading).

### 7. Config version decision

**`CurrentConfigVersion` stays 17.** Every new key is additive and optional; an absent key changes nothing and an older config loads unchanged. No migration is warranted. The two SQLite migrations are additive `ALTER TABLE ADD COLUMN` statements on the existing idempotent list.

### 8. #1409 advisory-comment revocation

#1409 stamped the metric advisory-only. This change revokes that **for gating and sizing**, in the same PR that makes it an input:

- `shared_tools/regime.py` — the "Advisory metric only — never read by `map_composite_label`, gating, or sizing" comment is rewritten to state that the metric is now a declared input to the Go-side `hurst_gate`, while remaining advisory for classification.
- `shared_tools/test_regime.py` — the `#1409` section header is updated to match.
- A wiring-inventory test asserts the revocation landed **and** that `map_composite_label` is still genuinely blind to the metric.
- The `#1410` report is a frozen historical artifact and is **not** edited. `indicators_core.py`'s docstring never restated the invariant, so it needed no change.

### 9. Backtest parity

`backtest/hurst_gate.py` mirrors the state machine, formula and validation. `backtester.py` computes rolling H from the #1409 `indicators_core.hurst_exponent` SSoT (never reimplemented) over `max(200, 2*maxPeriod−1+10)` closes — the live regime fetch depth — then shifts it one bar, so a signal at bar N reads H computed through N−1 and still fills at N+1's open.

`run_backtest.load_strategy_config` **rejects loudly** what the engine cannot model, following the `tiered_tp_atr_live_regime_dynamic` precedent: regime disabled, no windows, unknown window, non-composite window, a non-**primary** window (the engine classifies only the primary window, #1058), `type=options|manual`, or any validation failure. A disabled block passes through unchanged.

`parity_diff.py` gains per-bar `bt_hurst`, `bt_hurst_state`, `bt_hurst_holds`, `bt_hurst_size_mult`.

**Formula divergence note:** the #1410 study swept `m = clamp(1 + gain·e, 0, 1.5)`, which can *exceed* 1.0. This implements the issue's `clamp(|H−0.5|/0.15, floor, 1.0)`, which never does. The study shipped no recommendation, so the issue's formula governs — on both engines.

## Two findings worth flagging

**1. H is not bounded to (0, 1) at runtime.** Measured, not assumed: DFA returns **H ≈ 2.0033** on a perfectly smooth linear ramp. The *config* bounds are validated to (0,1) because that is the only range an operator can sensibly gate on, but both comparators and the size clamp are written to stay correct for any finite H, and tests pin that asymmetry on both engines so nobody "tightens" them on a false assumption.

**2. A silent CLI seam.** `run_backtest.py` copies only an explicit `stop_keys` allowlist of resolved config keys into the Backtester. A key missing from that list is dropped **silently** — `load_strategy_config` resolved the gate, every unit test passed, and the backtest still reported entries the daemon would have held. Caught by an end-to-end smoke run, not by the unit tests. `hurst_gate` is now in the allowlist and two regression tests pin both the allowlist entry and the receiving signature.

## Scope note

Manual actions (`manual.go`) are deliberately **not** gated. `type=manual` is rejected at config load, and an operator-initiated manual open on a perps strategy is a deliberate override, consistent with the gate being a signal-path filter (the #1150 pause shape) rather than a hard portfolio cap (the #1344 notional-cap shape, which does refuse manual opens).

## Verification

- `go build ./...`, `gofmt` clean on every touched file.
- `go test ./...` and `go test -race ./...` — green.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — green.
- `uv run --no-sync python -m pytest shared_scripts/test_regime_wiring.py` — green (run explicitly; not in `testpaths`).
- **End-to-end daemon smoke:** an adx-window config is refused at startup with the named `#1411` error; a composite-window config loads and renders the `hurst=?` status marker.
- **End-to-end backtest:** same config, gate off vs on → 57 trades / +25.56% vs 55 trades / +12.63%. The gate demonstrably changes behaviour through the real CLI path.
- `docs/backtesting-registry.md` row added for the new module, per the upkeep rule.

New tests: `scheduler/hurst_gate_test.go`, `scheduler/hurst_gate_wiring_test.go`, `backtest/test_hurst_gate_backtest.py`, plus additive extensions to `shared_scripts/test_regime_wiring.py`. No existing test assertion was weakened; the only edits to existing test files are mechanical argument updates for the executor signatures the plan anticipated.

## Dependency merge order

Both predecessors are already merged into `main`, so this branches directly off it:

1. #1409 → PR #1419 (merged) — Hurst estimator + metric surfacing.
2. #1410 → PR #1420 (merged) — calibration study, report only.

---
Created with LLM: Opus 5 | xhigh | Harness: milestone-pipeline


richkuo:
LGTM

### Recommended Optional

1. **Open-time stamp comments claim H lies in (0, 1) exclusive, contradicting the PR's own range note.**
   `scheduler/portfolio.go:126-133` (the `HurstAtOpen`/`HurstSizeMult` field comment) and the `scheduler/db.go` migration comment for `hurst_at_open` justify the 0-as-unstamped sentinel with "H is in (0,1) exclusive". The same PR's range note (`scheduler/hurst_gate.go:293-301`) and the pinned test `test_composite_hurst_is_a_finite_float_but_is_NOT_bounded_to_zero_one` establish that the runtime metric exceeds 1 on a smooth series (~2.0), and `stampHurstGateAtOpenIfOpened` stamps any known finite H, so a stored `hurst_at_open` above 1 is a legal value. Verified by reading both files at the head commit and running the Python test (passes, `saw_above_one` asserted). The sentinel itself stays sound because the DFA slope is positive for every valid reading; the justification should say "H > 0", and the comment's current wording invites a future edit that clamps or rejects stamps above 1.
   **Invariant:** A comment that justifies a sentinel value states the true range of the values sharing that column.
   **Must survive:** (1) a stamped H of 2.0033; (2) a future edit that validates stamps against the commented (0,1) range; (3) the unstamped 0 default.
   **Plain simple English:** Some code comments say the stored measurement is always between 0 and 1. Other parts of the same change show it can be larger than 1. The code itself is correct. Correct the comments so a future change does not trust the wrong range.

2. **`parity_diff.py` computes the Hurst frame depth without the regime windows spec, so it can diverge from both engines.**
   `backtest/parity_diff.py:733` calls `hurst_live_frame_bars(None, cfg.regime_period)`. The Backtester passes `self.regime_windows_spec` to the same helper, and the live fetch depth takes the max period over all configured windows (`scheduler/regime_multi_window.go:293-296` — verified against the formula the helper mirrors). For any regime window period above 95 (where `2*p-1+10 > 200`), the parity tool computes H over a different frame length than the engine and the daemon, and its `bt_hurst`/`bt_hurst_state` columns then report spurious disagreements — the opposite of the tool's purpose. `ParityConfig` has no `regime_windows_spec` field today, and `load_strategy_config` already returns one, so it can be threaded through `config_from_live_config`.
   **Invariant:** The parity tool replays the gate over the identical input series length the engine and the live daemon read.
   **Must survive:** (1) a config whose largest window period exceeds 95; (2) a non-primary window with a larger period than the primary; (3) the default single-window period-14 case, which must stay at 200 bars.
   **Plain simple English:** A comparison tool checks that the test engine and the live system agree. In some setups the tool measures the market data over a different length than the two systems it compares. Then it can report false disagreements. Pass the full window settings to the tool so all three use the same length.

3. **No test exercises the SIGHUP adoption arm for `hurst_gate` in `applyHotReloadConfig`.**
   Issue #1411's acceptance criteria ask for a "hot-reload adoption test matching the stated while-open policy". `TestHurstGateIsHotReloadableIncludingWhileOpen` proves the field is masked from the restart shape and passes `validateHotReloadCompatible`, and `TestCloneHurstGateConfigDeepCopies` covers the clone helper, but no test calls `applyHotReloadConfig` and asserts the new block lands on the live config (the arm at `scheduler/config_reload.go:265-274`). Verified by reading `scheduler/config_reload_test.go` at the head commit: its `applyHotReloadConfig` cases never touch `HurstGate`. A regression in that arm would silently ignore an operator's SIGHUP edit to a live entry gate.
   **Invariant:** Every hot-reloadable field has a test that drives the actual adoption path end to end.
   **Must survive:** (1) an edit while a position is open; (2) an edit that changes only `on_failure`, where the threshold key and latch are kept; (3) a threshold edit whose stale latch must be discarded on the next cycle.
   **Plain simple English:** The system can accept a settings change while it runs. The code for this exists and reads correctly. No test runs that exact code path. Add a test that changes the setting while a trade is open and checks that the system adopts it.

---
Validated with LLM: Fable 5 | high | Harness: milestone-pipeline

richkuo:
Addressed review feedback (richkuo · 2026-08-08T05:23:30Z) in `365ce39a`.

All three `Recommended Optional` findings were re-validated against the head commit before any edit. All three held. Nothing was refuted.

### Fixed

1. **Open-time stamp comments claimed H lies in (0, 1) exclusive** — corrected at three sites, one more than the review cited: `scheduler/portfolio.go:126-141`, `scheduler/db.go:649-653`, and `scheduler/trade_diagnostics.go:131-136`, which carried the same wording. Each now states the sentinel's real basis: it rests on the LOWER bound only, and `hurst_at_open` is explicitly NOT capped at 1, so a stored value above 1 is legal and must never be clamped or rejected.

   I measured the lower bound rather than assuming it. `hurst_exponent` returns the raw DFA regression slope, so positivity is not enforced by the code. On a maximally anti-persistent series (alternating returns) it reads ~0.007, on AR(-0.99) ~0.024, and a degenerate series returns NaN — so a finite reading is always > 0, but only just. The comments now say that instead of implying a code-enforced bound.

   New test `TestHurstStampsAcceptReadingsAboveOne` (`scheduler/hurst_gate_test.go`) covers all three of your Must-survive cases: a 2.0033 reading survives `stampHurstGateAtOpenIfOpened`, the `trade_diagnostics` row, and the SQLite round trip; and 0 still reads as unstamped.

2. **`parity_diff.py` computed the Hurst frame depth without the regime windows spec** — `ParityConfig` now carries `regime_windows_spec` (`backtest/parity_diff.py:187-198`), populated from `load_strategy_config` in `config_from_live_config` (`:271`), and fed to `hurst_live_frame_bars` at `:747`. The tool, the engine and the live daemon now read H over the same frame length.

   Scope note beyond the finding: `extract_fills` built its Backtester without `hurst_gate` **or** `regime_windows_spec` (`:845-853`), while `run_backtest` passes both. On a gated config `--fills` therefore listed entries the engine holds, and it would have recomputed the depth from `regime_period` again inside that Backtester — reintroducing the same divergence one layer down. Both kwargs are now passed. `hurst_gate` is `None` for any config that does not opt in, so the ungated baseline is unchanged.

   Three new tests in `backtest/test_hurst_gate_backtest.py` cover your Must-survive list: a 150-period window (depth 309, verified against a directly computed `rolling_hurst`), a non-primary window carrying the largest period, and the default single-window case staying at the 200-bar floor. Verified load-bearing — the first test fails against the old `hurst_live_frame_bars(None, ...)` call.

3. **No test exercised the SIGHUP adoption arm** — `TestApplyHotReloadConfigAdoptsHurstGateWhileOpen` (`scheduler/hurst_gate_wiring_test.go:383-500`) now drives `applyHotReloadConfig` end to end. Every case runs with a position open, matching the stated while-open policy. It covers your three cases exactly:
   - a threshold edit lands on the live config and is deep-copied, and both change lines reach the operator log;
   - an `on_failure`-only edit leaves the threshold key unchanged, so an ARMED latch survives — proven observably, since a discarded latch would reset to unknown and disarm on the same in-gap reading;
   - a threshold edit rewrites the key, so a stale ARMED latch is discarded and the gate re-derives from the next reading.

   Verified load-bearing: deleting the adoption arm at `scheduler/config_reload.go:274-277` fails all three sub-cases.

### Acceptance criteria unmoved

No fix touched an issue constraint. The gate stays default-off and opt-in only; `config.example.json` still ships no `hurst_gate` values (the #1410 study remains INCONCLUSIVE with its "Do not build a Hurst entry gate on this evidence" line, quoted in the PR description); `CurrentConfigVersion` stays 17; `allowed_regimes` / `applyRegimeGate` / `resolveRegimeGateOnFailure` semantics are byte-identical; no existing test was weakened. Every change is additive — three comment corrections and four new tests, plus the parity threading.

### Verification

`go build` OK. `go test ./...` ok (36s). `pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3064 passed, 1 skipped. `shared_scripts/test_regime_wiring.py` — 25 passed. `gofmt -l scheduler/` clean. No merge conflict with the base: `origin/main` (`a097d3a7`) is an ancestor of this branch.

---
Created with LLM: Opus 5 | xhigh | Harness: milestone-pipeline

richkuo:
Needs Updates

### Needs Fixing

1. **The backtester never applies the Hurst gate to scale-in adds — neither the hold nor the size multiplier — so a strategy combining `hurst_gate` with `allow_scale_in` silently diverges from live.**
   `backtest/backtester.py:2299` (`_try_scale_in_add`) computes the add quantity from `scale.base_open_notional` / `cfg.AddNotionalUSD` alone; it never reads `hurst_size_mult` or the gate hold. Both call sites (`backtest/backtester.py:2806`/`2812` and `:3227`/`3234`) invoke it unconditionally on `self.allow_scale_in and open_action == "..." and position > 0`, with no check of `regime_blocked` — the same variable that `hurst_blocked` is folded into at `:2491-2494`. On the live side, `scheduler/main.go` applies the gate's hold to a scale-in add through `pausedBlocksSignal`-gated `result.Signal = 0` *before* `perpsScaleInDecision` runs (so a disarmed gate makes `perpsScaleInDecision` return `"not a same-direction add"`, `scheduler/scale_in.go:215`), and separately scales the decided add quantity by `hurstDecision.OpenSizeMult()` (`scheduler/main.go:2674-2683`).
   I reproduced this directly rather than inferring it from the diff: running `Backtester` on the same signal/price series with `allow_scale_in=True, scale_in={"add_notional_usd": 50.0, "max_adds": 50}` gives `scale_in_adds=50, added_notional≈2494.97` identically whether `hurst_gate` is absent, set to `mode="size", size_floor=0.05` (which should shrink nearly every add), or set to `mode="gate", min=0.999, disarm_min=0.999` (a band the H series can essentially never satisfy, so the gate is disarmed for the entire post-warmup run and would hold every add live). All three runs produced byte-identical scale-in output. `backtest/test_hurst_gate_backtest.py` contains no `allow_scale_in`/`scale_in` reference at all, so this path has zero test coverage.
   This is exactly the failure class the PR's own docstring at `run_backtest.py`'s `_resolve_backtestable_hurst_gate` names as the thing to prevent ("the run would report entries the daemon would have held, and the operator would reasonably trust the number") and the one the "silent CLI seam" fix in this same PR was written to catch — it just wasn't caught for this path. `run_backtest.load_strategy_config` has no rejection for `hurst_gate` + `allow_scale_in` together, so a config combining both reaches the engine and reports numbers a re-run of the #1410-style study — the PR's stated reason for shipping this at all — would rely on.
   **Invariant:** every code path that can increase an open position live (fresh open, scale-in add, bidirectional flip) is gated and sized identically in the backtester.
   **Must survive:** (1) `mode="gate"` disarmed for the whole post-warmup window with `allow_scale_in=True` — adds must stop, not continue at the ungated rate; (2) `mode="size"` with a low `size_floor` and `allow_scale_in=True` — added notional must shrink, not match the ungated run; (3) the existing fresh-open and flip paths, which are already correctly gated, must stay unaffected by the fix.
   **Plain simple English:** The test tool that predicts strategy performance does not apply the new persistence filter to trades that add to an existing position. The live system does block or shrink those adds. So the test tool can report bigger and more frequent adds than the live system would ever allow, for any strategy that uses both features together.

---

richkuo:
Addressed review feedback (richkuo · 2026-08-08T05:49:26Z) in `09df881c`.

### Fixed

1. **Backtester never applied the Hurst gate to scale-in adds** — CONFIRMED, and re-derived from the code rather than taken on the reviewer's word. `_try_scale_in_add` (`backtest/backtester.py:2299` pre-fix) sized every add from `scale.base_open_notional` / `add_notional_usd` alone, and its four call sites (`:2806`/`:2812`, `:3227`/`:3234`) checked only `allow_scale_in` + signal + position sign. `regime_blocked` — the variable `hurst_blocked` was folded into at `:2495` — is read only at the four fresh-open sites (`:2666`, `:2734`, `:3075`, `:3147`), all guarded by `position == 0`, so the fold never reached an add. Live applies both arms: `pausedBlocksSignal` returns `true` for a same-side signal on an open position (`scheduler/pause.go:63`, asserted by the existing `"scale-in add on long"` case in `hurst_gate_test.go`), and `scheduler/main.go:2687` scales the decided quantity by `hurstDecision.OpenSizeMult()`. `evaluateHurstGate`'s own note at `scheduler/hurst_gate.go:398-400` states the intent plainly: "adds and flips are held". Both arms now sit **inside** `_try_scale_in_add` (`backtest/backtester.py:2371`, `:2382`) rather than at its call sites, so a fifth add site added later inherits the gate instead of bypassing it.

2. **Per-add default notional was double-scaled** (found while fixing item 1, not in the review) — `scale.base_open_notional` froze `shares * effective_price` from the already-multiplied `entry_fraction`. It is the backtest image of live's `defOpenNotional`, which live recomputes every cycle with **no** multiplier applied. Left as-is, applying the current bar's multiplier on top would have shrunk an add twice, and the open bar's multiplier would have persisted for the life of the position — an open at m=0.25 would keep shrinking adds by 0.25 long after H recovered. `_ungated_leg_notional` (`backtest/backtester.py:792`) strips it back out at all four fresh-open sites. `size_floor` is validated to (0, 1] on both engines, so the divisor is never zero.

### Not changed (deliberate, and load-bearing)

1. **The label gate still bypasses adds.** `_try_scale_in_add` does not consult `regime_blocked` for the `allowed_regimes` half, because live `regimeBlocksOpen` passes `posQty>0` (`scheduler/regime.go:102-110`). Only the Hurst arm was added — the Hurst gate carries no flat-only restriction outside its fail-closed arm. `allowed_regimes` / `applyRegimeGate` / `resolveRegimeGateOnFailure` semantics stay byte-identical.

2. **No load-time rejection of `hurst_gate` + `allow_scale_in`.** The tiered_tp_atr_live_regime_dynamic rejection precedent exists for what the engine genuinely cannot model. This path is modellable, so implementing parity beats refusing a legitimate combination.

### Verification

Reproduced the divergence before fixing (50 adds, added notional 2498.31, identical with and without `mode="size", size_floor=0.05`), then confirmed the same run reports 667.62 after. Six new tests in `backtest/test_hurst_gate_backtest.py` drive a **deterministic** H series through the real state machine, covering all three "Must survive" cases and the double-scaling regression:

1. `test_gate_mode_holds_scale_in_adds_once_disarmed` — disarmed for the rest of the run: adds stop at the exact armed bar, not at the ungated rate.
2. `test_size_mode_scales_scale_in_adds_by_the_multiplier` — added notional shrinks by exactly `clamp(|H-0.5|/0.15, floor, 1.0)`, and the same adds still fire.
3. `test_open_bar_multiplier_never_leaks_into_later_adds` — an add at m=1.0 is identical whether the position opened at m=0.2 or m=1.0.
4. `test_armed_gate_leaves_scale_in_adds_byte_identical` — an armed gate never scales.
5. `test_absent_hurst_gate_leaves_the_scale_in_baseline_byte_identical` — default-off stays bit-for-bit pre-#1411.
6. `test_scale_in_hurst_arms_live_inside_the_add_helper` — wiring inventory: both arms inside the helper, all four call sites gate-free.

Items 1-3 and 6 fail on the pre-fix `backtester.py`; items 4-5 pass either way by design (they are the "must not change" guards). `scheduler/hurst_gate_wiring_test.go` gains `TestHurstGateSizesTheScaleInAddQuantity`, a source-scan pin on the live half so the two engines cannot drift apart silently.

Suites: `go build` and `go test` in `scheduler/` ok; `pytest backtest/` 1640 passed under `-n auto`; `pytest shared_strategies/ shared_tools/ platforms/` 1431 passed; `shared_scripts/test_regime_wiring.py` 25 passed; `gofmt` clean. `docs/ARCHITECTURE.md` records the scale-in parity rule.

**Plain simple English:** The test tool did not apply the new persistence filter to trades that add to a position you already hold. The live system blocks or shrinks those adds. So the test tool could report bigger and more frequent adds than the live system allows. Both engines now behave the same, and new tests hold them together.

---
Created with LLM: Opus 5 | xhigh | Harness: milestone-pipeline

richkuo:
LGTM

---
Validated with LLM: Fable 5 | high | Harness: milestone-pipeline

PR #1423, opened 2026-08-08
feat(#1422): re-run Hurst gate calibration with cluster-aware power analysis [C66, Opus 5, high, fableplan]

### LINKED ISSUE #1422: [C66] The Hurst gate calibration cannot reach a verdict because its trade sample is too small and correlated
**Complexity: 66/100** — Capability 2 (Uncertainty 3 — the power method, the cluster-permutation null, and how much genuinely independent data exists all need design; Risk 1 — offline, report-only); Volume 16 (Scope 3; Coupling 2 — the report contract path shared with #1411/#1412 plus the #1410 reproducibility contract; Verification 3 — real `Backtester` legs with harness-identity assertions) — Opus 5, high · fableplan: yes

**Chain:** depends on #1409 (estimator), #1410 (the study this supersedes as live evidence), #1411 (the default-off gate, unaffected). **Blocks #1412** — its Stage 0 evidence gate reads the same report file, so this must land first.

## Problem

The #1410 study returned `INCONCLUSIVE` and told the reader not to build on it: "Do not build a Hurst entry gate on this evidence. Re-run this study before revisiting the question" (`backtest/research/hurst_gate_calibration.md:215`). 0 of 30 hypotheses reached Benjamini-Hochberg significance at alpha=0.05 (`hurst_gate_calibration.md:205`), while 16 met every economic condition and failed only on significance — their drawdown and chop-loss reductions come from trading less, not trading better (`hurst_gate_calibration.md:199`).

Three concrete properties of that study explain the null result, and none of them is "Hurst does not work":

1. **The sample is small and re-used.** The sweep runs 5 exemplars (`hurst_1410_gate_calibration.py:118-121`) over the 6 audit datasets (`backtest/eval_windows.py:67-75`, BTC/ETH/SOL x 1h/4h) and 5 windows (`eval_windows.py:79-87`) — 150 legs total. Every one of the 30 hypotheses is scored on that same trade pool, so nothing in the study is out-of-sample with respect to the sweep itself.

2. **The 30-hypothesis grid is self-inflicted.** 30 = 2 families x 3 Hurst windows (`hurst_1410_gate_calibration.py:141`) x (3 gate pairs (`:155-158`) + 2 size gains (`:167`)). The Benjamini-Hochberg denominator covers all 30 (`hurst_gate_calibration.md:193`), which is correct given the grid — but the grid was a convenience, not a hypothesis. A pre-registered primary set would carry a far smaller penalty.

3. **The null assumes trades are exchangeable when they are not.** `permutation_pvalue_group_diff` shuffles the suppressed/kept labels freely across all pooled trades (`hurst_1410_gate_calibration.py:459-493`), and `permutation_pvalue_weighted` shuffles multipliers freely (`:496-529`). Dedup only collapses the same physical entry seen through overlapping windows, keyed `(strategy, symbol, timeframe, entry_date)` (`:427-457`) — it never collapses across assets. A momentum entry on BTC 1h and the concurrent one on ETH 1h are two rows in the pool and are strongly correlated. Free shuffling treats them as independent, so the nominal sample size overstates the information in the pool, in an unquantified direction.

There is also an unrelated but load-bearing hazard in the current layout: `hurst_1410_gate_calibration.py:200` defaults its report output to `backtest/research/hurst_gate_calibration.md` — the exact contract path that `scheduler/hurst_gate.go:6`, `docs/ARCHITECTURE.md:22` and #1412's Stage 0 all cite. Any later `--render-only` of the #1410 study silently overwrites that file with the old verdict.

## Goal

The question "does a Hurst entry gate work?" gets a decisive answer, and the answer is trustworthy because the study says how much effect it could have detected. `backtest/research/hurst_gate_calibration.md` carries either a concrete recommended configuration per family, or an `INCONCLUSIVE` verdict with a stated, measured power ceiling that justifies closing the question rather than re-running again. The report also carries the joint ADX x Hurst bucket analysis #1412's Stage 0 requires, so #1412 reads evidence that was designed to answer it.

Report-only, exactly like #1410. No scheduler, config, gating, sizing, or live-path change. `#1411` stays default-off and byte-identical; `CurrentConfigVersion` stays 17.

## Approach

Correctness outranks diff size and run cost throughout. A study that reaches a confident wrong verdict is worse than the current honest null, because #1412 and a live entry gate both key off it.

### Placement — a new one-shot, not an edit of #1410's

The #1410 script, its JSON, and its `--render-only` reproducibility are a committed negative-result artifact (`docs/backtesting-registry.md:80`), and the repo keeps frozen research one-shots reproducible under the exact model they were graded with (`eval_windows.py:102-114`). So this lands as a new script, `backtest/research/hurst_1422_gate_power.py`, reusing the #1410 helpers rather than reimplementing them: `hurst_exponent` (`shared_strategies/open/indicators_core.py:251`, the #1409 SSoT), `benjamini_hochberg` (`backtest/regime_stats.py:44`), and the `eval_windows` harness constants and `run_leg`, keeping the data-source axis (`PLATFORM`) and fee-model axis (`FEE_PLATFORM`) uncoupled.

**Fix the path collision in the same PR.** Repoint `hurst_1410_gate_calibration.py:200` to `hurst_1410_gate_calibration.md`, commit the #1410 render there byte-identical, and let the new study own the contract path `backtest/research/hurst_gate_calibration.md`. Add a comment at both sites naming the contract. Without this, re-rendering the old study reverts the live evidence file.

### Stage A — measure the power before spending the sweep (fail-loud)

Compute what the #1410 design could have detected. The committed `hurst_1410_gate_calibration.json` strips per-trade rows (`hurst_1410_gate_calibration.py:1681` serializes `legs` without `trades`), so the pool must be regenerated by re-running the ungated legs; the per-leg `run_leg` mirror assertion (`:833-847`) is what proves the regenerated pool is the pool #1410 scored. Then inject a synthetic per-trade edge of magnitude `d` into the suppressed group, and find the smallest `d` at which the permutation test clears BH-corrected alpha over the same 30 hypotheses. Report that minimum detectable effect in percentage points per trade, beside the effect magnitudes the Part A buckets actually show.

This turns "inconclusive" into a number and sets the target the expansion below must beat. **If the required `d` stays implausibly large even under the full expansion, a second `INCONCLUSIVE` with a stated power ceiling is a legitimate and valuable outcome of this issue** — it converts #1412's Stage 0 from "maybe" into a defensible "no", and closes the question instead of licensing a third re-run. The issue is not required to produce a positive result.

### Stage B — add genuinely independent samples, and prove they are independent

Three expansion axes, in descending order of how much real information they add:

- **History.** The OHLCV cache currently starts 2023-01-01 for the audit datasets (verified by querying the `ohlcv` table in the cache DB resolved by `shared_tools/storage.py:29`). **`load_cached_data` will not backfill this** — it fetches only when the cached slice comes back empty (`shared_tools/data_fetcher.py:184-188`), and a 2020 request against a 2023-start cache returns non-empty rows and never fetches. Extending history needs an explicit backfill step calling `fetch_full_history(..., store=True)` (`data_fetcher.py:69`), which is idempotent because `store_ohlcv` upserts (`shared_tools/storage.py:170-173`). This adds structurally different regimes (2020 crash, 2021 bull, 2022 bear) rather than more of the same tape.
- **Symbols.** Beyond the audit six, the cache already carries BNB/USDT 4h and XRP/USDT 4h; further pairs need the same explicit backfill. Add symbols outside the BTC-beta cluster where the data supports it. XRP/USDT 4h starts 2023-07-14 in the cache, so pre-2023 XRP history is likely unavailable from this venue — missing cells must be dropped loudly, never scored on partial data.
- **Timeframes.** Cache coverage is uneven: 15m is present for BTC, ETH and SOL, but 30m and 2h exist only for BTC (ETH 30m holds 50 stray rows; SOL 30m and ETH/SOL 2h are absent). Any added timeframe needs backfill, and the choice must be justified — a 512-bar Hurst window at 15m spans about five days, which measures microstructure rather than the multi-week regime persistence this study is about.

**Publish effective sample size, not row count.** Compute a cross-dataset correlation matrix over per-bar returns and derive an effective-N estimate for each pooled trade set. The report must print effective N beside nominal N in every table that feeds a p-value, so no reader can mistake row inflation for statistical power. A dataset that only adds correlated copies must be visible as adding almost nothing.

### Stage C — an inference scheme that matches the dependence structure

Replace the free label shuffle with a **cluster/block permutation** that permutes at the level the dependence lives: calendar blocks (month or quarter) and dataset, so concurrent correlated trades move together under the null. Report both p-values side by side — the free-shuffle p (directly comparable to #1410's numbers) and the cluster p — and take the verdict from the cluster p. A large divergence between them is itself a reportable finding about how to read #1410's table.

### Stage D — a primary test that selection cannot contaminate

Shrinking the hypothesis grid lowers the BH penalty, but choosing which hypotheses to keep by looking at #1410's p-values is selection on the same data and would invalidate the result. The clean fix is disjoint data, not a promise:

- **The primary hypothesis set is scored on data #1410 never saw** — the newly fetched pre-2023 history and any newly added symbols. Selection can then draw freely on the #1410 evidence without contaminating the primary test, because the two data sets do not overlap.
- The primary set stays small and its selection rule is stated in the report.
- The full grid still runs on the #1410 windows and is still reported, labelled explicitly as exploratory, under its own correction. The two are never pooled into one denominator.
- The acceptance rule keeps #1410's shape: volume floors, significance, economic conditions on both protocol windows, and non-degradation on held-out windows (`hurst_gate_calibration.md:190-197`), with the volume floors re-derived for the larger pool rather than carried over unchanged.

### Stage E — joint ADX x Hurst buckets (the #1412 Stage 0 substrate)

Per-bar ADX is available from `compute_regime` (`shared_tools/regime.py:392`), whose frame carries an `adx` column (`regime.py:159`). **Warm-up bars are filled with `0.0`, not NaN (`regime.py:403`)**, so a naive stamp would label every warm-up bar low-ADX — mask bars before the ADX warm-up to NaN. Stamp each trade's entry ADX through the same shift convention already used for H — the decision series is `shift(1)` (`decision_series`, `hurst_1410_gate_calibration.py:678-687`) and the fill-bar entry stamp is `shift(2)` (`entry_stamp_series`, `:690-697`) — so both features share one look-ahead rule; the existing look-ahead unit test pattern extends to ADX unchanged.

Report, per family: high/low ADX (split at the composite default threshold 25.0, `regime.py:151`) crossed with the H buckets `<0.45` / `0.45-0.55` / `>0.55` / `NaN`, giving trade count, win rate, net PnL and max drawdown per cell. Answer #1412's exact question in the report text: does high ADX combined with low Hurst separate materially beyond high ADX alone, for momentum-style entries? Emit an explicit verdict, including a literal `NO JOINT SEPARATION` when it does not, so #1412 Stage 0 reads a decision rather than a table.

### NaN policy, unchanged

NaN stays its own bucket, never coerced to 0.5, never a size opinion, and it holds gate state rather than arming or disarming it (`hurst_1410_gate_calibration.py:137-138`, `:167-170`). The #1410 report notes the NaN bucket was empty only because the cache was deep enough (`hurst_gate_calibration.md:48`) — extending history back to 2020 changes warm-up depth at the new window starts, so the existing warm-up audit must stay fail-loud and its verdict must be printed.

### Out of scope / untouched

`scheduler/` behavior, `config.example.json`, `hurst_gate` semantics and defaults (#1411), `map_composite_label` and the composite thresholds (that is #1412's Stage 1), `CurrentConfigVersion`. The #1410 script keeps its logic; only its default report path and a contract comment change. Doc and comment text at `scheduler/hurst_gate.go:6`, `docs/ARCHITECTURE.md:22` and `docs/backtesting-registry.md:80` is updated only to match whatever verdict this study reaches.

## Acceptance criteria

- `backtest/research/hurst_1422_gate_power.py` committed and re-runnable via `uv run --no-sync python`, with a fixed seed so every p-value reproduces, and `--render-only` re-rendering the report byte-identical from the committed JSON.
- `backtest/research/hurst_gate_calibration.md` regenerated at the contract path, with `## Recommendation` last, carrying either a concrete per-family configuration or `INCONCLUSIVE` plus the measured minimum detectable effect.
- Stage A minimum-detectable-effect figure present in the report, in percentage points per trade, next to the observed bucket effect magnitudes.
- Every table feeding a p-value prints effective N beside nominal N, and the cross-dataset correlation matrix is in the report.
- Both the free-shuffle p and the cluster/block-permutation p are reported per hypothesis; the verdict is taken from the cluster p, and the report states this.
- The primary hypothesis set is scored only on windows and datasets absent from the #1410 run; the report names them and states the disjointness explicitly. The exploratory grid is reported separately and never pooled into the primary correction.
- Joint ADX x Hurst bucket section committed, with an explicit separation verdict (or the literal `NO JOINT SEPARATION`) that answers #1412's Stage 0 question.
- Harness identity: every ungated arm reproduces `eval_windows.run_leg` exactly, asserted per leg and the pass count reported, as #1410 does.
- Look-ahead: a unit test mutates bar N's close and proves neither the bar-N Hurst decision nor the bar-N ADX stamp moves, with a paired assertion that the unshifted series does move.
- `hurst_1410_gate_calibration.py` no longer defaults its report output to the contract path; the #1410 render is committed at `hurst_1410_gate_calibration.md` byte-identical, and a regression test asserts the #1410 default output path is not the contract path.
- Warm-up audit stays fail-loud over the extended history, and its verdict is printed in the report.
- `docs/backtesting-registry.md` row added for the new one-shot, in this PR, per the registry SSoT rule.
- Unit tests over the new pure helpers: cluster permutation, effective-N estimator, minimum-detectable-effect search, joint bucketing, and the mechanical recommendation function.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` green; `go -C scheduler build ./...` and `go -C scheduler test ./...` green (no Go behavior changed).
- PR description states the verdict, the measured power, the datasets and windows added, nominal vs effective N, and the joint-bucket answer for #1412.

## Execution

- Depends on: #1410
- Runs after: none
- Blocks: #1412
- Build model: Opus 5
- Build effort: high
- Validate effort: medium
- fableplan first: Yes
- Plan effort: high
- PR review: Opus 5 · high (first review)

---
Updated with LLM: Fable 5 | high | Harness: fableplan

richkuo:
## Plain simple English

We asked again if a market "trendiness" number can tell our bots when to skip a trade. The last study could not answer. This one uses five times more trades and better maths, and it measures how small an effect it could find. The answer: a big effect is ruled out, and a small one is still unknown because the study cannot see effects that small. So we ship no rule. Nothing in live trading changed.

## Summary

The #1410 study returned `INCONCLUSIVE` and could not say why. Three properties of its design explained the null rather than the data: one re-used sample, a 30-hypothesis grid that was a convenience, and a null that shuffled correlated concurrent trades as if independent.

This adds `backtest/research/hurst_1422_gate_power.py`, which fixes all three and measures what each design could have detected.

**Verdict: `INCONCLUSIVE` again, but with a measured bound instead of a shrug.** On the primary (confirmatory) cohort the design resolves 0.96 pp per trade, while that cohort's own anti-signal split separates by 0.37 pp. A separation under the limit is invisible to the design, so the null **excludes an edge of 0.96 pp or larger and says nothing either way about a smaller one**. Power is the binding constraint; absence is not proven. Best primary hypothesis reached cluster p=0.0485.

Separations and limits are compared **pool-matched** — measured on the same rows — because a whole-study separation read against a sub-cohort's limit compares two different samples. The report's `Resolvable?` column makes that reading explicit.

The better-powered pools outside the confirmatory cohort *do* resolve their separations, and those do not survive the cluster null: the exploratory grid at 0.99 pp against a 0.42 pp limit (cluster p=0.4923), and #1410's own cells at 1.04 pp against 0.38 pp (cluster p=0.4583). That contrast was not pre-registered and licenses no gate, but it is consistent with whole datasets moving together rather than with H sorting trades within them.

Report-only. Zero scheduler, config, gating, sizing, or live-path changes; `CurrentConfigVersion` stays 17; #1411's `hurst_gate` stays default-off with no recommended thresholds. The `scheduler/` diff is comment-only.

### What the study does differently

| Mechanism | Why it exists |
|---|---|
| **Measured detection limit** | Deterministic shift of the suppressed group, grid-searched to the rank-1 Benjamini-Hochberg bar. Turns "inconclusive" into a bound. **Raises** when `n_perm` cannot resolve that bar, so the report can never publish "no power" when the truth is "no permutations". |
| **Cluster-rotation null** | One shared circular calendar offset per draw, applied to every dataset's chronologically ordered label vector. Concurrent correlated trades move together, and a hysteresis run's autocorrelation survives. Short-span datasets excluded loudly; an unrotatable pool yields `p=None` and fails closed. The #1410 free-shuffle p is still reported for continuity; the verdict reads the cluster p. |
| **Effective N** | `N^2 / sum_ij rho_ij` over trades whose holding periods overlap. Anti-correlation is never credited, no matrix is inverted, bounds are `1 <= N_eff <= N` by construction. Printed beside nominal N, and the volume floors apply to **it**. |
| **Contamination-proof cohorts** | The 4 primary hypotheses are the per-(family, mode) smallest-raw-p configs in the committed #1410 JSON — pinned, re-derived with a hard assert — scored only on tape #1410 never saw. Primary and exploratory carry separate BH denominators. |
| **Coverage density floor** | A cell needs the required lead **and** at least 80% of the bars a complete cache would hold inside the window. |

### Findings worth reading

- **The correlation collapse is severe.** 6,238 momentum primary trades carry an effective sample size of ~295; 16,968 mean-reversion trades carry ~1,161. Roughly 93% of the rows were redundant. #1410 had no way to see this, and it is why more rows alone were never going to settle the question.
- **The density floor earned its place.** It dropped 15 cells, including the XRP delisting gap — 2021 present as 300 of 8,760 expected bars — plus late SOL and LINK listings. An unchecked "more than zero bars" rule would have weighted that year like a complete one.
- **#1412's Stage 0 is discharged.** The joint ADX x Hurst section reports the literal `NO JOINT SEPARATION` on both families, so fusing H into the composite classifier is not justified and the standalone gate remains the correct amount of Hurst.

### Live-evidence hazard fixed along the way

`hurst_1410_gate_calibration.py` defaulted its report output to the contract path `hurst_gate_calibration.md` — the file `scheduler/hurst_gate.go`, `docs/ARCHITECTURE.md` and #1412's Stage 0 all read. Any later `--render-only` of the superseded study would have silently reverted the live evidence to the old verdict. Its render now lives at `hurst_1410_gate_calibration.md`, committed byte-identical to the original, with a regression test asserting the #1410 default is not the contract path.

### Cohort rule — one deliberate tightening

`cell_cohort` keys on **symbol and window**, not the (dataset, window) grid. BTC 2h over a #1410 window is not in that grid, but it is the same tape #1410 mined at 1h and 4h over the same months. Scoring a hypothesis there that was chosen by reading #1410's p-values would be selection on the same data wearing a different timeframe. Only a genuinely new period or asset is clean.

### Known limitation, asserted as a test

The rotation null has little power against a perfectly periodic label pattern — labels alternating every trade are reproduced by any even rotation. Real gate labels come in runs, so this does not bite, and the test exists so a future reader does not mistake it for a bug.

### Files

| File | What |
|------|------|
| `backtest/research/hurst_1422_gate_power.py` | the study |
| `backtest/research/hurst_1422_gate_power.json` | aggregate output of the committed run |
| `backtest/research/hurst_gate_calibration.md` | the rendered report (contract path, now owned here) |
| `backtest/research/hurst_1410_gate_calibration.md` | #1410's render, byte-identical, moved off the contract path |
| `backtest/tests/test_hurst_1422_gate_power.py` | 113 unit tests over the pure helpers |
| `docs/backtesting-registry.md` | new row + corrected #1410 row, same PR per the registry SSoT rule |
| `docs/ARCHITECTURE.md`, `CLAUDE.md`, `scheduler/hurst_gate.go` | calibration-status text updated to the new verdict (Go is comment-only) |

### Verification

- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — **3226 passed** (155 in the study's own suite).
- Full run: **765 legs, 6,885 gated Backtester arms, 765/765** reproduced `eval_windows.run_leg` exactly.
- `--render-only` re-renders the contract report **byte-identically**; #1410's render matches its original committed report **byte-for-byte** (`cmp` against `HEAD`).
- `go -C scheduler build ./...` and `go -C scheduler test ./...` green; `git diff scheduler/` contains no non-comment lines.
- `uv run --no-sync python -m py_compile` on both study scripts.

### Note for the reviewer

Review findings and their dispositions are in the [disposition comment](https://github.com/richkuo/go-trader/pull/1423#issuecomment-5229962667). Two points deserve attention here.

The reviewer offered two remedies for the rotation-null defect. The obvious one — wrap each short dataset's offset independently — trades a conservative bias for an anti-conservative one: a short dataset then rotates by a different calendar amount than its concurrent neighbours, understating the co-movement the cluster null exists to preserve and pushing p **down**. The offset range is capped at the pool's shortest span instead, so every retained dataset hosts every offset. A false positive is the worse failure and this p-value is the whole conclusion.

Correcting the null raised the primary cohort's detection limit above its observed separation, which reached a branch of the report that had never executed. That branch reported a sub-limit separation as proof that "no edge of a size that would matter is present". It is the wrong inference — an effect under the design's own limit is invisible to the design — and two existing tests asserted it. The code and both tests were corrected; the tests now assert that the report never converts a power failure into a claim about the market.

Closes #1422

---


richkuo:
@claude opus review

richkuo:
## Review disposition — 8dcd639c

All six findings traced to current code and confirmed. None refuted. Fixing the null changed every published number, so the study was re-run end to end on the identical cached tape (`--skip-fetch`), and reading that output caught two further defects in the study's own conclusion.

### Fixed — review findings

**1. Cluster-rotation null left short datasets unrotated.** A dataset whose span was shorter than the drawn offset kept its observed label alignment on every draw, so the null inherited the very structure it exists to destroy and p came out too HIGH.

The reviewer offered two remedies. I implemented the wrap first, then rejected it: wrapping a short dataset rotates it by a *different* calendar amount than its concurrent neighbours, which understates the cross-dataset co-movement the cluster null preserves and pushes p **DOWN**. That is a false-positive failure, and this p-value is the study's entire conclusion. The offset range is now capped at the pool's **shortest** span (`_admissible_offsets`), so every retained dataset hosts every offset and concurrent datasets always shift together. A long dataset rotating by a smaller fraction of its span is fine — what a rotation must destroy is label autocorrelation, and the 30-day floor is long against a hysteresis run. `effective_offset_days` survives as a guard, with a test proving it never fires inside the capped range.

Invariant: under the null, concurrently-open trades in different datasets shift by the same calendar amount, and no retained dataset is ever the identity.
Must survive: a ragged pool where one dataset is far shorter than the rest; an equal-span pool (the fold must be a no-op); every admissible offset producing a non-identity shift on every dataset.

**Plain simple English:** The test that measures luck did not move the short data sets. This made the result look better than it was. The fix moves all data sets together by the same number of days. This keeps the test honest.

**2. Short datasets were excluded from the offset draw but not from the contrast.** Their rows still entered the observed statistic. `usable_cluster_rows` now drops them from both sides, and the free p-value, trade counts and effective-N floors score exactly the rows the cluster p scores. Counts of what was dropped are recorded per config.

**Plain simple English:** Some trades were counted in the result but not in the luck test. Now both use the same trades.

**3. Coverage density used a per-dataset last bar.** One shared reference bar is now computed across datasets, so a mid-window data gap and an early end are both caught.

**Plain simple English:** The check for missing price data now uses one common end date, so it finds gaps it missed before.

**4. Stage 0 joint verdict inherited a limit measured on another pool.** It now measures its own detection limit on its own retained rows. Consequence, documented and asserted by a test: with the limit measured on the same pool, a p-value that clears the bar forces the limit to zero, so the materiality condition is a floor that cannot bind while p passes. That is the correct implication of the invariant.

**Plain simple English:** The joint check now measures its own accuracy limit on its own data.

**5. Gate sense was assumed, not resolved.** `anti_signal_side` resolves the suppressed side per family; an unknown sense raises instead of silently defaulting and inverting an injected contrast.

**Plain simple English:** Each strategy family has its own idea of a bad signal. The code now asks which one, and stops if the answer is unknown.

**6. Scoped runs could overwrite full-run artifacts.** A scoped run refuses the committed JSON path and the contract report path; `--render-only` requires a `complete` scope stamp before it will write the contract path.

**Plain simple English:** A partial run can no longer overwrite the official report.

### Corrected scope — two defects the re-run exposed

These are mine, not the reviewer's. Both were latent and only became reachable once the corrected null raised the primary cohort's detection limit above its observed separation.

**7. The conclusion drew the inference backwards.** The newly-reachable branch reported a separation *below* the design's detection limit as evidence that "no edge of a size that would matter is present", and the closing guidance said "the absence is measured rather than assumed" and told the reader to treat the question as closed. An effect under the design's own limit is **invisible** to the design. That null bounds the edge from above and says nothing whatever about a smaller one. Published as written, the report would have claimed proof of absence on the strength of its own lack of power — and #1412 reads this file as an evidence gate.

Both the decision text and the report's closing guidance now state the bound actually earned and name power as the binding constraint.

Invariant: the report never converts a null result into a claim about the market without a limit below the effect it is dismissing.
Must survive: separation below the limit (bound only, no absence claim); separation above the limit (rule blamed, not the sample); no limit reachable at all (no conclusion in either direction).

**Plain simple English:** The report said the study found no effect. In truth the study was not accurate enough to see an effect that small. Those are different things. The report now says what it really proved: a large effect is ruled out, a small one is not.

**8. Separation and limit were measured on different samples.** The separation came from the whole trade pool while the limit came from the smaller primary cohort, so the two were never comparable. Separations are now recorded per pool on the same rows as that pool's limit (`mde.observed_separation_pp_by_pool`, schema v3), and the report gained a `Resolvable?` column making the like-for-like reading explicit. On the real data this moved the primary-cohort separation from 0.45 pp to 0.37 pp.

**Plain simple English:** The study compared a number from one group of trades against a limit from a different group. Now both come from the same group.

The pool-matched table also surfaced a result the single-number comparison hid, so the Recommendation now reports it: the better-powered pools outside the confirmatory cohort *can* resolve their separations, and those separations do not survive the cluster null. It is labelled exploratory and licenses no gate.

### Tests

Two existing tests asserted the wrong inference from item 7 in their own assertions (`"no edge of a size that would matter" in text`, `"Treat the question as closed" in text`). They were rewritten to assert the corrected claim. That is a correction of a test that encoded a defect, and it is called out here rather than left silent. No test was weakened to make code pass.

155 tests in `backtest/tests/test_hurst_1422_gate_power.py`, up from 146.

### Not changed

Nothing was refuted, and no finding was deferred to a follow-up issue.

### Result of the re-run

Verdict **INCONCLUSIVE**. 0 of 4 primary hypotheses significant; best cluster p=0.0485. Primary cohort resolves 0.96 pp/trade against a 0.37 pp separation, so power binds and absence is not proven. Exploratory grid 0.99 pp against a 0.42 pp limit, cluster p=0.4923; #1410's cells 1.04 pp against 0.38 pp, cluster p=0.4583. Stage 0 stays `NO JOINT SEPARATION` on both families (momentum p=0.8109, mean_reversion p=0.0629). `hurst_gate` ships no recommended thresholds and `config.example.json` stays clean.

### Verification

- `pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3226 passed.
- `go -C scheduler build ./...` and `go -C scheduler test ./...` — pass; `gofmt` clean.
- `git diff scheduler/` is comment-only, so no live path moves.
- `--render-only --write-report` round-trips byte-identically; the #1410 render is untouched.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-09 05:32:41 UTC) in 6ed256e3.

### Fixed

1. **Per-pool observed separations used rows the cluster limit excluded** — the cluster-usable subset is now applied once to family separations, cluster/free detection limits, and both zero-injection p-values; an excluded dataset cannot leave a one-sided contrast behind (`backtest/research/hurst_1422_gate_power.py:2945`, regression at `backtest/tests/test_hurst_1422_gate_power.py:1497`).
2. **Stage-0 significance and detection-limit resolutions could disagree at the boundary** — removed the separate joint MDE resolution and now drive both calculations with the same permutation count and seed; the report records that invariant and a boundary regression pins it (`backtest/research/hurst_1422_gate_power.py:1012`, regression at `backtest/tests/test_hurst_1422_gate_power.py:764`).
3. **`symbol_return_correlations` lacked direct coverage** — it now selects the true finest timeframe by parsed resolution, resamples daily last closes, enforces the overlap floor, skips undefined constant-series correlations, preserves negative values for downstream clipping, and emits stable pair keys (`backtest/research/hurst_1422_gate_power.py:1246`, regression at `backtest/tests/test_hurst_1422_gate_power.py:220`).

### Verification

- Focused #1422 suite: 161 passed.
- Full Python suite: 3232 passed.
- Go build and Go tests: passed.
- Full cached-tape study rerun: every published result remained byte-identical; only methodology text changed.
- Contract report render: byte-identical to a fresh render from the committed JSON.

No findings were refuted, partially applied, deferred, or left for a judgment call.

---
Created with LLM: GPT-5 | high | Harness: Codex

richkuo:
@claude sonnet review

PR #1425, opened 2026-08-10
feat(#1424): raise the Hurst gate study's power with a three-route resolution design [C68, Opus 5, high, fableplan]

### LINKED ISSUE #1424: [C68] Raise the Hurst gate study's statistical power so it can resolve the separation its own buckets show
**Complexity: 68/100** — Capability 2 (Risk 1 — report only, no live path, no config surface; Uncertainty 3 — the #1422 harness exists and the routes are enumerated; the open judgment is the route combination and the data feasibility, and the pools that CAN resolve a separation contradict it); Volume 18 (Scope 3 — one study script plus the contract-path report; Coupling 2 — research surfaces only; Verification 4 — the whole deliverable is a statistical argument) — Opus 5, high · fableplan: yes

**Chain:** follows #1409 (estimator), #1410 (first study), #1411 (standalone default-off gate), #1422 (power study). #1412 (classifier fusion) closed on its own Stage 0 rule and this issue does not reopen it.

## Problem

The #1422 power study returned INCONCLUSIVE because of power, and not because of an absent signal. Its own Recommendation section states this: the primary cohort resolves 0.96 pp per trade under the cluster null, the largest anti-signal separation on those same rows is 0.37 pp, so the design cannot see an effect of the size its buckets show. The null bounds the edge from above at 0.96 pp and says nothing about anything smaller.

Two facts set the size of the problem, both from `backtest/research/hurst_gate_calibration.md`:

- Correlation destroys most of the sample. The momentum primary pool holds 6238 trades and only 294.9 EFFECTIVE trades. The `mean_reversion` primary pool holds 16968 trades and 1160.6 effective.
- The strongest primary hypothesis reached cluster p=0.0485 (`momentum/gate/W512/arm0.52/dis0.48`), which did not survive Benjamini-Hochberg across 4 primary hypotheses.

So the question "does Hurst carry a usable edge" is still open, and the current design cannot answer it. Adding more crypto symbols over the same years adds raw rows and almost no effective rows, so a re-run of the same design at larger raw N stays blind.

## Goal

Produce a successor study whose measured detection limit on its OWN primary cohort falls BELOW the separation that cohort shows, so the verdict reads as a statement about the market. A null result then means the edge is absent, and a positive result then licenses a threshold. Report only. Nothing wires to the scheduler, to config, or to any live path, and `hurst_gate` stays default-off with no recommended thresholds unless the successor passes its pre-registered acceptance rule.

## Approach

Three routes lower the detection limit. They are independent and can combine. The design must choose and pre-register the combination BEFORE it reads any new numbers.

**Route 1 — fewer pre-registered hypotheses.** #1422 corrected 4 primary hypotheses together, so the rank-1 Benjamini-Hochberg threshold sat at alpha/4. One pre-registered hypothesis per family raises the bar to alpha and lowers the limit directly, with no new data. The cost is real: the study then tests one gate design and cannot rank alternatives. Pick the single design from #1422's already-published ranking, and state in the report that the pick is inherited and therefore not re-selected on new evidence.

**Route 2 — more INDEPENDENT calendar clusters.** Effective N is set by how many independent cluster spans the pool holds, and not by trade count. Candidates, all of which need a feasibility check first because data availability is unverified:
- Longer history than the 2020-01-01 floor, which needs a data source beyond the current `binanceus` cache.
- Instrument families whose calendars do not move with crypto — the TopStep futures and IBKR surfaces already exist in this repo.
Adding more crypto symbols inside the existing windows is explicitly NOT this route.

**Route 3 — a lower-variance scoring target.** The detection limit scales with the per-observation noise of the target, and per-trade net return is the noisiest target available. Score the same contrast on a lower-variance statistic — for example forward efficiency over a fixed horizon, or an MFE/MAE ratio — so the same rows resolve a smaller effect. A win on such a target is evidence about the mechanism and is NOT by itself a licence to ship a gate; the economics must still clear the acceptance rule on net return.

## Key risk, and it must be addressed head-on

The pools in #1422 that CAN resolve a separation say the separation does not survive the cluster null. The #1410 design's cells show 1.04 pp against a 0.38 pp limit at cluster p=0.4583. This study's exploratory grid shows 0.99 pp against a 0.42 pp limit at cluster p=0.4923. That pattern is consistent with whole datasets moving together, and not with H sorting trades inside them. A successor that only raises N may therefore confirm a null.

The successor must state this prediction BEFORE the run, and must report the outcome against it. If the higher-power design returns a null too, that outcome is the answer, and the issue closes with "no usable Hurst edge at or above the new limit" rather than with another request for more data.

## Acceptance criteria

- The successor script pre-registers its design constants at the top of the file, in #1422's style, and the report renders every number from the run's JSON.
- The report states the measured detection limit on the primary cohort AND that cohort's own observed separation, pool-matched. The successor is only valid when the limit falls below the separation. When it does not, the report says so plainly and the verdict stays a power statement.
- The look-ahead invariant is unchanged: rolling H at bar `i` reads closes `[i-W+1, i]`, the decision series shifts one bar, the fill-bar stamp shifts twice, and ADX carries the same two shifts.
- The estimator stays `hurst_exponent` from `shared_strategies/open/indicators_core.py` (#1409 SSoT) and is never reimplemented.
- NaN stays its own bucket for H and for ADX, and is never coerced to 0.5 or 25.
- The report records whether the key-risk prediction held.
- The contract path `backtest/research/hurst_gate_calibration.md` is handled the way #1422 handled #1410: the superseded render moves to its own issue-numbered file, and the superseding script owns the contract path. `scheduler/hurst_gate.go` and `docs/ARCHITECTURE.md` reference that path and must stay accurate.

## Untouched

`hurst_gate` semantics and its default-off state (#1411), the composite classifier and the label vocabulary (#1412 stays closed), every live path, and `config.example.json`.

## Execution

- Depends on: none
- Runs after: none
- Build model: Opus 5
- Build effort: high
- Validate effort: high
- fableplan first: Yes
- Plan effort: high
- PR review: Fable 5 · high (first review)

---

richkuo:
## Plain simple English

A study asks if a market memory number, called Hurst, can tell good trades from bad ones. The last study could not answer, because it was too weak to see an effect as small as the one in its data. This adds a stronger study. It is still too weak. So the answer stays open, and the feature stays off.

## Summary

Adds `backtest/research/hurst_1424_gate_resolution.py`, the successor to the #1422 power study, and hands it the live-evidence contract path `backtest/research/hurst_gate_calibration.md`.

#1422 returned INCONCLUSIVE because of POWER: its primary cohort resolved 0.96 pp/trade while separating by only 0.37 pp, so its null bounded the edge from above and said nothing about anything smaller. This study exists to make its OWN null readable, and it publishes a mechanical **validity gate** that decides — before any interpretation — whether the result is a statement about the market or about the design.

It combines all three routes the issue names:

1. **Fewer hypotheses.** Exactly ONE confirmatory hypothesis (`momentum/gate/W512/arm0.52/dis0.48`), derived as the smallest raw p over all 30 configs in the committed #1410 JSON, pinned as a constant and re-derived at run time with a hard assert. The Benjamini-Hochberg denominator is 1, so the rank-1 bar is alpha rather than alpha/4. The cost is stated in the report: this study cannot rank alternatives. #1422's interim look on a subset of the same rows is disclosed **before** the result, together with the direction of its bias.
2. **More INDEPENDENT calendar clusters.** Eight pre-2020 windows (`2013` … `2020H1`) from Bitstamp and Coinbase Exchange, venues the `binanceus` cache cannot reach — 16 windows and 26 datasets in total. Dataset identity is venue-qualified, a window-ownership matrix gives exactly one venue per (base asset, window), and same-base-asset pairs are credited correlation 1.0 so `BTC/USD` and `BTC/USDT` can never count as two independent markets. Every feasibility probe is recorded, including the refusals: binance global (HTTP 451 from this region), kraken (ignores `since` for deep history), TopStep (`get_ohlcv` takes only `limit`, never `since` — so the instrument-family route is not available in this repo today), IBKR (no OHLCV surface), and Coinbase's missing 4h granularity.
3. **A lower-variance target.** `signed_fixed_horizon_efficiency` over a 96-hour horizon fixed in calendar time, bounded in `[-1, 1]`. It is an OUTCOME exactly as PnL is, so no look-ahead invariant moves; a trade with too few bars left is EXCLUDED and counted, never truncated. `pnl_pct_net` is scored on the SAME rows as a continuity target. Significance reads the primary target and the economics rule reads net return, because a bounded-variance win is evidence about the mechanism and never on its own a licence to ship a gate.

### Result

**INCONCLUSIVE, and the validity gate FAILED.** On ~29.0k primary-cohort rows (against #1422's ~5.9k) the design resolves 0.008 efficiency units while that cohort separates by only 0.005 — still below its own limit. On net return the same cohort resolves 0.84 pp against a 0.23 pp separation, where #1422 read 0.96 pp against 0.37 pp: the limit fell, and the separation fell faster.

The verdict therefore stays a POWER statement and is not evidence that no edge exists. The pre-registered key-risk prediction is recorded as **UNRESOLVED**, because a design that still cannot see an effect the size its own buckets show cannot distinguish "whole datasets moving together" from "a small real effect".

The pinned hypothesis read cluster p=0.3557 on the primary target and 0.3387 on net return. #1422's 0.0485 on a subset of those rows did not reproduce, while the same config scored on the CONTAMINATED exploratory rows reads 0.0432 — the encouraging number lives on the rows that selected it, which is what the cohort split exists to expose.

#1412's Stage 0 is re-discharged on the larger pool: `NO JOINT SEPARATION` on both families (cluster p=0.9696 momentum, 0.7061 mean_reversion, against a Bonferroni bar of 0.025).

### Contract-path handover

Mirrors how #1422 handled #1410. The #1422 render moved to `backtest/research/hurst_1422_gate_power.md`, and that script now REFUSES any `--report-out` pointing at the contract path, so a stray `--render-only` there can never revert the live evidence.

### Shared surfaces touched

Both additive and behaviour-preserving:

- `shared_tools/data_fetcher.fetch_full_history` gained a `limit` parameter (default 500 keeps every existing caller byte-identical) so Coinbase's 300-candle page cap is respected. The loop already advanced by the LAST RETURNED candle, so a short page was always safe.
- `backtest/eval_windows.run_leg` gained an optional `exchange_id` so the harness-identity mirror loads the same venue's tape as the study.

### Nothing live changed

`hurst_gate` keeps its #1411 semantics and its default-off state, the composite classifier and label vocabulary are untouched, `config.example.json` still carries no `hurst_gate` block, and no thresholds are recommended anywhere. `scheduler/hurst_gate.go`, `docs/ARCHITECTURE.md`, `docs/backtesting-registry.md` and `CLAUDE.md` are updated to name the new owner and the actual verdict.

## Verification

- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — **3326 passed**, 92 of them new, covering all three routes, the ownership-collision assertions, the efficiency estimator's bounds and truncation rule, the protocol-window rule, the validity gate, the mechanical verdict vocabulary, and the contract-path protections on both scripts.
- `go -C scheduler test ./...` — ok. `go -C scheduler build .` — ok. `gofmt -l scheduler/` — clean.
- `--render-only` reproduces the committed report byte-for-byte from the committed JSON, for both #1424 and #1422.
- 860 of 860 ungated legs reproduce `eval_windows.run_leg` exactly, including every new-venue leg.
- Coverage audit: 172 of 189 owned cells scored; 17 dropped loudly with their reasons (late listing, delisting gap, insufficient warm-up lead). Minimum warm-up lead 627 bars against a required 514.

Closes #1424

---


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-10T04:56:35Z) in 69fd7d8a.

Both blocking findings validated against the code and were real, and both changed what the study is allowed to claim — so the full sweep was re-run and the contract report re-rendered from the new JSON. CI was green (`go`, `python`), so there were no CI findings.

**The verdict is unchanged (INCONCLUSIVE, gate FAILED, key risk UNRESOLVED), but the reason is different and stronger.** On the confirmatory family (`momentum`, 7,992 rows) the limit is 0.013 efficiency units / 0.94 pp and the separation is **-0.005 / -0.12 pp**. The trades the gate would have SUPPRESSED did BETTER than the ones it would have kept. The run carries no bound in either direction on that family.

### Fixed

1. **Validity gate compared a two-family pooled limit against a single-family separation** — `validity_gate` now reads `by_family_cluster[PRIMARY_FAMILY]` against `by_family_separation[PRIMARY_FAMILY]`, both on the confirmatory family's own rows (`backtest/research/hurst_1424_gate_resolution.py:867`). `measure_detection_limits` stores the new per-family separations on exactly the rows the per-family limits were measured on (`:1734`) — the pool-wide and per-family usability filters are not interchangeable, because a dataset rotatable in the combined pool can fall short inside one family's slice, so reusing `observed_separation_by_pool` would have been row-matched only by luck.
2. **Gate credited a separation pointing the way its one-sided test cannot reject** — the `abs()` is gone; the gate compares the SIGNED separation and records why it landed where it did in `gate["mode"]` (`:833`). `decide_recommendation` (`:953`) and the report's key-risk paragraph (`:1892`) branch on that mode: a too-small separation leaves a bound from above, a reversed one leaves no bound at all, and gluing one mode's sentence onto the other would publish a bound the run never measured. The test that pinned the magnitude behaviour encoded the defect and was replaced.
3. **Separations printed magnitude-only** (optional 1) — every separation in the report now carries its sign, with a per-family breakdown column beside it (`:2430`, `:2463`). `Largest separation` is the largest SIGNED value, and a reversed pool renders `n/a (reversed)` rather than a resolvability verdict the one-sided test cannot give. A new table prints the two row-matched numbers the gate actually reads with a `Reads the gate?` column, beside the pooled rows, so a reader can see they are different samples.
4. **Two stale ownership pointers** (optional 2) — `docs/backtesting-registry.md:81` no longer also claims #1422 OWNS the contract path, and `backtest/research/hurst_1410_gate_calibration.py:10` and `:210` name the current owner rather than a fixed predecessor, so the next handover cannot strand them again. A grep for `hurst_gate_calibration.md` now returns exactly one claimed owner.
5. **Committed-artifact guard covered only the narrowing flags** (optional 3) — `inference_deviations` (`:2791`) refuses the committed JSON and the contract path for any run whose `--n-perm`, `--n-perm-mde` or `--seed` differs from the pre-registered constant, or that passes `--no-mirror-check`. A value stated explicitly AT its pre-registered setting is not a deviation. The run stamps `scope.pre_registered_inference` and `--render-only` fails closed to the contract path without it.

### Resolved judgment calls

1. **Which family the gate belongs to** — the row-matching fix forced a choice the finding did not settle, and its must-survive case (c) pointed the way. The gate is evaluated on `PRIMARY_FAMILY` alone, because that is the family the single pre-registered hypothesis belongs to; another family's larger separation is not evidence about the confirmatory test's power (`:859`). Rejected: max over both families' row-matched pairs (lets an exploratory-only family decide the confirmatory verdict); requiring every family to pass (imports a hypothesis the study never pre-registered). Override if you would rather the gate speak for the whole primary cohort.
2. **Refreshing the committed decision without a third sweep** — `decide_recommendation` is a pure function of `(configs, mde)`, both of which the payload carries, so the decision block was recomputed in place and everything else asserted byte-identical. To make that provable rather than merely believed, `decision_payload` (`:2768`) now owns the projection and a test pins `payload["decision"] == decision_payload(decide_recommendation(configs, mde))`. Without that pin a stale JSON publishes a verdict today's rule would not reach, silently, because `--render-only` renders the STORED decision. Rejected: a third full sweep (identical output, seeded); relaxing `--render-only` to recompute the decision (hides JSON/code drift instead of failing on it).

**Plain simple English:** The study checks if it was strong enough to see an effect. Two things were wrong. It measured its strength on a bigger set of trades than the effect, and it ignored which way the effect pointed. Both are fixed. The answer stays the same, but the reason is now correct.

Verification: 3352 pytest tests pass (118 in the #1424 module); `go -C scheduler test ./...` ok; `gofmt` clean; `--render-only` reproduces the committed report byte-for-byte; 860 of 860 legs mirror-verified in the re-run.

---

richkuo:
@claude review

Issue #1285, opened 2026-07-09
[C45] Remove config-migration code for ancient config versions once every live deployment is confirmed on a recent version


richkuo:
**Complexity: 45/100** — scope: medium (config_migration.go ladder + DM-notice ladder + ~30 migration tests + operator checklist); risk: medium (config-loading path — a wrong cut could brick startup on an old config, but the design is fail-loud, never silent mis-load); uncertainty: medium (fleet config_version state unknown until the verification step runs)

## Problem

The config-migration surface still carries handlers for config versions that no live deployment should be running anymore:

- `CurrentConfigVersion = 16` (`scheduler/config_migration.go:13`).
- The field registry reaches back to **version 2** (`configFieldRegistry`, `scheduler/config_migration.go:25-40`: v2 `discord.owner_id`, v3 `portfolio_risk.warn_threshold_pct`).
- `MigrateConfig` runs a full ladder of legacy rewrites: v6/v7 DM-channel field removal + `translateV7DMChannels` (`config_migration.go:148-161`, `:267`), v8 dead-field cleanup (`:167-171`), `addV10SizingLeverage` (`:176-178`, `:311`), v12 ATR-stop knob (`:183-187`), `migrateV13StrategyShape` (`:197-199`, `:650`), `migrateV14Direction` (`:204-206`, `:610`), `migrateV15CloseKeys` (`:210-212`), `migrateV16UserDefaults` (`:217-221`).
- The operator-DM deprecation-notice ladder duplicates the same version checks (`config_migration.go:406-544`), with five multi-paragraph notice constants (v6-v15) kept alive alongside it.

Meanwhile the daemon **rewrites the config in place at the current version** on startup (`scheduler/main.go:522` triggers `MigrateConfig`; #1056 keeps the file writable at `/var/lib/go-trader[/<instance>]/config.json` precisely so this works). Any deployment that has run a recent build is therefore already at v16 — the old handlers are dead weight that every future migration change has to reason around, and `scheduler/config_migration_test.go` (30 tests) pins behavior for versions nobody runs.

A related but separable compat surface: the legacy `close_strategies` array (collapsed to single `close_strategy` in #842) is still accepted by `UnmarshalJSON` — `scheduler/config.go:547-548`, `:610-621` (len-1 lifted, len>1 rejected at `:1562-1571`), plus the wire type in `scheduler/strategy_composition.go:16`.

## Goal

[...]

PR #1434, opened 2026-08-17
feat(#1430): add live/paper config drift fleet audit [C37, Kimi, high]

### LINKED ISSUE #1430: [C37] Paper strategies trade at a different cadence and size than their live twins
**Complexity: 37/100** — Capability 1 (Uncertainty 2 — the paper sizing model was misspecified in the original draft and is corrected below); Volume 12 — Opus 5, high · fableplan: no

## Problem

When one `strategy_id` runs in both a live and a paper deployment (same strategy code, same `close_strategy`, same `regime_directional_policy`, same direction), the two per-strategy config blocks can drift in cadence and sizing. The drift is invisible until you compare `closed_positions` across the two `state.db` files and see live opening at hour T while paper opens at hour T+1, or the two legs holding very different quantities.

### Worked example: `hl-vwap-eth-60`

`hl-vwap-eth-60` runs in `hl-live` (live HL perps) and `hl-paper-hl-eth` (paper HL perps). Reported per-strategy differences:

| Field | Live | Paper |
|---|---|---|
| `args` mode | `--mode=live` | `--mode=paper` |
| `interval_seconds` | `300` (every 5 min) | `3600` (hourly) |
| Sizing | `leverage: 20`, `margin_per_trade_usd: 50` | `leverage: 1`, virtual capital only |
| `direction` | `long` | `long` |
| `close_strategy` | `trailing_tp_ratchet_regime` (defaults) | `trailing_tp_ratchet_regime` (defaults) |
| `trailing_stop_atr_regime` | 2.5 ATR across all 9 regimes | 2.5 ATR across all 9 regimes |
| `trailing_stop_min_move_pct` | `0.2` | `0.2` |
| `regime_directional_policy` | byte-identical | byte-identical |

Reported observation from `closed_positions`: on 2026-08-10T12:53Z both opened long ETH near 1908; live held ~10 min and exited at a $11.75 loss, paper held ~3h and exited at a $13.03 loss. Other entries in the last 7 days differ by hours or are missing on one side.

The table above is **not verifiable from this repository**. Deployment configs live out of the checkout at `/var/lib/go-trader[/<instance>]/config.json` (#1056) and the in-tree path is git-ignored, so the exact per-deployment values must be confirmed on the host before any edit.

### Two corrections to the original draft

1. **`initial_capital` does not size anything.** It is a fixed starting balance used for PnL display only. The field that funds a paper book, and therefore bounds paper open size, is `capital`; it seeds the strategy's virtual cash at state creation.
2. **`margin_per_trade_usd` is NOT ignored in paper mode.** Its resolver gates on strategy type `perps` alone, with no live/paper condition, and the paper perps open path sizes through the same notional function as live. Setting it on a paper strategy changes the size. This removes the draft's "the change may be a no-op" risk.

### Why `capital: 0` cannot be used

Config validation rejects a strategy with `capital <= 0` and no `capital_pct`. The only exemption is shared-wallet pool budgeting, and that opt-in is restricted to a cluster of two or more **live** perps strategies on a platform with a balance fetcher — a paper strategy never qualifies. Config load runs validation, so a paper block with `capital: 0` fails at startup and is rejected on SIGHUP (the daemon keeps the previous config). The paper virtual cash book must stay funded.

## Goal

Bring each paper strategy's cadence and sizing in line with its live twin, so the two books are comparable. Change only per-strategy config; keep `--mode=paper`.

### 1. Find candidate pairs (repo deliverable)

Add a read-only fleet audit script, `scripts/check-live-paper-config-drift.sh`, modeled on `scripts/check-config-versions.sh`:

- Discover deployments with `discover_deployment_dirs_from_systemd` (`scripts/update_helpers.sh`) and read each unit's `ExecStart --config` path, falling back to `<WorkingDirectory>/scheduler/config.json`.
- Group strategy blocks by `strategy_id` across deployments and classify each as live or paper from its `--mode=` argument.
- For every id present on both sides, print the drift in `interval_seconds`, `leverage`, `sizing_leverage`, `margin_per_trade_usd`, `capital`, `capital_pct`, and `initial_capital`, and flag any *other* differing field as a reason to leave the pair alone.
- Write nothing: no config rewrite, no daemon interaction. Exit non-zero when drift is found so the audit can gate an operator runbook.

A pair is a candidate only when the differences are limited to cadence, sizing, and `--mode`.

### 2. Apply the sync (operator action, per candidate pair)

On the paper block only:

1. Set `interval_seconds` to the live value.
2. Set `leverage` to the live value — **from flat only** (see ordering below).
3. Set `margin_per_trade_usd` to the live value, and keep `capital` at or above that number so the margin cap binds instead of the cash book. Remove `sizing_leverage` if present, because a positive `margin_per_trade_usd` takes priority over it.
4. Leave `capital` positive. Do not set it to `0`.
5. Keep `--mode=paper`.

### 3. Reload ordering

- `interval_seconds` and `margin_per_trade_usd` hot-reload on SIGHUP, and the scheduler tick is recomputed during the reload.
- A `leverage` change is **refused while the strategy holds an open position**; the reload is rejected as a whole. Flatten the paper leg first, or apply the leverage change during a flat window.
- Record the pre-change values so a single-file revert plus SIGHUP restores the previous state.

## Acceptance criteria

- `scripts/check-live-paper-config-drift.sh` exists, is read-only, discovers deployments through the shared systemd helper, and reports every live/paper pair with its cadence and sizing drift. It has a test alongside `scripts/test_update_helpers.sh`.
- The script's output for the fleet is recorded in the pull request as the candidate list, with a per-strategy before/after table of the values to be applied.
- No paper block ends with `capital: 0` or a removed `capital` key; every edited deployment loads cleanly (verify with the binary's config load before restart).
- Every leverage change is applied from a flat paper position, and the runbook states this.
- After the change, for each synced pair the paper leg opens within minutes of the live leg for the same `strategy_id`, over the next 5–10 closed trades.
- The regime-bundle and script-failure alert rate on the paper deployment is checked for the first hour after each cadence change and reported.

## Verification

Per updated paper strategy, after SIGHUP reload or restart:

- `sqlite3 <state.db> "SELECT opened_at, quantity, avg_cost FROM positions WHERE strategy_id='<id>'"` — paper opens within minutes of live opens for the same id.
- The next 5–10 rows of `closed_positions` show paper close times within 5–15 min of live close times.
- `close_reason` strings still differ (`hl_sync_stop_loss` on live vs `trailing_stop_loss_paper` on paper). This is expected.

## Known residual differences (out of scope to fix here)

Cadence and sizing sync narrows the gap; it does not make the two legs identical:

- **Exit timing.** Live rests the stop on the exchange, so it fills inside a bar at any moment. Paper tests the trigger only against the mark at each cycle, so the paper exit price and time are quantized to `interval_seconds`. A 5-minute paper cycle still cannot reproduce an intrabar live fill.
- **On-chain take-profit** is live-only; paper never places one.
- **Process phase.** The two deployments are separate processes with independent tick phases, so equal intervals still allow up to one interval of offset.
- **Fees.** Live books real fill fees; paper books modeled fees.

True decision-level shadowing is tracked in #1431.

## Risks

- **Cadence load.** The scheduler tick is the minimum of all strategy intervals in a deployment, so lowering one strategy from 3600 to 300 speeds up the whole paper deployment roughly 12× and multiplies regime-bundle and balance fetches. Issues #1432 and #1433 already report regime-bundle 502 responses and timeouts being mis-classified as script failures at the current rate; a 12× increase raises that exposure. Watch the alert channel for the first hour per deployment.
- **Risk profile.** Raising paper `leverage` from 1 to 20 changes the paper book's drawdown and circuit-breaker behavior, not only its trade size. Confirm `max_drawdown_pct` and the circuit-breaker settings are still sensible for the new size.
- **Cross-deployment isolation.** Do not sync a paper strategy that has a documented reason for a different cadence or size (a long-history research variant, for example). The audit script's "other fields differ" flag is the filter; record any pair left alone and why.

## Plain simple English

Some strategies run twice: once with real money and once on paper. The paper copy is set to trade less often and in a much smaller size, so the two results cannot be compared. This issue makes the paper copy use the same timing and size as the live copy, and adds a check that finds these mismatches across all deployments.

---

---
Validated with LLM: Opus 5 | high | Harness: validate-issue

richkuo:
## Plain simple English

Some strategies run twice: once with real money and once on paper. The paper copies can drift to a slower schedule and smaller size, so the two results cannot be compared. This adds a read-only check that scans every deployment and reports each live/paper pair whose timing or size has drifted. Changing the configs stays a human-run step.

## Summary

Closes #1430. Repo deliverable only — the config sync itself is the operator runbook the issue specifies, not a code change.

- New `scripts/check-live-paper-config-drift.sh`, modeled on `check-config-versions.sh`:
  - Discovers active go-trader systemd units via the shared helpers (`update_systemd_unit_globs`, `update_execstart_config_path`), reading each unit's `ExecStart --config` path and falling back to `<WorkingDirectory>/scheduler/config.json`; explicit deployment dirs also accepted.
  - Groups strategy blocks by `id` across deployments and classifies each as live/paper from its `--mode` arg, mirroring `isLiveArgs` (`--mode=live` / `--mode live` win; `--mode=paper` explicit; anything else is `unset` and never pairs).
  - For every id on both sides, prints drift in `interval_seconds`, `leverage`, `sizing_leverage`, `margin_per_trade_usd`, `capital`, `capital_pct`, `initial_capital`; any **other** differing field (mode-stripped `args` included) flags the pair `SKIP` — leave it alone.
  - Read-only (no config writes, no daemon interaction). Exit 1 when syncable drift is found or a deployment is unreadable/missing, exit 2 on an empty audit, exit 0 when pairs are in sync or only SKIP-flagged.
- Tests added alongside the #1285 fleet-audit tests in `scripts/test_update_helpers.sh` (drifted pair → exit 1 + CANDIDATE; in-sync → exit 0 + IN SYNC; other-fields-only → SKIP, no gate; drift+other → SKIP, still gates; unreadable/missing config → exit 1; read-only assertion).

## Verification

- `bash scripts/test_update_helpers.sh` → `OK: update_helpers tests passed` (new tests were run red first against the missing script).
- `shellcheck` on the new script: no new findings (only the shared SC1091 source-resolution note the existing scripts also carry).
- Auto-discovery mode on a host without `systemctl` exits 2 with a clear error, as designed.

## Fleet output

The acceptance criterion asks for the audit's fleet output as the candidate list. This session ran on a macOS dev checkout with no go-trader systemd units, so the real fleet run must happen on the production host: `bash scripts/check-live-paper-config-drift.sh` (no args). Sample output against a fixture matching the issue's `hl-vwap-eth-60` worked example:

```
PAIR hl-vwap-eth-60
  DRIFT  interval_seconds       live=300   paper=3600
  DRIFT  leverage               live=20    paper=1
  DRIFT  margin_per_trade_usd   live=50    paper=-
  DRIFT  capital                live=100   paper=10000
  VERDICT: CANDIDATE — differences limited to cadence/sizing/--mode
```

## Judgment calls (not in the issue text)

- A pair whose **only** differences are outside cadence/sizing (e.g. a documented research variant) reports `SKIP` but does **not** fail the exit-code gate; the gate fires on syncable drift or unreadable deployments, so an intentional difference doesn't permanently block the runbook gate. SKIP pairs with cadence/sizing drift still gate (exit 1).
- Strategies with no `--mode` token are reported as `unset` and never paired, rather than assumed paper.

LLM: Kimi | high | Harness: Kimi Code CLI

### DIFF
--- a/scripts/check-live-paper-config-drift.sh
+++ b/scripts/check-live-paper-config-drift.sh
@@ -1,0 +1,264 @@
+#!/usr/bin/env bash
+# check-live-paper-config-drift.sh — READ-ONLY fleet audit of live/paper
+# per-strategy cadence + sizing drift (#1430).
+#
+# When one strategy id runs in both a live and a paper deployment, the two
+# per-strategy config blocks can drift in cadence (interval_seconds) and
+# sizing (leverage, sizing_leverage, margin_per_trade_usd, capital,
+# capital_pct, initial_capital), making the two books incomparable. This
+# script finds every live/paper pair across the fleet and prints the drift.
+#
+# A pair is a sync CANDIDATE only when its differences are limited to
+# cadence, sizing, and the --mode arg; any OTHER differing field (script,
+# args beyond --mode, close_strategy, …) flags the pair SKIP — leave it
+# alone, it has a documented reason to differ or needs human review.
+#
+# Usage:
+#   bash scripts/check-live-paper-config-drift.sh                    # auto-discover active systemd deployments (#1055)
+#   bash scripts/check-live-paper-config-drift.sh /opt/go-trader ... # audit explicit deployment dirs instead
+#
+# Discovery mirrors update.sh --all: active go-trader systemd units, reading
+# each unit's ExecStart --config path (the #1056 out-of-tree location) and
+# falling back to <WorkingDirectory>/scheduler/config.json (the transition
+# symlink, or the legacy in-tree file).
+#
+# The script never writes anything — no config rewrite, no daemon
+# interaction. Mode classification mirrors the daemon: a strategy is live
+# when its args carry --mode=live (or "--mode live"), paper when they carry
+# --mode=paper; anything else is "unset" and never forms a pair.
+#
+# Exit codes:
+#   0 — every live/paper pair in sync on cadence/sizing (SKIP-only flags do
+#       not gate: those pairs are deliberately left alone)
+#   1 — cadence/sizing drift found, or a deployment unreadable/missing
+#   2 — nothing to audit (an empty audit is NOT a verified fleet)
+set -euo pipefail
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=update_helpers.sh
+source "${SCRIPT_DIR}/update_helpers.sh"
+
+rows_file=$(mktemp)
+trap 'rm -f "$rows_file"' EXIT
+rows=0
+
+add_row() {
+    local source="$1" cfg_path="$2"
+    printf '%s\t%s\n' "$source" "$cfg_path" >> "$rows_file"
+    rows=$((rows + 1))
+}
+
+if [[ $# -gt 0 ]]; then
+    for dir in "$@"; do
+        dir=$(canonicalize_deployment_dir "$dir")
+        add_row "$dir" "${dir}scheduler/config.json"
+    done
+else
+    if ! command -v systemctl >/dev/null 2>&1; then
+        echo "ERROR: systemctl not available and no deployment dirs given — pass dirs explicitly" >&2
+        exit 2
+    fi
+    globs=()
+    while IFS= read -r g; do
+        [[ -n "$g" ]] && globs+=("$g")
+    done < <(update_systemd_unit_globs)
+    units=()
+    while IFS= read -r unit; do
+        [[ -n "$unit" ]] && units+=("$unit")
+    done < <(systemctl list-units --type=service --state=active --no-legend --plain "${globs[@]}" 2>/dev/null | awk '{print $1}')
+    if [[ ${#units[@]} -eq 0 ]]; then
+        echo "ERROR: no active go-trader systemd units found — an empty audit is not a verified fleet" >&2
+        exit 2
+    fi
+    for unit in "${units[@]}"; do
+        execstart=$(systemctl show "$unit" -p ExecStart --value 2>/dev/null || true)
+        cfg_path=$(update_execstart_config_path "$execstart")
+        if [[ -z "$cfg_path" ]]; then
+            wd=$(systemctl show "$unit" -p WorkingDirectory --value 2>/dev/null || true)
+            if [[ -z "$wd" ]]; then
+                # Record an unreadable row so the audit fails loudly instead of
+                # silently skipping a deployment.
+                add_row "$unit" "-"
+                continue
+            fi
+            cfg_path="${wd%/}/scheduler/config.json"
+        fi
+        add_row "$unit" "$cfg_path"
+    done
+fi
+
+if [[ "$rows" -eq 0 ]]; then
+    echo "ERROR: nothing to audit" >&2
+    exit 2
+fi
+
+python3 - "$rows_file" <<'PY'
+import json
+import sys
+
+# Cadence + sizing fields the #1430 sync runbook is allowed to touch.
+WATCHED = [
+    "interval_seconds",
+    "leverage",
+    "sizing_leverage",
+    "margin_per_trade_usd",
+    "capital",
+    "capital_pct",
+    "initial_capital",
+]
+MISSING = object()
+
+
+def classify(args):
+    # Mirrors scheduler/state_presence.go isLiveArgs: --mode=live or
+    # "--mode live" wins; then explicit paper; anything else is unset and
+    # never forms a pair.
+    for i, a in enumerate(args):
+        if a == "--mode=live":
+            return "live"
+        if a == "--mode" and i + 1 < len(args) and args[i + 1] == "live":
+            return "live"
+    for i, a in enumerate(args):
+        if a == "--mode=paper":
+            return "paper"
+        if a == "--mode" and i + 1 < len(args) and args[i + 1] == "paper":
+            return "paper"
+    return "unset"
+
+
+def strip_mode(args):
+    # Remove the --mode token(s) so a pair whose args differ ONLY by
+    # --mode=live vs --mode=paper is not flagged as other-drift.
+    out = []
+    skip = False
+    for a in args:
+        if skip:
+            skip = False
+            continue
+        if a in ("--mode=live", "--mode=paper"):
+            continue
+        if a == "--mode":
+            skip = True
+            continue
+        out.append(a)
+    return out
+
+
+def fmt(v):
+    if v is MISSING:
+        return "-"
+    return json.dumps(v)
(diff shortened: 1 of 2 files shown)


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-17T12:20:27Z) in 51b9231.

### Fixed
1. **Watched+other pair classified SKIP bypassed the exit-1 gate** — confirmed at `scripts/check-live-paper-config-drift.sh:237`: the `watched_diffs and other_diffs` branch incremented only `skip_pairs`, so a single such pair exited 0. Now any pair with a watched (cadence/sizing) difference increments the gating counter while keeping the SKIP label (`scripts/check-live-paper-config-drift.sh:237-244`); other-fields-only pairs still exit 0. Header exit-code doc and DRIFT summary wording updated to match. New regression test uses a single live/paper3 pair with `interval_seconds` drift plus a differing `close_strategy`, asserting exit 1 — verified red before the fix (`scripts/test_update_helpers.sh:402-414`). `bash scripts/test_update_helpers.sh` passes; CI go/python were green on the prior head.

LLM: Kimi | high | Harness: Kimi Code CLI

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-17T13:23:16Z) in 80a98ee.

### Optional implemented
1. **No-mode paper twin classified `unset` and silently dropped** — confirmed: the daemon runs any non-`--mode=live` args as paper (`scheduler/config.go:517`, `scheduler/main.go:3837-3840`), so the audit's paper notion was narrower than the daemon's. Implemented the reviewer's middle ground, extended one step: an unset block whose id has a live twin but no explicit `--mode=paper` block now **pairs as the paper twin** with a header annotation (`scripts/check-live-paper-config-drift.sh:206-211`), so its drift is actually computed and gates — a bare warning would surface the block but still hide the drift. When an explicit `--mode=paper` twin already pairs the id, the unset block is flagged `UNPAIRED (unset mode)` with a remediation note instead of a second pair (no double-report); unique-id unset blocks still produce nothing. Rejected alternative: warning-only even when no explicit twin exists — surfaces the gap but never the drift, which is the failure this tool exists to catch. Tests: no-mode twin pairs + gates, unique-id unset produces nothing, explicit+unset coexistence yields exactly one pair plus UNPAIRED (`scripts/test_update_helpers.sh`, all verified red first). `bash scripts/test_update_helpers.sh` passes.

LLM: Kimi | high | Harness: Kimi Code CLI

richkuo:
@claude sonnet review

Issue #1147, opened 2026-06-30
[C62] Per-trade per-strategy diagnostics store + on-demand tuning report (diagnostics-only)


richkuo:
**Complexity: 62/100** — scope: medium-large (new `state.db` table + idempotent migration, a close-time capture hook that computes MFE/MAE from the hold-window OHLCV, a `diagnostics` CLI subcommand, and a diagnostics→hypothesis mapping); risk: low-moderate (read/log only — never touches live trades or config — but the capture hook runs on the close path and needs hold-window candles); uncertainty: moderate (whether the diagnostics yield actionable tuning, MFE/MAE computation correctness, and sample-size thresholds).

## Problem

go-trader records each trade's realized P&L in the `trades` table, and `closed_positions` already stores entry/exit price, side, close reason, and hold duration per full close — but nothing captures **trade quality**: how good the entry timing was (MAE), how much of the favorable move the close strategy actually captured (MFE / capture ratio), or expectancy split by regime at open. `closed_positions` also lacks `position_id` and the regime label, so it can't join cleanly to `trades` or answer "which regimes should be gated out." Strategy tuning today is ad-hoc against the backtester with no systematic evidence of *what* to tune.

## Goal

A **diagnostics-only** system that logs trade-quality metrics for **every** closed position into `state.db`, then surfaces per-strategy findings and concrete, backtestable tuning hypotheses through an **on-demand CLI report**. It never touches live trades or config and never auto-tunes — it tells the operator what to investigate and gives the exact backtest command to validate a fix.

This is independent of #1137 (the LLM entry analysis). The deterministic metrics need no LLM; #1137, when enabled, just enriches the same record with an optional verdict column.

## Architecture (three independent pieces)

[...]

PR #1435, opened 2026-08-18
feat(#1431): add live-to-paper decision replay log [C72, Kimi, high]

### LINKED ISSUE #1431: [C72] Add a live-to-paper decision replay log so paper deployments shadow live trades
**Complexity: 72/100** — Capability 2 (Risk 3 — rows are written on the live trade-persistence path, so a write bug is a data-integrity risk on a real-money ledger; Uncertainty 3 — the mechanism is known but the cross-deployment coordination shape still needs discovery); Volume 22 (Scope 4, Coupling 4, Verification 3) — Opus 5, high · fableplan: yes

## Problem

When a strategy runs in both a live and a paper deployment (same `strategy_id`, same strategy code, same `close_strategy`, same `regime_directional_policy`), the config-sync issue (#1430) closes the cadence and sizing gap but does not make paper a *true* shadow of live. Two fundamental residual gaps remain:

1. **Decision divergence**: even with matching cadence and sizing, live and paper can open different trades if their indicators differ at the moment of evaluation (price snapshot, regime window state, OHLCV cache hit/miss, scale-in timing). Live and paper are separate deployments with separate `state.db` files; paper executes in-process against its virtual book while live executes through the exchange subprocess, so there is no shared decision point.
2. **Execution divergence**: live stop-loss exits are booked by the balance reconciler as `hl_sync_stop_loss` only after the exchange-side SL fill is confirmed; paper trailing-stop exits are simulated and booked as `trailing_stop_loss_paper` (paper also exits via close-strategy evaluation and other paper reasons such as `stop_loss_atr_paper`). The exit price can differ by minutes and several dollars per coin.

### Worked example: hl-vwap-eth-60

On 2026-08-10T12:53Z both `hl-live` and `hl-paper-hl-eth` opened long ETH near 1908, but live held ~10 min and exited $11.75 loss, paper held ~3h and exited $13.03 loss. Closer than usual but still divergent — and this is the *same* strategy running on the same exchange.

**Verifiability caveat:** the worked example and per-deployment config values are runtime deployment data. Deployment configs and state live out of the checkout at `/var/lib/go-trader[/<instance>]/` (#1056), so they are not verifiable from this repository and must be confirmed on the host.

## Goal

Make paper a faithful shadow of live for any strategy that opts in: every live open (full, scale-in, partial close, full close) drives the same trade in paper, sized to live's actual filled position, with the same entry/exit timestamps. Paper's only legitimate differences become (a) no real money moves, and (b) exit price may differ when live's fill is asynchronous.

## Approach

A shared **decision log** that live writes to and paper replays from. The mechanism is generic — any strategy that runs in both modes can opt in by setting a `replay_sharing` flag.

### Storage

- A new SQLite decision-log database at an **operator-configured path outside both deploy trees** (new config knob, e.g. `replay_log_path`, pointing at a `StateDirectory`-compatible location such as `/var/lib/go-trader/shared/`). The log must NOT live under the code checkout (e.g. `shared_tools/`): deployed instances run with `ProtectSystem=strict`, the checkout is read-only, and writable state is restricted to the state directory (#1056). Live and paper deployments on the same host both open this path.
- Schema, roughly:
  - `decision_id INTEGER PRIMARY KEY`
  - `strategy_id TEXT NOT NULL`
  - `decision_type TEXT NOT NULL` — `open` | `scale_in` | `partial_close` | `full_close`
  - `decided_at TEXT NOT NULL` — ISO 8601 UTC of live's decision
  - `symbol TEXT NOT NULL`
  - `side TEXT NOT NULL`
  - `quantity REAL NOT NULL` — signed qty in coin units, **live's actual filled quantity**
  - `reference_price REAL NOT NULL` — live's actual fill price (VWAP) where the fill resolver provides it
  - `close_reason TEXT NOT NULL` — for close decisions: the live close reason
  - `replay_status TEXT NOT NULL DEFAULT 'pending'` — consumption state for the single configured mirror
- Index on `(strategy_id, decided_at)`.

### Live write path

- Decision rows are written **after fill resolution** (the fill resolver in `hyperliquid_fills.go`), not at decision time — live's actual filled quantity and price are only known post-fill, so a decision-time row would record intended rather than actual size.
- The write-set covers every site that opens or closes live exposure: open execution, scale-in (`scale_in.go`), partial close, close-strategy full close, the stop-loss reconciler (`hl_sync_stop_loss` in `hyperliquid_balance.go`), trailing/ratchet closes, and kill-switch closes. The clean choke point is the shared position/trade recording helpers, not each call site.
- Fail-safe: the decision insert runs in its **own transaction immediately after** the position/trade write commits — never in the same transaction. A same-transaction insert would let a log failure roll back trade state, contradicting the fail-safe requirement. If the log insert fails, the live trade and its bookkeeping stand; alert on consecutive failures.

### Paper replay path

- New paper-side flag: `replay_sharing: live_mirror` on the strategy config block (default `none`). As a new per-strategy flag it needs the full flag surface: config field + unknown-key guard, wizard/init option, and a `validateHotReloadCompatible` rule — toggling `replay_sharing` is **blocked while a position is open** (paper's open book would desync from live's log mid-trade), hot-reloadable while flat.
- On each paper tick, instead of computing its own decision, scan the decision log for `replay_status='pending'` rows for the relevant strategy_id, replay the same open/scale-in/close actions against paper's local state with live's filled quantities, and mark `replay_status='applied'`.
- When `replay_sharing=live_mirror`, paper does *not* run the check script's open logic — it only re-evaluates closes (so it can still apply `trailing_stop_loss_paper` if live missed a stop).
- When `replay_sharing=none` (default), paper behaves exactly as today.
- Replayed closes book `close_reason = replay_live_mirror`, which must be registered in the close-reason display mapping (`stopLossCloseDetailsPrefix` in `portfolio.go`) and any reason-keyed dashboards.
- **Hedge interaction (#1159):** no separate hedge replay is needed — paper's own state-derived hedge reconciler converges the hedge leg from the replayed primary position, same as it does for a natively-opened one. Validate this in the shadow-week rollout; if it diverges, `live_mirror` is rejected for hedge-enabled strategies in v1.

### Open design questions

- **Source of truth for paper sizing**: resolved — replay live's actual filled quantity. The point is shadowing, not parallel sizing.
- **What happens during live downtime**: paper sees a gap in the log. Options: (a) hold last replayed state, (b) flat-out, (c) resume on next live decision. Recommendation: (c) resume, with the watchdog alerting on the gap. Needs a final policy.
- **Replay window**: resolved — full history on paper startup so a fresh paper deployment can resync from any replay-safe point.
- **Multiple mirrors**: resolved for v1 — **one mirror per live strategy**. The single `replay_status` column can only track one consumer; multiple mirrors would need per-mirror consumption rows (e.g. a `replay_cursor` table keyed by mirror deployment) and are deferred to a follow-up.
- **Opt-in vs opt-out**: resolved — explicit per strategy. Forces operators to think about whether shadow behavior is desired.

## Out of scope

- Config sync (#1430) — should land first, so the replay layer is built on top of matched cadence/sizing.
- Replacing the paper close-reason (`trailing_stop_loss_paper`) with a live-mirror exit price. Even with replay, paper's *close execution* stays simulated; only the *decision* is mirrored.
- Replay for strategies that have no live / paper pair (single-mode strategies).
- Multiple paper mirrors of the same live strategy (v1 is single-mirror; see Approach).

## Rollout

1. Land the #1430 config-sync changes first for all candidate strategies.
2. Implement the decision log database + live write path behind a feature flag.
3. Implement the paper replay path with `replay_sharing=none` as the default.
4. Opt in one strategy (suggest `hl-vwap-eth-60`) and run paper in shadow mode for a week.
5. Compare paper's `positions` / `closed_positions` against live's at the end of each day. If drift is within the verification window, flip the flag for more strategies.
6. Document the operator playbook for enabling replay per strategy.

## Acceptance criteria

- After deploy, for every strategy with `replay_sharing=live_mirror`, paper's `positions` and `closed_positions` rows match live's within 1 minute of each open/close decision.
- Paper's realized PnL is within ~0.1% of live's over a 5-trade window per strategy (only acceptable drift is fill-price timing).
- Paper's `close_reason` appears as `replay_live_mirror` for replayed closes, and `trailing_stop_loss_paper` only when paper's own trailing SL fires before live's mirroring reaches that close.
- A live trade whose decision-log insert fails still executes and books normally; consecutive insert failures raise an alert.
- Toggling `replay_sharing` via SIGHUP is refused while the strategy has an open position, and applies cleanly while flat.

## Risks

- **Replay dependency**: paper's correctness now depends on live's decision log integrity. A corrupted log = paper losing state. Need a watchdog that alerts on `(strategy_id, currently_open)` mismatch between live and paper.
- **Replay slippage**: paper's fill price on a replayed close is paper's last-seen price, not live's actual fill. Drift acceptable; black-swan moves unacceptable.
- **Single-source-of-truth question**: if paper ever needs to make an independent decision (e.g., live missed a stop), the policy has to be explicit. Recommend an escalation log that records paper-side overrides.
- **Adoption risk**: if a strategy gets `replay_sharing=live_mirror` enabled by accident, paper will stop computing its own signal and silently start mirroring live. That's the desired behavior, but it should be loud in the operator's audit trail so it's not surprising.

## Plain simple English

Paper deployments run the same strategies as live but make their own decisions, so their trades drift away from live. This change lets a paper deployment copy every live trade decision — open, add, close — with live's real sizes, so paper becomes a faithful shadow used to judge whether a strategy truly works.

---
Validated with LLM: kimi-for-coding | high | Harness: Kimi Code

richkuo:
## Plain simple English

Paper Hyperliquid deployments can now copy a live deployment's actual trades — same size, same fill price, same timing — instead of computing their own entries that drift away from what live really did. This is off by default; turn it on with one config flag per strategy pair. Paper still manages its own stops and exits as a safety net.

## Summary

Closes #1431.

Implements the live→paper decision replay log:

- **`scheduler/replay_log.go`** — shared SQLite `DecisionLogDB` at the new root `replay_log_path` (WAL, busy_timeout, single writer conn; opened at startup by both deployments, fatal if configured-but-unopenable). `ReplayDecision` rows carry decision type (open/scale_in/partial_close/full_close), actual filled qty + VWAP, EntryATR/regime stamps, and mirrored timestamps. `replayLiveSources` registry rebuilt at startup and on SIGHUP.
- **Per-strategy flag** `replay_sharing` (`"none"` default | `"live_mirror"`), HL perps only, Go-only (never forwarded to check scripts — same class as `paused`), hot-reloadable flat-only; `replay_log_path` itself is restart-required.
- **Live write hooks** at the recording choke points — `recordPositionOpen` (open vs scale_in by position growth), `recordClosedPosition` (funnels every close incl. SL reconciler / kill-switch / CB / external), `bookPerpsPartialCloseWithFillFee`, and both signal-driven partial branches in `executePerpsSignalWithLeverage`. Hedge legs and zero-qty rows skipped. Consecutive insert failures DM the owner at 3, once per streak (`decisionLogPersistWarn`).
- **Paper mirror (`scheduler/replay_mirror.go`)** — a paper HL perps strategy with the same `id` suppresses its own position-increasing signals (a `replayMirrorPaperActive(sc) && pausedBlocksSignal(...)` arm in the HL perps dispatch, after the #1270 exposure-cap block) and replays pending rows each cycle (`applyReplayedLiveDecisions`: fetch outside `mu`, apply inside, mark applied after). Opens book the exact live qty at live VWAP via `ExecutePerpsSignalWithLeverageDeferredOpen` with EntryATR/regime stamping; scale-ins via `applyPerpsScaleIn`; partial closes via `bookPerpsPartialCloseWithFillFee`; full closes via `bookPerpsClose` at paper's current mark with reason `replay_live_mirror`. Paper's own close re-evaluation and trailing SL keep running as a backstop.
- **Drift handling:** open-while-holding, close-while-flat, and side-mismatch rows are consumed with a WARN, never retried — a wedge would block the whole queue. In-memory `replayMirrorProgress` high-water prevents double-apply if `MarkDecisionsApplied` fails.
- Hedge legs need no replay rows: paper's state-derived #1159 hedge reconciler converges from the replayed primary.
- Docs: AGENTS.md guardrail bullet, ARCHITECTURE.md mechanics entry, `config.example.json` knob. `CurrentConfigVersion` stays 17 (precedent #1411).

No implementation plan was posted on the issue (zero comments despite `fableplan: yes` metadata); the diff is built from the issue body and acceptance criteria directly.

## Verification

- 22 new tests in `scheduler/replay_log_test.go` + `scheduler/replay_mirror_test.go`: DB round-trip/pending/mark-applied, registry rebuild + predicates, write-hook gating (source off, nil writer, hedge legs, zero-qty), failure-streak alerting, config validation vocabulary/scope/required-path, restart-required + flat-only hot-reload rules, mirror apply for all four decision types, drift-row consumption, high-water double-apply guard.
- `go -C scheduler build .` + `go vet` clean; `gofmt -l scheduler/` empty.
- Full Go suite: `go -C scheduler test ./... -count=1` → `ok trading-scheduler` (~40s).
- Python suite: `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` → green (Go-only change; sanity).

## Unfiled follow-ups (named in the issue's Risks, not acceptance criteria)

These remain **unfiled** and need their own issues:

1. Watchdog/alert for live-vs-paper position mismatch (replay drifts silently today beyond the per-row WARN).
2. Multi-mirror cursors (multiple paper deployments shadowing one live source need per-consumer applied-cursors).
3. Paper-side override escalation log (when paper's own backstop SL fires before the mirrored close arrives).

---
Created with LLM: Kimi | high | Harness: Kimi Code


richkuo (commented):
Needs Updates

### Needs Fixing

1. **Paper apply is not atomic with persisted state, so a restart can drop a mirrored trade or book the same add twice.**
The log's `replay_status` is treated as the durable cursor (`scheduler/replay_mirror.go:24-29`, `scheduler/main.go:2894-2901`), but paper positions are only persisted at cycle-end `SaveStateWithDB` (`scheduler/main.go:3425`). `MarkDecisionsApplied` runs immediately after in-memory apply, minutes of other-strategy work later. `replayMirrorProgress` is process-memory only (`scheduler/replay_mirror.go:30-33`) and is cleared on restart; `TestMirrorReplayHighWaterPreventsDoubleApply` never restarts. SQLite WAL allows only one writer at a time (https://www.sqlite.org/wal.html §2.2), so a live `InsertDecision` can make paper's mark return busy (`busy_timeout=5000`) in that window. Verified against the PR files and sqlite.org, not the PR body.

**Invariant:** A replayed decision must change paper's persisted book at most once, and must stay pending until that persist succeeds.

**Must survive:** (1) apply + mark-applied, then kill before cycle-end SaveState — restart must re-apply, not drop the row; (2) apply + SaveState, mark fails, then restart — must not double a scale-in; (3) an open then a scale-in in the same pending batch across that restart.

**Plain simple English:** Paper copies a live trade into memory, then tells the shared log "this row is done" before that copy is saved to disk. A restart in that gap can lose the copy, or book the same add twice.

2. **Shipped systemd instances cannot write the documented shared log path, so enabling the feature fails closed on the intended live+paper layout.**
`systemd/go-trader@.service` sets `ProtectSystem=strict` (line 46), `StateDirectory=go-trader/%i` (line 20), and `ReadWritePaths` only for that instance's checkout `scheduler/` and `logs/` (line 51). systemd.exec: `ProtectSystem=strict` mounts the whole tree read-only except `/dev`, `/proc`, `/sys`; only `StateDirectory=` / `ReadWritePaths=` are excluded (https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html). Two template instances therefore cannot write `/var/lib/go-trader/shared/`, which `scheduler/config.go` and `docs/ARCHITECTURE.md` name as the example. `OpenDecisionLogDB` then fails and `scheduler/main.go:193-194` exits 1. `PrivateTmp=true` also blocks `/tmp` as a shared fallback. This PR does not change either unit file. Verified against freedesktop systemd.exec, not a URL from the diff.

**Invariant:** Both the live writer and the paper reader must be able to create and write `replay_log_path` under the systemd units this repo ships.

**Must survive:** (1) `go-trader@hl-live` and `go-trader@hl-paper` both set `replay_log_path` to `/var/lib/go-trader/shared/replay.db`; (2) `ProtectSystem=strict` still blocks the checkout and other `/var` paths; (3) `PrivateTmp=true` so `/tmp` is not a shared fallback.

**Plain simple English:** Live and paper run as two named services. Those services cannot write one shared folder with the unit files we ship. Turning this on with the path in the notes makes both services fail to start.

### Recommended Optional

1. **Agent-facing docs say each log row stores live EntryATR and regime stamps; the table and struct do not.**
`docs/ARCHITECTURE.md` and `CLAUDE.md` describe `ReplayDecision` rows as carrying EntryATR/regime stamps. `scheduler/replay_log.go` `ReplayDecision` and `replayLogSchemaDDL` have no such columns. Paper stamps from its own check payload (`scheduler/replay_mirror.go:193-195`). An agent following those files will think live stop geometry is already in the log. Either drop the claim, or persist live EntryATR/regime if paper's trailing stop should match live's distance.

**Invariant:** Every field the architecture notes say a decision row carries must exist on `ReplayDecision` and in `decisions`, or the notes must not claim it.

**Must survive:** (1) a later change that reads EntryATR from the row; (2) paper ATR missing at replay-open time; (3) live and paper ATR disagreeing on the same bar.

**Plain simple English:** The notes say each log row stores live's stop-distance number. The table has no column for it. Paper uses its own number. Fix the notes, or store the live number if paper stops should match live.

2. **Pending-row query will scan every historical row for a strategy.**
`PendingDecisions` filters `strategy_id` + `replay_status='pending'` and orders by `decision_id` (`scheduler/replay_log.go`). The only index is `(strategy_id, decided_at)`. Applied rows are never pruned, and this query runs every paper cycle. Add a covering index `(strategy_id, replay_status, decision_id)` (and/or prune applied rows) in this PR.

**Invariant:** The per-cycle pending read must not grow with the count of already-applied rows for that strategy.

**Must survive:** (1) years of applied rows for one `strategy_id`; (2) a paper instance catching up after a long outage; (3) live still inserting while paper queries pending.

**Plain simple English:** Every few minutes paper asks "which live trades are still waiting?" The database still walks every old finished row for that strategy to answer. Add an index so that lookup stays cheap as the log grows.

richkuo:
Addressed review feedback (richkuo · 2026-08-18T19:38:31Z) in 1c80bdcb.

### Fixed
1. **Paper apply not atomic with persisted state** — confirmed at the old `scheduler/main.go` replay block (mark ran right after in-memory apply; the book only persisted at cycle-end save). Reworked to a two-layer cursor: apply under `mu` → `SaveStateWithDB` persists book mutations AND the new `strategies.replay_mirror_watermark` in one transaction → only then `MarkDecisionsApplied` (`scheduler/main.go` replay block; `scheduler/replay_mirror.go:62-90`; column + save/load in `scheduler/db.go`, field in `scheduler/state.go:174`). All three must-survive cases now hold: kill after mark → watermark re-marks without re-applying; save failure → rows stay pending; open+scale-in batch across a restart → `TestMirrorReplayPersistedWatermarkSurvivesRestart`.
2. **Template instances cannot write the shared log path** — confirmed (`systemd/go-trader@.service:46` `ProtectSystem=strict` + only per-instance `StateDirectory`). Added `go-trader/shared` to the template unit's `StateDirectory` so every instance gets `/var/lib/go-trader/shared` writable; the single-instance `go-trader.service` needs no change (its `StateDirectory=go-trader` already covers the whole tree). Locked by `TestSystemdTemplateGrantsSharedReplayDir`.

### Recommended Optional (implemented)
1. **Docs claimed EntryATR/regime on rows; schema didn't have them** — implemented the persist option (better than dropping the claim: paper's trailing-stop geometry now matches live's distance). Rows carry live's open-time `entry_atr`/`regime` (idempotent ALTERs in `scheduler/replay_log.go:86-96`); the mirror seeds them on replayed opens and falls back to its own payload stamps on legacy 0/"" rows (`scheduler/replay_mirror.go` `replayBookOpen`). Live/paper ATR disagreement resolves to live's — the faithful mirror.
2. **Pending-row query scans applied history** — added covering index `idx_decisions_strategy_status_id (strategy_id, replay_status, decision_id)` to the DDL (`scheduler/replay_log.go:84-90`); presence locked by `TestDecisionLogPendingIndexExists`.

Verification: `go build` + `go vet` clean, `gofmt` clean, full Go suite green (`go -C scheduler test ./... -count=1`), 10 new regression tests pass. Note on red→green: the new tests reference the new fields/columns, so they could not compile against the pre-fix tree; the restart-simulation test exercises the exact double-apply/drop scenarios from the finding.

---
Created with LLM: Kimi | high | Harness: Kimi Code

richkuo (commented):
LGTM

### Recommended Optional

1. **A kill during the new mid-cycle SaveState can duplicate paper trades even though the position re-applies correctly.**
`applyReplayedLiveDecisions` still goes through `recordPositionOpen` / `bookPerpsClose` / `bookPerpsPartialCloseWithFillFee`, and those call `RecordTrade` which eager-inserts via `tradeRecorder` (`scheduler/state.go:42-65`, `persisted=true`). `SaveStateWithDB` then skips those rows (`scheduler/db.go:1411-1416`) and only commits positions + `replay_mirror_watermark`. A SIGKILL while that save transaction is open rolls the watermark back (`scheduler/db.go:1568-1579`) and leaves the log row `pending`, so the next start re-applies and `InsertTrade`s a second copy. Verified against the apply path in `scheduler/replay_mirror.go` and the two persist paths, not the PR body. Positions restore; trade history and `#T` do not.

**Plain simple English:** Paper copies a live trade into the trade list right away, then saves the position in a second step. If the process dies in that second step, the position is copied again and the trade list shows the same trade twice.

**Invariant:** A replayed decision must persist paper's position, watermark, and trade row at most once, and all three must commit together or not at all.

**Must survive:** (1) open then scale-in in one pending batch, kill during the mid-cycle `SaveStateWithDB`, restart must not insert a second add trade; (2) eager `InsertTrade` succeeds and `SaveState` rolls back; (3) `InsertTrade` itself fails (`persisted` stays false) so the save's unpersisted-trade flush is the only persist.

---

richkuo:
Addressed review feedback (richkuo · 2026-08-18T20:10:37Z) in 589fce24.

### Fixed
1. **Kill during mid-cycle SaveState can duplicate paper trades (Recommended Optional)** — `applyReplayedLiveDecisions` now suspends the #289 eager `InsertTrade` hook (`scheduler/replay_mirror.go:77`) via `suspendEagerTradePersist` (`scheduler/state.go:34`). `RecordTrade` still appends in-memory with `persisted=false`; the caller's `SaveStateWithDB` inserts those rows in the same transaction as positions + `replay_mirror_watermark`. A kill during that save rolls all three back; restart re-applies once. Tests cover no eager insert during apply, kill-before-save open+scale-in does not duplicate, and SaveState is the only persist. Remaining narrower leftover: `captureTradeDiagnostics` still eager-inserts on a replayed full_close (#1147; not flushed by SaveState; outside the must-survive cases).

---
Created with LLM: Grok 4.6 | high | Harness: Cursor

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-18T20:51:07Z) in dc2ab9db.

### Fixed
1. **Full-fleet SaveState while holding the lock** — replay persist is now `SaveStrategyBookWithDB` (`scheduler/main.go` replay block; `scheduler/db.go` `SaveStrategyBook`). It writes only the mirrored strategy's row (watermark), positions, unpersisted trades, closed-position buffer, and deferred diagnostics. Unrelated strategies are not DELETE/reinserted. First-cycle saves seed `app_state` with `ON CONFLICT DO NOTHING` so a restart can load the book. Locked by `TestSaveStrategyBookDoesNotRewriteUnrelatedStrategies` and `TestMirrorSaveBeforeMarkWiring`.
2. **Replayed full-close can orphan a `trade_diagnostics` row** — apply now `suspendEagerDiagnosticsPersist` (`scheduler/replay_mirror.go:88`). `captureTradeDiagnostics` buffers on `pendingTradeDiagnostics`; `SaveStrategyBook` / `SaveState` insert those rows in the same transaction as the close. A kill during the save rolls the diagnostics row back with the watermark; restart re-applies once. Locked by `TestMirrorReplayFullCloseDefersDiagnosticsUntilSave` (two aborted attempts, shared `position_id`, one committed row).

### Deferred to follow-up
1. **File the disclosed watchdog/alert for live-vs-paper replay drift** — out of scope (alert thresholds/throttling/DM design). Filed https://github.com/richkuo/go-trader/issues/1436.

---
Created with LLM: Grok 4.6 | high | Harness: Claude Code

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-18T21:16:04Z) in e7c1cebb.

### Fixed
1. **Replayed decision can silently drop an earlier same-cycle trade's notification text** — confirmed at `scheduler/main.go:2910-2911` (was `detail = strings.Join(replayDetails, "; ")` after paper trailing/fixed-ATR SL had already set `detail`). Replay now `mergeTradeDetails(detail, replayDetails...)` (`scheduler/replay_mirror.go`; call site in the HL perps replay block) so a same-cycle SL close stays in the digest next to the replayed action. Empty replay parts leave native `detail` unchanged. Locked by `TestMergeTradeDetails*` and `TestMirrorSuppressionWiring`.

---
Created with LLM: Grok 4.6 | high | Harness: Claude Code

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-18T23:11:04Z) in ca6fb583.

### Corrected scope (partial)
1. **SaveStrategyBook inserts into positions without deleting/replacing the existing row** — `INSERT OR REPLACE` on `strategies` already CASCADE-deleted children in this SQLite driver (the two-cycle case passed before the patch), so a UNIQUE failure on `(strategy_id, symbol)` was not the live bug. Relying on that implicit CASCADE is still the wrong persist contract: a full close inserts nothing, and a stale child row would resurrect on `LoadState`. `SaveStrategyBook` now `DELETE FROM positions` / `option_positions WHERE strategy_id = ?` before rewrite (`scheduler/db.go` after the strategy upsert). `TestSaveStrategyBookReplacesOpenPositionAcrossCycles` covers open→scale-in, open→partial-close, and open→close→reopen (zero `positions` rows after the close).

---
Created with LLM: Grok 4.6 | high | Harness: Claude Code

richkuo:
@claude review

PR #1439, opened 2026-08-19
feat(#1436): notify owner on paper replay book drift [C37, Grok 4.6, high]

### LINKED ISSUE #1436: [C37] Paper does not notify the owner when live-vs-paper replay drifts
**Complexity: 37/100** — Capability 1 (Risk 2 — operator-facing drift alert, not a trading path); Volume 12 — Opus 5, high · fableplan: no

## Problem

#1431 paper replay consumes live/paper book-drift rows with a per-row `logger.Warn` and never retries them. The three WARN skip sites in `applyReplayedLiveDecisions` (`scheduler/replay_mirror.go`) are:

- open-while-holding — `ReplayDecisionOpen` when `pos != nil && pos.Quantity > 0`
- scale-in-while-mismatched — `ReplayDecisionScaleIn` when `pos == nil || pos.Quantity <= 0 || pos.Side != row.Side` (paper flat or wrong side, not side-only)
- partial-close-while-flat — `ReplayDecisionPartialClose` when `pos == nil || pos.Quantity <= 0`

Each skip calls `markApplied` so the pending queue does not wedge. `StrategyLogger.Warn` writes stdout/file only. A skip books no trade, so it never enters the cycle Discord digest.

Consecutive decision-log *insert* failures already DM the owner at 3 via `decisionLogPersistWarn` (`decisionLogAlertThreshold`, once per streak). Trade persist failures also DM via `tradePersistWarn`, but on every failure with no 3-strike. Replay drift does not DM. An owner watching Discord will not learn that paper's book no longer matches live unless they grep logs.

PR #1435's body named this as an unfiled follow-up. The #1435 review (2026-08-18 20:51) asked that it be tracked as its own issue rather than folded into the persist/atomicity work.

Close-while-flat (`ReplayDecisionFullClose` when paper is already flat) is `logger.Info` only — paper's own stop beating the mirror is the trailing-stop carve-out, not book drift.

## Goal

When paper skips a live decision because the books already disagree, the owner gets a throttled Discord DM. The skip still consumes the row. The same skip does not DM again inside the alert wait.

## Approach

1. Record a drift at the three WARN skip sites only (open-while-holding, scale-in-while-mismatched, partial-close-while-flat). Leave "booked no trade" and unknown-type WARNs as logs; those are apply failures, not book disagreement. Close-while-flat stays INFO and does not Record. The high-water / watermark re-mark path (`row.DecisionID <= lastApplied`) does not Record — that row was already skipped or applied this process.

2. In-memory tracker (process heap, not SQLite). Key `(strategy_id, kind)` with kind fragments exactly `open-while-holding`, `scale-in-while-mismatched`, `partial-close-while-flat`. Store `lastNotifiedAt`. Fire when the slot is empty or `now.Sub(lastNotifiedAt) >= effectiveAlertThrottleInterval()` (config `alert_throttle_interval`, empty → 6h). Do **not** reuse `shouldNotifyDrainFailure` — that helper also re-alerts on every 10th event and would violate the once-per-window rule. Restart clears the map; the first post-restart drift always DMs (same lifetime as `LiveExecFailureThrottle`).

3. `applyReplayedLiveDecisions` keeps `logger.Warn` at the skip sites, still `markApplied`, and returns the already-throttled DM texts as an extra result. Stamp `lastNotifiedAt` when Record decides to notify (under `mu`), before the send.

4. The HL perps replay block holds `mu` through apply and `SaveStrategyBookWithDB`, then `mu.Unlock()` (`scheduler/main.go`). Send the DMs **after** that unlock via a nil-safe `replayDriftWarn` hook wired next to `decisionLogPersistWarn` when `notifier.HasOwner()` (prefix `[replay]`). Do not call `SendOwnerDM` inside `applyReplayedLiveDecisions`. Tests/subcommands with a nil hook still apply and mark.

5. Do not retry the skipped row and do not change `MarkDecisionsApplied`, the watermark, or eager-persist suspension.

6. Tests: first drift of each kind fires; a second identical `(strategy_id, kind)` inside the throttle window does not (including two same-kind skips in one pending batch); a different kind or strategy still fires; high-water re-mark does not fire; close-while-flat / booked-no-trade / unknown-type do not fire; nil hook does not panic and the row is still marked; structural check that the replay block sends after `mu.Unlock()`. Update the ARCHITECTURE.md drift sentence (today it lists close-while-flat as a WARN; code is INFO, and this issue adds the owner DM on the three WARN skips).

## Acceptance criteria

1. Open-while-holding, scale-in-while-mismatched, and partial-close-while-flat each produce an owner DM on the first occurrence in a throttle window.
2. A repeated same-kind drift on the same strategy inside `alert_throttle_interval` does not send another DM.
3. Drift rows stay consumed (not retried); the pending queue does not wedge.
4. Tests/subcommands with a nil warn hook still apply and mark the row.
5. Close-while-flat remains INFO-only and does not DM.
6. High-water / watermark re-marks do not DM.
7. Owner DMs leave the `mu.Lock` region before `SendOwnerDM` runs.
8. "Booked no trade" and unknown-type WARNs stay log-only.

## Plain simple English

If paper's book no longer matches live, today the only sign is a log line. Nobody is messaged. An owner watching Discord will not know the pair is out of sync.

---
LLM: Grok 4.6 | high | Harness: Cursor

richkuo:
## Plain simple English

If paper's copy of a live trade no longer matches, the owner now gets a Discord message. The same mismatch does not message again for hours. The skipped row is still marked done so later trades can run.

## Summary

- Record book drift only at the three WARN skip sites (`open-while-holding`, `scale-in-while-mismatched`, `partial-close-while-flat`). Close-while-flat stays INFO. Booked-no-trade, unknown-type, and high-water re-marks do not DM.
- In-memory throttle keyed `(strategy_id, kind)` fires on first sighting or after `alert_throttle_interval` (empty → 6h). Does not reuse `shouldNotifyDrainFailure` (that helper also re-alerts every 10th event). Restart clears the map.
- `applyReplayedLiveDecisions` still WARNs, still `markApplied`, and returns already-throttled DM texts. The HL perps replay block sends them through nil-safe `replayDriftWarn` (prefix `[replay]`) after `mu.Unlock()`.
- Drift rows stay consumed. `MarkDecisionsApplied`, the watermark, and eager-persist suspension are unchanged.

Closes #1436.

## Verification

- New tracker and apply-path tests cover first-of-each-kind, same-kind batch throttle, different kind/strategy, high-water re-mark, close-while-flat / booked-no-trade / unknown-type exclusions, nil hook still marks, apply does not invoke the hook, and send-after-unlock wiring.
- `go -C scheduler test -count=1 .` passed (37.0s).
- `go -C scheduler build` passed; `gofmt -l` clean.

---
Created with LLM: Grok 4.6 | high | Harness: Cursor


richkuo:
LGTM

---
Validated with LLM: Composer 2.5 | high | Harness: Cursor

PR #1445, opened 2026-08-22
fix(#1444): give manual HL strategies a live mark each cycle [C35, Opus 5, high]

### LINKED ISSUE #1444: [C35] Manual strategies on Hyperliquid get no live price each cycle, so the trailing stop-loss walker stops
**Complexity: 35/100** — Capability 1 (Risk 2 — auto-protective trailing stop-loss path, contained and reversible); Volume 10 — Opus 5, high · fableplan: no

## Problem

A strategy of `type=manual` whose position lives on Hyperliquid perps never receives a live mark in the scheduler's per-cycle `prices` map. The on-chain trailing stop-loss walker and the take-profit ratchet then no-op on every cycle the position is open.

### Root cause

`collectPerpsMarkSymbols` (`scheduler/risk.go:70-72`) skips every strategy whose `Type != "perps"`, so a manual strategy contributes no coin to the Hyperliquid mark rail. The companion collectors filter the same way — `collectPriceSymbols` on `Type != "spot"` (`risk.go:39`) and `collectFuturesMarkSymbols` on `Type != "futures"` plus `Platform != "topstep"` (`risk.go:156-160`). Manual strategies therefore contribute to no mark rail at all.

The per-cycle init calls `collectPerpsMarkSymbols` (`main.go:961`) and fetches mids only for the returned coins (`main.go:985-996`), so a manual-only coin never enters `prices`.

### What breaks

At the manual live-management block (`main.go:3113`):

- `mark := prices[sc.Symbol]` (`main.go:3122`) reads `0.0` for a missing key.
- The ratchet gate `mark > 0` (`main.go:3126`) fails, so `applyTrailingTPRatchet` never runs.
- The walker gate `mark > 0` (`main.go:3134`) fails, so `runHyperliquidTrailingStopUpdate` never runs.
- `Position.RatchetFallbackNormalizePending` therefore never clears on a live manual position: the only live-path clears are inside the mark-gated walker (`hyperliquid_trailing_stop.go:685,692`). The other two clear sites (`main.go:2612`, `trailing_tp_ratchet_sl.go:106`) are paper-only.

`runPostTPStopLossAdjustment` is a partial exception. It is invoked unconditionally at `main.go:3121`, and the `breakeven` and `atr_offset` `sl_after` modes compute from `avgCost`/`EntryATR` with no mark (`post_tp_sl.go:214-229`). Only `trail_from_here` requires `currentMark > 0` (`post_tp_sl.go:230-232`), and the validator blocks `trail_from_here` for `sc.Type=="manual"` (`post_tp_sl.go:1229`). The fully dead paths are the trailing walker and the ratchet.

### Why it is silent

- A missing key in a `map[string]float64` reads `0.0`. No error, no panic.
- The `[WARN] No HL perps mark for %s` loop (`main.go:991-995`) iterates `hlPerpsCoins` only. The manual coin is absent from that slice, so the warning cannot fire for it.
- `runHyperliquidProtectionSync` needs no live mark: `buildHyperliquidProtectionPlan` reads `pos.riskAnchorPrice()` and `pos.EntryATR` (`hyperliquid_protection.go:77-78`, gate at `:38`). The stop-loss seeded at open is correct, so only the cyclical updates stop. This makes the fault look like a ratchet problem instead of a missing-mark problem.
- The Python close evaluator keeps logging a real price. The manual dispatch calls `runManualCloseEval` (`main.go:3062`), which calls `runHyperliquidCheck(&sc, nil, ...)` with a nil prices map (`manual.go:1057`) and discards the returned price. `check_hyperliquid.py` self-fetches the mark, so `Signal: HOLD | <COIN> @ $<real>` (`main.go:4125`) prints every cycle and masks the Go-side state.

### Reproduction conditions

1. A `type=manual` strategy with `Platform=="hyperliquid"` holding an open live position (`hyperliquidIsLive(sc.Args)` true — a record-only manual config never runs the walker by design, `main.go:3113`).
2. No other strategy donates that coin to the Hyperliquid mark rail. Two donor classes exist, and either one masks the fault:
   - a `Type=="perps", Platform=="hyperliquid"` strategy on the same coin, and
   - a hedge leg on the same coin, which `collectPerpsMarkSymbols` adds through `hedgeCoinsForStrategies` (`risk.go:88-96`, #1159).

With a donor present, `prices[coin]` is populated as a side effect and the manual strategy appears to work. With no donor, the walker dies.

### Observable symptom

The strategy log shows only `Signal: HOLD` and status lines. These log strings never appear:

- `Updating trailing SL for %s: ...` (`hyperliquid_trailing_stop.go:757`)
- `Trailing SL trigger updated oid=...` (`hyperliquid_trailing_stop.go:694`)
- `post-TP SL adjustment for %s: tier %d cleared, ...` (`post_tp_sl.go:1231`)

## Goal

Give every coin that only manual strategies trade a live mark on each cycle, so the trailing stop-loss walker and the ratchet operate again. Add a loud warning when an open position has no live mark, so a future strategy type cannot regress this silently.

### 1. Include manual strategies in the perps mark collector

Relax the type filter in `collectPerpsMarkSymbols` (`scheduler/risk.go`) to accept `type=manual` alongside `type=perps`. `Platform` is set correctly for manual strategies, and the existing `switch sc.Platform` already routes the coin to the Hyperliquid or OKX set. The coin comes from `sc.Args[1]`, which `loadConfig` fills from `sc.Symbol` for manual strategies (`config.go:1466-1468`).

This is the single choke point. All three consumers inherit the fix with no further edits:

- the per-cycle init (`main.go:961`),
- `augmentMarksBestEffort` (`main.go:5217`, used by the `--once` path at `main.go:3626` and `main.go:5265`),
- `NewStatusServer` (`server.go:102`).

The failure policy is unchanged: the fetch stays best-effort, `mergePerpsMarks` skips non-positive marks and never overwrites a strategy-published mark (`risk.go:119-128`), and valuation still falls back to `pos.AvgCost`.

Do **not** relax `collectFuturesMarkSymbols`. `type=manual` is rejected at load for any platform other than hyperliquid (`config.go:2080-2082`), and that collector also filters `Platform != "topstep"` (`risk.go:158`), so the edit would be dead code.

**Named side effect:** manual coins will also join `PortfolioValue`, exposure and drawdown inputs, and the status page at live mids instead of frozen `AvgCost`. This is the wanted direction, and it matches the reason the #1159 hedge-coin block was added (`risk.go:88-96`).

### 2. Add a missing-mark regression guard

Add a warning for any strategy that holds an open position on a coin with no usable mark. Two lifecycle facts shape the placement:

- Perps strategies publish their own mark inside the strategy loop (`prices[result.Symbol] = price`, `main.go:2482`), after an init-time guard would run. A guard at init would warn spuriously for them.
- A record-only (non-live) manual position never runs the walker by design (`main.go:3113`), so it must not warn.

Place the guard after the per-cycle strategy loop, when `prices` is complete. Extract the decision as a pure helper — it takes the strategy list, the per-strategy open positions, and the `prices` map, and returns the strategies and symbols to warn about — so it is testable without spawning Python, per the CLAUDE.md testing rule. Read positions through `state.Strategies[sc.ID].Positions[sym]` under `mu` (the pattern at `main.go:1876` and `main.go:3131`); `sc.SymbolFromArgs()` does not exist.

### Rejected alternatives

- **Weaken the `mark > 0` gate at `main.go:3134`.** The same guard exists one layer deeper at `hyperliquid_trailing_stop.go:716` and at `trailing_tp_ratchet.go:576,598`, so the inner functions still return early. The data is wrong, so the fix belongs in the data.
- **Fall back to `pos.AvgCost` when the mark is zero.** The walker would compute trigger prices from the entry price and widen the stop-loss on cycle 1. An entry price is the wrong semantic for a trailing stop.

## Acceptance criteria

1. `collectPerpsMarkSymbols` includes the coin of a `type=manual, platform=hyperliquid` strategy in `hlCoins`, and a `type=manual` strategy on any other platform contributes nothing. `collectFuturesMarkSymbols` is unchanged.
2. Existing hedge-coin inclusion (#1159) and the perps and spot membership rules are unchanged; unit tests in `scheduler/risk_test.go` cover inclusion and exclusion.
3. The missing-mark guard is a pure helper with unit tests. It warns for a live manual strategy holding an open position with no mark, and stays silent for a record-only manual position, for a flat strategy, and for a perps strategy that published its own mark during the cycle.
4. On a live manual strategy on a coin with no donor strategy, `Updating trailing SL for <coin>` appears within a few cycles, `Trailing SL trigger updated oid=` follows on a real tighten, and `RatchetFallbackNormalizePending` clears.
5. A manual strategy on a coin that already has a donor keeps its current behavior; the mark is fetched once, and `mergePerpsMarks` still lets a strategy-published mark win.
6. `go -C scheduler build .` and `go -C scheduler test ./...` pass (the module lives in `scheduler/`).

## Plain simple English

Some positions on Hyperliquid are traded by hand. The system did not get a fresh price for them each cycle. Without a price, the automatic stop-loss could not follow the market, and the profit ratchet could not move. The stop stayed where it started, and nothing warned the operator.

---

---

richkuo:
## Plain simple English

Some Hyperliquid positions are traded by hand. The system did not get a fresh price for them on each cycle. Without a price, the automatic stop-loss could not follow the market and the profit ratchet could not move, so the stop stayed where it started. Nothing warned the operator. Both now work.

## Summary

`collectPerpsMarkSymbols` filtered on `Type != "perps"`, and its two siblings filter on `spot` and `futures`+topstep, so a `type=manual` strategy contributed to **no** mark rail. With no perps or hedge donor on the same coin, `prices[sc.Symbol]` read `0.0` every cycle and the `mark > 0` gates in the manual dispatch killed both the Hyperliquid trailing stop-loss walker and the take-profit ratchet. It was silent because a map miss is not an error, and the existing `[WARN] No HL perps mark for %s` loop only iterates the coins that were collected.

Changes:

1. `collectPerpsMarkSymbols` now emits `type=manual` coins alongside `type=perps`. `collectFuturesMarkSymbols` is unchanged — relaxing it would be dead code.
2. New pure helper `collectMissingMarkPositions(strategies, openSymbols, prices)` runs once per cycle **after** the strategy loop and `[WARN]`s for every open position with no positive mark.

All three `collectPerpsMarkSymbols` consumers inherit the fix with no further edits: the per-cycle init, `augmentMarksBestEffort` (the `--once` and leaderboard paths), and `NewStatusServer`. The failure policy is unchanged — the fetch stays best-effort, `mergePerpsMarks` still skips non-positive marks and still lets a strategy-published mark win, and valuation still falls back to `pos.AvgCost`.

### Deviation from the issue's Approach section

The issue proposed reading the manual coin from `sc.Args[1]`, on the grounds that `loadConfig` fills `Args` from `sc.Symbol`. That derivation only runs when `len(sc.Args) == 0` (`config.go`), so a hand-written `args` list can name a different coin. Every manual consumer — the dispatch, the position map, the walker — keys off `sc.Symbol`, so `Args[1]` would park the mark under a key nobody reads while the real key stayed empty. This branch reads `sc.Symbol`, which validation guarantees non-empty for `type=manual`. A regression test asserts this by making `Args[1]` deliberately disagree with `Symbol`.

The same paragraph also required that a manual strategy on a non-hyperliquid platform contribute nothing (acceptance criterion 1). Falling through to the shared `switch sc.Platform` would have routed a `manual`+`okx` entry into the OKX rail, so the manual branch checks the platform itself and skips.

### Guard scope

Two exclusions, both documented at the helper:

- **Options strategies.** Their value comes from `OptionPositions` (`CurrentValueUSD`, marked in Phase 5); any spot leg parked in `Positions` has no collector feeding it, so including them would emit a standing warning about a pre-existing valuation path.
- **Record-only (non-live) manual strategies.** The manual ratchet and walker are gated on `hyperliquidIsLive` by design, so a missing mark breaks no management path there.

Placement is after the strategy loop rather than at init because perps strategies publish their own mark **inside** the loop; an init-time guard would warn for them spuriously.

### Named side effect

Manual coins now join `PortfolioValue`, the exposure and drawdown inputs, and the status page at live mids instead of a frozen `AvgCost`. This is the wanted direction and matches the rationale for the #1159 hedge-coin block. It does mean a losing manual position now contributes real unrealized PnL to the portfolio drawdown inputs, where before it contributed exactly zero — the risk was always there, and it is now visible to the auto-protective path.

## Verification

- New collector test proves red on the unfixed code: `hlCoins = [BTC], want [BTC ETH HYPE]`. Green after the change.
- `gofmt -l scheduler/` — clean.
- `go -C scheduler build .` — passes.
- `go -C scheduler vet ./...` — clean.
- `go -C scheduler test ./...` — `ok trading-scheduler 36.980s`.

Tests added in `scheduler/risk_test.go`:

1. `TestCollectPerpsMarkSymbols_IncludesManualHyperliquid` — live and record-only manual both donate; dedup against a perps donor on the same coin; `Args[1]` disagreeing with `Symbol` proves the key choice; manual+okx contributes nothing to either slice; empty `Symbol` ignored.
2. `TestCollectFuturesMarkSymbols_IgnoresManual` — pins the decision not to relax the CME rail.
3. `TestCollectMissingMarkPositions` — warns for a live manual holding an open position with no mark and for a perps strategy that lost its mark; silent for a record-only manual, a flat strategy, a strategy whose mark was published during the cycle, a spot position with a price, and an options leg. A present-but-zero price still counts as a miss.
4. `TestCollectMissingMarkPositions_SortsSymbolsPerStrategy` — deterministic operator output when one strategy holds a primary and a hedge leg.
5. `TestCollectMissingMarkPositions_NoOpenPositions` — no allocation on the all-flat cycle.

`docs/ARCHITECTURE.md` records both mechanics in the `risk.go` row.

Closes #1444

---


richkuo:
@claude opus review

richkuo:
## Review disposition — all three findings addressed

Reviewed: `github-actions[bot]`, 2026-08-22T00:10:30Z. Commit `55996f5e`.

### ❓ Judgment resolved — Requires Human Review 1 (peak basis vs. `PeakValue`)

Confirmed against the code: `risk.go` puts manual coins in `prices`, `main.go` feeds `totalPV` to `checkPortfolioRiskWithEquityAvailability`, and that divides by a sticky ratchet-up-only peak accumulated under the old basis. One correction to the framing of the recommended solution: the "just reset it afterwards" option does not exist, because `planKillSwitchClose` force-closes live positions *before* the latch is written. A first-cycle fire is irreversible, so the migration has to happen before the check, not after the fire.

**Implemented, in preference to both the pre-deploy manual step and a blanket re-baseline:** a one-shot, *measured* basis migration, run in-cycle before `checkPortfolioRiskWithEquityAvailability`.

The shift is measured, never assumed. The cycle recomputes the same total against `pricesWithoutSymbols(prices, manualOnlyMarkSymbols(cfg.Strategies))` — deletion, not zeroing, because `PortfolioValue` falls back to `AvgCost` only on a *missing* key — and applies `manualMarkBasisPeakAdjustment`:

```
newPeak = oldPeak + (liveTotal - legacyTotal)
```

Why the delta and not a re-baseline to the current total: a blanket re-baseline would erase any drawdown that had already accumulated under the old basis and disarm a legitimately armed kill switch. The delta form corrects the units and leaves the real drawdown exactly where it was. `manualOnlyMarkSymbols` also excludes any manual coin that a perps, spot, hedge or futures rail already donated, since those coins were live-marked before this PR and their basis never moved. The shared-wallet dedup path is handled for free: when a real balance supplies the total, both recomputes agree and the delta is zero.

Guards: only while `PortfolioRisk.ManualMarkBasisRebaselined` is false, and only when every open position has a positive mark (an incomplete mark set would make the two totals agree for the wrong reason and under-migrate). The latch is persisted in `portfolio_risk.manual_mark_basis_rebaselined`, so a restart cannot move the peak twice, and it is stamped whenever the measurement ran, including when no adjustment was needed. The move emits a `basis_rebaseline` `kill_switch_events` row and one owner DM naming both totals and stating that the drawdown reading was not reset.

Rejected alternatives: a hand edit of `peak_value` before deploy — no audit trail, must be repeated per deployment, needs the service stopped; a startup-time hook next to `ClearLatchedKillSwitchSharedWallet` — no `prices` map exists yet at that point, so the shift cannot be measured there.

### ✅ Confirmed and fixed — Recommended Optional 1 (log-only tripwire)

Real. The sibling silent-failure guards escalate this class to a throttled owner DM. A live miss now raises one, via `missingMarkAlerts` in the new `missing_mark_alerts.go`.

- **Invariant:** a condition that disables an auto-protective mechanism on a live position reaches the operator through a channel they watch.
- Must survive (a): the throttle is keyed `(strategy_id, symbol)` and fires once per `alert_throttle_interval`, so a multi-cycle outage across several positions cannot emit one DM per position per cycle.
- Must survive (b): record-only positions carry `Live=false` and never enter the DM path.
- Must survive (c): `Retain()` prunes every slot whose position is no longer missing, so a returning mark stops the alert and a *later* outage notifies immediately, with no restart.

### ✅ Confirmed and fixed — Recommended Optional 2 (record-only exclusion)

Real, and traced: `computeSubsetPortfolioValue` virtual-sums every strategy not folded into a dedup set, and `sameAccountLiveManualMembers` folds in live manual only, so a record-only manual position reaches `totalPV` through `PortfolioValue`, where a missing key silently reverts it to `AvgCost` inside the drawdown input.

The exclusion is removed for every type except options. The `Live` flag now carries the live/record-only split instead, which satisfies this finding and must-survive (b) of the previous one at the same time.

- **Invariant:** every position whose valuation feeds a portfolio-level protective gate is covered by the tripwire, whatever its mode.
- Must survive (a): a record-only manual miss warns; an options leg stays silent.
- Must survive (b): a position stored under a symbol other than `sc.Symbol` still warns — the walk is over the position map, not the mark rail.
- Must survive (c): a flat record-only manual strategy allocates nothing and stays silent.

### Verification

`go build`, `go vet`, `gofmt` clean, and the full `go test ./...` suite passes. New tests cover the tripwire's `Live` split and all three must-survive cases per finding, `manualOnlyMarkSymbols` donor exclusion, `pricesWithoutSymbols` delete-vs-zero semantics, every branch of `manualMarkBasisPeakAdjustment` (including a case pinning that a pre-existing real drawdown survives the migration), the throttle and re-arm behaviour, and the persistence of the one-shot latch across a save/load round trip.

One existing test asserted that record-only manual stays silent. That assertion encoded the defect this review reported, so it was updated to the corrected behaviour rather than worked around; the replacement assertions are strictly stronger.

## Plain simple English

Hand-traded positions now count at the live price. The automatic shut-off compares the total against the highest total ever recorded, which was measured the old way, so it could have closed everything on the first check. The recorded high is now corrected once, by the exact amount the counting changed. Real losses still count.

---

richkuo:
@claude review

richkuo:
## Review disposition — both optional findings fixed

Addressed review feedback (`github-actions[bot]` · 2026-08-22T05:18:53Z) in `31af68ee`. Both were `Recommended Optional`, and both validated as real against current code.

### Fixed

1. **Peak-migration gate deferred on any missing mark, not just the manual ones** — traced and confirmed. `computeSubsetPortfolioValue` (`scheduler/shared_wallet.go:504`) is additive over `PortfolioValue` plus price-independent real wallet balances, and `legacyPrices` removes only `manualOnlyMarkSymbols`, so a missing spot / OKX-perps / futures mark is absent from **both** maps, falls back to `pos.AvgCost` in both, and cancels out of the live-minus-legacy delta exactly. The dedup structure depends on configs alone, so no branch flips either — the cancellation is exact, not approximate. The gate is now the new pure `missingManualOnlyMarks` (`scheduler/risk.go:414`), consumed at `scheduler/main.go:1388`.

   One refinement past the suggested fix: it walks the **open-position map** rather than the config list. Requiring a positive mark for every *configured* manual-only coin would stall the migration indefinitely on a coin the book does not hold, which trades one over-deferral for another. Holding is what makes a coin able to move the delta.

   - **Invariant:** the migration runs as soon as the shift it measures is measurable — every *held* manual-only coin marked — regardless of unrelated positions' mark status.
   - Must survive (a): manual mark present, futures/OKX/spot marks missing → runs, delta correct (`TestMissingManualOnlyMarks_IgnoresNonManualOutage`).
   - Must survive (b): a manual coin's own mark missing → still defers, including the present-but-zero shape (`TestMissingManualOnlyMarks_DefersOnManualOutage`).
   - Must survive (c): no manual-only coins → runs immediately, delta zero, latch stamped (`TestMissingManualOnlyMarks_NoManualOnlyCoinsRunsImmediately`).
   - Two edges beyond the reported case: a manual coin a perps rail already donates never gates (its basis never moved), and a flat manual strategy never gates.

2. **Live missing-mark DM hard-coded Hyperliquid mechanisms for every venue** — traced and confirmed. `collectMissingMarkPositions` marks any live spot/perps/futures/manual miss `Live=true`, and the OKX perps branch runs `runOKXCheck`/`runOKXExecuteOrder` and neither the walker nor the ratchet; spot and TopStep futures run neither either. `effectiveTrailingStopPct` (`scheduler/hyperliquid_trailing_stop.go:96`) returns 0 for any other platform or type, which is the same boundary. The bullets now come from the pure `markGatedManagers` (`scheduler/risk.go:220`), rendered by `formatMissingMarkDM` (`scheduler/missing_mark_alerts.go:114`); the stdout `[WARN]` carries the same scoping.

   Every live miss still raises a DM, on every venue. Only the text narrowed: the universal consequence — `PortfolioValue` falling back to `AvgCost`, so the portfolio kill switch reads a stale valuation — holds everywhere, so a venue with no mark-gated manager still carries a true claim rather than losing its alert.

   Keyed on **type and venue** rather than on per-strategy close knobs, deliberately. The knob-level gates are position-dependent: `effectiveTrailingStopPct` reads `pos.EntryATR`, and a post-TP `sl_after: trail_from_here` stamp arms the walker on a strategy that configures no trailing field at all. A config-only test would therefore under-report a genuinely disabled protection, which is the worse error of the two.

   - **Invariant:** an alert that names specific auto-protective mechanisms only names mechanisms that apply to that position's type and venue.
   - Must survive (a): live HL perps and live HL manual misses still name the walker and the ratchet (two tests).
   - Must survive (b): live BinanceUS spot miss claims neither.
   - Must survive (c): live OKX-perps miss claims neither, and the same subtest covers TopStep futures. Record-only misses stay log-only for all types, unchanged.

### Verification

`go build`, `go vet`, `gofmt` clean, and the full `go test ./...` suite green on an uncached run. New tests cover all three must-survive cases per finding, the donor-coin and flat-strategy edges on the gate, and the full six-row venue matrix for `markGatedManagers`. One existing test compared `missingMarkPosition` by struct equality, which the new slice field makes non-comparable; it now compares the three fields it was actually pinning, with no assertion weakened.

## Plain simple English

Two fixes to the automatic shut-off path. The one-time correction to the recorded high-water mark used to wait for every price in the book. It only needs the hand-traded ones, and waiting left the very gap it was meant to close. Separately, the missing-price alert claimed a stop-loss had stopped even on venues that never run one. It now names only real safeguards.

---

richkuo:
@claude sonnet review

Issue #1441, opened 2026-08-20
[C74] Keep Python workers warm for read-only strategy and regime checks


richkuo:
**Complexity: 74/100** — Capability 2 (Risk 3 — read-only results drive live trading decisions; Coupling 4 — Go/Python protocol, worker pool, regime dispatch, startup, shutdown, and health); Volume 24 — Opus 5, high · fableplan: yes

## Problem

Each call that reaches the shared Python runner starts `.venv/bin/python3`, waits for one result, and exits the process. Repeated strategy signal checks and regime calculations therefore pay interpreter and import startup costs on every call.

The regime store starts one calculation for each distinct due-strategy signature. It runs up to four calculations concurrently. Strategy signal checks then run through the same four-slot subprocess limit. The current repository proves this process topology. It does not contain reproducible evidence for the fleet call counts, startup cost, processor use, or improvement targets that were previously stated in this issue.

The shared runner also serves live orders, position closes, balance and position fetches, dashboard work, simulations, backtests, backfills, command-line operations, and user-configured Python script paths. These paths have different timeout, shutdown, isolation, and failure requirements. A single shared worker queue would combine these failure domains and would serialize the current four-way regime calculation.

The current venue scripts also change `sys.path` and import generic module names such as `adapter` and `strategies`. One interpreter can therefore reuse the wrong cached module when it serves different venues or both OKX spot and swap checks.

## Goal

Keep Python interpreters warm for an explicit set of repeated read-only strategy signal and regime checks. Preserve decision semantics, four-call regime concurrency, per-call deadlines, shutdown behavior, current configuration reload behavior, and the existing fork-and-exec path.

Measure the result with a reproducible benchmark before setting an operational performance target.

### 1. Limit the worker route to repeated read-only checks

Add a process-owned pool of four persistent Python workers. Route only these methods through the pool:

- `check_strategy` signal evaluation.
- `check_hyperliquid` signal evaluation.
- `check_okx` signal evaluation.
- `check_topstep` signal evaluation.
- `check_robinhood` signal evaluation.
- `check_options` signal evaluation, including its stdin position context.
- `check_regime` bundle calculation.

Keep live execution modes, protection changes, closes, balance and position fetches, price and candle fetches, simulations, backtests, backfills, large language model analysis, tuning, command-line operations, and user-configured scripts outside this allowlist on their current execution lanes.

[...]

PR #1446, opened 2026-08-22
perf(#1442): batch hyperliquid strategies sharing market data [C74, Opus 5, high, fableplan]

### LINKED ISSUE #1442: [C74] Batch Hyperliquid strategies that read the same market data into one Python call
**Complexity: 74/100** — Capability 2 (Risk 3 — batched results drive live strategy dispatch); Volume 24 (Scope 4 — cycle loop, HL check lane, executor, backtester parity; Coupling 4 — regime store, replay choke points, 6-phase locking; Verification 4 — decision parity vs unbatched) — Opus 5, high · fableplan: yes

## Problem

A deployment runs many strategies against a single coin. The per-strategy dispatch loop (`scheduler/main.go:1875`) is sequential and spawns one Python check subprocess per due strategy through `runPython` (`scheduler/executor.go:151`), with concurrency capped at 4 by `pythonSemaphore` (`scheduler/executor.go:28`).

**Two dedupe mechanisms already exist, and this issue must not claim credit for them.**

1. **Regime is already computed once and injected.** The regime store computes one bundle per distinct `(platform, symbol, timeframe, windows-spec)` signature per cycle (`scheduler/regime_store.go:55–67, 513–530`) and injects it via `--regime-payload-json` (`scheduler/strategy_composition.go:154–160`). With an injected payload the check script never recomputes the regime, even when the payload is empty (`shared_tools/regime.py:700–704`). There are no redundant regime computations to remove.
2. **Candle fetches are already deduplicated per instance.** The `#839` on-disk OHLCV cache keyed `(symbol, interval, limit)` with a 60s TTL means the first strategy of a cycle fetches and the rest read from cache (`platforms/hyperliquid/adapter.py:36–49, 596–607`). Its own comment records the motivating shape: "~20 strategies per instance sharing a handful of asset+timeframe combos".

What remains duplicated per strategy is therefore narrower than a naive reading suggests: **interpreter start, module imports, DataFrame construction from the cached candles, and indicator computation**. Batching removes the DataFrame and indicator share; issue #1441 removes the interpreter and import share. The two issues attack adjacent halves of the same residual cost and must be sequenced together.

Per-strategy argv is also not uniform. Beyond `sc.Args`, the HL builder appends position context, an on-chain-protection-filtered close-strategy ref set, the HTF filter flag, per-strategy regime windows, the resolved ATR method, and a cycle-local mark price (`scheduler/main.go:4019–4046`). A batch key must be derived from the inputs that genuinely determine the shared market state, and every other argument must stay per-slot.

**Unmeasured quantities.** Deployment config lives out of tree (`/var/lib/go-trader/<instance>/config.json`, `#1056`), so strategy counts and interval distributions cannot be verified from this repository. The in-repo open registry declares 50 strategies, 32 of them `deprecated_m5`. Any count used to justify this work must come from a captured snapshot of the running fleet, recorded in the pull request.

## Goal

For the Hyperliquid perps check lane, evaluate all due strategies that share the same market-data inputs in one Python call that builds the shared candle frame and indicator base once and returns one decision per strategy.

Preserve exactly: the per-strategy decision values, the 6-phase locking cycle, per-strategy risk checks and execution, the `#1431` replay recording and paper-side suppression choke points (`scheduler/main.go:2548, 2888`), per-strategy error isolation, and the regime store's existing behavior.

Measure the result against the true baseline above before claiming any speedup.

### 1. Target the Hyperliquid perps lane only

The motivating deployment is HL perps. It dispatches `runHyperliquidCheck` (`scheduler/main.go:2481`) → `RunHyperliquidCheck` (`scheduler/executor.go:323`) → `shared_scripts/check_hyperliquid.py`, which loads `shared_strategies/open/futures`. The spot lane (`check_strategy.py` / `RunSpotCheck`), and the OKX, TopStep, Robinhood, and options lanes, are explicitly out of scope for this issue.

### 2. Group by shared market-data inputs

A pure helper in the scheduler partitions due strategies by a key that determines the shared frame:

- data-source platform (the lane's actual OHLCV source, mirroring why `regimeBundleKey` carries `Platform` — the default spot dispatch fetches BinanceUS data regardless of `sc.Platform`, `scheduler/regime_store.go:55–59`)
- symbol
- timeframe
- OHLCV limit
- resolved ATR method

Everything else — position context, close-strategy refs, HTF filter, per-strategy regime windows, mark price, strategy params — travels per slot inside the batch, never in the key.

The partition runs over `dueStrategies` only. A strategy whose interval has not elapsed is never added to a batch. The grouping is a pure function over the already-built due list; note that the due loop itself runs lock-free (`scheduler/main.go:913–926`) and only the interval computation takes the read lock (`:906–910`), so grouping introduces no new lock interaction.

### 3. Batch entry point on the HL lane

Extend `check_hyperliquid.py` with a batched read-only signal mode that accepts N strategy slots sharing one key, builds candles and the indicator base once, then evaluates each slot with its own arguments and position context. Each slot is wrapped in its own error handling so one failing strategy returns an error for that slot while the rest of the batch returns normally.

A shared-state failure (candle fetch failure, adapter initialization failure) must return a distinct "shared state failed" sentinel for the whole batch, so an operator can tell it apart from a per-strategy fault. Today's per-strategy isolation is the behavior being traded away here, and the sentinel is what keeps the loss legible.

**Dispatch ownership is Go, not Python.** The scheduler decides whether a group has one member or many, and calls the single-strategy path or the batched path accordingly. The Python script never re-dispatches itself.

### 4. Composition with #1441

Issue #1441 adds a persistent worker pool and a request protocol. This issue and that one must agree, before either builds, on whether the worker protocol method is single-strategy or multi-strategy. Land the sequencing decision as an explicit contract in whichever issue builds first, and gate this issue's speedup claims on #1441's committed benchmark so the residual win is measured rather than asserted.

### 5. Backtester parity

Add a batched dimension to `backtest/parity_diff.py`, which is the backtest-vs-live per-bar decision parity tool (`#906` D7.4) covering `signal`, `open_action`, `close_fraction`, and `regime`. It is not a registry-wiring check; registry and regime wiring are covered by `shared_scripts/test_regime_wiring.py` and the registry tests.

### 6. Config

No new config keys. `CurrentConfigVersion` stays at 17; no migration.

### Future

Extend batching to the spot, OKX, TopStep, Robinhood, and options lanes only after the HL lane is measured and stable. An explicit operator override that forces two strategies into one batch is deferred until a real case appears.

## Acceptance criteria

- A pure grouping helper partitions due strategies by data-source platform, symbol, timeframe, OHLCV limit, and ATR method, and is unit-tested against fixtures including single-member groups and multi-symbol inputs.
- Every argument outside the key travels per slot. A test proves two strategies with different position context, close refs, HTF filter, or regime windows still batch together and still receive their own values.
- The HL batched mode exists, accepts the documented argv and returns the documented JSON, and its results parse through the existing `HyperliquidResult` contract.
- Decision parity: for a fixed candle fixture and strategy set, every parsed decision field is identical between the batched path and the current per-strategy path. Tests control or separately validate timestamp fields.
- Per-strategy error isolation holds inside a batch: one failing slot returns an error for that slot only, and the other slots return normal decisions.
- A shared-state failure returns a distinct batch-level sentinel that the operator-facing logs and alerts render differently from a per-strategy failure.
- Strategies whose interval has not elapsed are never evaluated. A test asserts the batch contains exactly the due set.
- The `#1431` replay recording and paper-side suppression choke points behave identically under the batched path, with tests covering an open, a scale-in, and a full close.
- The 6-phase locking cycle is unchanged; grouping adds no new lock acquisition.
- `backtest/parity_diff.py` gains a batched dimension and reports zero diff on the standard fixture set.
- A committed benchmark reports raw per-cycle wall time, process starts, and processor time for the batched and unbatched paths on the same host, configuration, and candle fixture. **No speedup target is accepted without it.** Any target stated later must be arithmetically reachable for the group size it names.
- The pull request records a captured snapshot of the motivating deployment's strategy count, coin, and interval distribution, replacing the unverified counts this issue previously carried.
- Live rollout takes paper first, with one full trading day of decision parity before any live deployment adopts the path.
- `go test ./...` and the Python suites pass, including the new batch, parity, isolation, and replay tests.
- Negative criteria: no behavior change for single-member groups; no change to cycle interval or polling cadence; no change to the regime store; no change to the existing per-strategy backtest mode.

## Plain simple English

Many strategies trade the same coin in one bot. Each strategy starts its own Python process and loads the same libraries and indicators again. The market condition and the candles are already shared, so the remaining waste is the repeated startup and calculation. Group strategies that read the same data into one call.

---

---

Validated with LLM: MiniMax-M3 | high | Harness: validate-issue

richkuo:
## Plain simple English

Many strategies trade the same coin. Each one used to start its own Python
program and build the same price table again. Now the strategies that read the
same market data run together in one program call. The trading decisions do not
change. If the shared data fails, the bot says so with a special message and
goes back to the old one-at-a-time method.

## Summary

Closes #1442.

Due Hyperliquid perps strategies that share a market-data key now run in one
`check_hyperliquid.py --batch-check` call instead of one Python process each.
The regime bundle (#879) and the OHLCV disk cache (#839) already deduplicated
the regime computation and the candle fetch, so what this removes is the
remaining per-strategy interpreter start, module imports, DataFrame build and
indicator base.

**Python.** `check_hyperliquid.py` splits the signal check into
`build_shared_signal_state` (fetch, frame, ATR base, display price, fetch memos)
and `evaluate_signal_slot` (per strategy, on its own frame copy). Single mode
and `--batch-check` both run that pair, so batched and unbatched decisions come
from one implementation. A shared-state failure raises a typed error and emits
one `error_scope="shared_state"` sentinel with no results; a slot exception
stays inside that slot.

**Go.** `scheduler/hl_batch.go` groups due strategies by
`{DataPlatform, Symbol, Timeframe, OhlcvLimit, ATRMethod}`. Everything off-key
travels per slot. `runHyperliquidCheck` gains a batch parameter: a map hit skips
the spawn, a map miss is the untouched spawn path, and both finish through the
same `finishHyperliquidCheck` pipeline (#779 directional policy, #907
divergence, inversion, price fallback) — so risk, execution, the #1431 replay
choke points and hedge sync cannot tell which path produced the decision.

### Invariants this keeps

- **Grouping is pure** — no locks, no globals, no I/O — and runs over
  `dueStrategies` only, so a strategy whose interval has not elapsed is never
  batched.
- **The slot snapshot is one `mu.RLock` read placed AFTER
  `reconcileHyperliquidAccountPositions`.** Folding it into the earlier due-list
  read lock would hand close evaluators pre-reconcile position contexts — exit
  logic running against a position the exchange already closed. The issue's
  "grouping adds no new lock acquisition" criterion is met for the grouping;
  the snapshot adds exactly one read-lock scope, no lock is held across the
  subprocess, and `strategiesMu` is never involved, so the lock order is
  unchanged.
- **A group of one produces no batch.**
- **Eligibility is strict**: HL perps on the stock check script, with an argv of
  exactly the three positionals plus an optional `--mode`. An unknown configured
  flag would have to be translated into the slot envelope and could be silently
  dropped, so an unrecognized argv shape takes the per-strategy path.

### Correctness guards

- **Slot fingerprint.** The exact slot inputs plus the cycle mark price are
  rendered as canonical JSON at snapshot time and rechecked at dispatch. A
  position that moved, a params change, or a mark price written back by an
  earlier member discards the cached slot and that strategy spawns its own
  check. "The batched decision used this strategy's current inputs" is a checked
  fact rather than an argument about what else a cycle can touch.
- **A slot-level error clears the cached Result.** A slot error payload carries
  signal 0, so a surviving Result would have read as an ordinary HOLD; the
  member now takes the per-strategy soft-error branch and skips its cycle. The
  regression test was confirmed red against the unfixed code.
- **A shared outage freezes the members' own trackers** (neither recorded nor
  cleared) and fires ONE alert on a synthetic group identity as
  `**HL BATCH SHARED STATE FAILED**`, visually distinct from
  `**SIGNAL SCRIPT FAILING**`, reusing the existing 3-strike / transient /
  recovery machinery. Recording per member would storm N duplicate DMs and fake
  N recoveries.
- **Safety valve.** Three consecutive shared-state failures revert that group to
  per-strategy dispatch, retried every ten cycles. Without it, one batch-path
  defect would blind every strategy on the coin indefinitely.
- **Rollback.** `GO_TRADER_HL_BATCH=0` disables batching with no rebuild and no
  config key. `CurrentConfigVersion` stays 17; no migration.
- A batch timeout still matches the transient regex, so it classifies exactly
  like a per-strategy timeout.

## Benchmark

`docs/benchmarks/hl_batch.md` (+ raw per-repetition records and the pinned
candle fixture), produced by `scripts/bench_hl_batch.py`. Both arms are
network-free via a read-only `GO_TRADER_HL_OHLCV_FIXTURE` seam.

| N | unbatched wall median | batched wall median | process starts |
|---|---|---|---|
| 2 | 1.0649 s | 0.5247 s | 2 → 1 |
| 5 | 2.8055 s | 0.5458 s | 5 → 1 |
| 10 | 5.0720 s | 0.7066 s | 10 → 1 |
| 20 | 10.3598 s | 0.5169 s | 20 → 1 |

Read honestly: this ran on a developer workstation, **not** the 2 GB / 2 CPU
deployment class, and the artifact says so. What it establishes is the shape —
a fixed per-process cost of about 0.52 s and a per-slot marginal cost below this
run's noise floor. Peak child RSS shows no regression and **cannot** show a
reduction here, because both arms run their children sequentially; the artifact
does not claim one. The combined figure with #1441 is measured only after both
land, using that issue's own benchmark.

## Verification

- `go build`, `go vet`, `go test -race ./...` — green.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3356 passed.
- `uv run --no-sync python -m pytest shared_scripts/ -n auto` — 224 passed (26 new).
- `py_compile` on every touched Python file.
- Startup probe accepts the new `--batch-check` argv against real Python
  (`go-trader probe` exit 0).
- Live `--once` cycle batched two BTC strategies into one call; both decisions
  came through the normal per-strategy pipeline. `GO_TRADER_HL_BATCH=0` on the
  same config produced no batch.

### Tests added

Go (`scheduler/hl_batch_test.go`): grouping over mixed coins / timeframes / ATR
methods with custom-script, non-HL and unknown-argv exclusions; due-set-only
batching; a purity test that holds the write lock while partitioning; slot-vs-argv
equivalence including the zero-float omission rule; shared-argv leak check;
batch slots parsing identically to the single-mode `HyperliquidResult`; per-slot
isolation; missing and duplicate slots; shared-state tracker freeze and the
"do not falsely clear a prior member streak" case; the fallback threshold, retry
cadence and recovery; the environment kill switch; cached-slot consumption and
fingerprint rejection; and a replay-choke-point test driving open, scale-in and
full-close shapes through both paths and asserting identical suppression.

Python (`shared_scripts/test_check_hyperliquid_batch.py`): batched-vs-solo
decision parity over heterogeneous slots, timestamp validation, frame-mutation
isolation, fetch and funding memoization, per-slot error isolation, the
shared-state sentinel, envelope parsing and rejection, the argv/stdout wire
contract, `--probe-only` not reading stdin, and the single-strategy error and
short-history payload shapes staying unchanged.

`backtest/parity_diff.py` gains an opt-in `--batched` dimension: each compared
bar is evaluated alone and as one slot of a multi-slot batch, and `solo_*` vs
`batch_*` must agree. Four tests cover zero diff bare and composed, off-by-default,
and — so a clean report proves something — that a deliberately perturbed batch
slot is caught.

## Deviations from the posted plan

The diff follows the [Fable 5 implementation plan](https://github.com/richkuo/go-trader/issues/1442#issuecomment-5376246343)
with three deliberate deviations:

1. **A slot-input fingerprint was added.** The plan argued from per-strategy
   state isolation that the pre-pass snapshot cannot go stale before a member's
   own loop iteration. That argument is probably right, but it is an argument;
   the fingerprint makes it a check, and it also removes the plan's documented
   mark-price divergence (item 5) by sending that member to its own check
   instead.
2. **The eligibility rule is stricter than the plan's script-path test**: the
   argv must be the three positionals plus an optional `--mode`. The slot
   builder can only carry flags it knows about, so an unrecognized configured
   flag must not be silently dropped.
3. **The `parity_diff.py` dimension diffs batched against solo**, not against
   the tool's existing `live_*` columns. Those columns come from the spot check
   script's flow with Go-side signal transforms applied, so diffing them against
   the Hyperliquid evaluator would manufacture mismatches that are not batching
   defects. Solo-vs-batch is the claim #1442 actually rests on, and the added
   perturbation test proves the dimension can fail.

## Rollout

Paper first, one full trading day of decision parity before any live deployment
adopts the path. `GO_TRADER_HL_BATCH=0` is the instant rollback.

---


richkuo:
@claude opus review

richkuo:
Addressed review feedback (`github-actions[bot]` · 2026-08-22T00:28:40Z) in 64bbf8cc.

All four findings validated against the code and confirmed. No refutations.

### Fixed

1. **Rounded price write-back invalidated every peer's fingerprint** — dropped the cycle mark price from the slot signature (`scheduler/hl_batch.go:392-415`) and re-stamp the dispatch-time mark on a COPY of the cached result instead (`scheduler/main.go:4038-4062`). Traced and confirmed: `shared_scripts/check_hyperliquid.py:477` emits `round(price, 2)`, `scheduler/main.go:2503` writes it back into `prices[symbol]`, and `--mark-price` reaches only `price_override` → `output["price"]` (`shared_scripts/check_hyperliquid.py:247-249, 451-452`) while the decision reads `market_ctx["mark_price"] = df["close"].iloc[-1]` (`shared_scripts/check_hyperliquid.py:399`). The new `hyperliquidBatchDisplayPrice` rounds to two decimals so the batched member reports — and paper-fills at — exactly what its own spawn would have; a slot carrying an error keeps `Result == nil` and is never resurrected as a hold.

2. **Shared-state failure blanked protective maintenance for the whole group** — `hlBatchApplySharedFailure` now leaves every member a map MISS (`scheduler/hl_batch.go:657-681`), so each takes its own spawn in the same cycle and close evaluation, trailing-SL cancel+replace, the ratchet, protection sync and hedge sync all still run. The group alert and the 3-strike fallback are unchanged and member trackers stay frozen. The unreachable `SharedFailure` field, the `sharedFailure` parameter on `finishHyperliquidCheck` and the `scriptFailureSharedState` mode are removed rather than left as a trap.

3. **`GO_TRADER_HL_OHLCV_FIXTURE` in the live adapter** — removed from `platforms/hyperliquid/adapter.py` entirely; the adapter now reads no benchmark variable at all, so the invariant holds structurally rather than by a runtime check. The pinning moved into `scripts/bench_hl_batch.py`, which writes a `sitecustomize.py` onto the children's `PYTHONPATH` to replace `HyperliquidExchangeAdapter.get_ohlcv`. Taking the seam out entirely beats the alternative the review offered (declare symbol + interval, fall through on mismatch, `[CRITICAL]` on every hit): a loud override is still an override a stray setting can reach. Serving the fixture through the #839 disk cache was rejected because that cache's TTL is bounded at 60 s (`platforms/hyperliquid/adapter.py:51, 246-252`), so a long run would expire mid-arm and silently hit the network. `_verify_injection` runs a preflight child and aborts unless it confirms the patch took effect, so a failed injection cannot publish a network-bound artifact.

4. **Futures-registry fast path could not tell the spot registry apart** — now compares `os.path.realpath(mod.__file__)` against `FUTURES_STRATEGIES_PATH` (`shared_scripts/check_hyperliquid.py:144-170`). Confirmed: `shared_strategies/open/spot/strategies.py:57` defines `apply_strategy` too. The options registry does not, so that must-survive case was already safe, but the file check covers it uniformly.

### Verification

`go build`, `go vet`, `go test ./...` (full scheduler suite); `pytest shared_scripts/` (227), `platforms/` (328), `backtest/ shared_tools/ shared_strategies/ -n auto` (3028) — all pass. `scripts/bench_hl_batch.py` ran end to end on the new injection path and reproduced the committed artifact's shape.

New Go tests: the rounded write-back keeping peers batched on a low-priced coin, dispatch-time mark adoption with the cached slot left unmutated, a whole-number mark still hitting, a failed slot not resurrected by the price path, and shared-failure / timeout / unparseable-envelope each leaving members as misses. New Python tests: registry resolution against the spot shim, against a module with no `__file__`, and the no-regression case where `strategies` already is the futures registry.

`CLAUDE.md`, `docs/ARCHITECTURE.md` and `docs/benchmarks/hl_batch.md` are updated for all three behaviour changes.

## Plain simple English

Four problems were found and fixed. The group check used to throw away its own work as soon as the first strategy rounded the price, so the bot did the group work and all the single work. A failed group check used to stop stop-loss and take-profit upkeep for every strategy on that coin; now each one just checks by itself. A test-only setting that could have made live trading run on fake saved prices is removed from the trading code. And a shortcut that could pick the wrong strategy list now checks the file.

---

richkuo:
Correction to the line references in my previous comment — several were the review's pre-fix numbers or drifted with the edits. The dispositions and the code are unchanged; only the anchors below are right.

**Finding 1** — fingerprint: `scheduler/hl_batch.go:402` (`hyperliquidBatchSlotFingerprint`) and `scheduler/hl_batch.go:421` (`hyperliquidBatchDisplayPrice`); dispatch-time re-stamp: `scheduler/main.go:4035-4062`. Evidence traced: `shared_scripts/check_hyperliquid.py:487` emits `round(price, 2)`; `scheduler/main.go:2492` is the HL perps write-back into `prices[symbol]`; `--mark-price` reaches only `price_override` (`shared_scripts/check_hyperliquid.py:254`) → `output["price"]` (`shared_scripts/check_hyperliquid.py:457`), while the decision reads `market_ctx["mark_price"] = df["close"].iloc[-1]` (`shared_scripts/check_hyperliquid.py:404`).

**Finding 2** — `hlBatchApplySharedFailure` at `scheduler/hl_batch.go:673`.

**Finding 3** — the #839 cache bound that rules out serving the fixture through it: `OHLCV_CACHE_TTL_S = 60` at `platforms/hyperliquid/adapter.py:51`, applied by `_ohlcv_cache_ttl` at `platforms/hyperliquid/adapter.py:222`.

**Finding 4** — `_futures_strategies_module` at `shared_scripts/check_hyperliquid.py:144-170`; the collision it now survives is `apply_strategy` at `shared_strategies/open/spot/strategies.py:57`.

---

richkuo:
@claude review

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T04:16:43Z) in c9935ec0.

### Fixed

1. **Missing or duplicated slot cached a crash outcome for that member** — `hlBatchApplySlots` now leaves that member a map MISS so its own check runs in the same cycle, exactly like the shared-failure path (`scheduler/hl_batch.go:706-742`). The function returns a drift count, and the caller records it against `hlBatchFallback` and fires the group-identity alert, so persistent envelope drift trips the group back to per-strategy dispatch after 3 cycles instead of repeating forever (`scheduler/hl_batch.go:648-668`). A genuine per-slot `error` payload is untouched and stays a cached soft error. Tests: `TestMissingOrDuplicateSlotLeavesThatMemberAMiss`, `TestRepeatedSlotDriftTripsTheFallback`, `TestSlotErrorPayloadStaysACachedSoftError`.

2. **`parity_diff.py --batched` never parsed `strategy_refs`** — `_hl_batch_slot` now routes the slot through `check_hyperliquid.parse_batch_slots`, the same translation the `--batch-check` entrypoint applies, so `open_strategy` / `close_strategies` / `params` / `close_params_by_name` reach `evaluate_signal_slot` (`backtest/parity_diff.py:501-534`). Two tests were added: one asserts the composed keys are present on the slot, one asserts `batch_close_fraction` can be non-zero with a real open position. Note the pre-existing composed-path test used `profit_pct: 0.9`, which is a 90% tier and could never be hit — the new test uses 0.02.

3. **Group stderr stamped on every member** — the batch call's combined stderr is now logged once under the `hl-batch <key>` line and attached only to a member whose own slot carried an error, where its traceback actually is (`scheduler/hl_batch.go:640-647`, `scheduler/hl_batch.go:727-740`). A healthy member's outcome carries no stderr. Test: `TestGroupStderrIsNotStampedOnHealthyMembers`.

4. **Benchmark verified only the unbatched arm** — added `_verify_batched_arm`, which asserts the envelope parses and returns one clean result per slot before any batched timing is taken, and `_run` now aborts the benchmark on any non-zero timed child (`scripts/bench_hl_batch.py:145-190`, `scripts/bench_hl_batch.py:212-231`). Verified: a malformed envelope exits 1, and `--sizes 2 --reps 1` still completes and reports.

`docs/ARCHITECTURE.md` was updated for all three behavior changes — the old text stated a missing slot was that member's `scriptFailureCrash`.

### Verification

`go test ./...` in `scheduler` passes (37s). Full pytest suite (`shared_strategies/ shared_tools/ platforms/ backtest/`, `-n auto`) passes: 3358 tests. `gofmt` clean. Benchmark harness smoke-run completes with the new preflight.

## Plain simple English

If the group check comes back without an answer for one strategy, that strategy now just runs its own check, so its stop-loss and take-profit work stays on time. If this keeps happening, the system stops grouping that coin. Error text from one strategy no longer appears in another strategy's log. Two test tools now really test what they claim.

---

richkuo:
@claude review

Issue #296, opened 2026-04-17
Portfolio-level kill switch uses equity drawdown, understates margin drawdown for all-perps accounts


richkuo:
## Context

Follow-up from PR #294 review. PR #292 / #294 fixed per-strategy drawdown for perps to use deployed margin as the denominator. However, `CheckPortfolioRisk` at the portfolio level still uses equity-based drawdown:

https://github.com/richkuo/go-trader/blob/main/scheduler/risk.go#L335-L336

```go
if prs.PeakValue > 0 {
    prs.CurrentDrawdownPct = (prs.PeakValue - totalValue) / prs.PeakValue * 100
    ...
}
```

## Problem

For an all-perps account, `totalValue` is dominated by cash + unrealized PnL rather than deployed margin. A 50% drawdown on the margin staked in a leveraged position will show up as a much smaller percentage of total account value, so the portfolio-level kill switch fires far later than a margin-aware view would suggest. This is the same category of issue that PR #292 fixed at the strategy level.

## Suggested Fix

Mirror the `perpsMarginDeployed` logic from `risk.go` `CheckRisk` (#292) at the portfolio aggregate:

- Sum `perpsMarginDeployed` across all strategies.
- When perps margin is deployed, use peak-margin-relative drawdown for the perps portion.
- For mixed accounts, either (a) run two independent kill switches (spot/options equity vs perps margin), or (b) use a blended denominator.

Option (a) is cleaner and matches the per-strategy model.

## Out of scope in #294

#294 explicitly kept portfolio-level behavior unchanged to limit blast radius. This issue tracks the portfolio-level follow-up.

## Acceptance criteria

- [ ] All-perps portfolio with 50% margin drawdown triggers kill switch at the configured limit (currently it does not).
- [ ] Mixed spot+perps portfolios don't regress (spot equity drawdown still honored).
- [ ] Test coverage for both all-perps and mixed scenarios.

richkuo:
@claude

PR #1449, opened 2026-08-22
fix(#1448): let equity drawdown own the portfolio kill-switch latch [C89, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1448: [C89] Portfolio kill switch latches the whole book on one position's margin drawdown before equity loss reaches the limit
**Complexity: 89/100** — Capability 3 (Risk 4 — money/auto-protective kill-switch path); Volume 14 (Scope 2 + Coupling 2 + Verification 3) — Opus 5, xhigh · fableplan: yes

## Problem

The portfolio kill switch (`checkPortfolioRiskWithEquityAvailability`, `scheduler/risk.go:919`) fires on either of two signals against the same limit `MaxDrawdownPct` (`scheduler/risk.go:954`):

1. Equity drawdown — `(peak - totalValue) / peak`.
2. Perps margin drawdown (#296) — `perpsUnrealizedLoss / perpsMargin` (`scheduler/risk.go:942`).

The margin signal is intentional and documented (`scheduler/risk.go:890-906`): for an all-perps account, equity drawdown understates a leveraged loss. The defect is proportionality, and it has two parts:

1. The margin signal shares one limit with equity. A book that deploys only a small share of its value as perps margin trips the **portfolio** latch on a loss that is small against book equity. The two signals diverge exactly when deployed margin is a small share of the book — and in that case the maximum loss the margin trip can avert is bounded by that small margin, while the latch cost is the full book.
2. The trip latches and force-closes the **whole book**, though a per-position margin guard already exists: the #292 per-strategy circuit breaker measures the same margin-drawdown ratio and force-closes that one strategy (`scheduler/risk.go:1953-1985`).

Margin aggregation counts only `Type == "perps"` strategies (`scheduler/risk.go:867-878`); `manual` positions add nothing to the margin signal but are still force-closed when it trips.

Latch semantics: with a DM owner configured, the latch holds until the owner resets (`scheduler/main.go:1776`, `kill_switch_reset_dm_timeout`, #1368, default 6h). With no owner, it auto-resets after a confirmed flat close (`scheduler/main.go:1671-1695`). In both cases every position is force-closed first (`planKillSwitchClose`, `scheduler/kill_switch_close.go:362` → `forceCloseKillSwitchPositions`, `scheduler/main.go:1667`).

## Evidence (live incident, 2026-08-22 05:10:50 UTC)

From the `go-trader-live` deployment state DB (rows are on that host; they are listed here for the record):

| rowid | strategy | close price | unrealized PnL | configured SL floor |
|---|---|---|---|---|
| 1205 | hl-vwap-eth-60 | $2,432.95 | -$31.62 | $2,426.90 |
| 1204 | manual-eth | $2,432.95 | -$35.78 | ~$2,375 (2.5×ATR) |
| 1203 | manual-btc | $76,747.00 | $0.00 | n/a |

Kill-switch event (rowid 12202539):

```
portfolio perps margin drawdown 65.3% exceeds limit 30.0%
  (unrealized loss=$31.62, margin=$48.42, value=$914.97, peak=$1014.25)
```

- Margin DD: $31.62 / $48.42 = 65.3% (limit 30%) → trips. Only the perps strategy fed the ratio; the manual positions are excluded from margin aggregation.
- Equity DD at the same moment: $1,014.25 → $914.97 = 9.79%, far under the 30% limit.
- Both ETH closes executed **above** their SL floors: the configured stop losses never triggered, because the kill switch closed the positions first. The realized loss was smaller than the SL floor would have given, but the positions lost any chance to recover, and the whole book stopped trading.
- manual-btc was flat on PnL and was closed only because the kill switch closes every position.

## Goal

The portfolio latch fires only on real book-equity loss when equity data is available. Margin drawdown keeps its warning role, keeps the trip on the pooled-wallet path where equity is unavailable, and per-position margin protection stays with the existing #292 per-strategy circuit breaker. The kill switch never again flattens and latches a book at single-digit equity drawdown.

## Approach

Change the trip condition in `checkPortfolioRiskWithEquityAvailability` (`scheduler/risk.go:954`):

- **Equity available:** the latch trips on `equityDD > cfg.MaxDrawdownPct` only. Margin DD no longer trips the latch; it stays in the warning band (`scheduler/risk.go:989-1026`) with its existing `source="margin"` event labeling, so operators still get early margin alerts.
- **Equity unavailable** (pooled shared wallet without a trustworthy balance, `equityAvailable=false`): margin DD keeps the trip exactly as today (`scheduler/risk.go:968-976`) — it is the only kill-switch signal on that path and must survive (regression test `scheduler/risk_test.go:2575`).
- Per-position protection is unchanged: the #292 per-strategy circuit breaker (`scheduler/risk.go:1953-1985`) already force-closes a single strategy on margin drawdown with a cooldown, without latching the book.
- Rewrite the #296 rationale comment block (`scheduler/risk.go:890-906`) to the new semantics.
- No config version change; no new config field. The change makes the portfolio latch strictly less sensitive, and the equity guard still bounds real book loss at `MaxDrawdownPct`.

## Acceptance criteria

- With equity available, margin DD above `MaxDrawdownPct` alone produces a warning (source `margin`) and never latches, force-closes, or blocks strategies.
- With equity available, equity DD above `MaxDrawdownPct` latches with unchanged close, event, and reset semantics (owner DM reset and no-owner auto-reset both covered).
- With `equityAvailable=false`, margin DD above the limit still trips the latch; the existing regression tests stay green.
- The #292 per-strategy circuit breaker behavior is untouched by the diff.
- Margin-trip unit tests (`scheduler/risk_test.go:2521` onward) are rewritten to the new semantics; warn-band tests keep asserting margin warnings; the #1444 PeakValue rebaseline interaction is covered.
- The `KillSwitchEvent.Source` field still records `margin` for margin-driven warnings.
- The #296 doc comment matches the shipped behavior.

## Plain simple English

The emergency stop for the whole account can trigger when one leveraged position shows a loss that is large only against its own small margin. The account itself has lost very little. The stop then closes all positions and blocks all trading, and the normal per-position stop losses never get to operate.

---

richkuo:
## Plain simple English

The emergency stop for the whole account could trigger when one leveraged position showed a loss that was large only against its own small margin, while the account itself had lost very little. It then closed every position and blocked all trading. Now only real account loss can trigger it. The margin signal still warns, and each strategy keeps its own protection.

## Summary

Closes #1448

`checkPortfolioRiskWithEquityAvailability` still computes both drawdown lenses, but only one of them can latch the portfolio on a cycle:

```go
equityGuardArmed := equityAvailable && prs.PeakValue > 0   // read AFTER the peak ratchet
armed     -> trip on equityDD > MaxDrawdownPct, Source is always "equity"
not armed -> trip on marginDD > MaxDrawdownPct, Source is always "margin"
```

- **`PeakValue == 0` folds into "not armed"** rather than forming a third case. The peak ratchet runs before the trip check, so any cycle with a positive `totalValue` arms the guard in the same call. `PeakValue == 0` at the check therefore means no equity snapshot has ever been recorded, which is the same "the equity guard cannot measure" condition as a pooled wallet with no trustworthy balance. **Corrected in review round 1:** this is not "a cold-start account that blows up margin on bar 1 stays protected". Any first cycle carrying a positive total arms the guard through the ratchet above, so from that cycle on margin cannot fire the latch. The margin arm on the `equityAvailable` path is reachable only while `totalValue` has been non-positive on every cycle, which is what `TestCheckPortfolioRisk_PeakZero_MarginCanStillFire` pins with `totalValue = 0`. Margin's standing backstop is the `equityAvailable == false` path plus the #292 breaker.
- **The margin-wins tie-break is deleted, not reordered.** The two arms are mutually exclusive, so no tie can exist.
- **Margin drawdown above the limit is now a reachable warn state** and gets its own reason wording naming who owns the latch and where per-position protection lives, instead of the false "approaching kill switch limit".
- **Per-position margin protection is untouched.** The #292 per-strategy circuit breaker measures the same ratio and force-closes that one strategy on a cooldown, without latching the book. `git diff` does not touch it.
- No config field, no schema change; `CurrentConfigVersion` stays 17. `CurrentMarginDrawdownPct` computation and persistence, `planKillSwitchClose`, `AutoResetConfirmedFlatKillSwitch` and the DM-reset path are all untouched.

### Why the portfolio latch should not follow the margin signal

The latch force-closes the whole book and holds it until an operator resets, so its cost is the entire book, including `manual` and spot positions that contribute nothing to the margin ratio (`AggregatePerpsMarginInputs` counts `type == "perps"` only). The two signals diverge exactly when deployed margin is a small share of the book, and in that regime the loss a margin trip can avert is bounded by that small margin.

The live incident on 2026-08-22 latched the fleet at 65.3% margin drawdown on $48.42 of deployed margin while equity drawdown was 9.8% against a 30% limit. It closed an ETH position that was still above its configured stop-loss floor and a flat-PnL manual BTC position.

## Verification

| Check | Result |
|---|---|
| `gofmt -l scheduler/` | empty |
| `go -C scheduler vet ./...` | clean |
| `go -C scheduler build .` | passes |
| `go -C scheduler test -count=1 ./...` | passes (full suite) |
| `go -C scheduler test -count=1 -run 'KillSwitch\|SharedWallet\|NotionalCap' .` | passes |
| `grep -rn "whichever signal breaches\|tie-break to margin\|Ties go" scheduler/` | no matches |

**Red before green.** `scheduler/risk.go` was reverted to `HEAD` with the new tests in place; all six new or rewritten tests failed, then passed once the fix was restored:

```
--- FAIL: TestCheckPortfolioRisk_AllPerps_MarginDrawdownWarnsWithoutLatch
--- FAIL: TestCheckPortfolioRisk_MixedAccount_EquityGovernsWhenBothBreach
--- FAIL: TestCheckPortfolioRisk_Incident1448_MarginTripAvertedWhenEquityHealthy
--- FAIL: TestCheckPortfolioRisk_BothWarnMarginAboveLimit_ReasonNamesEquityGovernance
--- FAIL: TestCheckPortfolioRisk_MarginAboveLimit_WarnBookkeepingAcrossCycles
--- FAIL: TestCheckPortfolioRisk_AfterManualMarkBasisRebaseline_MarginDoesNotLatch
```

### Test changes

- `TestCheckPortfolioRisk_AllPerps_MarginDrawdownFires` → `..._MarginDrawdownWarnsWithoutLatch`: same #296 inputs, inverted expectation, plus warn-band bookkeeping assertions.
- `TestCheckPortfolioRisk_MixedAccount_MarginFiresFirst` → `..._EquityGovernsWhenBothBreach`: both signals over the limit now latch on equity, with both lenses still persisted separately.
- New `TestCheckPortfolioRisk_Incident1448_MarginTripAvertedWhenEquityHealthy`: replays the live incident numbers, then proves the equity guard still bounds real book loss at the same limit.
- New `TestCheckPortfolioRisk_BothWarnMarginAboveLimit_ReasonNamesEquityGovernance`: covers the new both-warn / margin-over-limit branch.
- New `TestCheckPortfolioRisk_MarginAboveLimit_WarnBookkeepingAcrossCycles`: `WarnBandEnteredAt` stamped once, per-cycle delta tracking, band clears on recovery, no latch at any point.
- New `TestCheckPortfolioRisk_AfterManualMarkBasisRebaseline_MarginDoesNotLatch`: the #1444 interaction. A margin arm that could still latch would re-introduce the first-cycle flatten that migration exists to prevent.
- `TestCheckPortfolioRiskMissingPooledEquitySuppressesOnlyEquityArm` extended with `Source == "margin"` on the equity-unavailable trip.
- Unchanged and green: `_PeakZero_MarginCanStillFire`, `_SpotEquityStillHonored`, `_NoPerps_EquityBehaviorUnchanged`, `_MarginWarning`, `_MarginWarning_FieldsPopulated`, `_BothSignalsBreachWarn_ReasonIncludesBoth`, `TestAggregatePerpsMarginInputs*`.

## Adopted plan

Built to the Fable 5 implementation plan posted on the issue: https://github.com/richkuo/go-trader/issues/1448#issuecomment-5379065178

The diff follows it step for step. One deviation:

- **`scheduler/portfolio_warning.go` is modified, though the plan lists it as untouched.** The plan assumed the new margin-over-limit reason string reaches the operator, but `BuildPortfolioWarningMessage` accepts `Reason` and never renders it. On the warn path the only operator-facing text is `Distance to kill switch: X% equity / Y% margin`, which after this change would tell an operator mid-incident that a fleet flatten is imminent when margin can no longer latch. The margin figure is now rendered as `perps margin Y% from limit`, which is true on both the armed and the unarmed path. `TestBuildPortfolioWarningMessage_IncludesTriageSections` updated to the new label.

## Behavior notes

- Margin drawdown above the limit now warns instead of latching once. **Revised in review round 1:** the send is throttled (`portfolio_warning_throttle.go`) because the band can now persist indefinitely — a per-cycle cadence would be roughly 144 identical messages per channel per day. The 50-entry kill-switch event ring buffer stays protected by the `portfolioWarnBandEntered` gate in `main.go`, which appends only on the transition into the band, and that append is not throttled.
- `effectiveStrategyIntervalSeconds` reads per-strategy drawdown, not the portfolio warn band, so the longer-lived portfolio warn state does not accelerate check intervals.
- The warn-event `Source` in `main.go` still names the larger of the two signals, so an audit row carries the headline number. That is a labeling choice for the warning band only, and its stale "mirrors CheckPortfolioRisk's tie-break" comment is rewritten.

## Review round 1 (all four Recommended Optional findings)

The `@claude` review returned `LGTM` with four non-blocking findings. Each was traced to current code and confirmed, then fixed.

1. **Untrusted equity totals.** `pooledEquityComplete` says the total EXISTS, not that it is trustworthy: `usedPVFallback` (`scheduler/shared_wallet.go`, a failed shared-wallet balance fetch substituted by `sum(member PV)`) and `usedStaleRiskBalance` (a one-generation-old cached balance) both leave it true, while `scheduler/main.go` already distrusts the same total enough to roll the peak back. The risk check now takes `equityTrusted` as well.

   The latch **stays with equity** on those cycles rather than folding the flags into `equityAvailable`. `sum(member PV)` prices every position at the same live marks and a one-generation-old balance is a real balance, so equity still measures real book loss — and handing the whole book to the margin signal over a transient fetch blip is exactly the failure this PR exists to remove. What an untrusted total must never do is make a drawdown disappear, so two things gate on `equityTrusted` instead: the peak ratchet is skipped inside the check (so `CurrentDrawdownPct` is no longer written off a peak the caller is about to undo), and `equityDD` is floored at the last recorded reading. The floor can never latch by itself — every stored reading came from a cycle that did not latch, so it is at or below the limit.

2. **`circuit_breaker: false`.** The #292 per-strategy breaker now owns margin protection, and it is opt-outable (#1048) at a looser perps default (50 against the portfolio default of 25). `recordCircuitBreakerSuppression` now raises a throttled owner DM (`scheduler/circuit_breaker_suppression_alert.go`) instead of a log line only: queued under `mu` inside `CheckRisk`, drained by the main loop after `mu.Unlock()` per the #880 no-notifier-I/O rule, throttled by the existing once-per-episode `circuitBreakerSuppressedWarned` latch. Scope is every strategy type, not perps only — futures share the breaker and a spot strategy with it off is equally unhalted.

3. **Cold-start comment.** Reworded in `scheduler/risk.go`, both doc bullets, and the bullet above.

4. **Warn-DM cadence.** `portfolioWarningShouldNotify` (pure) throttles on the `alert_throttle_interval` floor while keeping immediate escapes for the first cycle of a band, a signal that was not in band at the last notification, and a rise of 1.0pp or more on either lens. A suppressed cycle returns the previous state unchanged, so a slow creep still accumulates past the escalation threshold. `portfolioWarningAlertsReset()` runs whenever the band is not active, so a cleared or latch-ended band notifies again on its next entry. `portfolioWarnBandSignals` is now the single definition of band membership, read by both the risk check and the throttle.

**New tests** (`scheduler/portfolio_risk_review_1449_test.go`), each covering the finding and its inverse: `TestUntrustedEquity_FreezesPeakAndFloorsDrawdown`, `TestUntrustedEquity_FloorAloneCannotLatch`, `TestUntrustedEquity_LatchStaysWithEquity`, `TestColdStartWithPositiveEquity_MarginCannotFire`, `TestPortfolioWarningThrottle`, `TestPortfolioWarnBandSignals_SharedDefinition`, `TestCircuitBreakerSuppression_QueuesOwnerDM`. `gofmt` clean, `go vet` clean, `go -C scheduler test ./...` passes.

---
Updated with LLM: Opus 5 | xhigh | Harness: Claude Code


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T08:02:15Z) in 8c9d2f0d. All four `Recommended Optional` findings traced to current code, confirmed, and fixed. `gofmt` clean, `go vet` clean, `go -C scheduler test ./...` passes (full suite).

### Fixed

1. **Cold-start comment claimed protection the ratchet makes unreachable** — reworded to the condition the code actually tests: any first cycle with a positive total arms the guard, so the `equityAvailable`-path margin arm is reachable only while `totalValue` has been non-positive on every cycle (`scheduler/risk.go:1001`). Same correction applied to the `CLAUDE.md` / `docs/ARCHITECTURE.md` bullets and the PR body. New `TestColdStartWithPositiveEquity_MarginCannotFire` pins bar-1-with-cash; the retained `TestCheckPortfolioRisk_PeakZero_MarginCanStillFire` pins the `totalValue = 0` case.

2. **`circuit_breaker: false` left a leveraged blow-up with no automatic protection at any level** — the interaction is now named in the `risk.go` #1448 comment and both doc bullets, and `recordCircuitBreakerSuppression` raises a throttled owner DM instead of a log line only (`scheduler/circuit_breaker_suppression_alert.go`). Queued under `mu` inside `CheckRisk`, drained by the main loop after `mu.Unlock()` per the #880 no-notifier-I/O rule (`scheduler/main.go:2166`), throttled by the existing once-per-episode `circuitBreakerSuppressedWarned` latch. Scope is every strategy type, not perps only: futures share the same breaker and a spot strategy with it off is equally unhalted, so narrowing would hide the identical gap elsewhere. `TestCircuitBreakerSuppression_QueuesOwnerDM` covers the fire, the same-episode suppression, and the re-enable/re-disable re-arm.

3. **Unbounded warn DMs while the band persists** — `portfolioWarningShouldNotify` (`scheduler/portfolio_warning_throttle.go`) throttles on the `alert_throttle_interval` floor (the `missingMarkAlerts` knob) with immediate escapes for the first cycle of a band, a signal that was not in band at the last notification, and a rise of 1.0pp or more on either lens. A suppressed cycle returns the previous state unchanged, so the escalation baseline stays anchored to the last reading sent and a slow creep still fires once it accumulates. `portfolioWarningAlertsReset()` runs whenever the band is inactive, so a cleared band — or one ended by the latch — notifies again on its next entry. The `RecentTrades` query and message build are skipped on a throttled cycle; the `[WARN]` stdout line and the band-entry `kill_switch_events` append are not throttled. `TestPortfolioWarningThrottle` covers all three of your **Must survive** cases.

### Resolved judgment calls

1. **Equity guard armed on cycles whose total the same code flags as substituted or stale** — confirmed exactly as reported: `usedPVFallback` (`scheduler/shared_wallet.go:554`) and `usedStaleRiskBalance` (`scheduler/shared_wallet.go:184`) both leave `pooledEquityComplete == true`, and `scheduler/main.go:1420` distrusts the same total enough to roll the peak back while the latch ran solely off it.

   You offered two remedies. I took neither verbatim and implemented a third, because folding the flags into `equityAvailable` would contradict this PR's own principle. `sum(member PV)` prices every position at the same live marks, and a one-generation-old balance is a real balance, so on those cycles the equity signal **can** still measure real book loss — and #1448's rule is that equity owns the latch whenever it can measure at all. Folding them in would hand the whole book to the margin signal on a transient balance-fetch blip, which is the failure this PR exists to remove, and it would do so in both error directions: a substitute can understate equity as easily as overstate it, so the margin arm could spuriously flatten a healthy book.

   What an untrusted total must never do is make a drawdown *disappear*. So a third state, `equityTrusted = pooledEquityComplete && !usedPVFallback && !usedStaleRiskBalance` (`scheduler/main.go:1414`), gates two things instead of latch ownership: the peak ratchet is skipped inside the check (`scheduler/risk.go:990`) so `CurrentDrawdownPct` is no longer written off a peak the caller is about to undo, and `equityDD` is floored at the last recorded reading (`scheduler/risk.go:1005`). That closes the fail-open direction you named. The floor can never latch by itself — every stored reading came from a cycle that did not latch, so it is at or below the limit; it can only stop an overstated substitute from masking a loss already visible.

   Your **Must survive** case (a) therefore resolves differently from your remedy: a non-pooled shared wallet with a failed balance fetch and margin far above the limit does **not** latch on margin. It latches on equity as soon as the equity measurement — floored, and re-measured for real on the next trusted cycle — crosses the limit. Cases (b) and (c) hold as written: a stale balance whose real value has fallen past the limit latches on equity because the floor never lowers the reading and the next trusted cycle measures it directly, and a clean full balance read with margin over the limit still does not latch (the #1448 incident case). `TestUntrustedEquity_FreezesPeakAndFloorsDrawdown`, `TestUntrustedEquity_FloorAloneCannotLatch` and `TestUntrustedEquity_LatchStaysWithEquity` cover these plus the `equityAvailable == false` inverse.

   Rejected alternatives: fold the flags into `equityAvailable` (contradicts the PR's principle, re-opens the small-margin flatten on a fetch blip, wrong in both error directions); document the exclusion and change nothing (leaves the fail-open masking you identified). Override if you want the stricter hand-back to margin.

### Also

- `portfolioWarnBandSignals` (`scheduler/risk.go:1140`) is now the single definition of warn-band membership, read by both the risk check's reason switch and the new throttle, so "newly entered the band" cannot drift between them.
- Docs: `CLAUDE.md` and `docs/ARCHITECTURE.md` carry the `equityTrusted` third state, the `circuit_breaker: false` caveat, the corrected cold-start reach, and the throttle.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T08:27:43Z) in 57a17e3c. All three findings traced to current code and confirmed. `gofmt` clean, `go vet` clean, `go -C scheduler test ./...` passes (full suite).

### Fixed

1. **The untrusted-cycle floor could latch the portfolio on its own** — confirmed exactly as reported, and your soundness objection is the correct one: `prs.CurrentDrawdownPct = equityDD` runs before the latch check, so the latching cycle persists its over-limit measurement, `db.go` writes it to SQLite, and the DM reset left it. The comment claiming "every stored reading came from a cycle that did not latch" was false.

   You offered clamp and/or clear-in-the-DM-reset. I did both, because they close different halves. The clamp is the structural guard: `priorEquityDD` is clamped to `cfg.MaxDrawdownPct` (`scheduler/risk.go:1059`) while the latch test stays a strict `>`, so a carried reading can raise the reported drawdown but can never fire the kill switch alone — including on the restart path, where no reset ran at all. Clearing alone would not cover that. It is a **read-time** clamp, never stored, so a `max_drawdown_pct` change re-derives it instead of truncating the record, and a non-positive limit yields a non-positive floor that cannot raise a drawdown already floored at 0.

   The clearing half is `ResetPortfolioKillSwitchManual` (`scheduler/risk.go:817`), now the sole body of the DM reset (`scheduler/main.go:1833`). It clears both drawdown readings and the marker like `ClearLatchedKillSwitchSharedWallet` and `AutoResetConfirmedFlatKillSwitch`, and returns the pre-reset reading so the `reset` audit event still records the drawdown the operator reset from. It deliberately does **not** re-baseline `PeakValue` — unlike the two auto-resets, a manual reset makes no claim that the book is flat or the total verified, so the real high-water mark is retained.

   All three **Must survive** cases are pinned by `TestUntrustedEquity_StoredOverLimitFloorCannotLatch`: (a) stored reading exactly at the limit does not latch; (b) a genuine this-cycle 40% measurement still latches on an untrusted cycle, so the clamp is not a fail-open; (c) the restart shape — reading above the limit reloaded with `KillSwitchActive` already false, first cycle untrusted. I verified (c) reproduces your report by reverting only the clamp: `portfolio drawdown 40.0% exceeds limit 25.0% (value=$10000.00, peak=$10000.00)` — a portfolio at full peak value latching. `TestManualKillSwitchReset_ClearsStaleReadings` covers the reset half, and `TestUntrustedEquity_FloorAloneCannotLatch`'s doc comment now records the corrected argument rather than the disproved one.

2. **The floored value breaks the type's arithmetic invariant on the operator surfaces** — confirmed at every site you named. You offered a decision-time local or an explicit marker; I took the marker.

   Keeping the floor as a local and persisting the raw quotient restores the pure invariant, but it splits the displayed number from the decided number: during an untrusted run `/status` and the warning message would show 2% while the latch and the warn band were acting on 12%. Since `portfolioWarnBandSignals` reads the persisted field and is documented as the single definition of band membership, that split would also have to be threaded back through the warn path. So `CurrentDrawdownPct` keeps the floored decision value — the stored number always equals the one that cycle decided on — and `PortfolioRiskState.DrawdownReadingSubstituted` (`scheduler/risk.go:617`) marks it, persisted as the additive column `portfolio_risk.drawdown_reading_substituted`.

   **Must survive:** (a) your raw-0%/floored-10% case is `TestDrawdownReadingSubstitutedMarker`, and the warning message renders the label rather than an unqualified 10% loss at peak value; (b) the restart reload is in the `db_test.go` round-trip; (c) the inverse holds — a trusted cycle writes the exact quotient and clears the marker, and an untrusted cycle whose own measurement exceeds the floor is a real measurement and is not marked either.

   One consequence worth recording: a latch can never carry the marker. The floor is clamped to the limit, so a latching `equityDD` is always this cycle's own unclamped measurement. That is why the `/status` latched-drawdown line (`scheduler/discord_commands.go:302`) needs no change — it prints only on the latched path and shows no peak/total beside it.

3. **The warning message labeled margin "from limit" on the path where margin owns the latch** — confirmed, including the reachability: `portfolioWarnBandSignals` (`scheduler/risk.go:1207`) computes `marginInBand` independently of `equityInBand`, so the warn band is reachable with the equity guard unarmed. Threaded `EquityGuardArmed` through `PortfolioWarningMessageInputs` (`scheduler/portfolio_warning.go:27`) as you recommended, set at the call site from the same condition the risk check evaluated (`scheduler/main.go:1780`).

   Armed keeps the #1448 labeling verbatim. Unarmed swaps them, and two things beyond the swap: the equity figure renders `n/a (guard not armed)` instead of the stale reading the check did not update this cycle, and the trend line omits the equity delta — the check forces `WarningEquityDeltaPct` to 0 there, so printing it showed a reassuring `+0.0%` for a signal that was never measured. `TestPortfolioWarningLabels_FollowTheArmedGuard` covers your three cases: pooled wallet with no trustworthy balance, `PeakValue == 0` cold start with a leftover non-zero reading, and the armed inverse.

### Also

- Docs: the `CLAUDE.md` and `docs/ARCHITECTURE.md` #1449 bullets carry the clamp, the corrected soundness argument in place of the disproved one, `ResetPortfolioKillSwitchManual`, the substituted marker, and the label swap. The bullet's old "No config/schema change" now reads "no config change (`CurrentConfigVersion` stays 17); one additive SQLite column".

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T08:53:21Z) in 7d797752. `gofmt` clean, `go vet` clean, `go -C scheduler test ./...` passes (full suite).

### Fixed

1. **An untrusted total that reads too LOW can still latch the whole book on the equity arm** — confirmed at the exact lines you cite. The floor at `scheduler/risk.go` only ever raises `equityDD`, and the latch test carried no `equityTrusted` gate, so an inflated reading from a substituted or aged total reached it unimpeded. Your framing of it as the mirror of the guarantee the downward floor already gives is the right one, and that is the shape of the fix.

   **I implemented the deferral you recommended, with one deliberate change: it has a deadline.** A straight "only a trusted measurement may latch" rule opens a worse hole than it closes, and the reason is that your two untrusted sources have different lifetimes.

   `usedStaleRiskBalance` is structurally **one cycle**. The stale branch in `resolveSharedWalletRiskBalances` (`scheduler/shared_wallet.go:181`) requires `snapshot.Generation == generation-1` and does **not** write the snapshot back to the cache, so a second consecutive miss fails that test, `equityComplete` goes false, and the cycle lands on the `equityAvailable == false` path where margin owns the latch. That source can never hold the deferral open.

   `usedPVFallback` has no such bound. `computeTotalPortfolioValue` substitutes `sum(member PV)` for as long as the balance fetch keeps failing, and `pooledEquityComplete` stays true throughout. A pure veto would therefore leave the portfolio latch disarmed for the entire length of a balance-endpoint outage — and an outage is exactly when a genuine crash is least likely to be caught by an operator watching a dashboard. Your **Must survive (c)** allows "once a trusted cycle lands", but on this source a trusted cycle is not guaranteed to land at all.

   The two error directions are not symmetric. A spurious flatten costs slippage, fees, and a manual reset; it is bounded and recoverable. A missed latch during a real crash with a broken balance endpoint is unbounded. The unbounded error governs, so the latch waits the untrusted reading out and then fires.

   Mechanically: `PortfolioRiskState.UntrustedOverLimitSince` (`scheduler/risk.go:651`, persisted as the additive column `portfolio_risk.untrusted_over_limit_since`) stamps the start of an **unbroken** run of cycles on which an untrusted `equityDD` read above the limit. While the run is shorter than `untrustedEquityLatchDeferral` (15m, `scheduler/risk.go:585`) the equity arm is held via `equityLatchDeferred` (`scheduler/risk.go:1196`); once it elapses the latch fires with the reason naming the untrusted basis and the escalation. Every non-qualifying cycle clears the window — trusted, at-or-below the limit, or equity guard unarmed — so the clock only ever measures an unbroken run and a one-cycle dip cannot accumulate toward an escalation it never earned. The firing latch and all three reset paths clear it too. It is persisted specifically so a crash-restart loop cannot keep reopening a fresh window and defer forever, which is the failure mode an unpersisted counter would have.

   `cfg.MaxDrawdownPct <= 0` is excluded, matching the `priorEquityDD` clamp: that degenerate config already means "latch on any drawdown" and keeps its meaning rather than being silently redefined.

   The bound is not configurable. It is a safety bound on an auto-protective mechanism, and exposing it would let a config edit reintroduce either hole. `CurrentConfigVersion` stays 17.

   Your three **Must survive** cases are pinned by `TestUntrustedEquity_OverLimitLatchIsDeferredNotVetoed`: (a) untrusted cycle, real equity healthy, substituted total showing 40% against a 25% limit — does not latch, and the window opens; (b) a trusted 40% measurement latches on the same cycle, opens no window, and its reason does not claim an untrusted basis; (c) an aged window escalates to a real latch. Beyond your cases: `TestUntrustedEquity_DeferralWindowClearsOnEveryNonQualifyingCycle` covers the inverse — a window that must not survive a trusted cycle, an at-limit cycle, or an unarmed guard; `TestUntrustedEquity_DeferralDoesNotDisarmTheMarginArm` pins that the `equityAvailable == false` path still latches on margin, since that path plus #292 is the standing protection the deferral leans on; `TestUntrustedEquity_DeferralWindowSurvivesReset` covers the reset paths; `TestUntrustedEquity_DegenerateLimitKeepsExistingMeaning` covers the non-positive limit; and the `db_test.go` round-trip covers restart persistence.

   You asked for a loud interim, and it is loud on five surfaces: a `latch_deferred` `kill_switch_events` row on entry (once per window, not per cycle, so the 50-entry ring buffer is not evicted), the stdout line moving from `[WARN]` to `[CRITICAL]` (`scheduler/main.go:1811`), the warn DM **bypassing** the `portfolioWarningShouldNotify` throttle for as long as the deferral is active (`scheduler/main.go:1750`) — the throttle exists so a permanent band cannot spam, but silence here would hide the one window in which a manual decision still changes the outcome — the warning message replacing the `Distance to kill switch: 0.0% equity` line (`scheduler/portfolio_warning.go:118`), which on every other path means a flatten is imminent, with the deferral, its deadline and a pointer at #292, and a `Portfolio latch DEFERRED` line on `/circuit-breakers` (`scheduler/discord_commands.go:305`).

   One consequence worth recording: a deferred cycle persists its **own unclamped measurement** in `CurrentDrawdownPct`, not a value truncated to the limit, and is not marked `DrawdownReadingSubstituted`. The reading is a real measurement; only the decision was deferred, and that is recorded in `UntrustedOverLimitSince`. So the operator sees the true 40% rather than a 25% that would understate the incident.

### Test assertions I changed — please check this

Two round-2 assertions asserted a genuine over-limit measurement latches **immediately on an untrusted cycle**, which is the behavior your finding says must not happen. They existed to prove the `priorEquityDD` clamp was not a fail-open — that the clamp cannot swallow a real measurement — and that property is intact and now pinned on a **trusted** cycle, where the clamp is the only mechanism in play. `TestUntrustedEquity_StoredOverLimitFloorCannotLatch` case (b) and `TestUntrustedEquity_FloorAloneCannotLatch` now also assert the untrusted reading survives unclamped and opens the window, so the clamp's property is still checked on the untrusted path; only the immediacy of the latch moved. Flagging it explicitly because it is a reduction in how fast an untrusted over-limit reading reaches protection, which is the trade the finding asks for.

### Also

- Docs: the `CLAUDE.md` and `docs/ARCHITECTURE.md` #1449 bullets carry the deferral, why it is not a veto (the two sources' lifetimes), the clearing rules, the persistence rationale, the five operator surfaces, and the moved test assertions. The schema line now reads two additive columns.

---

richkuo:
@claude review

Issue #1428, opened 2026-08-10
[C68] Test the Hurst exponent as a sizing and exit input rather than an entry gate


richkuo:
**Complexity: 68/100** — Capability 2 (Coupling 3 — bar-for-bar parity with the shipped `hurst_gate` sizing form, the close-evaluator ATR rules, and the contract-path handover — forces Capability at least 2; Risk 1 — report-only, no live path; Uncertainty 2 — the sizing form is already fixed by #1411, the exit form is not); Volume 18 (Scope 3 — a research harness plus an exit-side arm that drives the Backtester's close path rather than only masking entries; Coupling 3; Verification 3 — pytest, a full sweep, and parity against the live formula) — Opus 5, xhigh · fableplan: no

## Problem

Every Hurst study so far asks one question: should this trade be allowed to happen. #1410, #1422, #1424, #1426 and #1427 all score an ENTRY decision, and #1424's confirmatory family answers with a separation of -0.005 efficiency units against a 0.013 limit — pointing away from the hypothesis, in a direction its one-sided design cannot detect.

Two things that question never reaches:

1. **Sizing.** #1411 ships a `mode: size` arm — `clamp(|H-0.5|/0.15, size_floor, 1.0)`, applied where size is computed. #1410 swept a DIFFERENT sizing form (`clamp(1 + gain*e, 0, 1.5)`, which can exceed 1.0), so no study has ever scored the form the code actually implements. The shipped multiplier is currently uncalibrated by construction.
2. **Exits.** Nothing has tested Hurst on the exit side at all. A statistic that fails to predict whether a trade wins can still predict how far it runs before it reverses, and that is an exit question rather than an admission question. Persistence is, on its face, a claim about how long a move continues.

## Sibling status

Both remaining entry-side studies have landed and both DEFER the live-evidence contract path:

- #1426 landed via PR #1452. It is exploratory-only, and its script refuses the contract path unconditionally (`hurst_1426_two_sided_sort.py:1656-1663`).
- #1427 landed via PR #1473 with `CONTRACT_PATH_CLAIMED = False`. Its committed artifact passes the supersede clause to #1428 alone, on the ground that the shipped gate reads the LEVEL of H and never its change, so nothing #1427 measured can calibrate it (`hurst_1427_change_sort.py:259-275`).

The clause is therefore no longer shared. This study is the only remaining one that scores the sizing form the shipped gate implements, and at most one study may claim the path.

[...]

Issue #1427, opened 2026-08-10
[C64] Test whether the change in the Hurst exponent sorts entry outcomes


richkuo:
**Complexity: 64/100** — Capability 2 (Scope 2 — one research harness plus tests; Coupling 2 — reuses the #1424 harness but adds a derived series and its warm-up; Risk 1 — report-only, no live path; Uncertainty 3 — the transform, the lookback and the bucket edges are all undetermined and each one costs power; Verification 3 — pytest, a full sweep, and a byte-identical committed-artifact round-trip); Volume 14 — Opus 5, xhigh · fableplan: no

## Problem

Every Hurst study so far — #1410, #1422, #1424 — buckets the LEVEL of the Hurst exponent at entry: below 0.45, 0.45 to 0.50, 0.50 to 0.55, at or above 0.55, and NaN. On the confirmatory rows of the best-powered study, that level did not sort trades. #1424's confirmatory family separates by -0.005 efficiency units against a 0.013 limit, and the sign points away from the hypothesis.

Nothing has tested whether the CHANGE in H carries information. Those are different claims. "This market is trending" is a statement about a level. "This market is becoming trending" is a statement about a derivative, and a persistence estimate that is rising through 0.5 describes a regime that is forming rather than one that is already priced.

## Goal

Measure whether the change in H over a fixed lookback sorts entry outcomes, under the same power discipline #1424 established: a row-matched, directional validity gate, a cluster-rotation null, effective sample size, and a measured detection limit reported beside the pool-matched separation.

## The constraint that shapes the design

Power is still the binding constraint, and this study makes it worse before it makes it better:

- A lookback is a free parameter. Sweeping several multiplies the Benjamini-Hochberg family and raises the rank-1 bar, which is the exact trap #1424's Route 1 escaped by pinning ONE hypothesis. The design must fix one lookback in advance, on a stated rationale, or budget for the family it creates.
- The bucket edges on a difference are not inherited from anywhere. #1410's level edges came from H's own 0.5 midpoint; a difference has no such landmark, so the edges are a pre-registration decision with no prior.
- A difference inherits H's warm-up and adds the lookback on top, so the required lead per dataset grows and the coverage audit will drop more early cells. Expect the pre-2020 tape #1424 added to lose its earliest windows.

State all three before the run, not after.

## Approach

[...]

PR #1452, opened 2026-08-22
feat(#1426): test the Hurst entry hypothesis in both directions [C62, Opus 5, high, fableplan]

### LINKED ISSUE #1426: [C62] Test the Hurst entry hypothesis in both directions, so a reversed effect becomes a finding
**Complexity: 62/100** — Capability 2 (Uncertainty 3 — the cohort-option design decision is open; Scope 2 — one research harness plus tests; Coupling 2 — contract-path coordination across three study scripts; Risk 1 — report-only, no live path; Verification 2 — pytest plus a full sweep); Volume 12 — Opus 5, high · fableplan: yes

## Problem

Three studies have now tested one hypothesis: for momentum-style entries, trades taken at high Hurst persistence are better than trades taken at low persistence. #1410 could not resolve it, #1422 could not resolve it, and #1424 (merged in #1425) could not resolve it either.

#1424 is the one that matters here, because it measured WHY. On its confirmatory family (`momentum`, 7,992 rows) the row-matched detection limit is 0.013 efficiency units and 0.94 pp of net return. The observed separation on those same rows is **-0.005 and -0.12 pp**.

The sign is the finding. `_separation` is `mean(kept) - mean(suppressed)`, and every null and every injection in that study is ONE-SIDED in that orientation — they can only ever reject "the suppressed trades were worse". A negative separation means the trades the gate would have SUPPRESSED did BETTER than the ones it would have kept. That is the direction the design cannot detect at any magnitude, so the observation is currently unusable as evidence: it is neither a rejection nor a confirmation, and #1424's report says so explicitly.

More data does not fix this. The hypothesis, and the test built around it, points one way while the data points the other.

## Goal

A pre-registered study that can reject in BOTH directions, so a reversed effect becomes a finding instead of a blind spot. The question it answers: does Hurst at entry sort trades at all, in either direction, at an effect size this design can see?

## The hazard that must be settled before any code runs

We have already seen the sign. That is an interim look of the worst kind for a confirmatory re-test on the same rows, and it is worse than #1424's disclosed look because this time the look is what motivated the hypothesis.

Two honest ways out, and the study must choose ONE and state it at the top of the file:

1. Reserve rows the #1424 render never scored — new calendar windows, or venue/window cells outside its ownership matrix — and make those the confirmatory cohort. This preserves a real pre-registration but shrinks the pool, which is the constraint that has bound all three studies.
2. Declare the study exploratory-only, run it on the full pool, and make the verdict say plainly that no confirmatory claim is available. This keeps the power but forbids a recommendation.

Picking silently, or claiming a pre-registration the data cannot support, is the failure mode. The choice, its cost, and the reason belong in the report.

## Approach

Reuse the #1424 harness wholesale — same datasets, windows, cohorts, estimator, targets, cluster-rotation null, effective-N model, coverage floors and warm-up audit. What changes is the direction of the test:

- Two-sided permutation and cluster-rotation p-values, with the one-sided functions unreachable from the two-sided path.
- A two-sided injection in the minimum-detectable-effect search, so the limit describes a symmetric test.
- A validity gate that compares the MAGNITUDE of the separation only because the test is genuinely two-sided. #1424's gate reads the SIGNED value on purpose, and relaxing that is only legitimate once the null underneath it is symmetric. The gate must stay ROW-MATCHED on the confirmatory family either way.
- The Benjamini-Hochberg family stays as small as the design allows. Power is still the binding constraint, and a two-sided test costs power relative to a one-sided one at the same alpha.

## Acceptance criteria

- Design constants pre-registered at the top of the file; the Recommendation is a mechanical render, never hand-written.
- A test asserts the two-sidedness: no one-sided p-value function is reachable from the confirmatory path, and a synthetic reversed effect of sufficient size is DETECTED rather than ignored.
- The interim-look disclosure states that the sign was known before the design was fixed, and names which of the two options above was taken.
- The validity gate stays row-matched on the confirmatory family, and `decision` in the committed JSON equals `decision_payload(decide_recommendation(configs, mde))`, pinned by test as #1424 does.
- Look-ahead invariant unchanged: rolling H at bar `i` reads closes `[i-W+1, i]`, decision series shifted one bar, fill-bar stamp shifted twice.
- The estimator stays `hurst_exponent` from `shared_strategies/open/indicators_core.py` (#1409 single source of truth) and is never reimplemented.
- NaN stays its own bucket for both H and ADX.
- Contract-path handling follows the precedent #1424 set for #1422: if this study supersedes #1424 as the live evidence, #1424's render moves to its own issue-numbered file and its script is refused the contract path; if it does not supersede it, this study never writes the contract path at all. State which, up front. The supersede decision must also be coordinated with #1427 and #1428, which carry the same clause — at most one of the three open studies may claim the contract path, and the study that claims it names the siblings that defer.

## Untouched

`hurst_gate` semantics and its default-off state (#1411), the composite classifier and label vocabulary (#1412 stays closed), every live path, and `config.example.json`. This is report-only research. No thresholds ship from it.

## Plain simple English

Three studies tried to show that trades opened during strong price trends do better. Each test could detect an effect in only one direction, and the data pointed the other way. This study makes the test work in both directions, so a reversed effect becomes a real finding and stops being invisible.

---

richkuo:
## Plain simple English

Three earlier studies asked whether trades opened during strong price trends do better. Each test could only find an effect in one direction, and the data pointed the other way, so the result could not be used. This study makes the test work in both directions. The answer is still "no clear effect", but a reversed effect is now visible instead of invisible.

## Summary

Closes #1426.

#1410, #1422 and #1424 all tested one hypothesis ONE-SIDED. #1424 measured the consequence: on its confirmatory family the row-matched detection limit is 0.013 efficiency units while those same rows separate by **-0.005**. `_separation` is `mean(kept) - mean(suppressed)` and every null and every injection in that study points one way, so a negative value is an effect the design cannot detect at any magnitude. The observation was neither a rejection nor a confirmation.

This study inherits #1424's design wholesale — datasets, windows and the ownership matrix, cohorts, estimator, targets, cluster-rotation null internals, effective N, coverage floors, warm-up audit, acceptance rule — and changes only the DIRECTION of the inference.

**What changed**

- Every confirmatory p is a doubled one-tailed permutation p, `p2 = min(1, 2*min(p_ge, p_le))`, with both tails counted over the draws the null ACTUALLY produced rather than the requested count. The doubled tail is used rather than the `|stat| >= |obs|` form because neither statistic is symmetric under its own null: the group difference is not, under unequal group sizes, and the weighted sizing statistic is not centred at zero at all.
- The detection limit injects BOTH directions at every grid point and scores it by `max(p(+d), p(-d))`, so the published limit describes an effect detectable whichever way it points. Taking the minimum would publish the easier direction as the test's resolution — the one-sided error in a new costume.
- The resolvability floor moves from `1/(n+1)` to `2/(draws+1)` and raises loudly below it. This is not cosmetic: the binding bar is the 30-hypothesis pools' `alpha/30`, not the confirmatory family's `alpha`, and doubling the tail doubles the draws needed to reach it. A draw count that sufficed for #1424 fails loudly here, which is what stops an underpowered run publishing "no power" when the truth is "no permutations".
- The validity gate compares the MAGNITUDE of the separation. That is legitimate ONLY because the null and injection underneath it are now symmetric, and the module and report both say a future edit making any p one-sided again must restore the signed comparison in the same change. The gate stays ROW-MATCHED on the confirmatory family and the sign is preserved and reported everywhere. `MODE_REVERSED` ceases to exist, because a symmetric test has no untested direction.

**The two decisions the issue required, settled and stated up front**

1. **Cohort — Option 2, exploratory-only over the full pool.** Option 1 (reserve rows #1424 never scored) is not available in this repository: #1424's windows already run open-ended through the tape at run time, its committed feasibility probes record every other reachable venue as INFEASIBLE, and effective N is set by independent calendar clusters rather than rows, so any strict subset carries a strictly WORSE limit. Option 1 would buy a pre-registration guaranteed to be inconclusive. The cost is enforced mechanically rather than promised: no `VERDICT_CONFIG` constant exists in the module, `decide_recommendation` has no branch that promotes a configuration, and every verdict string carries the exploratory-only sentence.
2. **Contract path — #1426 DEFERS.** An exploratory-only study cannot be the live evidence behind a shipping gate, so `hurst_1424_gate_resolution.py` keeps `hurst_gate_calibration.md`. `main` refuses that path unconditionally and BEFORE every other check, so no combination of flags routes around it. **The supersede clause remains available to #1427 and #1428, and at most one of them may claim it.**

**One disclosed exception to two-sidedness.** The joint ADX x Hurst Stage 0 verdict is inherited from #1422 verbatim and stays ONE-SIDED on net return, so the verdict recorded against #1412 stays comparable across studies. It sits outside the confirmatory path and a test pins that nothing in the config sweep, the detection limits, the gate or the verdict can reach it.

## Result

**Verdict: `inconclusive`.** The two-sided confirmatory cluster p is **0.4878** against a rank-1 bar of 0.05. The validity gate FAILED: on `momentum` (8,009 row-matched rows) the two-sided limit is **0.013** and the separation is **-0.0046**, below it. No winner in either family, `n_passing` zero.

This closely reproduces #1424 on a tape about eleven days longer (its limit 0.013, separation -0.0051, 7,992 rows), and the two-sided limit came out equal to rather than above the one-sided one, which the coarse grid permits.

The honest reading, which the report states mechanically: the symmetric test removed the BLIND SPOT — the reversed direction is now testable at all, and it tested as not significant — but it did not add POWER, so the run carries **no bound in either direction**. Nothing ships. `#1411`'s `hurst_gate` stays default-off and `config.example.json` carries no `hurst_gate` block.

## Verification

- 106 tests in the new module pass with **zero skips**, including every committed-artifact pin.
- The blind spot is demonstrated directly: on the same reversed synthetic rows the inherited one-sided cluster null returns p = 1.0 while the two-sided one returns 0.001.
- `--render-only` reproduces the committed report **byte-for-byte**.
- The committed `decision` equals `decision_payload(decide_recommendation(configs, mde))`, pinned as #1424 does.
- All **394** existing #1410/#1422/#1424 tests pass unchanged; the full pytest suite passes at **3,464**; safe under CI's `pytest -n auto`; the Go build is unaffected.
- Diff audit: the contract report, all three predecessor scripts, their tests, `scheduler/` and `config.example.json` are byte-unchanged.

## Deviations from the adopted plan

The [Fable 5 implementation plan](https://github.com/richkuo/go-trader/issues/1426#issuecomment-5378370725) was followed step for step. Three deviations, each driven by evidence rather than preference:

1. **Added a `docs/backtesting-registry.md` row.** The plan specified "zero edits to any existing file", but `test_every_harness_has_exactly_one_row` fails without a row, and CLAUDE.md requires one in the same PR that adds a harness. The repo's own rule and its own test outrank the plan's file-count goal. A test now pins that the row records both the two-sidedness and the deferral.
2. **Corrected the resolvability sentence in the report.** My first draft said the draw count sat "comfortably under the rank-1 bar of 0.05". That cites the wrong bar: the binding constraint is the 30-hypothesis pools' `alpha/30 = 0.00167`, where 2,000 draws give `2/2001 = 0.00100` — a 1.67x margin, not a comfortable one. Two tests now pin the binding bar.
3. **Rewrote `KEY_RISK_PREDICTION`.** My first draft predicted `INCONCLUSIVE` **and** a both-directions bound. Those cannot co-occur: a bound requires the validity gate to pass, which yields `resolved_null`. The corrected text says plainly that a failed gate carries no bound and that a symmetric test removes the blind spot without adding power. Three tests now pin that only a passing gate ever claims a bound. The sweep was re-run so the committed JSON and the module constant agree rather than hand-editing the artifact.

The plan's step 12 also called for a coordination comment on the issue; it is posted there so #1427 and #1428 can rely on both settled decisions.

## Note on the data

The OHLCV cache in this checkout only reached back to 2023, so every dataset was backfilled from the venues first (Bitstamp to 2012, Coinbase Exchange to 2015, Binance.US to 2020) using #1424's own `--fetch-only` procedure with identical datasets and history floors. The committed numbers therefore come from a freshly fetched pool rather than the exact tape #1424 scored. The design, not the row set, is what was frozen, and the eleven-day drift is visible in the row counts above.

---


richkuo:
@claude review

Issue #1315, opened 2026-07-10
[C45] Fee model doesn't match Hyperliquid reality — stale taker rate, BinanceUS audit fees, no maker rate


richkuo:
**Complexity: 45/100** — scope: medium (parity constants, the M-series audit fee model, maker constant, and re-screen work on prior verdicts); risk: medium (fee modeling is the live fallback when real fill fees can't be fetched, and changing the audit model invalidates recorded research baselines); uncertainty: low (fee schedule verified against Hyperliquid docs)

## Problem

Fee modeling is wrong in three ways, all overstating what we actually pay on Hyperliquid:

1. **The live/backtest Hyperliquid taker constant is stale.** `scheduler/fees.go` (`HyperliquidTakerFeePct = 0.00035`) and `backtest/backtester.py` (`PLATFORM_FEE_PCT["hyperliquid"] = 0.00035`) both say 0.035%. Hyperliquid's current base-tier perps taker fee is **0.045%** (maker 0.015%), per the official fee schedule (https://hyperliquid.gitbook.io/hyperliquid-docs/trading/fees). Real costs exceed the modeled fee by ~29% wherever the real fill fee can't be fetched (`hyperliquid_fills.go` fallback, `backfill_hl_fees.go`, `backfill_trade_ledger.go`) and in every backtest that uses the hyperliquid platform fee.

2. **The M-series research harnesses don't use Hyperliquid fees at all.** `backtest/eval_windows.py` pins `PLATFORM = "binanceus"` (0.1%/side + 5 bps slippage, ~0.3% round-trip) as the audit fee model, inherited by `fee_audit.py` (M5), `gross_edge_noise.py`, and `auto_suggest.py`. Live trading runs on Hyperliquid perps, whose real round-trip cost at base tier is ~0.09% + slippage — the audits charge more than 2× the real fee. Strategies have been deprecated for being "fee-consumed" under a fee model we don't actually pay: #1282 retired `tema_cross` and `regime_adaptive` longs with real gross edges (~0.3–0.4%/trade, permutation p ≤ 0.0002) that were net-negative only under BinanceUS-level fee drag. *Caveat: those were spot-audited legs, so the BinanceUS model was a defensible spot-deployment assumption at the time — the re-screen below makes the perps-deployment assumption explicit rather than claiming the old grading was an error.*

[...]

PR #1453, opened 2026-08-22
feat(#1443): screen HYPE from Hyperliquid and harden the cert family [C62, Opus 5, high, fableplan]

### LINKED ISSUE #1443: [C62] Add HYPE to the directional certification screen and rerun the full screen so configured directional policies can fire
**Complexity: 62/100** — Capability 2 (the emitted artifact can flip live trade direction — money path; plus statistical design judgment on the correction family); Volume 12 (Scope 2, Coupling 2, Verification 2) — Opus 5, high · fableplan: yes

## Problem

`regime_directional_policy` is evidence-gated default-OFF (#1085). The artifact `backtest/research/regime_directional_certifications.json` ships with `"certified": []` — every configured policy block resolves to base direction (fail-closed loader in `scheduler/regime_directional_certification.go`). Two coupled gaps keep it from ever firing:

1. **HYPE is not in the cert universe.** The cert screen loads OHLCV with `exchange_id=PLATFORM` where `PLATFORM = "binanceus"` is hardcoded (`backtest/eval_windows.py:100`, consumed at `backtest/research/regime_1076_directional_premise.py:103`). HYPE does not trade on Binance US; it trades as an HL perp (ccxt market `HYPE/USDC:USDC`). The deployment config (out of tree, `/var/lib/go-trader/config.json`) has `regime_directional_policy` configured on `hl-mr-hype-60` (added 2026-08-20), `hl-vwap-eth-60`, and `hl-rmc-eth-live`; the HYPE cell is unreachable by the producer today.

2. **The cert screen has been empirically negative on every reachable cell.** Per the docstring on `regime_1076_certify.py`: "Under the current universe NOTHING survives global correction (#1076: 0/2121)". `(ETH, 1h, composite)` is inside the default screened universe (`DEFAULT_SYMBOLS`/`DEFAULT_TIMEFRAMES`/`DEFAULT_CLASSIFIERS`) and does not pass global Benjamini-Hochberg (BH) at FDR=0.05, so ETH 1h stays inert.

The question this issue answers: can a non-empty cert artifact ever ship, and what is the honest way to test that?

## Goal

Reach a position where the certification question is answered for both target cells, with the statistical guarantees of #1085 intact:

1. **(HYPE, 1h, composite)** — requires plumbing HL-sourced HYPE OHLCV into the cert data loader.
2. **(ETH, 1h, composite)** — already screened; either the cell has no edge (then inert is the correct answer) or the negative verdict is re-confirmed on the refreshed run.

If a cell certifies, the live policy fires after the regenerated artifact is deployed and the daemon reloads it (the artifact is read at startup and on SIGHUP — `scheduler/main.go:128,798`). No live strategy-config change is required.

### 1. Per-symbol exchange mapping in the cert data loader

`load_cached_data` (`shared_tools/data_fetcher.py:173`) already accepts any `exchange_id` and falls back to a ccxt fetch that caches into the storage layer. The gap is only that the premise script pins every symbol to `PLATFORM`. Extend the `--symbols` contract with an optional per-symbol exchange (for example `HYPE/USDC:USDC@hyperliquid`); a bare symbol keeps the `PLATFORM` default. Constraints:

- **Do not repoint `PLATFORM` or cache HL data under the `binanceus` namespace.** The `eval_windows.py` axis-separation comment (#1315) pins `PLATFORM = "binanceus"` because every committed regime baseline was computed on that series; HL rows are cached under `exchange_id="hyperliquid"`.
- **No fee flag.** The premise screen is a gross forward-return permutation test; it prices no fees, and `FEE_PLATFORM` is already `"hyperliquid"` (#1315) for the harnesses that do. There is nothing to flip.
- The artifact key is exchange-agnostic: `normalize_cert_asset` reduces `HYPE/USDC:USDC` to `HYPE`, matching the Go side.

### 2. Re-run the producer over the FULL family — never a narrowed one

`certify()` applies global BH over only the rows of the current invocation (`regime_1076_certify.py:89-91`). A run restricted to `--symbols HYPE,ETH --timeframes 1h --classifiers composite` would shrink the correction family from the full 2121-row screen to a handful of rows, which inflates the pass probability — the same pooled-limit trap #1424 documents for the Hurst gate. The run therefore MUST cover the full default universe (BTC/ETH/SOL × 1h/4h × adx/composite × all windows) PLUS the HYPE rows in ONE invocation, so the BH family is a superset of the original screen.

Window coverage for HYPE (HL listing Nov 2024): `2025H1`, `is`, and `oos` are fully inside its history; `2024` is partial and `2023` is empty. `_load` already drops windows with too few bars; the run output must state which windows contributed HYPE rows.

Harden the artifact against this trap: record the screened family (row count and per-symbol data source) in the artifact `criteria`, so a narrowed-run artifact is detectable by inspection.

If both target cells still fail global BH, the artifact stays empty and the policy stays inert — that is the expected outcome, and the resolution is "verified: no cert exists, policy correctly inert".

### 3. If a cell certifies, validate the live flip

- Confirm the regenerated artifact carries `default_ttl_days` (currently 90) and parses through `LoadDirectionalCertSetFailClosed` without falling back to the empty set.
- Deploy the artifact and reload (restart or SIGHUP).
- Open a position under a certified regime and verify `Position.DirectionCertifiedStatesAtOpen` carries the cert map, and that the policy flips the side (for example short in `trending_down_clean`).

### 4. Decide what to do with already-configured policies

If no cert appears, the configured `regime_directional_policy` blocks on the three deployment strategies are decorative. Two paths: remove the blocks (cleaner config), or keep them with a comment (preserves the intent for the day a cert arrives). Decide and record the choice.

### 5. Boundaries

- The FDR=0.05 global-BH gate is the #1085 evidence bar and stays as-is. Relaxing the threshold to let a cell pass would defeat the gate's purpose; any proposal to change the bar is a separate issue with its own justification.
- Whether HYPE-HL fetches become a general storage-layer fixture or stay a cert-run cache entry is documented in the run README either way.

## Acceptance criteria

- The premise/certify scripts accept a per-symbol exchange mapping and run end-to-end on `HYPE/USDC:USDC@hyperliquid` alongside the full default universe in one invocation.
- The regenerated artifact's `criteria` records the screened family size and per-symbol data sources.
- A run report (premise `--out` rows or the research README) states the per-cell verdict — certified or failed, with the failing criterion — for (ETH, 1h, composite) and (HYPE, 1h, composite). The artifact schema itself lists only certified cells and stays at `schema_version: 1`.
- If a cell certifies: a live open under that cell's regime populates `DirectionCertifiedStatesAtOpen` as expected after deploy + reload.
- If no cell certifies: the #1085 `[WARN]` is documented as expected for these strategies, and the config-block decision (keep-with-comment or remove) is made and recorded.

## Related

- #1085 — the evidence-gate default-OFF
- #1076 — the empirical "regime doesn't predict direction" premise study
- #1087 — companion to #1085
- #1025 — backtest parity for the resolver
- #1166 — first concrete cert-requiring study (momentum_pro directional gate), M1 negative result
- #1315 — fee-model/data-source axis separation (`PLATFORM` vs `FEE_PLATFORM`)
- #1424 — the pooled-limit lesson the run design must respect
- `backtest/research/regime_1076_certify.py` — the SSoT producer
- `backtest/research/regime_directional_certifications.json` — the empty artifact

## Plain simple English

Some live strategies are configured to trade short when the market trends down. A safety gate keeps this off until statistics prove the edge is real. The proof tool cannot read data for one of the coins, so the gate can never open for it. This work connects that data and reruns the proof correctly.

---

richkuo:
## Plain simple English

Some live strategies can trade short when the market trends down. A safety gate keeps this off until statistics prove the edge is real. The proof tool could not read data for one coin, HYPE, so the gate could never open for it. This work connects that data and reruns the proof. The result is negative: no edge, gate stays closed.

## Summary

Closes #1443.

The certification screen loaded every symbol from one exchange (`eval_windows.PLATFORM`, `binanceus`), so HYPE — a Hyperliquid perp — was unreachable and the `regime_directional_policy` configured on `hl-mr-hype-60` could never become evidence-backed. `--symbols` now accepts an optional per-symbol data source, `SYMBOL[@exchange]`.

`PLATFORM` itself is never repointed: mapped rows cache under their own `exchange_id` namespace, so the #1315 axis separation and every committed `binanceus` baseline are untouched. Verified: zero HYPE rows exist under the `binanceus` namespace after the run.

### Three hardenings

1. **Window clipping in `_load`.** `load_cached_data`'s empty-cache fallback fetches from `since=start_date` and returns the whole history **unsliced**, so an asset listed after a window would have later data screened under the earlier window's label. The clip is a no-op on the cached path, so every already-screened cell is unchanged. This is the one place the run could have gone silently wrong.

2. **Family-integrity refusal.** `certify()` corrects across the rows of ONE invocation, so the BH family *is* the run's universe. A narrowed run shrinks it and inflates every cell's pass probability — the pooled-limit trap #1424 records. Writing the repo artifact from a non-superset universe is now refused **before any data is touched**. `--allow-narrowed-family` overrides for research artifacts written elsewhere.

3. **Provenance and per-cell verdicts.** The artifact `criteria` records the actual screened family size (computed from the rows, never claimed), the per-symbol data sources, the universe axes, `n_perm` and the permutation p-floor. A new run report gives every screened cell its **first failing criterion**, best p-value, and the BH critical value it needed. All new metadata nests inside `criteria` because the Go loader parses with `DisallowUnknownFields` — a new top-level key would make the live daemon fail closed on a valid artifact. `schema_version` stays 1; no production Go change.

## Result of the regenerated screen

One invocation over the full default universe plus HYPE, `--n-perm 30000 --seed 0` (per the maintainer sign-off on the plan's decision point 1):

**0 of 16 screened cells certify. Every cell fails at the global-BH gate.**

The resolution caveat does **not** apply here. The p-floor `1/(n_perm+1)` = **3.333e-05** sits *below* the rank-1 BH critical value `q/m` = **3.791e-05** over the 1319-row family, so a single genuinely-separating state was arithmetically able to pass. It did not. At the previous default of `n_perm=500` the floor would have sat ~53× above that bar, where no single row could certify at any effect size.

| Target cell | Verdict | best p | bar it needed | best row | sign |
|---|---|---|---|---|---|
| (ETH, 1h, composite) | `fails_global_bh` | 0.002167 | 1.137e-04 (rank 3/1319) | `trending_down_clean` @ `2025H1` h1, gap +0.00923 | wrong-signed |
| (HYPE, 1h, composite) | `fails_global_bh` | 0.072631 | 5.004e-03 (rank 132/1319) | `trending_up_clean` @ `oos` h48, gap −0.06573 | wrong-signed |

Both miss by more than an order of magnitude and are additionally wrong-signed, so a relaxed bar would not yield a profitable certification either. Closest cell overall was (HYPE, 4h, adx) at p=7.000e-04 against a 7.582e-05 bar — still ~9× short, also wrong-signed.

`certified` stays `[]`, so the #1085 `[WARN]` printed by `hl-mr-hype-60`, `hl-vwap-eth-60` and `hl-rmc-eth-live` is the **expected, verified** outcome.

### HYPE data-availability limit (issue decision point 2)

Hyperliquid's candle endpoint serves a fixed **5000-candle retention window per interval, anchored at now**; `since` is ignored beyond it. Verified at fetch time — requests with `since` at 2023-01-01, 2024-01-01, 2025-01-01 and 2025-06-10 all returned the same 1h start.

| HYPE window | 1h | 4h |
|---|---|---|
| `2023` | empty | empty |
| `2024` | empty | 113 composite bars (Dec only) |
| `2025H1` | empty | 1039 |
| `is` | **empty** | 1183 |
| `oos` | 4955 (from 2026-01-26) | 1353 |

(HYPE, 1h, composite) therefore rests entirely on `oos`. That is a held-out-forward window, so the cell was still eligible to certify. **No other venue's HYPE series was substituted**, per the issue boundary — the shortfall is recorded in the coverage table and run report instead.

## Verification

- `go -C scheduler test . -count=1` → `ok` (full package); `gofmt -l scheduler/` clean
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -q -n auto` → **3409 passed, 1 skipped**
- The clip regression test was proven **red → green**: against the unfixed `_load`, a HYPE `2023` request labels 7152 post-listing bars as 2023, and both clip tests fail
- Narrowed run to the default `--out` exits 1 before touching data, leaving the artifact unmodified
- Regenerated artifact parses through `parseDirectionalCertSet` (Go) and `load_certifications` (Python) with no warning; index is empty because `certified` is empty
- `sqlite3` check: 0 HYPE rows under the `binanceus` namespace

## Adopted plan

Built to the Fable 5 plan posted on the issue (https://github.com/richkuo/go-trader/issues/1443#issuecomment-5379028890) and the maintainer's follow-up sign-off fixing `--n-perm 30000` (https://github.com/richkuo/go-trader/issues/1443#issuecomment-5379041270). Deviations, both additive:

1. **`family_is_superset` also checks `horizons`.** The plan named four axes (symbols/timeframes/windows/classifiers). Dropping horizons shrinks the BH family exactly as dropping symbols does, so leaving it unchecked would leave the guard bypassable by the narrowing it exists to prevent. The parameter is optional, so the plan's four-axis call shape still works.
2. **`cell_verdicts` keys rows by family position, not `id()`.** The plan did not specify the lookup mechanism; identity keying misbehaves if a caller passes the same row object twice, which would attach a cell's reported bar to a different hypothesis. Regression-tested.

Also added beyond the plan's literal text: an explicit branch for a zero-row family (a coverage failure, which must not print as a negative result), and the empty-window enumeration reads the *requested* axes rather than the axes present in the results.

## Follow-on work — unfiled

- Removing or keeping the `regime_directional_policy` blocks on the three deployment strategies is recorded as **keep-with-comment** in `README_1076_directional_premise.md`, with the reasoning. That config lives out of tree (`/var/lib/go-trader/config.json`) and is an operator action, not a repo change.
- Live-flip validation (deploy + SIGHUP + `DirectionCertifiedStatesAtOpen`) is moot while nothing certifies; it stays an operator runbook item for the day a cert appears.

---

### DIFF
--- a/backtest/research/README_1076_directional_premise.md
+++ b/backtest/research/README_1076_directional_premise.md
@@ -116,6 +116,91 @@ using `labels[t]` to trade the move *into* bar `t` — produced impossible Sharp
 committed `_book` decides the side at bar `t` and holds it over `t→t+1`, verified by a
 buy-and-hold book reproducing the asset return exactly.)
 
+## Refreshed run — 2026-08-22, HYPE added (#1443)
+
+The #1085 producer was re-run over the full default universe **plus** HYPE sourced
+from Hyperliquid, in **one invocation**, so the Benjamini-Hochberg family is a
+superset of the previous certify screen. Command:
+
+```bash
+uv run --no-sync python backtest/research/regime_1076_certify.py \
+    --symbols "BTC/USDT,ETH/USDT,SOL/USDT,HYPE/USDC:USDC@hyperliquid" \
+    --n-perm 30000 --seed 0
+```
+
+Full per-cell output: `regime_1443_run_report.json`. Artifact:
+`regime_directional_certifications.json`.
+
+### Verdict — still 0 certified, and this time the arithmetic permits one
+
+**0 of 16 screened cells certify. Every cell fails at the first gate, global BH.**
+
+The important difference from earlier runs is the *resolution*.
+`per_state_significance` reports `p = (ge+1)/(n_perm+1)`, so its smallest possible
+p-value is `1/(n_perm+1)`. At `n_perm=30000` that floor is **3.333e-05**, and the
+rank-1 global-BH critical value over the 1319-row directional family is
+`q/m = 0.05/1319 =` **3.791e-05**. The floor sits *below* the bar, so a single
+genuinely-separating state was arithmetically able to certify. It did not.
+
+This run is therefore a real negative at the tested resolution, and not the
+"could never have passed anyway" artifact a lower `n_perm` would have produced —
+at the previous default of 500, the floor (1.996e-03) sat ~53× **above** the
+rank-1 bar, so no single row could have certified at any effect size.
+
+### The two target cells
+
+| Cell | Verdict | best p | BH bar it needed | Best row | Sign |
+|---|---|---|---|---|---|
+| **(ETH, 1h, composite)** | `fails_global_bh` | 0.002167 | 1.137e-04 (rank 3/1319) | `trending_down_clean` @ `2025H1`, h1, gap **+0.00923** | wrong-signed |
+| **(HYPE, 1h, composite)** | `fails_global_bh` | 0.072631 | 5.004e-03 (rank 132/1319) | `trending_up_clean` @ `oos`, h48, gap **−0.06573** | wrong-signed |
+
+ETH's best row misses its bar by ~19×; HYPE's by ~15×. Both are additionally
+**wrong-signed** — their separation points *against* the side the policy would
+bet — so even a relaxed significance bar would not produce a profitable
+certification for either.
+
+The closest cell in the whole screen was `(HYPE, 4h, adx)` at p=7.000e-04 against
+a rank-2 bar of 7.582e-05 — still ~9× short, and also wrong-signed.
+
+### HYPE window coverage — a hard data-availability limit
+
+Hyperliquid's candle endpoint serves a fixed **5000-candle retention window per
+interval, anchored at *now***; `since` is ignored for anything older. Verified at
+fetch time (repeated requests with `since` at 2023-01-01, 2024-01-01, 2025-01-01
+and 2025-06-10 all returned the same 1h start). The consequence:
+
+| HYPE window | 1h | 4h |
+|---|---|---|
+| `2023` | empty | empty |
+| `2024` | empty | 113 composite bars (Dec only) |
+| `2025H1` | empty | 1039 |
+| `is` | **empty** | 1183 |
+| `oos` | 4955 (from 2026-01-26) | 1353 |
+
+So **(HYPE, 1h, composite) rests entirely on `oos`**, and its `oos` starts
+2026-01-26 rather than the window's nominal 2026-01-01. `oos` is a held-out
+forward window, so the cell was still eligible to certify; it simply is not close.
+Per the issue boundary, no other venue's HYPE series was substituted.
+
+`_load` clips every loaded frame to its eval window before labeling. Without that
+clip, HYPE's empty windows would have ingested the fetch fallback's post-listing
+history under the earlier window's label — a silent statistical corruption, and
+the one place this run could have gone quietly wrong.
+
+### Consequence for the deployed strategies
+
+`certified` stays `[]`, so `regime_directional_policy` resolves to base direction
+on `hl-mr-hype-60`, `hl-vwap-eth-60` and `hl-rmc-eth-live`. The #1085 `[WARN]`
+those three print at config load is the **expected, verified** outcome, not a
+defect — the gate is doing exactly what it exists to do.
+
+The configured policy blocks are left in place (the `keep-with-comment` option in
+the issue's Approach 4) rather than removed. Removing them buys tidier config but
+loses the recorded intent, and re-adding a block is blocked by SIGHUP while a
+position is open — so the safe moment to remove one is from flat, at the
+operator's choosing. That deployment config lives out of tree
+(`/var/lib/go-trader/config.json`); this repo cannot and should not edit it.
+
 ## Action taken (#1076 scope 3)
 
 The premise holds **nowhere** in the tested universe, so the directional-gating surface must
--- a/backtest/research/README_1085_certification.md
+++ b/backtest/research/README_1085_certification.md
@@ -13,6 +13,20 @@ gate proves real, multiplicity-honest directional edge.
 Nothing in the tested universe currently qualifies, so the shipped artifact
 certifies nothing and every `regime_directional_policy` runs default-off.
 
+**Latest run: 2026-08-22 (#1443)** — full default universe plus HYPE sourced from
+Hyperliquid, one invocation, `n_perm=30000`, `seed=0`. 0 of 16 screened cells
+certify; every cell fails at the global-BH gate. At this permutation count the
+p-floor (3.333e-05) sits *below* the rank-1 BH critical value (3.791e-05) over the
+1319-row family, so a single genuinely-separating state was arithmetically able to
+pass — it is a real negative, not a resolution artifact. Per-cell verdicts, the
+HYPE window-coverage limit and the two target cells are written up in
+`README_1076_directional_premise.md` § "Refreshed run — 2026-08-22"; the machine
+-readable form is `regime_1443_run_report.json`.
+
+The `[WARN]` the three configured deployment strategies print at config load is
+therefore the **expected, verified** outcome. The configured policy blocks are
+kept with this record rather than removed; see the same section for why.
+
 ## Single source of truth
 
 The statistical test lives in ONE place — the Python research harness. Go never
@@ -46,9 +60,21 @@ direction only when a directional state for that direction:
 ```json
 {
   "schema_version": 1,
-  "generated_at": "2026-06-19T00:00:00Z",
+  "generated_at": "2026-08-22T00:00:00Z",
   "generator": "backtest/research/regime_1076_certify.py",
-  "criteria": { "global_correction": "benjamini-hochberg", "fdr_q": 0.05, ... },
+  "criteria": {
+    "global_correction": "benjamini-hochberg",
+    "fdr_q": 0.05,
+    "require_sign_aligned": true,
+    "require_held_out_forward": true,
+    "held_out_windows": ["is", "oos"],
+    "screened_family_size": 1319,
+    "n_perm": 30000,
+    "permutation_p_floor": 3.3332e-05,
+    "data_sources": { "BTC/USDT": "binanceus", "HYPE/USDC:USDC": "hyperliquid" },
+    "universe": { "symbols": [...], "timeframes": [...], "windows": [...],
+                  "classifiers": [...], "horizons": [...] }
+  },
   "default_ttl_days": 90,
   "certified": [
     {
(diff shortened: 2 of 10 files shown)


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T10:15:56Z) in 0ef58b5b. All five findings validated against current code first; all five held, all five implemented.

### Fixed

1. **Family-integrity refusal ignored the data source** — confirmed: `_bare_symbols` strips `@exchange` before the superset check, so `BTC/USDT@kraken,ETH/USDT,SOL/USDT` passed and wrote the live-consumed artifact. `family_is_superset` is unchanged (its bare-symbol comparison is correct for width, and `test_family_is_superset_ignores_the_exchange_suffix` still pins it); the source check is a separate gate, `baseline_source_violations` (`backtest/research/regime_1076_certify.py:124`), wired at `:518`. Every `DEFAULT_SYMBOLS` entry must resolve to `eval_windows.PLATFORM`; added symbols carry any venue. All three must-survive cases covered — explicit `@binanceus` allowed, added `@hyperliquid` allowed, two repointed defaults both named in one refusal.

2. **Degenerate runs overwrote the live artifact and exited 0** — confirmed: the write sat at `:428`, ahead of the zero-row and p-floor diagnostics, and returned 0 either way. `certify()` is pure, so the write now happens after the diagnostics and after a degenerate-run gate (`:597`); a refusal leaves the existing file byte-identical (pinned by a sentinel-file test). `--n-perm` default raised 500 → 30000 (`:398`). Override is a **separate** `--allow-degenerate-run` flag, not the existing one: narrowing a research family must not also unlock erasing the live artifact from a run that measured nothing. A full-width, well-resolved run that legitimately certifies nothing still writes and exits 0.

3. **`--report-out` was outside the refusal** — confirmed: `:408` inspected only `args.out` while `:482` wrote the committed run report unconditionally. Both guards now compute a `repo_targets` list covering `--out` *and* `--report-out` (`:503`), so a narrowed run that redirects only `--out` is refused, naming the report. `--report-out ""` still skips cleanly. The README research example redirects both.
   One deviation to flag: this makes finding 2's must-survive (c) — narrowed, non-default `--out`, default `--report-out` — refuse rather than write. Finding 3's invariant ("every repo-tracked evidence file is subject to the same refusal") is explicit and its own suggested fix says to add `--report-out` to that example, so the two are reconciled in favor of the stricter reading. Redirecting both, or `--report-out ""`, writes without refusing.

4. **Two symbols normalizing to one certified cell** — confirmed: `certify()` keys by `normalize_cert_asset`, and `resolve_data_sources` keeps only the last exchange for a repeated symbol string. Split across the two layers that own each problem: `premise.parse_symbols_arg` refuses a repeated symbol string (every symbol-keyed surface downstream is a dict, so it hurts the screen too), and `cert_asset_collisions` (`regime_1076_certify.py:140`) refuses two entries sharing a normalized asset, before any data access. `HYPE/USDC:USDC@hyperliquid` alongside the three defaults stays accepted.

5. **`bh_threshold` could contradict its own verdict** — confirmed against `backtest/regime_stats.py:69-74`: BH is step-up, rejecting every `p <= p_(k_max)`, so a row missing its per-rank bar can still be rejected. Rather than rename the field, it now reports the bar the verdict was actually decided against. `bh_step_up_cutoff` = `q*k_max/m` (`bh_step_up_cutoff`, `:180`); that value is exactly equivalent to the procedure for every row in the family, since no p-value can lie in `(p_(k_max), q*k_max/m]` — one there would satisfy the rank-`k_max+1` bar and contradict `k_max`'s maximality. Entries carry `bh_rank_threshold` (per-rank), `bh_step_up_cutoff` (`null` when nothing was rejected) and a decision-consistent `bh_threshold`; the printed line says which of the two it is. With nothing rejected the per-rank value stays, as the must-survive case requires.
   The committed `regime_1443_run_report.json` is backfilled with the two new fields rather than regenerated. That is derivation, not new measurement: every one of its 16 cells recorded `n_survive_global_bh: 0`, so the family rejected nothing, the cutoff is `null`, and every existing number is unchanged.

### Verification

`uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3433 passed, 1 skipped. `go build ./...` and `go test ./...` clean.

One existing test changed intent rather than being relaxed: `test_repo_artifact_run_over_the_full_family_is_allowed` stubbed the screen to return zero rows, which is now precisely the degenerate case finding 2 asks us to refuse. It now returns a directional row that fails the gate, so it pins the stronger property the finding actually calls for — a full-width run that legitimately certifies nothing must still write and exit 0. The zero-row case gets its own refusal test.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T10:39:16Z) in 3be38d3a. LGTM with one Recommended Optional; validated against current code first, and it held.

### Fixed

1. **`best_row` could contradict its own verdict** — confirmed: `best_i = min(idxs, key=lambda i: pvals[i])` picked the cell's globally lowest-p directional row before any of the gates ran, while the verdict came from the nested `passing` → `aligned` → `held` filters. The two are now derived from one ladder: the verdict names the deepest tier the cell reached, and the displayed row is that tier's lowest-p member (`backtest/research/regime_1076_certify.py:290-340`). A new `best_row_basis` field names the tier, so the diagnostic states which set of rows its evidence came from.
   All three must-survive cases pinned: (a) a wrong-signed minimum next to an aligned held-out row — CERTIFIED displays the aligned one; (b) a historical-window minimum next to a held-out row — CERTIFIED displays the held-out one; (c) single-row-per-cell prints that row unchanged at every verdict.

   **Bug-class expansion.** The same contradiction was reachable one rung down, which the finding did not name: `not_held_out_forward` asserts the cell survived BH *and* alignment, yet the displayed minimum could be wrong-signed. The tier ladder covers it, and it has its own test. `wrong_signed` and `fails_global_bh` cannot diverge — BH step-up rejects every p at or below the largest survivor, so a cell's minimum is in `passing` whenever anything in that cell is.

   **`min_p_value` and the BH fields deliberately stay anchored to the cell minimum.** The name is literal, and for a failing cell that value is the honest "how far short did it fall" — the field the previous review round was about. Since the displayed row is no longer always that row, `_format_verdict_line` now prints the displayed row's own `p=` and its `basis=`, but only when they differ from the minimum, so today's output is byte-identical.

   Committed `regime_1443_run_report.json` is **backfilled** with `best_row_basis`, not regenerated. All 16 cells recorded `n_survive_global_bh: 0`, so every one sits in the all-directional tier and no displayed row, p-value, or threshold moves — the diff is 16 pure insertions.

### Verification

`uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3441 passed, 1 skipped (was 3433; +8 new tests). `go -C scheduler build ./...` and `go -C scheduler test ./...` clean. No existing test relaxed or deleted.

---

richkuo:
@claude sonnet review

Issue #1455, opened 2026-08-22
[C89] Circuit-breaker closes keep the estimated profit and loss instead of the real fill


richkuo:
**Complexity: 89/100** — Capability 3 (Risk 4 — realized profit and loss plus cash accounting on the circuit-breaker path); Volume 14 (Scope 2, Coupling 2, Verification 3) — Opus 5, xhigh · fableplan: yes

Child of #1451 (umbrella). Independent of the sibling manual-strategy kill-switch part.

## Problem

When a per-position circuit breaker fires on a live Hyperliquid perps strategy, the recorded `realized_pnl` is a mark-derived model estimate and it is never corrected to the real exchange fill.

**Scope: sole-owner (and hedge) coins only.** `shouldForceCloseAllPositionsOnCircuitBreaker` is `!hyperliquidCircuitBreakerHasSharedCoin` (`scheduler/risk.go:1758-1759`), and `setHyperliquidCircuitBreakerPending` returns early on a shared coin (`scheduler/risk.go:1719-1720`). A strategy sharing its coin with a peer takes the operator-required path and writes no model-only row from this arm.

The ordering is the cause. `CheckRisk` calls `forceCloseAllPositions` at fire time (`scheduler/risk.go:2308` for the drawdown arm, `scheduler/risk.go:2332` for the loss-streak arm), which is before the reduce-only close is submitted, let alone filled. `forceCloseAllPositions` values the position at the cycle mark, writes a close row labelled `model-only reconciliation adjustment; no exchange fill` with `FeeSourceReconcileAdjustment` and no `ExchangeOrderID` (`scheduler/risk.go:2035`, `scheduler/risk.go:2046-2056`), books the estimated profit or loss into cash, and deletes the virtual position.

When the real fill arrives on a later cycle, `applyHyperliquidCircuitCloseFill` looks for that virtual position, finds it gone, and takes the defensive branch: it records a trade row with the real fill price, fee, and order ID but `RealizedPnL: 0`, because there is no `AvgCost` basis left to derive profit or loss from (`scheduler/hyperliquid_balance.go:2770-2802`).

The estimate is therefore permanent. It propagates into per-strategy statistics, the win/loss ledger, the daily profit-and-loss aggregate that feeds the daily loss limit, and the portfolio drawdown basis, and the cash balance carries a value the exchange never paid.

[...]

Issue #1451, opened 2026-08-22
[C93] Force-closed positions can record model estimates instead of real exchange fills


richkuo:
**Complexity: 93/100** — Capability 3 (Risk 4 — kill-switch / circuit-breaker money path, realized PnL and cash accounting); Volume 18 — Opus 5, xhigh · fableplan: yes

**Umbrella issue — tracking only.** Validation against `main` @ `135b97c5` confirmed the symptom and found three root causes plus one missing alert. The work is filed as two child issues:

- [ ] #1454 — [C91] The kill switch does not close or book fills for manual Hyperliquid positions (root cause 1; root cause 3 folds in here)
- [ ] #1455 — [C89] Circuit-breaker closes keep the estimated profit and loss instead of the real fill (root cause 2)
- [ ] Folded into whichever child lands first: throttled owner DM when a model-only close row books (the missing alert, below C41)

Close this umbrella when both children merge. Build against the children, not this issue.

## Problem

When the portfolio kill switch or a per-position circuit breaker force-closes positions, go-trader writes a "Circuit breaker close" trade row whose detail reads `model-only reconciliation adjustment; no exchange fill`. The row carries an empty `exchange_order_id`, a mark-derived close price, and a model-estimated `realized_pnl`.

`forceCloseAllPositions` (`scheduler/risk.go:1984`) is the sole production writer of that label (`scheduler/risk.go:2035`); those rows set no `ExchangeOrderID` and use `FeeSourceReconcileAdjustment` (`scheduler/risk.go:2046-2056`). On-chain closes stamp the real fill order ID instead (`scheduler/hyperliquid_balance.go:2757, 2824`), which is why the two row classes are distinguishable in the ledger.

### What validation refuted

Two of the original hypotheses do not hold for the **portfolio kill-switch** path:

[...]

PR #1457, opened 2026-08-24
fix(#1454): bring live manual positions inside the kill-switch close and fill chain [C91, ox-alpha, high]

### LINKED ISSUE #1454: [C91] The kill switch does not close or book fills for manual Hyperliquid positions
**Complexity: 91/100** — Capability 3 (Risk 4 — kill-switch money path: changes which on-chain positions get liquidated); Volume 16 (Scope 2, Coupling 3, Verification 3) — Opus 5, xhigh · fableplan: yes

Child of #1451 (umbrella). Independent of the sibling PnL-reconciliation part.

## Problem

A live Hyperliquid position held by a `type=manual` strategy is invisible to the portfolio kill switch, in both directions: the kill switch will not close it, and it will not book its fill if something else does.

`hlLiveAll` is built perps-only (`scheduler/main.go:1110-1112`). That is deliberate per #576 — reconcile lists extend to `type=manual`, but kill-switch, trailing-stop, and risk math stay perps-only. Three consequences follow:

- `forceCloseHyperliquidLive` builds `tradedCoins` from `hlLiveAll` plus held hedge coins (`scheduler/hyperliquid_balance.go:1988-2001`). A coin held **only** by a manual strategy fails the `tradedCoins` test and is skipped as "unowned" with no error, no report entry, and no latch (`scheduler/hyperliquid_balance.go:2006-2011`).
- `applyHyperliquidKillSwitchCloseFill` returns early unless `sc.Type == "perps"` (`scheduler/hyperliquid_balance.go:2190-2191`). A manual strategy therefore **always** falls through to `forceCloseAllPositions`, which writes a model-only reconciliation row (`scheduler/risk.go:2035`) with an empty `exchange_order_id` and a mark-derived `realized_pnl` — even when a real fill exists in the close report.
- The unconfigured-coin latch that would otherwise catch an unowned on-chain position only arms when there are **zero** live perps strategies (`scheduler/kill_switch_close.go:432`). A mixed fleet — the normal case — never reaches it.

Net effect: on a mixed fleet the kill switch can report `OnChainConfirmedFlat`, flatten the perps book, mark the manual positions closed in the database, and leave those manual positions open on Hyperliquid with no `[CRITICAL]` line and no latch. This is the hidden-exposure path.

### Live evidence

From the 2026-08-22 05:10:51 kill-switch incident (see #1451 for the full table): rows 1204 (`manual-eth`) and 1203 (`manual-btc`) both cleared with `exchange_order_id=''` and the `model-only reconciliation adjustment; no exchange fill` detail, alongside the perps row 1205.

## Goal

A live Hyperliquid position owned by a `type=manual` strategy must be inside the kill switch's close scope and its fill-booking scope. If the kill switch cannot close it, that must flip `OnChainConfirmedFlat` to false and latch, exactly as it does for a perps coin.

## Approach

Keep #576's separation where it is correct — trailing-stop and risk math stay perps-only — and widen only the kill-switch surface.

- Introduce a kill-switch roster that includes live `type=manual` strategies alongside live perps, built with `isHLLiveReconcilable` (`scheduler/main.go:4211`) exactly as `runPendingHyperliquidCircuitCloses` already builds `hlCircuitPeerAll` (`scheduler/hyperliquid_balance.go:2284-2293`). Leave `hlLiveAll` itself perps-only — trailing-stop and risk math depend on that meaning — but the new roster must **replace** `hlLiveAll` throughout the whole kill-switch fill chain, not only in `tradedCoins`:
  - `KillSwitchCloseInputs.HLLiveAll` (`scheduler/kill_switch_close.go:42`), which feeds `forceCloseHyperliquidLive`'s `tradedCoins` construction.
  - The fill-share peer set — `hlLiveStrategiesForCoin` (`scheduler/hyperliquid_balance.go:2050`) as read by `hyperliquidKillSwitchFillShare` (`scheduler/hyperliquid_balance.go:2149`).
  - The virtual-quantity snapshot `snapshotHyperliquidVirtualQuantities` (`scheduler/hyperliquid_balance.go:2084`, called from `scheduler/main.go:1640`).
  - The resting SL/TP trigger-OID cancel snapshot (`scheduler/main.go:1614-1628`).

  Key coins with `hyperliquidConfiguredCoin` (`scheduler/hyperliquid_balance.go:2067-2078`), which resolves `sc.Symbol` for `type=manual` — the #1444 precedent — rather than `hyperliquidSymbol(sc.Args)`, which works for manual only because `scheduler/config.go` auto-fills `Args`.

  **Partial widening is worse than none.** Relaxing the type gate while leaving these surfaces perps-only produces a new mis-booking rather than a fix: a manual strategy missing from `virtualQty` or from the peer list hits the fail-closed `0, 0` branch (`scheduler/hyperliquid_balance.go:2171-2177`) and still clears model-only, and on a coin shared by one perps and one manual strategy the perps-only peer list has length one, so `hyperliquidKillSwitchFillShare` returns the **entire** fill to the perps strategy including the manual share (`scheduler/hyperliquid_balance.go:2151-2152`). The current type gate accidentally prevents that.
- Relax the `sc.Type == "perps"` gate in `applyHyperliquidKillSwitchCloseFill` so a manual strategy books the real fill price, fee, and order ID rather than falling through to the model-only row.
- Move the unowned/unconfigured-coin detection into the live-perps branch of `planKillSwitchClose` so an on-chain position no configured strategy owns raises its `[CRITICAL]` line and blocks `OnChainConfirmedFlat` regardless of how many live perps strategies exist.
- Widen or bound `hlKillSwitchNoFillRecoveryLookback` (currently 30s, `scheduler/kill_switch_close.go:17`) so an already-flat position whose fill landed before the window is still recoverable instead of clearing model-only.

Load-bearing invariants that must survive: shared-coin virtual-quantity splitting (`hyperliquidKillSwitchFillShare` fails closed), the #620 rule that a perps circuit breaker never closes a manual peer's wallet exposure, and the #1159 hedge-ownership matrix that lets shared-coin mechanisms stay blind to hedge coins.

## Acceptance criteria

- A live `type=manual` Hyperliquid position is closed by the kill switch, and its close row carries a non-empty `exchange_order_id`, the fill price, the exchange fee, and fill-derived `realized_pnl`.
- A coin held only by a manual strategy is no longer skipped as "unowned". A close failure on it emits a `[CRITICAL]` line, flips `OnChainConfirmedFlat` to false, and latches for retry next cycle.
- On a mixed fleet with at least one live perps strategy, an on-chain position that no configured strategy owns still raises the unconfigured-coin `[CRITICAL]` line and blocks flat confirmation.
- A fill that landed before the no-fill recovery window is still booked rather than clearing model-only.
- Shared-coin virtual-quantity splitting still fails closed; the #620 peer-scope rule and the #1159 hedge-ownership rules are unchanged.
- Manual resting stop-loss and take-profit trigger orders are cancelled before the kill-switch close fires, so closing a manual coin leaves no orphan triggers on Hyperliquid's open-order cap (#421 / #479).
- On a coin shared by a perps and a manual strategy, each books only its virtual-quantity share of the fill — no strategy claims the whole fill, and the fail-closed share path is not reached for a correctly configured pair.
- The #1448 / #1449 portfolio latch-ownership arms are unchanged (`equityGuardArmed`, `scheduler/risk.go:1152`).
- Trailing-stop and risk math remain perps-only — #576's separation is preserved outside the kill-switch surface.
- Regression tests cover: manual-only coin closed and booked; manual coin close failure latches; mixed-fleet unconfigured coin detected; shared perps+manual coin split by virtual quantity rather than awarded whole to one side; manual resting triggers cancelled; hedge coin still not adopted.

## Plain simple English

The emergency stop can miss positions that an operator opened by hand. It closes the automatic positions, tells the operator that everything is flat, and leaves the hand-opened positions live on the exchange with no warning. This work brings those positions inside the emergency stop and records their real exchange results.

---

richkuo:
## Plain simple English

The emergency stop could miss positions an operator opened by hand: it closed the automatic ones, reported everything flat, and left hand-opened positions live on the exchange with no warning. Their close rows also recorded estimates instead of real exchange fills. This change brings those positions inside the emergency stop, books their real fills, and alerts the owner when any close still has no exchange order behind it.

## Summary

Implements #1454 (root cause 1 + root cause 3 of umbrella #1451, plus the folded model-only alert).

- **Kill-switch roster** (`hlKillSwitchAll`, main.go): live HL perps + live `type=manual` via `isHLLiveReconcilable` (#620 peer-scope precedent). It replaces `hlLiveAll` across the whole kill-switch fill chain — `KillSwitchCloseInputs.HLLiveAll`, the resting SL/TP trigger-OID cancel snapshot (extracted to testable `collectHLKillSwitchStopOIDs`), `snapshotHyperliquidVirtualQuantities`, the `hyperliquidKillSwitchFillShare` peer set, and `forceCloseKillSwitchPositions`. `hlLiveAll` itself stays perps-only — trailing-stop arming and risk math keep their #576 meaning.
- **Coin keys are RAW, not normalized.** Exchange-side surfaces (`report.Fills`, `stopLossOIDsByCoin`, virtual-position maps) stay keyed by the raw on-chain ticker because HL lower-case-prefix tickers (`kPEPE`) are case-sensitive and `hyperliquidConfiguredCoin` upper-cases. Manual coins resolve via `sc.Symbol` (#1444 precedent). Normalization is used only for traded-coin membership comparison inside `forceCloseHyperliquidLive`. The full suite's three kPEPE already-flat tests pin this.
- **Unowned-coin detection** now lives in `forceCloseHyperliquidLive`: non-zero on-chain positions on coins no roster strategy trades surface as `plan.Unconfigured` with a `[CRITICAL]` line and block `OnChainConfirmedFlat` on a mixed fleet too — previously only an empty perps roster reached that check.
- **No-fill recovery lookback** widened 30s → 10m (`hlKillSwitchNoFillRecoveryLookback`) so a fill that landed just before the switch fired is recovered from userFills rather than clearing model-only; ambiguity still fails closed.
- **Folded #1451 alert**: every model-only close row now queues a throttled owner DM keyed `(strategy_id, symbol)` at the `alert_throttle_interval` floor, drained outside `mu` at two sites.

Preserved invariants: shared-coin virtual-quantity splitting fails closed; #620 perps-CB peer scope untouched; #1159 hedge ownership matrix untouched (declared-but-flat hedge coin stays unowned but is now reported); #576 trailing-stop/risk separation intact; #1448/#1449 latch-ownership arms untouched.

## Verification

- New `scheduler/kill_switch_manual_test.go` (8 tests): manual-only coin closed + booked with real OID/fill-derived PnL (no model-only row); manual close failure latches; mixed-fleet unconfigured coin blocks flat; perps+manual shared coin split by virtual quantity; manual triggers collected for cancellation incl. raw lower-case symbol; declared-but-flat hedge coin unowned-but-reported; model-only DM throttle/drain.
- Red→green: the manual-booking and mixed-fleet tests were run against unfixed `origin/main` and failed there exactly as the issue describes (empty `exchange_order_id`, mark price, `reconcile_adjustment`, confirmed-flat despite the foreign position); they pass with the fix.
- Full scheduler suite green: `go vet ./...`, `gofmt` clean, `go test ./...` (all packages) pass. No Python changes.

Closes #1454. Progress toward umbrella #1451 (root cause 2 is #1455).

---
Created with LLM: ox-alpha | high | Harness: opencode

### DIFF
--- a/scheduler/hedge_test.go
+++ b/scheduler/hedge_test.go
@@ -1642,7 +1642,7 @@ func TestEveryHedgeLegCarriesTheHedgeTradeType(t *testing.T) {
 	t.Run("circuit-breaker virtual force-close sweep", func(t *testing.T) {
 		s := hedgeTestState("eth-long")
 		s.Positions["BTC"] = hedgePos(0.4, "short", 10)
-		forceCloseAllPositions(s, map[string]float64{"BTC": 51000}, silentStrategyLogger("eth-long"))
+		forceCloseAllPositions(s, nil, map[string]float64{"BTC": 51000}, silentStrategyLogger("eth-long"))
 		assertAllHedgeRows(t, s, 1)
 	})
 
@@ -2007,7 +2007,7 @@ func TestCircuitBreakerFireWithFailedFetchThenRecoveryClosesBothLegs(t *testing.
 	if !shouldForceCloseAllPositionsOnCircuitBreaker(&sc, assist) {
 		t.Fatal("setup: a sole-owner strategy still force-closes virtually")
 	}
-	forceCloseAllPositions(s, map[string]float64{"ETH": testPrimaryPx, "BTC": testHedgePx}, silentStrategyLogger("eth-long"))
+	forceCloseAllPositions(s, nil, map[string]float64{"ETH": testPrimaryPx, "BTC": testHedgePx}, silentStrategyLogger("eth-long"))
 	if len(s.Positions) != 0 {
 		t.Fatalf("setup: the sweep must clear both virtual legs, got %d", len(s.Positions))
 	}
--- a/scheduler/hyperliquid_balance.go
+++ b/scheduler/hyperliquid_balance.go
@@ -1970,6 +1970,12 @@ type HyperliquidLiveCloseReport struct {
 	AlreadyFlat []string
 	// Errors is non-nil so coin-keyed writes don't panic; len() works on nil maps too.
 	Errors map[string]error
+	// #1454: on-chain positions on coins no live HL strategy in the kill-switch
+	// roster (live perps + live type=manual) trades. Surfaced by
+	// planKillSwitchClose as plan.Unconfigured so a MIXED fleet gets the same
+	// CRITICAL line + flat-confirmation block an all-paper fleet already had,
+	// instead of the silent skip that hid manual-only coins.
+	Unconfigured []HLPosition
 }
 
 // ConfirmedFlat reports whether every configured live HL coin reached a
@@ -2030,9 +2036,16 @@ func forceCloseHyperliquidLive(ctx context.Context, positions []HLPosition, hlLi
 
 	tradedCoins := make(map[string]bool)
 	for _, sc := range hlLiveAll {
-		sym := hyperliquidSymbol(sc.Args)
-		if sym != "" {
-			tradedCoins[sym] = true
+		// #1454: resolve sc.Symbol for type=manual so a coin held only by a
+		// live manual strategy is inside the close scope instead of being
+		// skipped as unowned. Membership is compared CASE-INSENSITIVELY (HL
+		// tickers like kPEPE carry a lower-case prefix; hyperliquidConfiguredCoin's
+		// upper-casing is safe for peer matching but NOT for exchange-side keys),
+		// while report.Fills stays keyed by the RAW on-chain coin.
+		if sym := hyperliquidSymbol(sc.Args); sym != "" {
+			tradedCoins[strings.ToUpper(sym)] = true
+		} else if sc.Type == "manual" && sc.Symbol != "" {
+			tradedCoins[strings.ToUpper(sc.Symbol)] = true
 		}
 	}
 	// #1159: a held hedge leg is exposure this scheduler opened and therefore
@@ -2046,11 +2059,18 @@ func forceCloseHyperliquidLive(ctx context.Context, positions []HLPosition, hlLi
 	}
 
 	for _, p := range positions {
-		if !tradedCoins[p.Coin] {
+		if !tradedCoins[strings.ToUpper(p.Coin)] {
 			// Unowned position — kill switch only acts on coins this scheduler
 			// is configured to trade. An on-chain leftover from a different
 			// system (manual trade, another bot) is the operator's problem to
 			// reconcile, not the scheduler's to liquidate.
+			// #1454: a NON-ZERO unowned position is reported (report.Unconfigured
+			// → plan.Unconfigured → latch) rather than silently skipped — the
+			// silent skip is what left manual-held exposure running after a
+			// "confirmed flat" kill-switch pass on a mixed fleet.
+			if p.Size != 0 {
+				report.Unconfigured = append(report.Unconfigured, p)
+			}
 			continue
 		}
 		if p.Size == 0 {
@@ -2140,8 +2160,18 @@ func snapshotHyperliquidVirtualQuantities(strategies map[string]*StrategyState, 
 		if ss == nil {
 			continue
 		}
-		if coin := hyperliquidSymbol(sc.Args); coin != "" {
-			if pos := ss.Positions[coin]; pos != nil && pos.Quantity > 0 {
+		// #1454: resolve the coin for type=manual so their virtual quantities
+		// join the same fill-share snapshot as perps; otherwise a shared
+		// perps+manual coin awards the whole fill to the perps side and the
+		// manual leg fails closed into a model-only row. RAW key — this must
+		// match the key applyHyperliquidKillSwitchCloseFill books under.
+		coin := hyperliquidSymbol(sc.Args)
+		if sc.Type == "manual" {
+			coin = sc.Symbol
+		}
+		if coin != "" {
+			pos := hlVirtualPositionFor(ss, sc, coin)
+			if pos != nil && pos.Quantity > 0 {
 				record(coin, sc.ID, pos.Quantity)
 			}
 		}
@@ -2159,6 +2189,26 @@ func snapshotHyperliquidVirtualQuantities(strategies map[string]*StrategyState, 
 		return nil
 	}
 	return out
+}
+
+// hlVirtualPositionFor looks up a strategy's virtual position for a kill-switch
+// coin. The kill-switch chain keys coins through hyperliquidConfiguredCoin
+// (normalized upper-case), while type=manual positions are keyed by the raw
+// configured symbol — the fallback covers an operator-entered lower-case
+// symbol so the normalized fill-share key still finds the virtual leg.
+func hlVirtualPositionFor(ss *StrategyState, sc StrategyConfig, coin string) *Position {
+	if ss == nil {
+		return nil
+	}
+	if pos, ok := ss.Positions[coin]; ok && pos != nil {
+		return pos
+	}
+	if sc.Type == "manual" && sc.Symbol != "" && sc.Symbol != coin {
+		if pos, ok := ss.Positions[sc.Symbol]; ok && pos != nil {
+			return pos
+		}
+	}
+	return nil
 }
 
 // computeHyperliquidCircuitCloseQty returns the unsigned coin quantity for a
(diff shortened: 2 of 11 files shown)


richkuo:
@claude opus review

richkuo:
@claude opus review

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T08:32:35Z) in 5f503fac.

### Fixed
1. **The close scope resolves a `type=manual` coin from `Args[1]` first while every other new site resolves `sc.Symbol` first, so a divergent pair leaves live manual exposure open.** — Extracted `hyperliquidRawCoin` (symbol-first for manual, the #1444 precedent) and routed all four sites through it: close scope (`scheduler/hyperliquid_balance.go:2053`), virtual-quantity snapshot (`scheduler/hyperliquid_balance.go:2168`), fill booking (`scheduler/hyperliquid_balance.go:2286`), resting-trigger cancel (`scheduler/kill_switch_close.go:658`). Tests pin the divergent-args case end-to-end: close scope, fill adoption under the symbol key, and the resolver unit test.
2. **The new model-only close DM has no live/paper gate, so paper strategies alert the owner about exchange positions that cannot exist.** — `forceCloseAllPositions` now takes the `StrategyConfig` and queues the DM only when `isLiveArgs(sc.Args)` (`scheduler/risk.go:2044`). Paper HL is silent; live HL perps and live OKX spot still alert; corrupt-position closes stay silent via the existing `details` branch. Covered by `TestForceCloseAllPositions_ModelOnlyAlertLiveGate`.
3. **`hlVirtualPositionFor`'s manual fallback is unreachable and its comment states the opposite of the key convention the PR relies on.** — Dead branch removed (`scheduler/hyperliquid_balance.go:2200`) and both it and the `main.go` comment rewritten to state the actual convention: RAW case-preserved keys throughout the kill-switch chain; normalizing either side would silently miss `kPEPE`-style lookups.
4. **`hlKillSwitchAll` duplicates `hlReconcileAll` exactly, creating two rosters that must stay identical by hand.** — `hlKillSwitchAll := hlReconcileAll` (`scheduler/main.go:1247`) with a comment directing any future divergence to get its own named predicate.
5. **Widening the no-fill recovery window to 10 minutes weakens the match without adding any side or time filter, so one unrelated same-size fill can be booked as the close.** — The recovery matcher now restricts candidates to REDUCING fills (non-zero `closedPnl`; Hyperliquid reports 0 on opening legs) whose event time is not before the query's since bound (`scheduler/kill_switch_close.go:333`), so an opening same-size fill inside the wider window fails closed to model-only instead of being booked. New tests cover opening-fill rejection and pre-since rejection; the two ambiguity fixtures gained non-zero closedPnl so they still pin genuine ambiguity failing closed.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T09:18:18Z, LGTM) in 72b55539.

### Fixed
1. **An unowned on-chain coin now blocks the entire fleet's kill-switch bookkeeping on every platform, discarding real fills the HL closer already returned.** — Implemented, with one correction to the proposed remedy: running the full sweep while `OnChainConfirmedFlat=false` would model-only-close positions whose exchange leg is still live on errored coins — the exact failure the #341 confirmed-flat gate exists to prevent. Instead, a held latch now books only POSITIVELY settled legs (`scheduler/kill_switch_close.go`, `applyKillSwitchSettledLegsWhileLatched`): confirmed HL fills book with real price/fee/OID (the #954 same-OID dedupe makes already-flat recovery re-finding the same fill idempotent), and non-HL coins listed in a venue's `ClosedCoins` book model-only at mark via `forceCloseSettledPositions` (`scheduler/risk.go`), platform-gated so one venue's coin set can never touch another's same-named symbol. Unsettled legs stay untouched for the retry. Both call sites share one booking implementation (`closeVirtualPositionAtMark`, extracted from `forceCloseAllPositions`). The latch itself is unchanged.
2. **Two comments still assert the kill switch is perps-only, which is exactly what this PR changed.** — Both rewritten (`scheduler/main.go` reconcile-roster comment and the `isHLLiveReconcilable` doc): the kill switch consumes the perps+manual roster; trailing-stop arming and risk math remain the perps-only consumers and hlLiveAll itself must not be widened (#576).

### Deferred to follow-up
1. **A resting manual limit order is not cancelled by the kill switch and its fill adoption has no latch gate, so live exposure can re-open after a "confirmed flat" flatten.** — out of scope, basis reviewer-routed to `### Create Follow-up Issue`; filed as #1462.

---

richkuo:
@claude review

Issue #1009, opened 2026-06-14
Bug: partial-close path stores negative position quantity on direction reversal


richkuo:
## Summary

When a strategy reverses position direction (long→short or short→long) via a single fill that both closes the existing position and opens the opposite side, the `partialClose` code path in `executePerpsSignalWithLeverage` sets `closeQty = fillQty` without capping at `pos.Quantity`. This stores a **negative-quantity position** that corrupts state, triggers perpetual shared-wallet drift alerts, and is deleted on restart (leaving an orphaned on-chain position).

## Precondition

This only triggers under `direction="both"`. The live order sizer `perpsLiveOrderSize` flip-sizes (`posQty + newSize`) **only** when `flipping = (direction == "both" && opposite-side signal)` — and that predicate ignores `CloseFraction` (`scheduler/portfolio.go:819,864`). Under a directional gate (`long`/`short`) an opposite-side signal is close-only and the sizer returns `posQty * closeFraction` (≤ `posQty`), so no negative quantity is possible.

Note: `rsi_macd_combo` is **not** in `bidirectionalPerpsStrategies` (`scheduler/init.go`), so it ships long-only by default. Reproducing this requires the strategy to be explicitly configured `direction="both"`.

## Root Cause

The deeper defect is a **sizer-vs-executor intent mismatch**: for the same reversal signal, `perpsLiveOrderSize` evaluates "full flip" (sizes `posQty + newSize`) while `executePerpsSignalWithLeverage` evaluates "partial close" (`closeFraction ∈ (0,1)` → `closeOnlyAction`). The two predicates are computed independently and never mutually exclude each other (`scheduler/main.go:3083,3127`).

In `scheduler/portfolio.go`, `executePerpsSignalWithLeverage`, both the close-short (signal=1) and close-long (signal=-1) paths:

```go
closeQty := pos.Quantity
if partialClose {
    if fillQty > 0 {
        closeQty = fillQty  // ← NOT capped at pos.Quantity
    } else {
        closeQty = pos.Quantity * closeFraction
    }
}
```

When `partialClose` is true and `fillQty` exceeds `pos.Quantity` (e.g. a 1.192 flip fill against a 0.597 position), the subsequent `pos.Quantity -= closeQty` produces a negative value:

```
pos.Quantity = 0.597 - 1.192 = -0.595
```

## Impact

1. **Negative-quantity position persisted to DB** — side label retains the old direction (e.g. `side="long", quantity=-0.595`)
[...]

PR #1458, opened 2026-08-24
fix(#1455): reconcile circuit-breaker realized PnL to the real exchange fill [C89, ox-alpha, high]

### LINKED ISSUE #1455: [C89] Circuit-breaker closes keep the estimated profit and loss instead of the real fill

richkuo:
## Plain simple English

When one strategy hits its safety limit, the system recorded its own guess of the closing price and never replaced it with what the exchange paid. The stored profit or loss stayed wrong, and so did the cash balance and the statistics built on it. This change corrects the record when the real result arrives, in one transaction across every table the estimate reached.

## Summary

Implements #1455 (root cause 2 of umbrella #1451).

- **Reconcile, don't stack.** `reconcileModelOnlyCloseWithFill` (new `scheduler/model_only_reconcile.go`) runs in `applyHyperliquidCircuitCloseFill`'s no-virtual-position branch before the defensive zero-PnL row. It finds the most recent uncorrected model-only close row (empty OID + `reconcile_adjustment` fee source + model-only detail marker; #1009 corrupt rows excluded by their distinct detail), recovers the basis (quantity / avg cost / side / multiplier) from the in-memory closed-positions buffer or the persisted `closed_positions` row, and corrects the trade row to the real fill price/fee/OID/`userfills` source with fill-derived gross PnL.
- **Every surface the estimate reached is corrected**: trades row, cash (delta), `RiskState.DailyPnL` (delta only — the #1269 daily-loss limit reads the corrected figure), the buffered `ClosedPositions` entry, the persisted `closed_positions` row, and the #1147 `trade_diagnostics` capture — the latter three in ONE transaction (`(*StateDB).ReconcileModelOnlyClose`). The loss streak is never re-litigated; hedge legs stay streak-excluded (#1159).
- **Restart-safe, no new carried store.** Row identity is database-side (`(strategy_id, timestamp)` for trades per the `UpdateTradeStampedFields` precedent; `(strategy_id, symbol, close_reason, closed_at)` for closed_positions), so a restart between fire and fill still reconciles via `LoadModelOnlyCloseBasis`. `PendingCircuitClose` and the kill-switch latch remain the sole retry owners; the failed-fetch fire path that sets no pending entry is covered because the lookup never reads pending state.
- **Invariants preserved.** #954 one-fill-one-row: the corrected row takes the fill's OID so replays are no-ops and the reconciler-won-race case still short-circuits; the UPDATE re-asserts the uncorrected shape so a second pass can never match. No-basis fills keep the unchanged defensive branch. Paper strategies never reach the fill path (live-args gate in the drain). The portfolio kill-switch path is untouched.

Known bounded limitation: on a partial first fill the reconciliation covers min(basis qty, filled qty); the residual keeps retrying under PendingCircuitClose and books its own rows as before.

## Verification

- New `scheduler/model_only_reconcile_test.go` (8 tests): fire-then-fill reconciliation; no-basis defensive fallback; restart between fire and fill; loader-failure fallback; duplicate-OID proof; hedge-leg daily-PnL routing without streak impact; shared coin takes the operator-required path (no model-only row ever written); real-SQLite test asserting all three persisted rows are updated in one shot and a second reconciliation is refused.
- Red→green: an equivalent fire-then-fill assertion was run against unfixed `origin/main` and failed there (two rows, estimate kept); it passes with the fix. Full scheduler suite green: `go vet ./...`, `gofmt` clean, `go test ./...` all packages. No Python changes.

Closes #1455. Together with #1457 this completes both children of umbrella #1451 (each also carries its copy of the folded model-only owner alert).

---
Created with LLM: ox-alpha | high | Harness: opencode


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T04:40:22Z) in 8e289497.

### Fixed
1. **A partial first fill permanently drops the residual quantity's realized PnL from every surface.** — The reconciler is now slice-wise and incremental: each fill reconciles only its own slice against the estimate's exact linear share (`qty × mult × (estPx − avgCost)`, with `estPx` preserved in `trades.price` for the whole sequence), and the SAME trades row carries cumulative filled state until the persisted basis quantity is covered — then it takes the cumulative VWAP, the last OID, and `FeeSourceUserFills`. Over-fills clamp to the basis with a proportionally scaled fee; exhausted bases refuse and fall to the defensive branch (`scheduler/model_only_reconcile.go:143`). Covers all three Must-survive cases with new tests, including the two-stage different-price/fee fill and the short residual crossing AvgCost.
2. **A failed persist leaves cash corrected and the trades row holding the estimate, with no retry, no re-flush and no operator alert.** — Every in-memory mutation (row fields, Cash, DailyPnL + date, streak, closed_positions entry) is snapshotted before the DB correction and rolled back on error, with the failure raised through `tradePersistWarn` like `RecordTrade`'s identical failure class (`scheduler/model_only_reconcile.go:267`). Memory and DB now stay consistent, so the same fill retries cleanly after a restart instead of double-counting.
3. **The reconciler ignores the caller's close reason and puts no age bound on the row it matches.** — `closeReason` is threaded from `applyHyperliquidCircuitCloseFill` into both row matching and the closed_positions/diagnostics WHERE clauses plus the operator label, and `findModelOnlyCloseTrade` skips rows older than a 48h bound (`scheduler/model_only_reconcile.go:110`, `scheduler/hyperliquid_balance.go:2819`). A regime-flip/kill-switch fill can never correct a circuit-breaker row.
4. **A correction that arrives the next UTC day posts to the wrong day's daily-loss meter.** — The DailyPnL delta applies only when the fire landed on today's UTC day; cash carries the true-up either way (`scheduler/model_only_reconcile.go:246`).
5. **The diagnostics UPDATE discards its error and the transaction commits anyway.** — The statement's error now fails the whole transaction; only the legitimate zero-row hedge outcome is tolerated (`scheduler/db.go:1204`). A dropped-table test proves the trades UPDATE rolls back with it.
6. **The only basis lookup that runs in production is never tested against a real database.** — New end-to-end test drives fire → eager trade insert → real `SaveState` (buffer cleared) → restart-style loader recovery → reconcile, asserting all three persisted rows through `formatTime` round-trip against real SQLite, with no stubs on the production path (`scheduler/model_only_reconcile_test.go:663`).
7. **The consecutive-loss counter keeps the fire-time estimate even when the real fill flips its sign.** — Re-litigated exactly once per sign transition between the fire-time estimate sign and the cumulative realized net (decrement when a booked loss turns win, increment on the inverse; untouched rows read their prior net as the full fire-time estimate), hedge legs excluded in both directions (`scheduler/model_only_reconcile.go:254`). Both flip directions pinned by tests.

All seven findings live in code this PR adds or changes, so each was fixed here under fixer scope rule 1 rather than deferred.

Verification: `go build .`, `go vet .`, full `go test . -count=1` (38s, ok), gofmt clean.

---

richkuo:
@claude opus review

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T08:32:36Z) in eb7d6b1e.

### Fixed
1. **The new same-close-event guard does not exist on the production path — `LoadModelOnlyCloseBasis` hardcodes `close_reason='circuit_breaker'`.** — `closeReason` is now part of the basis-loader identity (`scheduler/model_only_reconcile.go:91`, `scheduler/db.go:1223`), so the guard holds on the only branch production executes; a `kill_switch`/regime fill with no matching persisted row takes the defensive zero-PnL branch instead of matching the CB row and failing the transaction. Pinned by a real-SQLite restart-path test with a stale CB row (`TestLoadModelOnlyCloseBasis_ReasonGuardHoldsOnProductionPath`). All six findings live in code this PR adds or changes, so each was fixed here under fixer scope rule 1 rather than deferred.
2. **A two-slice fill can move the consecutive-loss counter on a transient intermediate result and can never move it back.** — The streak now re-litigates exactly ONCE, when the sequence COMPLETES, comparing the full-basis estimate sign against the final pre-fee cumulative realized sign — never on an intermediate slice; hedge legs stay excluded in both directions (`scheduler/model_only_reconcile.go:301`). Both transient directions pinned by `TestModelOnlyClose_TwoSliceTransientStreakNeverMovesMidSequence` / `_TwoSliceTransientLossNeverOvercountsStreak`.
3. **An abandoned residual leaves a permanently half-corrected row and an unbacked estimate in cash, with no alert.** — The drain's flat-on-chain skip now detects an in-flight partial row for that coin and raises a throttled owner DM naming the inconsistency and the recovery (`warnAbandonedPartialModelClose`, `scheduler/hyperliquid_balance.go:2612`; 24h cooldown keyed strategy|symbol). Tested in `TestWarnAbandonedPartialModelClose_AlertsOncePerCooldown`.
4. **On a persist failure the fill is dropped permanently, and the operator alert tells the operator the opposite.** — The reconciler returns a dedicated persist-failed outcome; the caller books the defensive audit row so the fill's price/fee/OID land durably, while the untouched model-only row stays matchable for offline repair — and the DM names that actual recovery (`backfill trade-ledger`) instead of an automatic retry that never happens (`scheduler/model_only_reconcile.go:352`, `scheduler/hyperliquid_balance.go:2811`). Extended `TestModelOnlyClose_PersistFailureRollsBackAndAlerts` accordingly.
5. **A partial row carries no OID, so the per-fill dedup the design relies on cannot see it.** — Every applied slice's OID is recorded on the row the moment it lands (`oids=` token in Details, cumulative across the sequence and kept after completion) and `strategyHasCloseTradeForOID` scans it (`scheduler/portfolio.go:838`), restoring per-fill dedup from first application through completion (`scheduler/model_only_reconcile.go:250`). Covered by `TestModelOnlyClose_InFlightSliceOIDsAreDuplicateProof`.
6. **The offline `backfill trade-ledger` command can clobber an in-flight partial row and strand the residual.** — `tradeLedgerNoOIDReconcileMatches` skips rows carrying the fill-reconciled marker, so the offline repair never touches a reconciliation the running scheduler still owns; untouched pre-#1455 rows remain repairable (`scheduler/backfill_trade_ledger.go:108`). Tested in `TestTradeLedgerNoOIDReconcileMatches_SkipsInFlightPartialRows`.

Verification: `go build .`, `go vet .`, full `go test . -count=1` (38s, ok), gofmt clean.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T09:17:51Z) in 11d0ffb0.

Growth check: diff 1572 lines vs 609 at first push (2.6x); cycle 4. All four findings live in code this PR adds or changes, so each was fixed here under fixer scope rule 1.

### Fixed
1. **The streak re-litigation adjusts by ±1 where `RecordTradeResult` resets to 0, so a corrected close can under-count the #292 loss streak.** — The fire-time row now stamps its PRE-fire streak (`", pre-streak=<n>"`, `scheduler/risk.go:2038`), and the reconciler reconstructs exactly what `RecordTradeResult` would have produced for the TRUE sign: 0 on a win, pre-fire+1 on a loss; hedge legs stay excluded (`scheduler/model_only_reconcile.go:311`). All three Must-survive cases pinned, including the pre-fire-3→4 and pre-fire-4→0 directions and a hedge leg at streak 7 (`TestModelOnlyClose_PreFireStreakRecoveredFromRowStamp`).
2. **The abandoned-partial owner alert is wired to a branch a flat coin can never reach, so a half-corrected row is still dropped silently.** — Confirmed: flat coins are absent from the snapshot (`szi==0` dropped), so the check now ALSO rides the `alreadyFlat` closer outcome (`scheduler/hyperliquid_balance.go:2677`) and both guard sites that clear pending without a retry (non-live config, shared coin) via one helper, `checkAbandonedPartialModelClose`; the zero-size site is kept. Must-survive (c) is alerted rather than scoped out.
3. **The abandoned-partial alert directs the operator to `backfill trade-ledger`, which this PR teaches to skip exactly those rows.** — The first abandonment observation now tags the row `[reconcile-abandoned]` in memory AND persistently (`MarkModelOnlyCloseAbandoned`, `scheduler/db.go:1240`; idempotent), releasing scheduler ownership: the offline repair still skips LIVE in-flight rows but matches ABANDONED ones (`scheduler/backfill_trade_ledger.go:108`) — the named recovery now actually converges the row. Must-survive cases pinned in `TestWarnAbandonedPartialModelClose_MarksRowForOfflineRepair`, `TestTradeLedgerNoOIDReconcileMatches_SkipsInFlightPartialRows`, `TestMarkModelOnlyCloseAbandoned_PersistsIdempotentTag`.
4. **A basis-lookup database failure is indistinguishable from "no such row" and silently reinstates the old wrong behaviour with only a stdout line.** — `modelOnlyCloseBasisFor` now distinguishes wrapped `sql.ErrNoRows` (ordinary unexplained fill — silent defensive branch) from any other error, which raises `tradePersistWarn` like every other failed money persist while still degrading fail-safe (`scheduler/model_only_reconcile.go:158`). Both paths + restart-reconcilability pinned in `TestModelOnlyClose_BasisLookupErrorAlertsNoRowsStaySilent`.

Verification: `go build .`, `go vet .`, full `go test . -count=1` (38s, ok), gofmt clean.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T10:00:59Z) in a1b02901.

Growth check: diff 2009 lines vs 609 at first push (~3.3x); cycle 5. Both findings live in code this PR adds, so each was fixed here under fixer scope rule 1.

### Fixed
1. **A multi-slice partial fill loses the `pre-streak=` stamp, so the round-3 loss-streak fix silently regresses to the ±1 under-count it was meant to eliminate.** — Confirmed: the slice/complete rewrite replaced Details wholesale. The stamp is now re-attached on EVERY rewrite (`detailsSuffix` built from the value parsed once per call from the pre-mutation text, `scheduler/model_only_reconcile.go:285`), and `modelOnlyPreStreak` also cuts at `)` so the paren-embedded stamp parses on rewritten rows (`:512`). All three Must-survive cases pinned in `TestModelOnlyClose_PreStreakStampSurvivesMultiSliceRewrites`: two-slice est-win→real-loss ends at 4 (not 1), two-slice est-loss→real-win resets to 0, hedge multi-slice never touches the streak.
2. **`checkAbandonedPartialModelClose` mutates a shared trade row while holding only `mu.RLock()`, racing concurrent readers.** — Confirmed against `warnAbandonedPartialModelClose`'s in-place `t.Details +=` and the UI status handler's RLock copy of `TradeHistory`. The helper now holds `mu.Lock()` around the check-and-mark (`scheduler/hyperliquid_balance.go:2306`) — the mutation is a short in-place append plus one DB UPDATE; the non-abandoning fast path still takes only that brief exclusive window and no longer regresses anything else.

Verification: `go build .`, `go vet .`, full `go test . -count=1` (38s, ok), gofmt clean.

---

richkuo:
@claude review

richkuo:
Resolved the merge conflict with origin/main (#1457 merged) in b55ade51: `scheduler/risk.go` only. #1457 extracted `forceCloseAllPositions`'s per-position body into `closeVirtualPositionAtMark`; the merge takes that extraction whole and re-applies this PR's `pre-streak=` stamp inside the extracted function, keeping #1454's live-only model-only alert untouched. The test helper moves to the new `(s, sc, prices, logger)` signature with `sc=nil` (skips the live-only alert, as before). Full `go test . -count=1` ok after the merge.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-24T10:41:30Z) in e2310215.

Growth check: diff 2009 lines vs 609 at first push (3.3x); cycle 8. Scope test re-run on the implemented finding: it is a defect in code this PR adds, so scope rule 1 keeps it here.

### Fixed
1. **The corrected loss-streak is classified on GROSS PnL, but every other close classifies wins/losses on NET PnL, so a near-breakeven fill whose fee flips the sign under-counts the #292 loss-streak.** — Confirmed against scheduler/model_only_reconcile.go:376-390: the truth side compared pre-fee `cumGross` while `RecordTradeResult` (scheduler/risk.go:2463) classifies on the net pnl every real close passes it (e.g. scheduler/portfolio.go:2014). The switch now compares against `cumNet := cumGross - cumFee`, keeping the fire-time estimate's own pre-fee sign unchanged (the model-only row carries no fee, so gross == net there). Tests added for all three must-survive cases (`TestModelOnlyClose_StreakClassifiedOnNetPnL`, scheduler/model_only_reconcile_test.go): estimated loss with a gross-positive/net-negative fill keeps pre-fire+1 = 4; estimated win in the same band becomes pre-fire+1 = 3; a hedge leg in the band leaves the streak untouched.

Verification: full package suite `go test -C scheduler -count=1 .` ok after the change.

---

richkuo:
@claude review

PR #1469, opened 2026-08-29
fix(#1462): cancel resting manual limit orders in the kill-switch flatten [C91, Opus 5, xhigh]

### LINKED ISSUE #1462: [C91] Kill switch leaves resting manual limit orders live to refill after a confirmed-flat flatten
**Complexity: 91/100** — Capability 3 (Risk 4 — kill-switch flatten, live order cancellation and realized-PnL booking); Volume 16 — Opus 5, xhigh · fableplan: yes

## Problem

The portfolio kill switch cancels only the resting stop-loss and take-profit trigger orders that are attached to an open position. `collectHLKillSwitchStopOIDs` gathers `pos.StopLossOID` and `pos.TPOIDs` and returns nothing when the strategy holds no position (`scheduler/kill_switch_close.go:503-524`, `scheduler/hyperliquid_balance.go:1647-1652`). Those OIDs are then consumed only inside the per-position close loop, which skips any coin whose on-chain size is zero (`scheduler/hyperliquid_balance.go:1555, 1565-1567`), and the close script cancels only the OIDs it is handed (`shared_scripts/close_hyperliquid_position.py:136-150`).

A resting manual limit order with no position yet therefore survives the flatten untouched. No path in the kill-switch block reads `pending_limit_orders` (`scheduler/main.go:1069-1112`). `manual-close` already cancels resting limit orders before it flattens (`scheduler/manual_core.go:863` → `clearRestingLimitRemainderForPositionAction`, `scheduler/manual_core.go:247-315`), so the kill switch is the only flatten path in the codebase that leaves them live.

The command-line placement guard does not close the gap: it refuses a **new** `manual-open --limit-price` while the switch is latched (`scheduler/manual_limit.go:212-215`, mirrored at `scheduler/manual.go:93`) and never revisits an order already resting at the venue.

Confirmed-flat is computed from positions only, so `plan.OnChainConfirmedFlat` stays true with an order still working, and the operator is told the book is flat (`scheduler/kill_switch_close.go:540-566`).

### Where the exposure actually persists

The fill comes from the operator's own resting order; the scheduler places no new order. While the latch holds, `checkPortfolioRiskWithEquityAvailability` returns `allowed=false` on every cycle (`scheduler/risk.go:532-534`), so on a cycle with due strategies the close pass re-flattens an adopted position in the same cycle. Three windows escape that:

1. **Auto-reset with no DM owner.** A confirmed-flat close clears the latch outright (`CanAutoResetWithoutOwner`, `scheduler/kill_switch_close.go:93`; applied at `scheduler/main.go:1121-1147`). The resting order stays live with no kill switch behind it, and a later fill is a full re-entry.
2. **After the operator resets.** The reset clears the latch (`scheduler/main.go:1258-1266`) while the forgotten order still rests.
3. **Quiet ticks.** `reconcilePendingLimitOrders` runs on every tick (`scheduler/main.go:620`), above the `dueStrategies` gate; the kill-switch close pass runs only on cycles that have due strategies (`scheduler/main.go:640-700`). A fill adopted on a quiet tick stays open until the next due cycle.

The reconciler itself reads no kill-switch state at all (`scheduler/manual_limit.go:478-627`), so it adopts the fill in every one of those windows.

## Goal

A kill-switch flatten must leave no operator order able to put money back on the exchange. Cancel every resting limit order as part of the close pass, keep the latch and the confirmed-flat report honest when a cancellation cannot be confirmed, and keep booking any fill that already landed.

## Approach

1. Add an independent cancel pass over the persisted `pending_limit_orders` rows (`scheduler/db.go:2356, 2401`), keyed on the row itself, so it runs whether or not a position exists, and run it before the flatten in the kill-switch block. Folding the order OIDs into the existing `HLStopLossOIDs` map does **not** work: that map is consumed only inside `forceCloseHyperliquidLive`'s per-position loop, which `continue`s on a zero on-chain size, so in the exact scenario this issue describes the coin never reaches the closer.
2. Reuse the cancel contract that already exists at `clearRestingLimitRemainderForPositionAction` (`scheduler/manual_core.go:247-315`): cancel the OID, re-read `--limit-status`, refuse to finalize while the order is still resting or while open-order state is unknown, adopt any fill that landed before the cancel, then delete the queue row. It fails closed on every unknown.
3. Keep fill adoption exactly as it is today. Do **not** gate the reconciler on kill-switch state. New exposure is prevented by cancelling the order. Refusing to record a fill that already happened does not prevent it. Suppressing adoption would leave `o.FilledSize` at zero, so the terminal branch would report "limit order cancelled with no fill" and delete the queue row (`scheduler/manual_limit.go:532, 575-598`), orphaning the fill entirely; and it would put a live on-chain position outside the virtual book, where `forceCloseKillSwitchPositions` books nothing (`scheduler/main.go:1123-1127`) and the realized profit, loss and fees never reach the ledger — the model-only accounting class that #1454 and #1455 just closed.
4. On a cancellation that cannot be confirmed, set `plan.OnChainConfirmedFlat = false`. That holds the latch, suppresses `CanAutoResetWithoutOwner` (`scheduler/kill_switch_close.go:93`), routes the operator message to the latched-and-retrying wording, and names the order.
5. Cover both surfaces with tests: a resting order at latch time is cancelled and its row finalized; an unconfirmed cancellation blocks confirmed-flat and auto-reset; a fill that lands mid-latch is still adopted and booked; the limit-order lifecycle outside a latch is unchanged.

## Acceptance criteria

- A live Hyperliquid `type=manual` strategy with a resting limit order and no position gets that order cancelled by the kill-switch close pass. Only manual strategies can hold one: rows are inserted solely by `manual-open --limit-price` (`scheduler/manual_limit.go:260-272`, resolved through `findManualStrategy`, `scheduler/manual.go:590`) and the reconciler skips any strategy whose type is not `manual` (`scheduler/manual_limit.go:501`).
- A cancellation that cannot be confirmed clears `OnChainConfirmedFlat`, so the latch holds, auto-reset is suppressed, and the operator message names the order and its OID.
- A limit fill that lands before or during the cancellation is still adopted into the book and booked with its real fill price and fee, so the flatten closes it and its realized profit or loss reaches the ledger.
- The auto-reset path never clears the latch while any resting limit order is unaccounted for.
- The limit-order lifecycle outside a kill-switch latch is unchanged: fills adopt exactly as they do today.

## Plain simple English

A hand-placed order that waits for a price is not cancelled by the emergency stop. The stop reports all positions closed, but that order stays live at the exchange. If it fills later, real money goes back into the market with no protection. The stop must cancel these orders first.

---
Validated with LLM: Opus 5 | high | Harness: Claude Code

richkuo:
## Plain simple English

The emergency stop closed every open trade but ignored hand-placed orders that were still waiting for a price. It then told the operator the account was flat. If such an order filled later, real money went back into the market with no emergency stop behind it. The stop now cancels those orders first, and refuses to report "flat" until each one is confirmed gone.

## Summary

Closes #1462

The kill switch cancelled only the stop-loss and take-profit triggers attached to an open position. `collectHLKillSwitchStopOIDs` returns nothing when a strategy holds no position, and the OIDs it does return are consumed only inside `forceCloseHyperliquidLive`'s per-position loop, which `continue`s on a zero on-chain size. A resting `manual-open --limit-price` order on a flat coin therefore survived the flatten while `OnChainConfirmedFlat` stayed true.

**New `scheduler/kill_switch_limit_orders.go`** — an independent cancel pass keyed on the persisted `pending_limit_orders` row, so it runs whether or not a position exists. `planKillSwitchClose` runs it first, ahead of the flatten.

1. **Cancel contract reused verbatim** from `clearRestingLimitRemainderForPositionAction`: cancel the OID, re-read `--limit-status`, then refuse to finalize while the order still rests, while `OpenOrdersError`/`FillsError` leave book state unknown, while the order is absent from the status response, or while the exchange reports a fill the book has not adopted. Only a fully clean result deletes the queue row.
2. **Every unresolved row clears `plan.OnChainConfirmedFlat`.** That holds the latch, suppresses `CanAutoResetWithoutOwner` (now gated directly on `LimitOrderReport.ConfirmedClear()` as well, so the invariant does not depend on statement order), routes the operator to the latched-and-retrying wording, and names the order and its OID.
3. **An unadopted fill keeps the queue row.** The row is left for `reconcilePendingLimitOrders` to adopt on the next tick; the following kill-switch pass then flattens the resulting position.
4. **Rows we cannot act on fail closed.** A row whose strategy is not a Hyperliquid strategy with a script in this config is reported as manual-intervention, mirroring the existing `Unconfigured` on-chain-position precedent — an order we cannot cancel is real exposure. An unwired canceller and an unreadable queue are treated the same way.
5. **`killSwitchLimitOrderRoster` spans every Hyperliquid strategy**, paper included and any type, rather than the live kill-switch roster. `RunHyperliquidCancelOrder` always cancels a real on-chain order, so widening the roster strictly reduces both false latches and uncancellable rows.

**Fill adoption is unchanged and deliberately not gated on kill-switch state.** Cancelling the order is what prevents new exposure; refusing to record a fill that already happened does not. Suppressing adoption would leave `FilledSize` at zero, send the row down the "limit order cancelled with no fill" branch and delete it, orphaning a live on-chain position outside the virtual book where `forceCloseKillSwitchPositions` books nothing — the model-only accounting class that #1454 and #1455 closed.

### Deviation from the issue

The issue's acceptance criteria scope the roster to live `type=manual` strategies. Item 5 above widens it. The traced code supports this: `RunHyperliquidCancelOrder` sends `--mode=live` regardless of the strategy's own mode, and `manual-open --limit-price` (`scheduler/manual.go:54-122`) does not guard on `hyperliquidIsLive`, so a row can exist for a strategy the live-only roster would not resolve. Under the narrow roster such a row would latch the kill switch permanently with no recovery path; under the wide roster it is cancelled.

## Verification

- `go -C scheduler build ./...`, `go -C scheduler vet ./...` — clean.
- `go -C scheduler test ./...` — the same five `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` failures on the branch and on `origin/main` (macOS bash 3.2 has no `declare -A`); no other failure. Confirmed pre-existing by stashing the change and re-running.
- `go -C scheduler test -race -run 'KillSwitch|Limit' ./...` — pass.
- **Red → green:** with the three-line cancel pass removed from `planKillSwitchClose`, all 11 new plan tests fail; restored, all pass.

### Tests added (`scheduler/kill_switch_limit_orders_test.go`)

| Test | Pins |
|---|---|
| `CancelsRestingLimitOrderWithNoPosition` | order on a flat coin is cancelled and its row deleted; message names it; auto-reset stays available |
| `StillRestingLimitOrderBlocksFlatAndAutoReset` | latch holds, auto-reset suppressed, row kept, message names order + OID |
| `LimitCancelErrorBlocksFlat` | failed cancel blocks flat and never polls status |
| `LimitCancelUnverifiedBlocksFlat` (5 sub-cases) | status subprocess failure, `OpenOrdersError`, order absent, `FillsError`, `Resting == nil` |
| `UnadoptedLimitFillKeepsRowAndBlocksFlat` | row survives so the reconciler books the fill |
| `LimitRowDeleteFailureBlocksFlat` | off-book but uncleared row blocks flat |
| `LimitQueueLoadErrorBlocksFlat` | unreadable queue blocks flat and cancels nothing |
| `LimitOrderForUnknownStrategyBlocksFlat` | unresolvable row demands manual intervention |
| `LimitOrderCancellerUnwiredBlocksFlat` | wiring gap fails closed |
| `CancelsLimitOrdersBeforeFlatten` | cancel precedes the close call |
| `NoLimitOrdersLeavesPlanConfirmedFlat` / `NilLimitOrderLoaderIsSafe` | no behavior change without rows |
| `KillSwitchLimitOrderRoster` | roster spans every Hyperliquid strategy with a script |
| `ReconcilePendingLimitOrdersAdoptsFillWhileKillSwitchLatched` | a latched kill switch does not suppress adoption |

**No existing test was edited or removed.**

Docs: `docs/ARCHITECTURE.md` kill-switch section and the `CLAUDE.md` guardrail line record the row-keyed pass and the "never gate the reconciler on kill-switch state" invariant.

---


richkuo:

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T00:46:21Z) in d95736ee.

### Fixed

1. **The kill switch keeps an unadopted-fill row "so the scheduler books the fill" even when the reconciler can never book it, latching the kill switch forever and leaving a live on-chain position outside the virtual book.** — Confirmed. `reconcilePendingLimitOrders` (`scheduler/manual_limit.go:500-509`) `continue`s unless the strategy is present AND `type=="manual"` AND `hyperliquidIsLive(sc.Args)`, and `applyLimitFillProgress` / `UpdatePendingLimitOrderFill` have no other caller, so the deferral named a path that cannot run for such a row. New `killSwitchLimitOrderAdoptionBlock` (`scheduler/kill_switch_limit_orders.go:54-64`) mirrors that gate exactly and is carried on the candidate as `AdoptionBlock`; the unadopted-fill branch (`scheduler/kill_switch_limit_orders.go:196-206`) now defers to the reconciler only when the row is adoption-eligible, and otherwise escalates as manual-intervention naming the block and the recovery (restore the strategy as a Hyperliquid-live `type=manual` strategy). Neither shape is auto-deleted. Scope rule 1: the defect is in `cancelKillSwitchRestingLimitOrders`, which this PR adds.

2. **A pending-limit row whose strategy is absent from the config latches the kill switch permanently with no supported recovery path — even cancelling the order on the exchange never clears the row.** — Confirmed, including the remedy. Verified in `shared_scripts/check_hyperliquid.py:1483-1560` that `run_limit_status` and `run_cancel_order` take only `--symbol`/`--oid`/`--mode=live` and build an account-scoped `HyperliquidExchangeAdapter()` from env, and that `resting` comes from `adapter.open_order_oids(symbol)` with fills resolved by OID — so any Hyperliquid script answers for any OID. `killSwitchLimitOrderRoster` (`scheduler/kill_switch_limit_orders.go:43-52`) now spans every Hyperliquid strategy (script or not), and `collectKillSwitchLimitOrderCandidates` (`scheduler/kill_switch_limit_orders.go:66-95`) falls back to the first roster script for a row whose own strategy is gone. Only a config with no Hyperliquid script at all leaves a row uncancellable, and that keeps the manual-intervention wording. Scope rule 1: the refusal is in code this PR adds.

Two consequences worth calling out, both derived rather than stated in the findings:

- A row confirmed off-book with **no** unadopted fill is now deleted whatever its adoption eligibility. The row tracks the *order*; any position it left behind is already latched by the `Unconfigured` on-chain-position path in `planKillSwitchClose`, so keeping the row would block on a state another mechanism already owns.
- Your must-survive (a) on finding 2 works because `run_cancel_order` reports a failed cancel as `cancelled:false` + `cancel_error` with no `error` field (`shared_scripts/check_hyperliquid.py:1552-1558`), so an already-cancelled order falls through to the status re-read, which is the authority. Pinned by `TestPlanKillSwitchClose_OrphanedLimitRowCancelledViaFallbackScript`.

### Tests

Two tests this PR itself added were rewritten — **Outdated** under the CLAUDE.md rule, grounded in these two findings:

- `TestPlanKillSwitchClose_LimitOrderForUnknownStrategyBlocksFlat` → `..._NoHyperliquidScriptInConfigBlocksFlat`. Ground: finding 2 authorizes cancelling the unknown-strategy row, so the blocking case it asserted is no longer correct; the replacement asserts the same fail-closed behavior for the state that genuinely cannot be cancelled (no Hyperliquid script anywhere in config), including that no cancel is attempted.
- `TestCollectKillSwitchLimitOrderCandidates` and `TestKillSwitchLimitOrderRoster`. Ground: findings 1 and 2 change the roster and the candidate shape; the replacements assert that every row resolves a script while any Hyperliquid script exists, and that each adoption block names its reason.

Five cases added, one per must-survive: `OrphanedLimitRowCancelledViaFallbackScript` (2a), `OrphanedLimitRowWithFillEscalates` (2b), `NoHyperliquidScriptInConfigBlocksFlat` (2c), `AdoptionIneligibleLimitFillNamesTheBlock` with paper and type-change subtests (1a, 1b), and `AdoptionEligibleLimitFillStillDefersToReconciler` (1c). Plus `CollectKillSwitchLimitOrderCandidatesWithoutAnyScript`.

Verified red on the pre-fix code (all four rewritten/new plan tests fail with the roster, fallback and adoption branches reverted) and green after. `go -C scheduler build ./...` and `go vet` clean; full `go test ./...` shows only the five pre-existing `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` failures that also fail on `origin/main` (macOS bash 3.2 has no `declare -A`; both CI jobs are green).

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T01:20:39Z) in b69df9a1.

### Fixed

1. **The pre-flatten cancel pass has no aggregate time budget, so a degraded Hyperliquid API delays the live-position flatten by minutes on every latched cycle.** — Confirmed. `runPythonSideEffect` → `runPython` → `runPythonWithTimeout(..., scriptTimeout)` (`scheduler/executor.go:126-128, 213-217`), so each of the two calls per row is capped at 30s with nothing bounding the walk. `cancelKillSwitchRestingLimitOrders` now takes a budget (`scheduler/kill_switch_limit_orders.go:117`), resolved from the new `HLLimitOrderTimeout` through `platformCloseBudget` (`scheduler/kill_switch_close.go:288-290`; 60s at the `main.go` call site). It refuses to **start** a call unless `killSwitchLimitOrderCallReserve` (= `scriptTimeout`) of the budget remains (`scheduler/kill_switch_limit_orders.go:156-160, 176-185, 207-210`), so the pass cannot overrun rather than merely checking between rows; unreached rows and an issued-but-unverified cancel are reported unresolved, which already clears `OnChainConfirmedFlat`. `killSwitchLimitOrderBudget` floors at `2*scriptTimeout` (`:43-48`) so one complete row always fits — your must-survive (b) is the opposite failure and I did not want the budget to create it. Scope rule 1: the unbounded pre-flatten leg is what this PR adds.

2. **The pass deletes the queue row without first flushing the fill the reconciler adopted earlier in the same cycle, dropping the only durable record of a real on-chain position.** — Confirmed, including the reachability. `reconcilePendingLimitOrders` guards the identical step at `scheduler/manual_limit.go:586-596` (`SaveStateWithDB`, then `continue` on failure), and the adoption branch at `:530-565` persists only the watermark via `UpdatePendingLimitOrderFill` while the position stays in memory. New `deps.Flush` (`scheduler/kill_switch_limit_orders.go:24`), wired to a `mu`-held `SaveStateWithDB` in `scheduler/main.go:1111-1115`, runs before `Delete` whenever `o.FilledSize > 0` (`scheduler/kill_switch_limit_orders.go:255-260`); a failed flush keeps the row and blocks confirmed-flat, and a zero-fill row skips it so it cannot latch forever. Scope rule 1: the unguarded delete is in code this PR adds.

1. **The pass never marks the row `cancel_requested`, so a cancel it started is abandoned the moment the operator resets the latch.** — Confirmed. `reconcilePendingLimitOrders` cancels only on `o.CancelRequested || expired` (`scheduler/manual_limit.go:602-604`), so after a DM reset clears `killSwitchFired` nothing finishes an unverified cancel. `MarkPendingLimitOrderCancelRequested` now runs before the cancel (`scheduler/kill_switch_limit_orders.go:187-191`), matching `clearRestingLimitRemainderForPositionAction`. Implemented rather than filed: it is money-path and scope rule 1 — the half-cancelled state exists only because this PR issues the cancel. A failed mark does **not** stop the cancel (the cancel is the safety action); it is appended to that row's unresolved reason so the operator is told a reset would drop it (`:166-174`).

On your must-survive (c) for that item — an adoption-ineligible row: the mark is inert there. Such a row was already outside the reconciler's gate before this change, and `TestPlanKillSwitchClose_NoFillNeedsNoFlush` pins that it is still deleted once confirmed off-book, so the mark creates no state only the kill-switch pass could clear.

### Tests

The shared stub harness `killSwitchLimitStubs` gained an injected clock, a flush counter, a mark recorder and failure injectors. That is this PR's own helper rather than a landed test, so no CLAUDE.md test-edit case applies; calling it out because the diff touches existing test lines.

Ten cases added, one per must-survive:

| Case | Test |
|---|---|
| 1a every call times out, flatten still runs | `LimitCancelBudgetStillLetsTheFlattenRun` (asserts the closer ran and all four rows are unresolved) |
| 1b slow-but-successful row not abandoned | `LimitCancelBudgetAllowsOneSlowRow` |
| 1c zero rows, no latency | `NoLimitOrdersLeavesPlanConfirmedFlat` (now also asserts the clock is never consulted) |
| budget floor | `KillSwitchLimitOrderBudgetFloor` |
| exhaustion between cancel and verify | `LimitCancelBudgetExhaustedBeforeVerification` |
| 2a already-durable fill, delete proceeds | `FlushesAdoptedFillBeforeDeletingTheRow` |
| 2b flush fails, row kept, not flat | `FailedFlushKeepsTheRecoveryRow` |
| 2c zero-fill ineligible row still deleted | `NoFillNeedsNoFlush` |
| 3a mark precedes cancel | `RecordsCancelIntentBeforeCancelling` |
| 3b fill adopted on a `cancel_requested` row | `ReconcilePendingLimitOrdersAdoptsFillOnACancelRequestedRow` |
| failed mark surfaced | `FailedCancelIntentIsSurfaced` |

Verified red on the pre-fix code — with the two budget gates, the flush gate and the mark removed, all six new plan tests fail — and green after. `go build`, `go vet` and `go test -race -run 'KillSwitch|Limit'` clean; full `go test ./...` shows only the five pre-existing `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` failures that also fail on `origin/main` (macOS bash 3.2 has no `declare -A`; both CI jobs green).

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T02:30:46Z) in b69df9a1 (no new commit — the one finding was deferred, see below).

### Deferred to follow-up

1. **Adoption-ineligible resting orders are re-cancelled only while the latch is held, so a persistent cancel failure plus an operator latch reset can leave such an order live and no longer retried.** — Confirmed against current code: the cancel pass runs only from `planKillSwitchClose` (`scheduler/kill_switch_close.go:286`) while `killSwitchFired` holds, an owner DM reset clears that latch (`scheduler/main.go:1277-1297`), and `reconcilePendingLimitOrders` skips an absent / non-`manual` / non-HL-live row at `scheduler/manual_limit.go:501-508` before reaching its `o.CancelRequested || expired` cancel branch at `scheduler/manual_limit.go:604-620`. Out of scope, basis scope rule 2: the remedy is a cancel-only convergence lane the reconciler has never had — it needs roster and fallback-script resolution threaded in (`scheduler/kill_switch_limit_orders.go:63-113`), its own no-fill-booking row-retention rule to avoid the orphaned-fill class #1462 Approach step 3 forbids, and a throttled owner alert for the per-tick cadence. #1462's yardstick asks only that a kill-switch flatten cancel resting limit orders and keep the latch and the confirmed-flat report honest; its first acceptance criterion scopes the guarantee to Hyperliquid-live `type=manual` strategies. The hazard is pre-existing — `origin/main` cancelled no resting limit order at all — and this PR narrows it rather than creating it, so scope rule 1 does not match. Filed as #1470.

Not implemented here, and nothing else changed: CI is green on both jobs, no formal reviews and no open inline threads remain.

---

richkuo:
@claude sonnet review

PR #1471, opened 2026-08-29
chore(ci): track rk-skills prompts via -local appends [C25, Opus 5, high]

richkuo:
## Plain simple English

Our review and fix instructions for the @claude bot were full copies of the shared ones in the rk-skills repository. A copy stops the shared updates from reaching us, and ours had fallen behind. This change keeps only the few lines that are special to this repository. The shared text now updates by itself. It also deletes four script copies that nothing runs.

## Summary

`richkuo/rk-skills/.github/workflows/claude-run.yml@main` stages the shared prompts and comment-patching scripts into `$RUNNER_TEMP` on every run, and consults `.github/prompts/<name>.md` as a full REPLACEMENT or `.github/prompts/<name>-local.md` as an APPEND (both read from `origin/main`).

This repo carried full replacements, so the shared body was shadowed and frozen at whatever it said when the copies were made.

- `pr-review-format.md` → `pr-review-format-local.md`. Delta kept: backtests named in the no-execution ban, and this repo's auto-protective mechanisms named in the safety carve-out (kill switch, circuit breaker, stop-loss and trailing-stop replacement, liquidation clamping, reconciliation, position/fill/realized-PnL accounting).
- `fix-pr-review.md` → `fix-pr-local.md`. Delta kept: backtests in the no-execution ban, gofmt on edited Go files, tests written as code without running them. The old file was also dead under the current router — the legacy override path is consulted only for mode `fix-pr-review`, and this repo routes mode `fix-pr`, which looks for `fix-pr.md`.
- Deleted the unused `.github/scripts` copies of `compose_claude_comment.py`, `patch_claude_comment.sh`, `rewrite_create_pr_link.py`, `strip_llm_footer.py`.
- `.github/scripts/test_workflow_logic.py` stays, with its `testpaths` entry: it extracts and executes the real `run:` blocks out of this repo's own `claude.yml`, so it covers repo-owned routing rather than the shared scripts.
- `CLAUDE.md` and `docs/ARCHITECTURE.md` updated for the removed paths, and the documented field order corrected to the shared file's, where **Plain simple English:** is every finding's last field.

## Test edits

Removed `test_compose_claude_comment.py`, `test_patch_claude_comment.py`, `test_rewrite_create_pr_link.py`, `test_strip_llm_footer.py`. Case: **obsolete by deletion of the covered behavior**. Ground: the four scripts they import by name are deleted from this repo, and `claude-run.yml` runs those scripts from `$RUNNER_TEMP` staged out of rk-skills, never from this tree, so no code path here reaches them; rk-skills carries the same four tests beside its own copies. No replacement assertions here.

## Verification

- `pytest --collect-only` collects 3463 tests with no errors.
- `pytest .github/scripts` passes 31 tests.
- Both new prompt files are free of the `"`, `` ` ``, and `$` characters the `claude_args` guard rejects.

## Note

The prompt override is read from `origin/main`, so this takes effect once merged. rk-skills' own newer `test_workflow_logic.py` additionally pins a `PERL_STRIP_FENCED`/`SED_STRIP_INLINE` env-var form that this repo's `claude.yml` does not use; adopting it would mean changing the workflow, which is out of scope here.

---


richkuo:
@claude review

richkuo:
Disposition of the review on `0159294`.

### Resolved judgment calls

1. **Confirm rk-skills `claude-run.yml@main` actually appends `pr-review-format-local.md` / `fix-pr-local.md`, or the go-trader review delta is silently dropped.**
Confirmed against `richkuo/rk-skills` `origin/main` (`.github/workflows/claude-run.yml`), so option A. The route sets `PROMPT_FILE=$PROMPTS_DIR/pr-review-format.md` for `review` (line 142) and `PROMPT_FILE=$PROMPTS_DIR/fix-pr.md` for `fix-pr` (line 139). Line 167 derives `LOCAL_PATH=".github/prompts/$(basename "$PROMPT_FILE" .md)-local.md"`, which yields exactly `.github/prompts/pr-review-format-local.md` and `.github/prompts/fix-pr-local.md`. Line 182 appends it: `PROMPT="$PROMPT $(git show "$TRUSTED_REF:$LOCAL_PATH" | flatten)"`, where `TRUSTED_REF` is the fetched default branch, so the append takes effect once this merges to `main` and not before. The base names and the suffix are the ones this PR uses; no re-wiring is needed. You were right that this repo cannot verify it — the mapping lives entirely in rk-skills, and the review route has no fetch tool. Worth watching the next `@claude review` after merge anyway, since that is the only end-to-end proof.

### Fixed

1. **CLAUDE.md still documents "Plain simple English" NOT last, contradicting the field order this PR set in ARCHITECTURE.md.** — `CLAUDE.md:98`, commit `29f549b`. The sentence now reads: Needs Fixing/Optional require **Invariant:** + **Must survive:**; Requires Human Review requires **Recommended proposed solution:**; every finding then ends with **Plain simple English:** as its LAST field, after those where they apply. Both docs now state the shared file's order. Must-survive cases (b) and (c) hold: the wording puts Plain simple English after **Recommended proposed solution:** on a Requires Human Review item, and its "then ends with" phrasing is unconditional, so a finding carrying none of the other fields still ends with it.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T03:49:18Z) in `49b15ca`.

### Fixed

1. **`fix-pr-local.md` pins its addition to "Phase 4" by number, re-coupling the append to the shared file's structure this PR set out to decouple from.** — dropped the number; the line now reads "The no-execution ban covers backtests as well: do NOT run backtests, optimizers, tuning runs, or parity harnesses." (`.github/prompts/fix-pr-local.md:3`). Confirmed against `rk-skills` `origin/main`: `templates/claude-workflow/prompts/fix-pr.md` carries Phase 1 through Phase 8 today, and the ban sits in Phase 4, so the citation was accurate now and fragile later — which is the coupling you named. Must-survive cases all hold: (a) and (b) the append names no phase at all, so a renumber or a move to named sections leaves it correct; (c) the file remains free of the double-quote, backtick and dollar characters the `claude_args` guard rejects, and both local files now share the number-free convention.

Verification: `pytest .github/scripts` passes 31 tests; the guard-character check on both `-local.md` files returns no matches. Only the one prompt file changed.

---

richkuo:
@claude sonnet review

PR #1472, opened 2026-08-29
refactor(#1465): rename trailing_stop_atr_regime to trail_stop_atr_regime [C68, Opus 5, xhigh]

### LINKED ISSUE #1465: [C68] The regime trail-stop field name is easy to confuse with the ratchet field name
**Complexity: 68/100** — Capability 2 (Coupling 3 — Go config schema, Python backtest parity, the bankruptcy-bound preflight script, the migration ladder and SIGHUP reload; Risk 3 — the field owns the live Hyperliquid stop distance); Volume 18 — Opus 5, xhigh · fableplan: no

## Problem

`trailing_stop_atr_regime` and `trailing_tp_ratchet_regime` both start with `trailing_`. When you scan a config it is easy to take the wrong one. The two have different jobs: one sets the stop distance, the other holds the take-profit ladder that tightens that distance.

This is a readability problem only. No code path confuses the two.

### What the field actually does (validated 2026-08-28 against `origin/main` be0072eb)

An earlier version of this issue said the field is a *fixed opening stop distance* that the ratchet may later tighten, and that the stop does not chase price until a ratchet rung fires. **The code refutes both statements.**

- The field feeds `effectiveTrailingStopPct` (`scheduler/hyperliquid_trailing_stop.go:100-112`), which drives a true high-water walker: `candidateHighWater` advances with every favorable mark, the trigger is re-derived as `highWater * (1 ± pct)`, and the resting reduce-only order is cancelled and replaced on-chain (`scheduler/hyperliquid_trailing_stop.go:355-404`, `:502-520`).
- That walker runs on every due cycle for any open Hyperliquid perps position with `result.Signal == 0` — paper at `scheduler/main.go:1912`, live at `scheduler/main.go:1934`, manual at `scheduler/main.go:2265`. No ratchet is involved.
- A ratchet rung only changes the *width* of that trail, through `PostTPTrailingATRMult` (`scheduler/trailing_tp_ratchet.go:428-440`).
- The field is not scoped to the ratchet. It is a standalone stop owner on Hyperliquid perps; validation requires only Hyperliquid perps, or Hyperliquid `manual` with the ratchet close (`scheduler/regime_atr.go:660-665`), and `user_defaults.regime_atr` seeds standalone `use_defaults`-only owners (`scheduler/close_defaults.go:176-187`, #1134).

**Therefore `opening_stop_atr_regime` is rejected as the target name.** It would be less accurate than the current name, because the stop trails the high-water mark for the whole life of the position.

The per-regime default table is confirmed accurate (`scheduler/regime_atr.go:168-181`), and covers more labels than the earlier excerpt listed:

```
Trailing: {
  "trending_up":              2.5,   "trending_down":              2.5,
  "ranging":                  2.0,
  "trending_up_clean":        2.5,   "trending_down_clean":        2.5,
  "trending_up_choppy":       2.25,  "trending_down_choppy":       2.25,
  "ranging_quiet":            1.0,
  "ranging_volatile":         1.25,
  "ranging_directional":      1.5,
  "ranging_directional_up":   1.5,   "ranging_directional_down":   1.5,
}
```

## Goal

Remove the `trailing_` prefix collision between the stop-distance field and the ratchet close-evaluator key, without changing any runtime behavior, and without giving the field a name that contradicts what it does.

Keep the word **trailing** in the new name. The field owns a trailing stop.

### Choose the name

Preferred: `trail_stop_atr_regime` — drops the shared `trailing_` prefix, keeps the correct verb, and stays parallel with `stop_loss_atr_regime`.

Alternatives, both rejected:
- `opening_stop_atr_regime` — refuted above; describes behavior the code does not have.
- `sl_atr_regime` — collides conceptually with `stop_loss_atr_regime`, which is the *fixed* (non-trailing) owner.

Also rename the Go symbol `TrailingStopATRRegime` and the constant `userCloseDefaultTrailingStopATRRegimeKey` to match, so the JSON tag and the symbol never diverge.

### Accept the legacy key

The strategy-key allowlist is reflected from the `StrategyConfig` struct tags (`scheduler/config_unknown_keys.go:11-37`). Renaming the tag therefore makes `trailing_stop_atr_regime` an **unknown field** that fails the load outright (`scheduler/config.go:825-829`). A migration is mandatory, not optional.

Three independent layers need alias handling. Only the first is auto-derived from the struct tags:

1. **Go migration** — add a v18 handler beside `migrateV15StrategyRegimeBlocks` (`scheduler/config_migration_v15.go:76-82`), bump `CurrentConfigVersion` 17 → 18 (`scheduler/config_migration.go:11`), gate it in `MigrateConfig` as `oldVer < 18`, and add a `needsV18Migration` branch in `loadConfig` mirroring the v15/v16 branches (`scheduler/config.go:803-819`). It must run **before** `validateStrategyJSONKeys`, which it does at that position. Rewrite the key in the strategy block, in `user_defaults.regime_atr`, and in `user_defaults.close["trailing_tp_ratchet_regime"]` (#1133).
2. **Go string-literal allowlists** — `scheduler/close_defaults.go:176,180` and `:96-102` are hand-written key lists, not struct-tag derived. They need the new key plus their operator messages updated.
3. **Python `--config` reader** — `backtest/run_backtest.py` reads raw JSON and never runs the Go migration. It needs its own alias, or research fixtures must be rewritten in the same pull request.

### Migration writes to disk

`MigrateConfig` is not an in-memory pass. It rewrites the config file: `os.WriteFile(tmpPath, ...)` then `os.Rename(tmpPath, configPath)` (`scheduler/config_migration.go:168-174`).

Two consequences the work must handle:

- **Symlink clobber.** Where `scheduler/config.json` is still the #1056 transition symlink, `os.Rename` replaces the symlink with a regular in-tree file. This is the exact hazard `scripts/migrate-config-out-of-tree.sh` refuses to risk. Deployments already pointed at `/var/lib/go-trader[/<instance>]/config.json` through `ExecStart --config` are unaffected. The rollout note must say so.
- **No state.db churn.** Confirmed: no column or persisted field carries this key (`scheduler/db.go`, `scheduler/state.go` have no reference). `trade_protection_snapshot.go:29-31` stores only the *resolved multiple*, a number.

### Complete touch set

The earlier scope list named four Go files. The verified set is much larger.

**Go, runtime (18 files):** `config.go` (tags at `:218`, `:504`; 11 symbol sites), `regime_atr.go`, `close_defaults.go`, `closing_strategies.go`, `config_reload.go` (`:231-233`, `:585-591`, `:703`, `:954`, `:1007` — SIGHUP diff and blocked-field messages), `trailing_tp_ratchet.go`, `hyperliquid_trailing_stop.go`, `hyperliquid_liquidation_guard.go`, `risk_sizing.go`, `scale_in.go`, `manual.go`, `manual_core.go`, `manual_sl.go`, `regime_unified.go`, `trade_protection_snapshot.go`, `discord.go`, `inspect_cmd.go`, `ui_tuner.go` (`:687` payload key).

**Go, new:** the v18 migration handler.

**Go, tests (16 files):** `config_reload_test.go`, `close_defaults_test.go`, `regime_atr_test.go`, `closing_strategies_test.go`, `manual_ratchet_regime_default_test.go`, `regime_iszero_resolution_test.go`, `regime_atr_composite_test.go`, `hyperliquid_liquidation_config_test.go`, `config_audit_test.go`, `config_unknown_keys_test.go`, `hyperliquid_stop_loss_test.go`, `trailing_tp_ratchet_test.go`, `discord_test.go`, `config_test.go`, `manual_sl_test.go`, `hyperliquid_open_trailing_test.go`, plus a new migration round-trip test.

**Preflight script:** `scripts/check-hl-stop-bankruptcy-bound.sh:45,93,111`. It mirrors LoadConfig's stop-owner resolution and is pinned to the Go check by a parity test (#1456 round 4). Missing it makes the preflight clear a config the daemon then rejects.

**Python parity:** `backtest/run_backtest.py` (`stop_keys` allowlist `:394`, `--config` validation `:515-551`, `user_defaults.regime_atr` fan-out `:575-597`, `Backtester` keyword `:1004`, `:1123`, `:1539`), `backtest/backtester.py` (19 references), `backtest/exit_policy_ab.py`, `backtest/tune_live.py`, `shared_scripts/simulate_strategy.py`, `shared_strategies/close/test_regime_atr.py`, and 6 pytest files under `backtest/tests/`.

**Research fixtures:** 16 files under `backtest/research/` carry the key — 14 JSON configs (`regime_1152_runs/*.json`, `regime_1152_config_ratchet*.json`) plus `regime_1081_economic_gate.py` and `regime_1120_trail_validation.py`. These are read by `run_backtest.py --config`, which has no Go migration path.

**Tuning surface:** `scheduler/ui_tuning_apply.go` — a schema-v2 `promotion_baseline` recorded under the old key must still compare cleanly, or `POST /api/tuning/apply` refuses on false drift.

**Docs:** `CLAUDE.md`, `SKILL.md` (`:311`, `:797`, `:848`, `:851`, `:853`), `README.md:186`, `docs/ARCHITECTURE.md` (`:15`, `:37`, `:78`).

## Acceptance criteria

- [ ] The new key is accepted in strategy blocks, `user_defaults.regime_atr`, and `user_defaults.close["trailing_tp_ratchet_regime"]`.
- [ ] `trailing_stop_atr_regime` is still accepted, rewritten by the v18 migration, and logs one migration line per strategy.
- [ ] `CurrentConfigVersion` is 18; `MinSupportedConfigVersion` stays 13.
- [ ] The migration runs before `validateStrategyJSONKeys`, proven by a test that loads a v17 config carrying only the legacy key.
- [ ] Every site in the touch set above is updated, including the two hand-written allowlists in `close_defaults.go`.
- [ ] `scripts/check-hl-stop-bankruptcy-bound.sh` accepts both keys and its parity test against `LoadConfigForProbe` still passes.
- [ ] `backtest/run_backtest.py --config` accepts both keys; the 16 research fixtures still run.
- [ ] SIGHUP reload keeps its current blocked-field behavior under the new key, with an open position and with none.
- [ ] No state.db schema change.
- [ ] A live-shaped config that arms a trailing stop produces the same trigger price before and after the rename, proven by a test.
- [ ] Go build, `go test ./...`, and the full pytest suite pass; docs updated.

## Plain simple English

Two settings have almost the same name, so an operator can change the wrong one. An earlier version of this issue also described one setting incorrectly. It said the stop stays still until a profit step fires. In fact the stop follows the price up on every cycle.

---
Validated with LLM: Opus 5 | high | Harness: Claude Code

richkuo:
## Plain simple English

Two settings had almost the same name, so an operator could change the wrong one. One sets how far the stop follows the price. The other holds the profit steps that tighten that stop. The first one now has a new name. The daemon changes the name in your config file when it starts. Nothing about trading changes.

## Summary

Renames the per-regime trailing-stop field `trailing_stop_atr_regime` → `trail_stop_atr_regime`, so it no longer shares the `trailing_` prefix with the `trailing_tp_ratchet_regime` close evaluator. The Go field `StrategyConfig.TrailStopATRRegime` and the `userCloseDefaultTrailStopATRRegimeKey` constant move with the tag, so the JSON key and the symbol cannot diverge.

Runtime behavior is unchanged. The field still owns the high-water trailing stop on Hyperliquid perps.

### Migration (mandatory, not optional)

The strategy-key allowlist is reflected off the `StrategyConfig` struct tags, so renaming the tag alone would make `trailing_stop_atr_regime` an unknown field that fails the load outright.

- New `scheduler/config_migration_v18.go`. `CurrentConfigVersion` 17 → 18; `MinSupportedConfigVersion` stays 13.
- Rewrites the key in strategy blocks, `user_defaults.regime_atr`, `user_defaults.manual`, and every `user_defaults.close` entry, logging one line per site.
- `loadConfig` runs it **before** `validateStrategyJSONKeys`, mirroring the v15/v16 branches. `needsV18TrailStopKeyMigration` also fires on a v18-stamped config that still carries the legacy key.
- Two spellings holding different blocks is a load error. Identical blocks drop the legacy one.
- `config_migration_v15.go`'s raw-JSON key lists carry **both** spellings, because v15 runs before v18 inside `MigrateConfig`.
- Owner DM / stdout notice `v18TrailStopRenameNotice`.

### Layers that read raw JSON and never run the Go migration

- `backtest/run_backtest.py` normalizes at `load_strategy_config`, so the research fixtures and any un-migrated operator config still run.
- `scripts/check-hl-stop-bankruptcy-bound.sh` normalizes at load. Without this the preflight would block a healthy config the daemon migrates and accepts (verified: the un-normalized script reports `max_drawdown_pct = 15% >= bound 5%` on a legacy-key fixture).
- `shared_scripts/simulate_strategy.py` falls back to the legacy payload key.

### Other surfaces

- `promotionBaselinesEqual` compares baselines with the key normalized, so a schema-v2 `promotion_baseline` recorded under the old spelling does not read as false drift and refuse `POST /api/tuning/apply`. A genuine baseline change still reads as drift (pinned by test).
- The load-time v18 branch is keyed on the legacy key **being present**, never on `config_version < 18`. `MigrateConfig` writes the file, so a version-keyed branch made every sub-v18 load rewrite the config and turned a non-writable config path (a read-only deploy tree under `ProtectSystem=strict`, where #1056's transition symlink still lives) into a refused boot for a config that had always loaded. A write failure stays fatal only when the key is present — a config whose on-disk key cannot be canonicalized must not load.
- Because the rename always consumes the legacy key before any validator runs, **no unknown-key hint for it can be reached**; an earlier revision added three such hints plus an `inspect` alias and they were removed as dead.
- 18 runtime Go files, 16 Go test files, the Python parity layer, the 16 research fixtures, and `CLAUDE.md` / `SKILL.md` / `README.md` / `docs/ARCHITECTURE.md` updated. `docs/POST_UPDATE_HISTORY.md` gains a v17→v18 entry; its older bullets and `docs/research/1152-ranging-exit-geometry-m6.md` keep the old key because they are records of past states.

### Rollout note — #1056 symlink

`MigrateConfig` rewrites the config file (`os.WriteFile` + `os.Rename`). Where `scheduler/config.json` is still the #1056 transition symlink, the rename replaces the symlink with a regular in-tree file. Deployments already pointed at `/var/lib/go-trader[/<instance>]/config.json` through `ExecStart --config` are unaffected. The migration notice states this.

No `state.db` change: no column or persisted field carries this key; `trade_protection_snapshot.go` stores only the resolved multiple.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test ./...` and `go test -race ./...`: pass except 5 pre-existing failures (`TestUpdateShell*`, `TestUpdateHelpersEnvfileParsing790`) that also fail on `origin/main` in this environment — macOS bash 3.2 has no `declare -A`. Unrelated to this change.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto`: 3550 passed, 1 skipped. `shared_scripts/`: 250 passed. `py_compile` clean on every changed Python file.
- Both real research configs load through `run_backtest.py --config` with no legacy key surviving.
- `go-trader init --json` writes `"config_version": 18`.

### New tests (each run red against the unfixed code first)

`scheduler/config_migration_v18_test.go`:
1. A v17 config carrying **only** the legacy key loads, resolves the block, and is rewritten on disk at version 18 — this is what proves the migration runs before `validateStrategyJSONKeys`.
2. Same for `user_defaults.regime_atr` and `user_defaults.close[trailing_tp_ratchet_regime]`.
3. Conflicting spellings with different blocks are rejected, naming both keys.
4. Identical blocks drop the legacy key.
5. `needsV18TrailStopKeyMigration` fires on a v18-stamped config with the legacy key, and not on a clean one.
6. A live-shaped config produces the **same** `effectiveTrailingStopPct` before and after the rename, across all three regime labels.
7. A promotion baseline under the legacy key does not read as drift; a genuine change still does.

`scheduler/hyperliquid_liquidation_config_test.go`: the bankruptcy-bound preflight accepts a legacy-key config, and `LoadConfigForProbe` accepts the same file.

`backtest/tests/test_user_close_defaults_regime_atr.py`: `--config` accepts the legacy key; legacy and canonical spellings produce identical kwargs; the canonical key wins when both are present.

SIGHUP blocked-field behavior under the new key, with an open position and with none, is covered by the existing `config_reload_test.go` cases (renamed key, assertions unchanged).

## Changed existing tests

- `scheduler/regime_test.go`: `TestCurrentConfigVersion_IsSeventeen` → `TestCurrentConfigVersion_IsEighteen`. **Case: Outdated.** **Ground:** the issue's acceptance criterion, "`CurrentConfigVersion` is 18; `MinSupportedConfigVersion` stays 13." The replacement asserts `CurrentConfigVersion == 18`. `TestMinSupportedConfigVersionFloor` is untouched and still pins the floor at 13.
- 16 Go test files and 8 Python test files carry the renamed key/symbol only. No assertion, fixture value, or expectation was changed in any of them.

Closes #1465

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T05:12:01Z) in 56af5da3.

### Fixed

1. **The v18 migration notice is never delivered, because `loadConfig` bumps the in-memory `config_version` to 18 before `main.go`'s notice gate runs.** — Confirmed exactly as described, and PR-caused, so it is fixed here rather than deferred (scope rule 1: the unreachable gate is created by this PR's own `needsV18TrailStopKeyMigration` branch at `scheduler/config.go:838`, which fires for every sub-v18 config). `cfg.ConfigVersion` was the wrong signal — *any* load-time branch rewrites it — so the gate now reads `Config.MigrationBaseVersion()` (`scheduler/config.go:175`), the version read from disk **before** any branch fires, captured at `scheduler/config.go:807` and stamped at `:851`. `main.go:480` and every notice in `runConfigMigrationDM` read it.

Three things beyond the literal finding, all from the same root cause and none needing a mechanism this PR lacks:

- **The pre-existing v13/v14 hole is closed by the same change.** On `origin/main` a v13 or v14 config already lost the v17 notice, because `needsV15CloseMigration` rewrote it to 17 before the gate. Your finding named the v16→v17 path as the working baseline, which it was; the lower versions were already broken. Basing the gate on the pre-migration version fixes both (`scheduler/config_migration.go:286`).
- **A version-less config had to be pinned, or the fix would have introduced its own regression.** `rawConfigVersion` returns 0 for an unstamped file, which would newly hand it the v14 `allow_shorts` and v15 close-key notices. #1285 defines a version-less config as current-shape, so the base pins to `CurrentConfigVersion` (`scheduler/config.go:808`). Pinned by test.
- **`configMigrationNotices(base)` is now the single ordered notice list** (`scheduler/config_migration.go:65`). The three delivery paths each carried their own copy and the `len(fields)==0` path omitted `v15DeprecationNotice` — and since #1285 emptied `configFieldRegistry`, that is the only reachable path, so the v15 notice was dead code.

**Your "Must survive" cases, each pinned by a test in `scheduler/config_migration_v18_test.go`:**

| Case | Result |
|---|---|
| v13, v14, v15, v16, v17 → v18 each deliver the notice once | ✅ `TestMigrationBaseVersionSurvivesLoadTimeMigration` (subtest per version) |
| clean v18 emits no notice and no rewrite | ✅ `TestMigrationBaseVersionCurrentConfigSendsNoNotice` (also byte-compares the file) |
| v18-stamped config carrying the legacy key renames without a per-restart duplicate notice | ✅ `TestMigrationBaseVersionLegacyKeyAtCurrentVersionSendsNoNotice` |
| version-less config is current-shape, gets no upgrade notice | ✅ `TestMigrationBaseVersionVersionlessConfigIsTreatedAsCurrentShape` |
| notice list ordering and version scoping | ✅ `TestConfigMigrationNoticesAreOrderedAndVersionScoped` |

Every one was run red against the unfixed code before the fix landed.

**Verification:** `go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./...` and `go test -race ./...` pass except five `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` failures that reproduce identically on `origin/main` in this environment (macOS bash 3.2 has no `declare -A`) — unrelated to this PR. No Python file changed this cycle; the 3550-test pytest run from the previous push still stands. `docs/ARCHITECTURE.md` records the new load-path invariant.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T05:30:18Z) in 3a70c762.

### Fixed

1. **The three "unknown key — renamed to `trail_stop_atr_regime` in #1465" hint messages this PR adds are unreachable in every call path that could trigger them.** — Confirmed, and your reachability argument holds end to end. I re-derived it rather than taking it: the only production callers are `validateStrategyJSONKeys` / `validateUserDefaultsJSONKeys` at `scheduler/config.go:853-854`, `validateUserDefaults` at `scheduler/config.go:1370` (reached from `validateConfig`, which `loadConfig` runs last), and `loadStrategyExplicitKeys` at `scheduler/inspect_cmd.go:36` and `scheduler/main.go:121` — every one behind a `LoadConfig` that has already renamed the key or failed the load on a spelling conflict. All four additions removed (`config_unknown_keys.go`, the two in `close_defaults.go`, the `inspect_cmd.go` alias), and **the PR body line they backed is corrected** — thank you for catching a claim that did not match the code.

2. **Boot regression this PR caused: a clean sub-v18 config could not load from a non-writable config path.** — Found while checking your reachability argument, and fixed here under scope rule 1 (the hazard is created by this PR's own `needsV18TrailStopKeyMigration`). `MigrateConfig` **writes the file** — `os.WriteFile` of a `.tmp` beside the config, then `os.Rename` (`scheduler/config_migration.go:214-217`). Keying the load-time branch on `config_version < 18` therefore made *every* sub-v18 load rewrite the config on disk, so a read-only config directory turned a config that had always loaded into a refused boot:

   ```
   v18 trail_stop_atr_regime key migration: write tmp: open .../config.json.tmp: permission denied
   ```

   Reproduced against a read-only directory, and confirmed PR-caused: with the v18 branch disabled the identical config loads clean. The shape is reachable — `ProtectSystem=strict` makes the deploy tree read-only and #1056's transition symlink still sits in it, so the `.tmp` lands there. Before this PR nothing wrote during `loadConfig` for a v15+ config; the only writer was `main.go`'s `runConfigMigrationDM`, whose write failure is non-fatal by design.

   The branch is now keyed on **the legacy key being present** (`scheduler/config_migration_v18.go:14-20`). v15 and v16 keep their version gates because they do shape work; v18 only renames a key, so presence is the exact condition. A write failure stays fatal when the key *is* present — a config whose on-disk key cannot be canonicalized must not load, since the unknown-key guard would reject it anyway. A clean config is left untouched and `runConfigMigrationDM` applies the version stamp as before.

**Your "Must survive" case** — a config carrying only the legacy key loaded via `LoadConfig` / `LoadConfigForProbe` / `runInspect` / `applyStrategyConfigPatch`, with the migration consuming the key before any validator or alias branch runs — still holds after the presence-gating change, because the branch fires precisely when that key exists. The prior cycle's tests pass unchanged: their fixtures carry the legacy key. `docs/ARCHITECTURE.md` now records both the presence-gating rationale and the rule that no unknown-key hint for this key can be reached, so it is not re-added.

**New tests** (`scheduler/config_migration_v18_test.go`), the two behavioral ones run red against the unfixed code first:

| Case | Test |
|---|---|
| a clean sub-v18 config loads from a read-only directory and still reports base 17, so the notice fires | `TestLoadConfigCleanConfigNeedsNoWriteToLoad` |
| a legacy-key config still fails the load loudly when the rename cannot be persisted | `TestLoadConfigLegacyKeyStillFailsLoudWhenItCannotBeRewritten` |
| the branch keys on the legacy key, not the version | `TestNeedsV18MigrationIsKeyedOnTheLegacyKeyNotTheVersion` |

Both permission tests skip themselves under root or on a filesystem that does not enforce directory permissions, so they cannot pass vacuously in one environment and fail in another.

**Verification:** `go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./...` and `go test -race ./...` pass except five `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` failures that reproduce identically on `origin/main` here (macOS bash 3.2 has no `declare -A`) — unrelated to this PR. No Python file changed this cycle.

---

richkuo:
@claude sonnet review

PR #1473, opened 2026-08-29
feat(#1427): test whether the change in the Hurst exponent sorts entry outcomes [C64, Opus 5, xhigh]

### LINKED ISSUE #1427: [C64] Test whether the change in the Hurst exponent sorts entry outcomes

richkuo:
## Plain simple English

We asked a new question: does the Hurst number going up or down at the moment of entry tell you anything about how the trade turns out? Earlier studies only looked at whether the number was high or low. The test found a difference that is unlikely to be chance, but it could not measure how big that difference is, and the money result does not agree. Nothing was turned on. No setting shipped.

## Summary

Adds a report-only research harness that scores the **change** in the Hurst exponent at entry, where #1410, #1422, #1424 and #1426 all bucket its **level**. The derived series is `dH(i) = H(i) - H(i - W//2)` over the same rolling H those studies compute, so `hurst_exponent` stays the #1409 single source of truth and no new estimator enters the codebase.

The three parameters a change study has to invent are pre-registered as named constants at the top of the file and rendered from the run's JSON:

1. The lookback rule `L = W // 2`, the smallest lookback at which half the newer estimate's sample is new. It is never swept, so it adds no Benjamini-Hochberg multiplicity.
2. The bucket edges and gate pairs, which are #1410's committed level landmarks **re-centred** by subtracting 0.5. `delta_bucket_label` routes through `bucket_label(dh + LEVEL_ORIGIN)`, and `delta_size_multiplier` and `delta_anti_signal_side` delegate the same way, so an edit to #1410's half-open edge semantics can never leave the two studies disagreeing. Import-time asserts pin both the landmarks and the re-centred pin.
3. The inference direction, **two-sided**, because #1424's only measured effect on these rows pointed the way a one-sided design cannot detect.

Everything else is imported rather than restated: #1424's datasets, windows and ownership matrix, cohorts, targets, effective N, coverage floors and acceptance rule, plus #1426's two-sided cluster null.

NaN policy is load-bearing. An undefined change propagates from either endpoint, is its own bucket, never becomes 0.0, never transitions the hysteresis gate, and sizes at exactly 1. The warm-up audit and the coverage floor account for the lookback **on top of** the Hurst window: lead 770 bars = 514 Hurst + 256 lookback, which drops 18 of 189 owned cells (171 kept), each listed with a reason naming both halves. The look-ahead invariant is tested end to end on the derived series.

## What the committed run found

Verdict `change_sorts`.

| | Value |
|---|---|
| Confirmatory family | `momentum`, 7,971 row-matched rows |
| Two-sided cluster p | 0.010995 against a rank-1 bar of 0.05 |
| Separation (primary target) | **+0.009863** efficiency units, kept trades did better |
| `mean_reversion` | -0.007096 at p=0.001 on 21,075 rows |
| Detection limit | **0.000, degenerate** |
| Continuity target | +0.550492 pp net return at p=0.814593 against a 1.10 pp limit, **below it** |
| Pinned config `momentum/gate/W512/arm0.02/dis-0.02` | sep +0.008181, net-return sep **-0.308537** pp, p_cluster 0.047995, effN 461.4 |

**The validity gate passes trivially and corroborates nothing.** Its limit is 0.000 because the MDE grid search stops at its first point (d=0) when the zero-injection contrast already clears the bar, so the number restates the rejection instead of measuring this design's resolution. `limit_is_degenerate` carries that through `validity_gate` into `decide_recommendation`, which refuses the corroboration sentence on a degenerate limit. The effect **size** is therefore unestimated.

**The economics do not follow the finding.** `continuity_clause` appends the net-return result to every verdict's justification, and the report renders both targets side by side.

`prediction_audit` states mechanically that the pre-registered prediction did not hold, that a degenerate limit cannot discharge its falsifiable clause, and that the nearest non-degenerate comparison (pooled primary limit 0.007 here against #1426's 0.009 on the same calendar) came in **below**, so the prediction's stated mechanism was wrong.

## Scope guards

- **No #1412 Stage 0 verdict.** That question is defined on the level of H and #1424 and #1426 discharged it there, so the joint ADX x change-in-H table renders for description only, and a test pins that no verdict function is called.
- **No threshold, and no configuration can be recommended structurally**: the quantity this study scores is not the quantity `scheduler/hurst_gate.go` reads, so `decide_recommendation` has no promoting branch.
- **Defers the live-evidence contract path** `hurst_gate_calibration.md` to #1424 and refuses that path unconditionally in `main`, checked before any other work. The supersede clause passes to #1428.
- `decision` in the committed JSON equals `decision_payload(decide_recommendation(configs, mde))`, pinned by test, and `--render-only` reproduces the report byte-for-byte.
- `hurst_gate` semantics and its default-off state, the composite classifier, every live path and `config.example.json` are untouched. `CurrentConfigVersion` is unchanged.

## Verification

- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` gives **3677 passed, 1 skipped**.
- `backtest/tests/test_hurst_1427_change_sort.py` adds **130 tests** covering the re-centring derivation, the lookback rules, the NaN policy, the look-ahead invariant end to end, warm-up and coverage, a two-sided landmine fixture, Stage 0 unreachability, the validity gate and its degenerate branch, the continuity clause, the prediction audit, report content, the contract-path refusals, run guards, and the committed artifacts.
- `--render-only --write-report` re-rendered the committed report byte-identically against the run's own output.
- **No existing test was edited, weakened, skipped or removed.**

Closes #1427

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T05:58:06Z) in b7126c39.

### Fixed

1. **`delta_anti_signal_side` is the one re-centred delegator that does not honour the load-bearing NaN policy** — confirmed at `backtest/research/hurst_1427_change_sort.py:526-527`: `float(dh) + LEVEL_ORIGIN` raised `TypeError` on `None`, and a NaN reached `anti_signal_side`, where `nan < 0.5` and `nan >= 0.5` are both `False`, so an undefined change read as the ALIGNED side under either sense. The function now returns `Optional[bool]` and yields `None` for `None` and for any non-finite value (`:526-534`), so an undefined change has NO gate side — mirroring `delta_bucket_label`'s distinct `BUCKET_NAN` (`:499`) rather than folding the row into an arm of the comparison. Returning a definite `False` was rejected: that is the silent behaviour the finding names, and it would put undefined rows in the kept arm. Sense validation moves ahead of the guard, matching `size_multiplier` (`hurst_1410_gate_calibration.py:226-227`), so an unknown sense still raises on an undefined change instead of being short-circuited by the new early return. Defined values are unchanged and still delegate to #1422 at `dh + 0.5`.

Must-survive cases, all pinned by new tests in `backtest/tests/test_hurst_1427_change_sort.py`:
- `None` `dh` does not raise — `test_an_undefined_change_has_no_gate_side` (`None`, `nan` and `inf`, both senses).
- NaN resolves to the documented NaN behaviour, not a silent `False` — `test_an_undefined_gate_side_is_never_the_aligned_side` asserts `is not False`, which a `False` return would fail while a `None` return passes.
- A future caller reaching the function without `_split`'s pre-filter (`:1009-1010`) gets `None`, an explicitly undefined answer, rather than a contaminated arm.
- `test_the_gate_side_still_rejects_an_unknown_sense_on_an_undefined_change` pins that the new early return did not swallow the `ValueError`.
- The existing delegation test gains a sweep over defined values pinning `delta_anti_signal_side(dh, sense) == study1422.anti_signal_side(dh + 0.5, sense)`, so the re-centring property the study depends on is still enforced after the guard.

`docs/ARCHITECTURE.md:129` now states the third leg of the NaN policy explicitly, since "delegate the same way" was the line the finding read against.

**Study output is unchanged.** The only call site already pre-filtered `None`/non-finite `dh`, so no verdict, p-value, separation or coverage figure moves; the committed JSON and the report are untouched and `--render-only` still reproduces byte-for-byte (test-pinned).

### Verification

- `uv run --no-sync python -m pytest backtest/tests/test_hurst_1427_change_sort.py` — 133 passed (was 130; three added, none edited, weakened, skipped or removed).
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -n auto` — 3680 passed, 1 skipped.
- `go build ./...` clean. Five `TestUpdateShell*` / `TestUpdateHelpersEnvfileParsing790` Go tests fail identically on `origin/main` in this local environment (macOS bash 3.2 has no `declare -A`); GitHub CI's `go` job is green and this change touches no Go.

---

richkuo:
@claude sonnet review

PR #1481, opened 2026-08-29
fix(#1470): converge a kill-switch cancel on an adoption-ineligible limit order [C66, Opus 5, high]

### LINKED ISSUE #1470: [C66] A cancelled-by-kill-switch order can stay live after the operator clears the alarm
**Complexity: 66/100** — Capability 2 (Risk 3 — live order cancellation on an auto-protective path; Coupling 3 — reconciler ↔ kill-switch cancel pass ↔ `pending_limit_orders` schema ↔ manual command path); Volume 16 — Opus 5, xhigh · fableplan: no

## Problem

PR #1469 gave the kill-switch flatten a pre-flatten cancel pass over `pending_limit_orders`. That pass marks `cancel_requested = 1` on every row it touches (`scheduler/kill_switch_limit_orders.go:187`) and re-issues the cancel on every latched cycle, so a cancel that fails once is retried while the latch holds.

Two facts make that retry stop for one class of row:

1. The pass runs only from `planKillSwitchClose` (`scheduler/kill_switch_close.go:286`), which the main loop calls only while `killSwitchFired` is true. An owner DM reset clears the latch (`scheduler/main.go:1277-1297` → `ResetPortfolioKillSwitchManual`), and the pass then stops running.
2. `reconcilePendingLimitOrders` is the only other cancel driver. It skips any row whose strategy is absent from the config, whose type is not `manual`, or which is not Hyperliquid-live (`scheduler/manual_limit.go:501-508`) — before it reaches the `o.CancelRequested || expired` cancel branch at `scheduler/manual_limit.go:604-620`. So it is not a convergence path for a paper, orphaned, or type-changed row.

Combine them: an adoption-ineligible row whose cancel keeps failing (a Hyperliquid error burst is most likely at exactly the moment the kill switch fires) is driven to cancelled by nothing once the operator resets the latch. The order stays resting at the venue and can still fill, putting real money back on the exchange after an emergency stop.

The state is bounded and is strictly better than `origin/main`, which cancelled no resting limit order at all. The operator message names the unresolved order every latched cycle (`scheduler/kill_switch_close.go:581-582`), and the reset prompt embeds that message plus a verify-manually note (`scheduler/kill_switch_close.go:671-681`). The gap is that after the reset the software stops both trying and reporting: the reconciler's skip is a plain stdout line (`scheduler/manual_limit.go:503, 507`), not an operator alert.

Because the remedy needs a convergence mechanism the reconciler has never had, PR #1469 deferred it here rather than growing the pull request.

## Goal

An order the kill switch decided must not fill converges to cancelled, or stays flagged to the operator, whatever the adoption eligibility of its owning strategy and whether or not the latch was reset.

## Approach

Give `reconcilePendingLimitOrders` a cancel-only lane for a row with `cancel_requested = 1` whose strategy fails the adoption gate. The lane must run above that gate and must book no fill.

Three design points carry the risk:

1. **Which script issues the cancel.** The row's own strategy is gone, paper, or no longer `manual`, so it supplies no usable Hyperliquid-live script. The kill-switch pass already solves this with a roster plus a fallback script (`killSwitchLimitOrderRoster`, `collectKillSwitchLimitOrderCandidates`, `scheduler/kill_switch_limit_orders.go:63-113`). Reuse that resolution rather than writing a second one, and refuse the lane when no Hyperliquid strategy with a script remains.
2. **Row retention.** The lane must never delete a row on a cancel that reveals a fill the book does not hold. Parent issue #1462 Approach step 3 forbids that class: a deleted row orphans the fill, leaves a live on-chain position outside the virtual book, and keeps its realized profit, loss and fees off the ledger — the accounting class #1454 and #1455 closed. Keep the row as the recovery record and keep reporting it.
3. **Operator visibility.** A row this lane cannot resolve must reach the owner, not only stdout. Reuse the existing throttled-alert shape so the reconciler's per-tick cadence does not produce a direct message every tick.

## Acceptance criteria

- A `cancel_requested` row whose strategy is absent, is not `type=manual`, or is not Hyperliquid-live has its cancel re-issued by the reconciler on later ticks, with the latch cleared and the kill switch no longer firing.
- The lane books no fill and deletes no row while the exchange reports a filled size the book does not hold. The row survives as the recovery record.
- A row the lane cannot resolve raises a throttled owner alert that names the order and its order id.
- An eligible Hyperliquid-live `type=manual` row converges exactly as it does today, through the existing branch, with no second cancel issued for it.
- An ineligible row whose cancel succeeds and which shows no unbooked fill is finalized and its queue row cleared, with no retry loop left behind.
- No Hyperliquid strategy with a script left in the config: the lane refuses, reports, and does not delete the row.

## Plain simple English

The emergency stop cancels waiting orders, but it only keeps trying while the alarm is on. For an account setup the program no longer fully manages, if those cancel attempts keep failing and the operator then clears the alarm, nothing finishes the job. The order can still fill and put money back in the market. The software must keep trying, or keep telling the operator.

---

richkuo:
## Plain simple English

The emergency stop cancels waiting orders, but it only kept trying while the alarm was on. For an account setup the program no longer fully manages, a cancel that kept failing was abandoned when the operator cleared the alarm, and the order could still fill. The program now keeps trying by itself, and it tells the operator when it cannot finish.

## Summary

`reconcilePendingLimitOrders` gets a cancel-only lane above its adoption gate, so a row the kill switch marked `cancel_requested` converges whether or not its owning strategy can still adopt it, and whether or not the latch was reset.

The lane takes a row that is `cancel_requested` or past its TTL and whose strategy fails `killSwitchLimitOrderAdoptionBlock` (absent from the config, not `type=manual`, or not Hyperliquid-live). It:

- resolves the cancelling script through `killSwitchLimitOrderRoster` + `collectKillSwitchLimitOrderCandidates`, reusing the kill-switch resolution rather than adding a second one, and refuses when no Hyperliquid strategy with a script remains in the config;
- polls status **first** and issues the cancel only while the order is still resting;
- books no fill, and keeps the queue row as the recovery record on any unadopted fill, unreadable state, still-resting order, failed flush, or failed delete;
- flushes an already-adopted fill to the state DB before clearing the row;
- raises a throttled owner DM keyed `(strategy_id, symbol, oid)` that names the order and its order id for anything it cannot resolve, and clears that key when the row finalizes.

An eligible Hyperliquid-live `type=manual` row is untouched: it still converges through the existing branch, and the lane issues no second cancel for it.

### Two design points worth a reviewer's attention

1. **Status-first, not cancel-first.** The kill-switch pass cancels then verifies. Copying that here would have created a stuck loop: cancel succeeds, the order goes off-book, the row delete then fails, and every later tick's cancel returns "order not found", so the delete is never retried and the row alerts forever. Polling status first makes the lane idempotent — a retry that finds the order already off-book skips straight to finalizing.
2. **TTL expiry is in the lane's trigger.** The lane fires on `o.CancelRequested || expired`, the same condition the eligible branch already uses. Scoping it to `cancel_requested` alone would have left an identical money-on-exchange leak for an expired orphaned row. Every acceptance criterion in the issue concerns `cancel_requested`; this is a deliberate, disclosed widening rather than a criterion the issue set.

`CLAUDE.md` records the new lane on the existing `kill_switch_limit_orders.go` guardrail line.

## Verification

- `go -C scheduler build`, `go -C scheduler vet`, `go -C scheduler test -race ./...`.
- Five failures (`TestUpdateShellAll*`, `TestUpdateHelpersEnvfileParsing790`) are pre-existing on the untouched tree — macOS bash 3.2 has no `declare -A`. Confirmed identical with this change stashed.
- No Python changed, so the pytest suite is unaffected.

### Tests

New file `scheduler/orphan_limit_cancel_test.go`, covering each acceptance criterion:

- a row whose strategy is absent, is `type=perps`, or is paper has its cancel re-issued and its row cleared;
- an already off-book order is finalized with **no** second cancel (the idempotent-retry case);
- an off-book order with an unadopted fill keeps its row and books nothing;
- a failed cancel is retried on the next tick, with the owner DM throttled across ticks;
- a config with no Hyperliquid script refuses, alerts, and keeps the row;
- an ineligible row with no cancel queued issues nothing and alerts nothing;
- an expired ineligible row converges;
- an eligible row takes the existing branch exactly once and raises no orphan alert;
- an adopted fill reaches the state DB before its row is deleted;
- `Retain` drops throttle entries for rows that left the queue.

**Red → green:** every convergence assertion was run against the unfixed reconciler (change stashed, tests kept) and failed there. The two guard tests — eligible-row and no-cancel-queued — passed before and after, which is what they exist to protect.

**Changed existing tests:** none. No test was edited, weakened, skipped, or removed.

Closes #1470

---


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T22:34:06Z) in a234e68f.

### Fixed

1. **The orphan alert describes a still-resting order for the one case where the order has actually filled and left untracked exposure on Hyperliquid.** — Confirmed. `orphanLimitCancelOutcome` now carries an `orphanLimitCancelState` set at every return site from the path that produced it — unknown / resting / off-book-unadopted-fill / off-book-row-stuck (`scheduler/manual_limit.go:479-597`) — and `formatOrphanLimitCancelDM` renders one template per state (`scheduler/orphan_limit_cancel_alerts.go:98-130`). The unadoptable fill now reads "UNTRACKED HYPERLIQUID POSITION", states that nothing is resting and there is nothing left to cancel, names the missing stop-loss and take-profit, and directs the operator to restore the strategy or flatten the position; `orphanLimitCancelRetryNote` withholds the retry promise for that state alone (`scheduler/orphan_limit_cancel_alerts.go:91-96`). Your three "Must survive" cases are each covered: (a) and (b) both reach the off-book state and get the new template, because the state is set after `*st.Resting` is false whether the lane cancelled or found it already gone; (c) a genuinely still-resting order after a failed cancel keeps "IS STILL RESTING", the `Cancel oid=N on the Hyperliquid UI` instruction and the retry promise (`scheduler/manual_limit.go:554`). An unreadable on-chain state is now its own template saying the order MAY still be resting, which the old single template also got wrong in the other direction.

2. **The success alert claims the lane cancelled an order in cases where it issued no cancel at all.** — Confirmed. The outcome now records the two facts separately — `CancelIssued` and `AdoptedFill` — and `orphanLimitCancelResolvedMessage` composes them (`scheduler/orphan_limit_cancel_alerts.go:146`), so your three cases read differently: "cancelled the resting order oid=N", "found order oid=N already off-book and sent no cancel", and "with its fill of X already booked and kept in the book".

3. **A row the lane classifies as permanently unresolvable still drives an exchange poll on every reconcile cycle, forever.** — Confirmed, and kept in this PR under scope rule 1: this PR introduces the poll, because such a row previously hit the `continue` skip and cost zero subprocesses. A new `pending_limit_orders.operator_required_since` column (`scheduler/db.go:349`, migration at `scheduler/db.go:559`) is stamped when the lane observes the unadoptable-fill state (`scheduler/manual_limit.go:608-620`) and read by `orphanLimitPollDeferred` before the poll (`scheduler/manual_limit.go:653`), so the poll and the owner DM share one `effectiveAlertThrottleInterval()` cadence. Your three "Must survive" cases: (a) the gate sits inside the ineligible branch, so a restored strategy takes the eligible branch immediately, and that branch clears the marker (`scheduler/manual_limit.go:670`); (b) the marker is per-row, so two terminal rows back off independently; (c) it is persisted on the row, so a restart re-derives the backoff rather than losing it. A transient failure clears the marker rather than inheriting the terminal cadence (`scheduler/manual_limit.go:616`).

### Test edits

None. No existing test was edited, weakened, skipped, or removed. `scheduler/orphan_limit_cancel_test.go` adds: per-state alert wording tests, including a banned-phrase sweep asserting the off-book template never says "can still be resting", "IS STILL RESTING", "NOT cancelled" or "Cancel oid=9001 on the Hyperliquid UI"; the retry-note asymmetry across all four states; the three-way resolved-message test; the backoff, its window expiry, per-row independence, and both marker-clearing paths; and a schema-migration test that drops the new column, seeds a pre-migration row, and reopens the DB.

**Red to green:** the two backoff tests were run with the deferral gate disabled and failed there — 3 polls on the single-row case and 2 polls per row on the two-row case, against the expected 1.

**Verification:** `go build`, `go vet`, `go test -race ./...` in `scheduler/`. The five `TestUpdateShell*` / `TestUpdateHelpers*` failures are pre-existing on the untouched tree (macOS bash 3.2 has no `declare -A`) and were confirmed identical with the change stashed. Both CI checks (`go`, `python`) were green on the reviewed commit.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T22:58:38Z) in ac42e7ad.

### Fixed

1. **The untracked-position alert offers "flatten the position yourself" as a resolution, but a manual flatten never clears the queue row, so the CRITICAL owner alert re-fires every throttle window indefinitely.** — Confirmed, and traced independently: the exchange fill for an OID is order history, so it does not move when the position is flattened, and the lane's terminal test `st.FilledSize > o.FilledSize+limitFillEpsilon` (`scheduler/manual_limit.go:573`) keeps matching, re-stamping `operator_required_since` and re-firing the DM every window. I also confirmed no operator path to clear the row existed — `manual-cancel` resolves its strategy through `findManualStrategy`, which cannot reach a row whose strategy is absent, paper, or no longer `type=manual`. Kept in this PR under **scope rule 1**: the DM text and the row-retention loop are both code this PR adds, and this PR's advice is what steers the operator into the unstoppable state.

   New `go-trader manual-clear-limit-row <order-oid> --flattened [--dry-run]` (`scheduler/manual_limit.go:375`) discards the recovery record for exactly that state, and it is the only operator path that does so — it is not an automatic delete, so the "never auto-delete a row carrying an unadopted fill" rule from #1462 stands. Its safety envelope is `clearOperatorRequiredLimitRowRefusal` (`scheduler/manual_limit.go:348`): it refuses a row whose strategy is Hyperliquid-live `type=manual`, because the scheduler adopts that fill itself and clearing would orphan it; refuses a row the lane has not marked `operator_required_since`, because it is still converging; and refuses without `--flattened`, the operator's explicit assertion of the one fact the scheduler cannot verify for a strategy it does not own. It takes the manual action file lock and prints the record before discarding it. Registered in `knownSubcommands`, the `main()` dispatch, `agent_info.go` and the README.

   The DM now names only paths that stop the alert (`scheduler/orphan_limit_cancel_alerts.go:107-110`): restore as the path the scheduler can finish on its own, or flatten paired with the new command, with the repeat cadence stated. `orphanLimitCancelRetryNote` names the same clearing path. Your three "Must survive" cases: (a) the operator now has a command that stops the CRITICAL channel, and a passing test drives it end to end — a cleared row produces no further poll and no further DM; (b) the DM explicitly warns against flatten-then-restore without clearing the row, and following the documented path leaves nothing to adopt; (c) the eligible branch is untouched, so a restore without a flatten adopts and arms protection exactly as before — covered by the existing eligible-branch and marker-clearing tests.

### Deferred to follow-up

1. **The untracked-position alert offers "flatten the position yourself" as a resolution, but a manual flatten never clears the queue row, so the CRITICAL owner alert re-fires every throttle window indefinitely.** — the residual of your "Must survive (b)" only: an operator who flattens, ignores the DM's warning, and restores anyway still gets a phantom adoption. That path is pre-existing — it exists identically on `origin/main`, where a surviving orphaned row plus a restore reaches `applyLimitFillProgress` the same way — and this PR neither creates nor changes it. Its remedy is a live-exposure check on the adoption path, which is a new cross-cutting invariant this PR does not have, and shared-coin attribution makes it its own design problem, so **scope rule 2**; filed as #1483. The in-PR half of the finding is fixed above.

### Test edits

1. **`TestKnownSubcommandsMatchDispatch`** (`scheduler/main_test.go:1314`) — Outdated; ground: this change adds `manual-clear-limit-row` to `main()`'s dispatch, and the test is the roster sync guard whose own failure message names that as the reason to update the list ("update validateDaemonInvocation when adding/removing a subcommand in main()"). Now asserts the same exact roster with `manual-clear-limit-row` in its dispatch position, so the guard still fails on any future unregistered subcommand. No other existing test was edited, weakened, skipped, or removed.

New tests: every remediation the off-book DM names is asserted present and the bare-flatten resolution asserted gone; the refusal ladder across all four of its outcomes; `findPendingLimitOrderByOID`; and an end-to-end pass showing that once the row is cleared the reconciler drives no further exchange poll and sends no further DM.

**Verification:** `go build`, `go vet`, `go test -race ./...` in `scheduler/`. The five `TestUpdateShell*` / `TestUpdateHelpers*` failures are pre-existing on the untouched tree (macOS bash 3.2 has no `declare -A`). Both CI checks (`go`, `python`) were green on the reviewed commit `a234e68`.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-29T23:45:42Z) in 28f22c08.

### Fixed

1. **A less-urgent alert can permanently consume the throttle slot a subsequent, more-urgent "untracked position" alert needs, hiding an unprotected Hyperliquid fill from the operator for up to the full throttle window.** — Confirmed, and re-derived independently: `Record` keyed its slot only by `orphanLimitCancelKeyFor(o)` with no reference to the outcome, and `reportOrphanLimitCancel` called it for every unresolved outcome, so the first alert in a window took the slot whatever its severity and the escalation reached only the server console. Scope rule 1 — the tracker and the state model are both code this PR adds.

   The slot now carries the severity it last notified at alongside its timestamp. `orphanLimitCancelSeverity` (`scheduler/orphan_limit_cancel_alerts.go:32`) ranks the states by the exposure each describes: off-book-row-stuck (bookkeeping only, no exposure) < unknown (may still be resting) < resting (confirmed live order) < off-book-unadopted-fill (untracked position, unprotected, no automatic recovery). `Record` (`scheduler/orphan_limit_cancel_alerts.go:45`) now fires when the window has elapsed, **or** when the new severity is strictly higher than the one already notified in the open window, and re-arms the window at that higher severity so the escalated alert gets a full window of its own.

   Your three "Must survive" cases each have a test: (a) resting then unadopted-fill inside one window delivers both, asserted both at the tracker and end to end through `reconcilePendingLimitOrders` with a failing cancel followed by an off-book poll; (b) two consecutive unadopted-fill outcomes deliver one DM, with the repeat firing only after the window elapses; (c) unknown then unadopted-fill delivers both. Two further cases the invariant implies are also pinned: a de-escalation does not re-open the window, and the stored value is the window's peak rather than the last state, so a flapping severity does not re-alert — and the peak does not latch across windows, so a fresh window alerts on its own severity.

   One thing I did **not** change, on the record because your reachability paragraph raises it: a cancel failure is still classified `orphanLimitStateResting` from the pre-cancel poll rather than re-polled. The observation is honest for that cycle (the order *was* resting when last read), the next cycle re-polls and corrects it, and the operator's action — check the order on the Hyperliquid UI — is right under either reading. With this fix the correction now reaches the owner immediately instead of being swallowed, which is the hazard the finding names. Adding a post-failure re-poll would be a separate behavior change the finding does not ask for.

### Test edits

1. **`TestOrphanLimitCancelTrackerRetainDropsClearedRows`** (`scheduler/orphan_limit_cancel_test.go:463`) — Outdated; ground: this finding requires the outcome state in the throttle decision, so `Record` takes it as a parameter and the test's existing calls no longer compile. Now passes one fixed state through every call and asserts the same `Retain` semantics as before — a retained key keeps its window, a key dropped by `Retain` alerts again. No other existing test was edited, weakened, skipped, or removed.

**Red to green:** the four escalation assertions were run with the severity gate disabled and failed there — the untracked-position alert was swallowed after each of the three predecessor states, and the end-to-end pass delivered 1 DM instead of 2.

**Verification:** `go build`, `go vet`, `go test -race ./...` in `scheduler/`. The five `TestUpdateShell*` / `TestUpdateHelpers*` failures are pre-existing on the untouched tree (macOS bash 3.2 has no `declare -A`). Both CI checks (`go`, `python`) were green on the reviewed commit `ac42e7a`.

Thanks for the independent re-verification of the earlier cycles' fixes.

---

richkuo:
@claude review

PR #1484, opened 2026-08-30
feat(#1428): score the Hurst exponent as a sizing and an exit input [C68, Opus 5, high]

### LINKED ISSUE #1428: [C68] Test the Hurst exponent as a sizing and exit input rather than an entry gate

richkuo:
## Plain simple English

The trading tool can make a trade smaller when the market looks less persistent. Nobody had checked that this helps. Nobody had checked whether the same measurement can tell us when to close a trade. This study checks both. It found no usable effect, and it says clearly that the data was too weak to prove anything either way.

## Summary

Adds one research harness with two arms, plus its committed run, tests, and registry row. Report-only: it touches no scheduler, config, or live path, and it can recommend no configuration.

**Sizing arm.** Scores the form `scheduler/hurst_gate.go` actually implements — `clamp(|H-0.5|/0.15, size_floor, 1.0)` — by **calling** `backtest/hurst_gate.py`'s `HurstGate.size_multiplier` rather than re-implementing it, so a drift in the shipped curve moves the study with it. #1410 swept `clamp(1 + gain*e, 0, 1.5)`, a different curve that can exceed 1.0, so the shipped multiplier was uncalibrated by construction. Import-time assertions plus a test that reads the Go source pin the span, the floor, the 1.0 cap, and the NaN case.

**Exit arm.** One pre-registered form, never swept: the trailing stop distance scales by the persistence side of the entry bar (persistent x1.5, anti-persistent x0.5, neutral and NaN unchanged), off a 2x ATR base and #1410's committed 0.45/0.55 edges read through the family sense, on the single rolling-Hurst window #1424's committed primary hypothesis pins.

`Backtester` exposes only one scalar `trailing_stop_atr_mult` per run, so a persistence-conditional trail cannot live inside a single run. The issue's "real Backtester re-runs" criterion constrains the mechanism without picking it, so the arm uses **bucket-conditional re-runs**: entries are masked to one bucket and the Backtester runs twice on that masked series, once at the base distance and once at the scaled one. Every leg on both sides is a real `Backtester.run`, and both sides share their masking, so the contrast isolates the exit change rather than the masking. A research-local close hook was refused — it would score a code path the live system does not have.

**Per-arm validity contrast.** The gate's kept-versus-suppressed partition survives neither arm, so each states its own row-matched signed contrast:

1. Sizing splits `multiplier >= 1.0` against `< 1.0`. The shipped form's 1.0 cap collapses the kept side to the tail `|H-0.5| >= 0.15`, so a breach of the 30-effective-trade floor is published as a power statement, never as a null about the market.
2. Exit is **paired**: a bucket's masked baseline and masked scaled runs admit the same entries, so each entry bar carries both outcomes and the row value is `scaled - baseline`. The unscaled buckets are a structurally-zero control. An entry present on only one side is dropped and counted, never imputed.

Targets swap by arm — signed fixed-horizon efficiency stays primary for sizing, net return over the actual holding period is primary for the exit arm, since a fixed horizon is a poor target for an arm that changes when a trade ends.

**Contract path.** Whether the study claims `hurst_gate_calibration.md` is resolved mechanically from the verdict into `decision.claims_contract_path`, and the command line can never target that path. Only a confirmatory result on a non-degenerate limit may claim it, per the precedent #1426 set and the maintainer restated on the issue. #1426 (#1452) and #1427 (#1473) both defer, passing the clause here.

## Committed run

860 legs, 37,923 sizing rows, 69,853 paired exit rows, 6,188 entries dropped for appearing on only one side.

**Both arms return `inconclusive`, and both validity gates fail as `below_limit`.** The run is therefore a power statement in both directions and bounds nothing:

| Arm | Family | Separation | Limit | Kept eff. N |
|---|---|---|---|---|
| sizing | `momentum` | +0.047105 efficiency units | 0.099 | 52.5 |
| exit | `momentum` | -0.021650 pp net return | 0.180 | 1033.4 |

Nothing reaches Benjamini-Hochberg significance; the smallest cluster p is 0.035396 (`mean_reversion/sizing/W256/shipped`) against a rank-1 bar of alpha/6. The pre-registered key risk — that this tape's calendar clusters cannot resolve an effect of the size these buckets show — held.

The shipped form's 1.0 cap does visibly collapse the sizing kept side as the window grows: effective kept N falls 301.2 -> 158.8 -> 52.5 across W128/W256/W512 on `momentum`, and to 4.0 (under the floor) on the exploratory cohort at W512. That is the design risk the study pre-registered, measured.

The run **defers** the contract path to #1424.

## Verification

- 59 new tests in `backtest/tests/test_hurst_1428_sizing_exit.py`.
- The four Hurst suites pass at **416**; the full `backtest/` suite passes at **2,306 passed, 6 skipped**.
- `--render-only` reproduces the committed report **byte-for-byte** (verified through the CLI, not only in-process).
- `decision_payload(decide_recommendation(configs, mde))` re-derived from the committed configs equals the committed payload, pinned by test as #1424 does.
- History backfilled from Binance US, Bitstamp and Coinbase before the run; the coverage audit dropped only cells with real gaps (XRP 2021-2023) or no warm-up lead (Coinbase ETH/LTC 2016), each listed by name and reason in the report.

## Two defects found and fixed during verification

Both were mine, both caught before the committed run, and both carry regression tests:

1. The headline verdict returned `resolved_null` when an arm was significant behind a degenerate limit — a null claim that contradicted its own arms. It now returns `sorts` whenever any arm does, while the contract-path claim stays separately gated on a non-degenerate limit.
2. Exit and sizing rows carried no `exit_ns`, so `effective_n` silently dropped every row and both arms were structurally pinned at a zero-effective-N floor breach regardless of the data. Paired rows now take the **later** of their two exits, so overlap is never understated and the effective-N correction can only be conservative.

## Deviations from the issue

None material. The issue left the exit mechanism open between bucket-conditional re-runs and a research-local close hook; this picks the former and states why in the artefact.

## Tests

No existing test was edited, removed, or weakened. All changes are additive.

Closes #1428

---


richkuo:
@claude review

PR #1487, opened 2026-08-30
feat(#1474): compare a rescaled-range Hurst estimator against the DFA estimator [C35, Opus 5, high]

### LINKED ISSUE #1474: [C35] Compare a rescaled-range Hurst estimator against the DFA estimator on the #1424 pool
**Complexity: 35/100** — Capability 1 (Uncertainty 2 — the R/S bias and window behaviour on our data are unknown; Risk 0 — report-only, no live path); Volume 10 (Scope 2 — estimator SSoT module, one research harness, registry row, tests; Coupling 1 — reuses the #1410/#1424 helpers; Verification 2 — synthetic-series unit tests plus a full sweep) — Opus 5, high · fableplan: no

## Problem

The repo has exactly one Hurst estimator: `hurst_exponent` in `shared_strategies/open/indicators_core.py:146`, a detrended fluctuation analysis (DFA) over log returns with a 100-point minimum (`HURST_DFA_MIN_POINTS`, line 112) and 12 geometric scales from 8 to `n // 4` (`_hurst_dfa_scales`, line 117). Every Hurst study (#1410, #1422, #1424, #1426) and the live composite classifier (`shared_tools/regime.py:268`) read that one function. #1409 chose DFA over classic rescaled range (R/S) with the note "R/S is too noisy at the window lengths this system uses", but that claim was never measured on our data. The #1410 report also records that DFA itself carries a small upward bias at short samples (mean about 0.51 on memoryless data at n=101 and n=201, with single draws reading 0.65 to 0.70 at n=101; `backtest/research/hurst_1410_gate_calibration.md:42`).

Result: the #1424 verdict on the confirmatory family (separation -0.005 efficiency units against a 0.013 limit) rests on one estimator. There is no evidence whether a second, independent estimator of H agrees with DFA on the same rows, or whether the estimator choice moves the gate separation at all.

## Goal

A research-only study that adds an R/S Hurst estimator beside the DFA single source of truth (SSoT) and reports, on the #1424 pool: how closely the two estimators agree per row, how each is biased on a memoryless series at each window length, and what gate separation each produces on the confirmatory family. No live behaviour changes and no threshold ships.

## Approach

Correctness and safety outrank diff size; the study inherits #1424's power discipline so its numbers are comparable.

1. **Estimator SSoT — `shared_strategies/open/indicators_core.py`.** Add `hurst_rescaled_range(close, *, min_points=...)` beside `hurst_exponent`, following the same conventions: log returns, deterministic, NaN below the minimum and on degenerate input (constant price, non-positive price), never 0.5 as a sentinel, never raises. Block sizes from 16 up to `n // 4` on a geometric grid; per block compute the range of the mean-adjusted cumulative sum divided by its standard deviation; fit the log-log slope. Apply the Anis-Lloyd expected-R/S correction (`H = 0.5 + slope of log(R/S) - log(E[R/S])`) as the default, with a flag to return the raw slope so the bias table can show both. Unit tests in `shared_strategies/open/test_indicators_core.py` mirror the existing `test_hurst_*` set (lines 494 to 548): random walk near 0.5, persistent above 0.5, mean-reverting below 0.5, insufficient data NaN, constant NaN, deterministic, custom minimum.

2. **Research harness — `backtest/research/hurst_1474_rs_estimator.py`** (number to match the filed issue). Import the #1410/#1422/#1424 helpers exactly as `hurst_1426_two_sided_sort.py` does (`rolling_hurst`, cluster null, effective N, coverage floors, warm-up audit, row-matched validity gate). Three sections:
   - **Agreement.** Roll both estimators over the same closes at the live windows `HURST_WINDOWS = (128, 256, 512)` (`hurst_1410_gate_calibration.py:75`) and at a reference window of 1,000 bars; report per-window Pearson and Spearman correlation between DFA and R/S, mean signed difference, and the share of rows where the two disagree on which side of 0.5 they fall. The 1,000-bar window exceeds the live fetch depth `hurst_live_frame_bars` (`backtest/hurst_gate.py:43`, `max(200, 2*maxPeriod-1+10)`), so the report must state that this reference cannot run live without a deeper fetch.
   - **Bias.** On synthetic Gaussian random walks seeded per draw, report mean and interquartile range of DFA, raw R/S, and Anis-Lloyd R/S at n = 101, 128, 256, 512, 1,000, 2,000. This is the measurement that #1409's "too noisy" claim needs.
   - **Gate separation per estimator.** Re-run the pinned #1424 confirmatory hypothesis (`momentum/gate/W512/arm0.52/dis0.48`) with R/S substituted for DFA at each live window, under the #1426 two-sided doubled-p null, and print the row-matched separation and detection limit beside the DFA numbers. Read separations pool-matched and signed, never `abs()`, never against the pooled limit (the #1424 rule).
   - The script refuses to write the contract path `hurst_gate_calibration.md` unconditionally (the #1426 pattern) and writes `hurst_1474_rs_estimator.md` + `.json`; `--render-only` re-renders from the committed JSON.

3. **Registry.** Add the harness row to `docs/backtesting-registry.md` beside lines 80 to 83, marked `one-shot`.

4. **Non-goals, stated in the report header.** No `metrics["hurst_rs"]` key on the live payload; `shared_tools/regime.py` and `scheduler/hurst_gate.go` untouched; `config.example.json` untouched; no recommended threshold. A follow-up issue may promote R/S to the payload only if the agreement section shows a material, signed difference on the confirmatory family.

## Acceptance criteria

- `hurst_rescaled_range` exists beside `hurst_exponent`, with the seven mirrored unit tests green and a test that the Anis-Lloyd corrected value on a 2,000-point random walk sits closer to 0.5 than the raw slope.
- `hurst_exponent` is byte-identical; all existing `test_hurst_*` and `shared_tools/test_regime.py` hurst tests stay green.
- The harness runs end-to-end on the #1424 pool and renders the three sections with the numbers above; `--render-only` reproduces the Markdown from the committed JSON.
- The bias table reports DFA, raw R/S, and corrected R/S at every listed n.
- The separation section prints DFA and R/S side by side, row-matched, signed, with the two-sided p and the detection limit.
- The script cannot write `hurst_gate_calibration.md`, and a test pins that refusal.
- `git diff` shows no change under `scheduler/`, `shared_tools/regime.py`, or `config.example.json`.
- The registry row exists.

## Plain simple English

We measure trend persistence with one formula. A second, older formula exists, but we never compared them on our data. This study adds the second formula, checks how much the two agree, how much each drifts on random data, and if the choice changes our earlier result. Nothing in live trading changes.

---
Created with LLM: Opus 5 | high | Harness: new-issue

richkuo:
## Plain simple English

The system measures whether prices trend or mean-revert with one method. This adds a second, older method beside it and compares the two on the same data. The two agree more as the sample gets longer. The second method is wider at every sample size the live system can read. Swapping methods does not change the earlier result.

## Summary

Report-only comparison of a classic rescaled-range (R/S) Hurst estimator against the #1409 DFA single source of truth, on the #1424 pool.

Nothing live changes. `git diff` shows no change under `scheduler/`, `shared_tools/regime.py` or `config.example.json`; no `metrics["hurst_rs"]` payload key is added; no threshold is recommended; and `hurst_exponent` stays the single source of truth every live and backtest path reads.

### Estimator

`hurst_rescaled_range(close, *, min_points=100, corrected=True)` lands in `shared_strategies/open/indicators_core.py` by pure insertion, so `hurst_exponent` is byte-identical after this work. It reads log returns, walks a geometric block grid from 16 to `n // 4`, takes the range of the mean-adjusted cumulative sum over the block standard deviation, and fits the log-log slope. The Anis-Lloyd expected-R/S correction is the default; `corrected=False` returns the raw slope. It returns NaN below the minimum and on degenerate input (constant price, non-positive price), never 0.5 as a sentinel, and never raises.

### Harness

`backtest/research/hurst_1474_rs_estimator.py` imports the #1410/#1422/#1424/#1426 helpers rather than reimplementing them, and answers three questions.

1. **Agreement.** Pearson, Spearman, mean signed difference, and the share of bars on opposite sides of 0.5, at W128/W256/W512 plus a 1,000-bar reference window the report states cannot run live. Spearman rises from 0.4945 at W128 to 0.6401 at W1000 over about 1.04M shared bars.

2. **Bias.** Mean and IQR of all three estimators on 500 seeded Gaussian random walks at n = 101, 128, 256, 512, 1,000, 2,000. This is the first measurement of the #1409 claim that R/S is "too noisy" at the window lengths this system uses. R/S is wider than DFA at every window below 1,000 bars (at n=512, IQR 0.0710 against 0.0686) and becomes tighter only at n ≥ 1,000, which no live fetch reaches.

3. **Gate separation.** The pinned `momentum/gate/W512/arm0.52/dis0.48` hypothesis re-scored with each estimator substituted. Rows are matched so all three estimators split the identical row set and only the partition changes; separations are signed, never through `abs()`, and read against each estimator's own row-matched limit under #1426's two-sided doubled-p null.

### Verdict

**The estimator move is bounded.** At the pinned W512 hypothesis, swapping DFA for Anis-Lloyd R/S moves the row-matched separation by 0.005936 efficiency units, below the 0.012 detection limit those same rows carry. The #1424 verdict is therefore not an artefact of the estimator choice — a statement #1424 could not make about itself.

This bounds the instrument and says nothing about the market. Every estimator stays under its own limit, so none resolves an effect and no threshold ships.

### Why the raw slope is scored as its own estimator

The Anis-Lloyd correction is a constant shift at a fixed window, because the block grid depends only on the window length and least squares is linear. Raw and corrected R/S therefore carry identical Spearman and identical row ordering, and they differ only in where the 0.5 edge falls. That edge is exactly what the gate's `anti_signal_side` reads, so folding the two together would hide the only difference that matters to a gate. Two tests pin the property.

## Verification

- 27 new tests in `shared_strategies/open/test_indicators_core.py`; the file passes at 91.
- 88 new tests in `backtest/tests/test_hurst_1474_rs_estimator.py`, including an `_ONE_SIDED` landmine fixture copied from #1426's own test, which proves no one-sided null is reachable from this module at either the measurement or the section level.
- Full `backtest/` suite: 2,336 passed. `shared_strategies/` and `shared_tools/`: 1,126 passed.
- `--render-only` reproduces the committed report byte-for-byte from the committed JSON.
- The study refuses the contract path `hurst_gate_calibration.md` unconditionally, and a test pins that refusal.
- `docs/backtesting-registry.md` gains the `one-shot` row.

The full run scored 860 legs over 26 datasets in 16,135 seconds, with `--skip-fetch`, on the venue caches as they stood. The report discloses that choice in its run summary and in its "History backfill:" line.

No existing test was edited, removed, or weakened.

Closes #1474

---


richkuo:
@claude review

Issue #1485, opened 2026-08-30
[C68] Test suite is larger than the code it covers and makes every change cost more context


richkuo:
**Complexity: 68/100** — Capability 2 (Uncertainty 3 — which tests are load-bearing needs per-test discovery against current code; Risk 2 — deletions are reversible but touch guards on money and auto-protective paths); Volume 18 (Scope 4, Coupling 2, Verification 3) — Opus 5, xhigh · fableplan: no

## Problem

The test suite is now larger than the code it covers, and every change pays for it in agent context and tokens.

Measured on `main` at `a864c355`:

- Go: 165 `scheduler/*_test.go` files, 97,510 lines, 3,191 `func Test…` functions, 3.44 MB. The non-test Go source is 64,541 lines, 2.08 MB. Test bytes are 1.65× source bytes.
- Python: 158 `test_*.py` files (worktrees and `.venv` excluded), 47,983 lines, 3,716 `def test_` functions.
- Largest Go test files: `scheduler/hyperliquid_balance_test.go` (4,182 lines), `scheduler/risk_test.go` (3,680), `scheduler/discord_test.go` (2,997), `scheduler/config_test.go` (2,813), `scheduler/db_test.go` (2,495), `scheduler/hyperliquid_stop_loss_test.go` (2,490).
- Largest Python test files: five Hurst research files under `backtest/tests/test_hurst_14*.py` total about 6,000 lines for a feature that is default-off (`scheduler/hurst_gate.go`, see CLAUDE.md), plus `shared_scripts/test_check_hyperliquid.py` (1,455).
- CI runs the whole set on every PR: `.github/workflows/ci.yml:23` (`go test ./...`), `:52` (`uv run pytest -n auto` over `pyproject.toml:27` `testpaths`), `:55` (`uv run pytest shared_scripts/`).

Effects:

1. An agent that edits one feature reads and updates several thousand lines of tests, many of which assert small details (exact log strings, DM wording, per-variant repeats of one code path) rather than the behavior that protects money or state.
2. Some tests no longer match current behavior or the current environment. Observed example: on macOS `main`, five `scheduler/update_sh_feature_test.go` cases fail through `scripts/test_update_helpers.sh` (`update_sh_feature_test.go:50`, message `helper does not de-dupe ... (got=       0 want=0)`), which is a whitespace comparison defect in the shell test, so the Go suite is red locally even though the code is correct. Other stale tests are likely and need discovery; this is unverified.
[...]

PR #1488, opened 2026-08-30
test(#1485): merge and prune the Go protected-core test suite, phase A [C68, Fable 5, high]

richkuo:
## Plain simple English

The automatic checks for the money and safety parts of the trading program were larger than needed. This change merges checks that test the same thing into one table and removes checks of small details. Every check that protects money, saved data, or a safety stop stays and still passes. The set is now smaller and easier to change.

## Summary

Phase A of #1485 (Go protected core). Part of #1485; phases B–E follow in their own PRs.

Files: `risk`, `daily_loss`, `exposure_cap`, `notional_cap`, `kill_switch_*`, `hyperliquid_liquidation_guard`, `hyperliquid_fills`, `shared_wallet*`, `db`, `state`, `config*` tests.

| Metric | Before (`a864c355`) | After |
|---|---|---|
| Phase A test lines | 26,825 | 22,967 |
| Phase A `func Test` | 822 | 515 |
| All Go test lines | 97,510 | 93,652 |
| All Go `func Test` | 3,191 | 2,884 |
| Package statement coverage | 70.4% | 70.4% |

Per-file coverage on protected files (mean of per-function coverage from `go tool cover -func`): every protected file is unchanged. `config_reload.go` and `config_migration.go` briefly dropped after the formatter and no-`strategies`-key tests were removed; both were restored (`TestFormatFloatPtrVariants`, `TestMigrateConfigRetainsInertLegacyDiscordKeys/versionless_without_strategies`) and the files are back at 92.1% and 85.4%. The only delta in the package is `model_only_close_alert.go` 71.4% → 75.0% (up).

## Verification

- `go build`, `gofmt -l`, `go vet` clean.
- `go test -count=1 .` from `scheduler`: every test passes except five pre-existing `update_sh_feature_test.go` cases that fail on macOS because `scripts/update.sh:604` uses `declare -A`, which macOS bash 3.2 rejects. They fail identically on `main` and are Phase B scope (updater); CI Linux has bash 5.
- No non-test source changed. The accuracy audit (issue rule 1b) ran on every kept test; findings that led to no edit are in the "Kept-but-flagged" lists below.

## Test edits (disclosure)

Every edit below is a merge (Obsolete-redundant, all assertions carried into the named table) or a removal with the stated case and ground. No test was loosened on a money/state path; the only loosened cases are listed per group.

## Removed
None outright; every vanished name was merged.
## Merged (all assertions carried; several subtests assert more)
- 3 `TestRolloverDailyPnL_*` → `TestRolloverDailyPnL`; 2 `TestRecordTradeResult_*` → `TestRecordTradeResult`; 4 `TestPortfolioNotional_*` → `TestPortfolioNotional`; `TestCollectFuturesMarkSymbols_IgnoresManual` → `TestCollectFuturesMarkSymbols`; 2 `TestCollectPerpsMarkSymbols_*` → `TestCollectPerpsMarkSymbols`; 2 `TestClearLatchedKillSwitchSharedWallet_*NoOp` → `_NoOp`; 6 `TestPerpsMarginDrawdownInputs_*` → one table (tol 1e-3→1e-6); 4 `TestAggregatePerpsMarginInputs_*` → one table; 6 `TestCheckRisk_Perps*/SpotUnchanged` → `TestCheckRisk_DrawdownBasis`; 3 `TestCheckRisk_LiveHL*` → `_LiveHLCircuitBreaker_SharedCoinVsSoleOwner`; 2 `TestCheckRisk_LiveTopStepCB_*` → `_PendingFlatten`; 3 `TestCheckPortfolioRisk_MixedAccount*/PeakZero*` → `_LatchSource`; 4 `TestCheckPortfolioRisk_MarginWarning*/Both*` → `_MarginWarningReasons` (#1448 regression kept as subtest); 3 pending-circuit-close round trips → `_RoundTrip`; 4 unmarshal variants → `_Unmarshal`; 2 `TestCircuitBreakerStrategyLabel_*`; 6 `TestCollectMissingMarkPositions_*`; 2 `TestManualOnlyMarkSymbols_*`; 5 `TestMissingManualOnlyMarks_*`.
## Loosened / rewritten
None.
## Kept-but-flagged
- Warn-band lifecycle family, DM triage-section pins, `_CircuitBreakerDisabled_WarnsOncePerEpisode` log pins (operator-facing, kept), `TestTruncateWarningField_UTF8Safe`.
## Code workarounds found
None.

## Files: scheduler/daily_loss_test.go, scheduler/exposure_cap_test.go, scheduler/notional_cap_test.go   Before: 1270 lines / 51 Test funcs (354/18, 689/26, 227/7)   After: 1121 / 31 (327/8, 602/18, 192/5)

Measured with `wc -l` and `grep -c '^func Test'`. Targeted run: `go test -count=1 -run 'DailyLoss|NotionalCap|ExposureCap|ManualStateView|ManualCore|CheckPortfolioRisk_Notional|ConfigValidationDailyLoss|ValidateConfig_ExposureCapBounds|ManualExposure' .` green (79 tests/subtests PASS); `go vet .` clean.

## Removed
- `TestNotionalCapHoldPassesReduceAndManage` (notional_cap_test.go) — case: Obsolete-redundant — ground: the test evaluated `const notionalBlocked = true && pausedBlocksSignal(...)`, so its only real code path is `pausedBlocksSignal` (scheduler/pause.go:3-20); the notional wiring in main.go is not exercised by it — survivor: `TestPausedBlocksSignal` (scheduler/pause_test.go) carries every row: `manage_only_signal0` (0,0,1,long,true,true→false) = "hold open"; `long_sell_pure_close` (-1,0,1,long,true,false→false) = "long-only sell exit" (qty 2 vs 1, same `posQty>0`/`!allowsShort` branch); `long_close_fraction` (-1,0.5,1,long,true,true→false) = "open long partial close"; `flat_buy_fresh_open` (1,0,0,"",true,true→true) = "flat buy" (allowsShort differs, but pause.go:7 returns true on `posQty<=0` before either flag is read); `long_buy_scale_in` (1,0,1,long,true,true→true) = "open long buy add both"; `long_sell_flip_both` (-1,0,1,long,true,true→true) = "long flip to short". The notional-specific guard (main.go must gate on `notionalBlocked && pausedBlocksSignal`, never `continue`) remains in `TestNotionalCapNeverSkipsStrategyCycle` (kept, #1344).

## Merged
- `TestEvaluateDailyLossLimitUnconfigured`, `TestEvaluateDailyLossLimitUSDThreshold`, `TestEvaluateDailyLossLimitPctThreshold`, `TestEvaluateDailyLossLimitPctBasisExcludesSharedWalletPool`, `TestEvaluateDailyLossLimitBothArmsLowerWins`, `TestEvaluateDailyLossLimitStaleDayExcluded`, `TestEvaluateDailyLossLimitWinsOffsetLosses`, `TestEvaluateDailyLossLimitManualStrategyIncluded`, `TestEvaluateDailyLossLimitPctBasisMiss`, `TestEvaluateDailyLossLimitNilStateSkipped` → `TestEvaluateDailyLossLimit` (daily_loss_test.go), 18 subtests — assertions carried: Tripped on every row; Configured=false + LossUSD=900 (max_drawdown-only) and Configured=false (nil config); USD arm below/at threshold and multi-strategy aggregate (loss 550, threshold 500); pct arm basis 5000/threshold 250 tripped and under-threshold not tripped; shared-wallet pool excluded from basis (2000/100, tripped) and all-pool → basis 0 + PctBasisMiss + not tripped; both arms lower-wins (threshold 400 tripped; threshold 1000 not tripped); stale day counts 0 (loss 50); wins offset losses (loss 50; net-positive loss 0); manual type counts (loss 350 tripped); pct-only basis miss inert vs USD arm still enforces (threshold 500 tripped); nil state entry skipped (loss 150 tripped). Rows that formerly mutated `states` between calls are now independent rows with fresh inputs. `dlBool` helper added; `fp` reused from risk_sizing_test.go.
- `TestManualOpenCoreRefusesDailyLossHold`, `TestManualAddCoreRefusesDailyLossHold` → `TestManualCoreRefusesDailyLossHold` (daily_loss_test.go) — assertions carried: execute never called (t.Error in stub), error contains "daily loss limit tripped" for both manual-open and manual-add (add row carries the open long Position).
- `TestManualOpenCoreRefusesNotionalHold`, `TestManualAddCoreRefusesNotionalHold` → `TestManualCoreRefusesNotionalHold` (notional_cap_test.go) — assertions carried: execute never called, error contains "new opens blocked, exits continue" for both open and add.
- `TestEvaluateExposureCap_AllLongBookBlocksLongsOnly` (evaluate half), `TestEvaluateExposureCap_NettingPerAsset`, `TestEvaluateExposureCap_SameAssetNetsBeforeBucketing`, `TestEvaluateExposureCap_DisabledByDefault` (evaluate half), `TestEvaluateExposureCap_FailSafeExclusions`, `TestEvaluateExposureCap_AvgCostFallback`, `TestEvaluateExposureCap_ManualPositionsCounted` → `TestEvaluateExposureCap_Buckets` (exposure_cap_test.go), 8 subtests — assertions carried: Configured; LongUSD/ShortUSD exact values (19000/0, 10000/10000, 10000/0 same-asset netting, 10000 with exclusions, 18200 AvgCost fallback, 12000 manual); LongBlocked/ShortBlocked on every configured row (some rows previously asserted only one flag; both are now asserted with the values the code produces, verified green); disabled (zero thresholds, nil config) → Configured=false and no bucket blocked; SkippedPositions count=2 with both entry strings and `exposureCapSkippedWarning` "2 position(s) excluded"; no skips on AvgCost fallback (len 0 now asserted on every configured row). Helpers `perpsPosState`/`perpsCfg` replace repeated inline fixtures with identical field values.
- `TestExposureCapBlocksSignal_ManageAndReducePassThrough`, `TestExposureCapBlocksSignal_DirectionalIncreases`, plus the `exposureCapBlocksSignal` halves of `..._AllLongBookBlocksLongsOnly` and `..._DisabledByDefault` → `TestExposureCapBlocksSignal` (exposure_cap_test.go), 12 subtests — assertions carried: signal 0 / close action / pure-close sell on long / pure-close buy on short pass under both buckets capped; scale-in add blocked while longs capped; long→short flip passes when only longs capped and is held when shorts capped; fresh short blocked when shorts capped, passes when only longs capped; fresh long on the 19000/15000 book blocked with reason containing "new long opens blocked", "$19000.00", "$15000.00"; fresh short passes on that book; unconfigured status never blocks. The AllLong reason row uses a status equal to the evaluated one (Configured, CapUSD 15000, LongUSD 19000, LongBlocked); the reason string is produced by `exposureCapBlocksSignal`, the same path.
- `TestManualOpenCoreRefusesExposureCap`, `TestManualAddCoreRefusesConcentrationOnly` → `TestManualCoreRefusesExposureCap` (exposure_cap_test.go), 4 subtests — assertions carried: open long refused ("manual-open (long) blocked", execute not called); open short reaches execute while only longs capped; add long refused by concentration-only status ("manual-add (long) blocked", execute not called); add short reaches execute when ETH is long-over-concentrated.

## Loosened / rewritten
- None.

## Kept-but-flagged (audit findings, no edit)
- `TestDailyLossStartupSummaryLine`, `TestExposureCapStartupSummaryLine` — pin boot-log wording (`usd=$500.00`, `pct=5.00%`, `same_direction=$15000.00`, `capped-direction opens only`). Left alone: the line is the operator's boot confirmation of the armed thresholds and the empty-when-disabled branch is behavior; the exact token spelling drives no decision and could be loosened to "contains the threshold numbers" if the lead prefers.
- `TestDailyLossStatusNote`, `TestFormatDailyLossPctBasisMissDM`, `TestExposureCapStatusNote`, `TestExposureCapHoldDetail`, `TestExposureCapAlertMessage_*` — assert operator-surface labels (`TRIPPED`, `armed`, `CANNOT evaluate`, `fully inert`, `USD arm still enforces`, emoji prefixes). Kept as contract: these labels tell the operator whether the gate is enforcing or inert.
- `TestNotionalCapNeverSkipsStrategyCycle` — reads and AST-parses `main.go` relative to the test cwd; holds under `go test` because cwd is the package dir. #1344 regression; kept.
- `TestCheckPortfolioRisk_NotionalCapReasonMentionsExitsContinue` — overlaps `TestCheckPortfolioRisk_NotionalCap` (risk_test.go) on the `nb` flag, but the "exits continue" reason wording (#1344) has no counterpart there; kept.
- `TestExposureCapFieldsHotReloadable` — asserts only that adding both thresholds passes `validateHotReloadCompatible`; the `max_notional_usd` restart-required rejection lives in the hot-reload test file. Kept as-is.
- `TestManualStateViewDailyLossHold` / `TestManualStateViewNotionalHold` — same shape in two files; left separate because each reads a different `manualStateView` field and lives beside its gate.

## Code workarounds found (do NOT fix, report only)
- None found in daily_loss.go, exposure_cap.go, notional_cap.go for these tests.

## Files: scheduler/kill_switch_close_test.go, scheduler/kill_switch_manual_test.go, scheduler/kill_switch_reset_dm_test.go, scheduler/kill_switch_limit_orders_test.go   Before: 3891 lines / 108 Test funcs (close 2044/48, manual 546/17, reset_dm 304/13, limit_orders 997/30)   After: 3308 lines / 79 Test funcs (close 1560/28, manual 551/15, reset_dm 231/8, limit_orders 966/28)

All four files green before and after; `go -C scheduler vet .` clean; `gofmt -w` applied. Every assertion on `OnChainConfirmedFlat`, `CanAutoResetWithoutOwner`, fill-share splitting, limit-order cancel/adoption eligibility, reset gating, and operator-facing DM text survived. No file outside the four assigned test files was touched.

## Removed
Nothing was removed as a standalone deletion. Every retired `Test*` function was folded into a survivor carrying all of its assertions (see Merged).

## Merged
- `TestHyperliquidKillSwitchClose_SharedCoinSplitsFillByVirtualQuantity`,`TestHyperliquidKillSwitchClose_SharedCoinPeersSumToReport` → `TestHyperliquidKillSwitchClose_SharedCoinSplitsFillByVirtualQuantity` (kill_switch_close_test.go) — carried: 1 trade per peer; peer qty 1.5 / 0.5; qty sum == 2.0; fee sum == 4.0; both prices == 3000; peer fees 3.0 / 1.0; peer A `ClosedPositions` len 1; peer A `RealizedPnL == -153.0`. Ground: identical two-peer state and fill in both tests; the old "SplitsFill" test asserted a peer-A-only subset plus PnL/ClosedPositions, now in the survivor.
- `TestPlanKillSwitchClose_{OKX,Robinhood,TopStep}{HappyPath,CloseError,FetchFailure,UnconfiguredBlocksReset}` (12 funcs) → `TestPlanKillSwitchClose_PlatformLifecycle` subtests `<Platform>/HappyPath|CloseError|FetchFailure|UnconfiguredBlocksReset` (kill_switch_close_test.go) — carried per platform. HappyPath: confirmed-flat, fetcher called once, closer calls == [coin], DM contains `"<Label> closes: [coin]"`. CloseError: NOT confirmed-flat, `<Platform>CloseReport.Errors[coin]` non-nil, DM contains `LATCHED, RETRYING` + closer error text. FetchFailure: NOT confirmed-flat, closer never called, DM contains `LATCHED, RETRYING`. Unconfigured: NOT confirmed-flat, `<Platform>Unconfigured == [SOL|DOGE|NQ]`, closer calls == [coin], DM contains `manual intervention required`. Per-platform harness builds real `KillSwitchCloseInputs` with the same stub closers/fetchers, so the same code path is reached. Error strings became `<Platform> venue rate limited` / `<Platform> auth failed`; the contract (error text surfaces in DM) is unchanged.
- `TestPlanKillSwitchClose_{HL,OKX,RH,TS}FetcherUnwiredLatches` → `TestPlanKillSwitchClose_FetcherUnwiredLatches` (kill_switch_close_test.go) — carried: NOT confirmed-flat with nil fetcher and platform configured (HL: `HLAddr` set, `HLStateFetched=false`); log contains `<X>Fetcher unwired`.
- `TestPlanKillSwitchClose_DeterministicErrorOrder`,`..._RobinhoodDeterministicErrorOrder`,`..._OKXDeterministicErrorOrder` → `TestPlanKillSwitchClose_DeterministicErrorOrder` subtests `HL|Robinhood|OKX` (kill_switch_close_test.go) — carried: DM identical across 10 runs; coin entries alphabetical (HL BTC<DOGE<ETH<SOL; RH BTC<DOGE<ETH; OKX BTC<ETH<SOL). Order check now also fails when a coin is absent (`idx < 0`).
- `TestPlanKillSwitchClose_HLSuccessOKXFailureStillLatches`,`..._RobinhoodFailureStillLatchesAcrossPlatforms`,`..._TopStepFailureStillLatchesAcrossPlatforms` → `TestPlanKillSwitchClose_PeerPlatformFailureStillLatches` subtests `OKX|Robinhood|TopStep` (kill_switch_close_test.go) — carried: NOT confirmed-flat when the peer venue fails while HL succeeds; HL `ClosedCoins == [ETH]` (previously TopStep-only, now all three); DM contains `LATCHED, RETRYING` (previously OKX-only, now all three); DM contains peer error text (`okx err` / `rh err` / `venue down`).
- `TestPlanKillSwitchClose_HLAndOKXHappyPath` → folded into `TestPlanKillSwitchClose_HLAndOKXAndRobinhoodHappyPath` (kill_switch_close_test.go) — carried: confirmed-flat; HL closer calls == [ETH]; OKX closer calls == [BTC]; DM contains `HL closes: [ETH]` and `OKX closes: [BTC]`.
- `TestPlanKillSwitchClose_OKXSpotSurfacesGap`,`..._RobinhoodOptionsSurfacesGap` → `TestPlanKillSwitchClose_OperatorRequiredGapSuppressesAutoReset` subtests (kill_switch_close_test.go) — carried: presence does NOT block confirmed-flat; `OKXSpotPresent`/`RHOptionsPresent` true; `CanAutoResetWithoutOwner()` false; DM contains `OKX spot` / `Robinhood options`.
- `TestKillSwitchInstanceLabel_UsesConfigDirBasename`,`..._FallsBackWhenPathGivesNothingUseful` → `TestKillSwitchInstanceLabel` (kill_switch_close_test.go) — carried: `live`, `paper-hl-eth`, bare `config.json` → neither `"."` nor `""`.
- `TestFormatKillSwitchResetPrompt_{ConfirmedFlatIncludesContextAndIdentity,LatchedRetryingWarnsProtectionMayBeGone,OmitsAddressWhenHLNotConfigured}` → `TestFormatKillSwitchResetPrompt` (kill_switch_close_test.go) — carried verbatim as want/reject lists: instance label, HL address, reused reason, reused `HL closes: [ETH]`, `does not itself close or protect any position`, no `still retrying` on confirmed-flat; `still retrying` + `stop-losses may already be cancelled` on latched; no dangling `()` and label present when address empty.
- `TestCollectHLKillSwitchStopOIDs_IncludesManualTriggers`,`..._LowerCaseManualSymbolFallsBackToRawKey` → `TestCollectHLKillSwitchStopOIDs` (kill_switch_manual_test.go) — carried: BTC → [333], ETH → [111 222], `sol` → [444] under raw lower-case key; added: no extra coins collected.
- `TestParseKillSwitchResetDMTimeout_{DefaultsToSixHours,ParsesGoDuration,RejectsInvalid,RejectsNonPositive}` → `TestParseKillSwitchResetDMTimeout` (kill_switch_reset_dm_test.go) — carried: `""`→default; default == 6h (implies old `!= 30m`); `"30m"`→30m; `"nonsense"`,`"0"`,`"-1h"` → error.
- `TestLoadConfig_AppliesKillSwitchResetDMTimeout`,`TestLoadConfig_KillSwitchResetDMTimeoutDefaultsToSixHours`,`TestLoadConfig_RejectsInvalidKillSwitchResetDMTimeout` → `TestLoadConfig_KillSwitchResetDMTimeout` subtests (kill_switch_reset_dm_test.go) — carried: `cfg.KillSwitchResetDMTimeout == "45m"` + runtime 45m after apply; omitted → probe OK + empty `Config{}` apply → default; `"nonsense"` → probe error. New shared fixture `writeKillSwitchResetDMConfig(t, field)` replaces five copies of the same JSON (also used by `LoadConfigForProbe_DoesNotMutate...` and `ConcurrentProbeDoesNotRace`; their assertions untouched).
- `TestPlanKillSwitchClose_AdoptionEligibleLimitFillStillDefersToReconciler` → folded into `TestPlanKillSwitchClose_UnadoptedLimitFillKeepsRowAndBlocksFlat` (kill_switch_limit_orders_test.go) — identical setup; carried: NOT confirmed-flat/auto-reset, DM contains `queue row kept so the scheduler books the fill`, DM lacks `NO automatic path can book`. Survivor keeps: row not deleted, row retained with `FilledSize == 0`, DM contains `unadopted fill`.
- `TestPlanKillSwitchClose_OrphanedLimitRowWithFillEscalates` → third case of `TestPlanKillSwitchClose_AdoptionIneligibleLimitFillNamesTheBlock` (kill_switch_limit_orders_test.go) — carried: NOT confirmed-flat/auto-reset; row never auto-deleted; DM contains `NO automatic path can book`, `absent from this config`, `manual intervention required`; DM lacks `queue row kept so the scheduler books the fill`. That negative check now applies to all three ineligible cases and passes.

## Loosened / rewritten
None. Every exact-string check on DM/prompt text naming a state (`LATCHED, RETRYING`), a remediation (`manual intervention required`), or a booking promise was kept verbatim.

## Kept-but-flagged (audit findings, no edit)
- `TestPlanKillSwitchClose_ZeroInputsAreSafe` (close) — near-subset of `TestPlanKillSwitchClose_NoHLConfigured`; differs only by an empty `PortfolioReason`. Left alone as a distinct formatting input.
- `TestPlanKillSwitchClose_PlatformBudgetOverrides` (close) and `TestKillSwitchLimitOrderBudgetFloor` (limit_orders) — pin timeout arithmetic; kept because the floor guarantees one cancel+status round trip fits in the pre-flatten leg.
- `TestHyperliquidRawCoin_ManualSymbolWinsOverArgs` (manual) — looks like a getter pin, but decides which coin the kill switch closes for a `manual` strategy whose `Args[1]` diverges from `Symbol`; kept.
- `TestPlanKillSwitchClose_LimitCancelErrorBlocksFlat` / `_LimitRowDeleteFailureBlocksFlat` / `_LimitOrderCancellerUnwiredBlocksFlat` / `_LimitQueueLoadErrorBlocksFlat` (limit_orders) — same "block flat + auto-reset" shape but each wires a different failing dependency with a distinct side-effect assertion; left separate.
- `TestModelOnlyCloseAlert_ThrottledPerStrategySymbol`, `TestQueueModelOnlyCloseAlert_DrainsAndSkipsEmpty` (manual) — use package-level queue state; order-sensitive if `t.Parallel()` is ever added.

## Code workarounds found (do NOT fix, report only)
None found.

## Files: scheduler/hyperliquid_liquidation_guard_test.go, scheduler/hyperliquid_fills_test.go   Before: 2072 + 379 = 2451 lines / 60 + 13 = 73 Test funcs   After: 1672 + 339 = 2011 lines / 38 + 6 = 44 Test funcs

Counts measured with `wc -l` and `grep -c '^func Test'` against `git show origin/main:<path>` (before) and the worktree file (after). All assigned tests green (`go -C <wt>/scheduler test -count=1 -run '<names>' .`); `go vet .` clean.

Audit of the interrupted agent's diff: every removed test is a merge into a table-driven survivor. One non-compliance found and fixed (Loosened / rewritten, item 1). One dead helper branch removed (item 2). No other loosening on prices, quantities, fees, or clamp states.

## Removed
Nothing was removed outright. Every original test function that no longer exists by name is carried by a merged survivor listed below.

### hyperliquid_liquidation_guard_test.go
- `TestPlanHyperliquidLiquidationAuditClassification`, `TestPlanHyperliquidLiquidationAuditRearmsUnprotectedScalarOwner`, `TestPlanHyperliquidLiquidationAuditRefusesUnreconciledCoin` → `TestPlanHyperliquidLiquidationAudit` — carried verbatim: action count 2, deterministic sort `a-trailing` then `b-scalar`, every kind `hlAuditTighten`, clamped trigger `2340.5*(1+buffer)`; one `hlAuditRearm` job at trigger 2325; every action `hlAuditRefuse` on `BookConsistent:false`.
- `TestRunHyperliquidLiquidationAuditSkipsPaper`, `...RefusesPhantomOnSharedCoin`, `...NoOpsWithoutOnChainPosition`, `...SkipsWithoutSnapshot`, `...StaleSideSoleOwnerDoesNothing`, `TestAuditRearmSkipsStaleSideSoleOwner` → `TestRunHyperliquidLiquidationAuditNoOpCases` — carried: placement calls 0; `ImmediateFills 0` and `CloseDetails` empty; position `StopLossOID`/`StopLossTriggerPx` untouched (4242/2325, or 0/0 for zeroed-stop rows); no alert key stored where the original asserted it (paper, no-snapshot, both stale-side rows). Paper row now also asserts empty result, untouched trigger, and no alert (stricter).
- `TestRunHyperliquidLiquidationAuditIsIndependentOfSignalResults`, `...HealsConsistentNonDuePosition` → `TestRunHyperliquidLiquidationAuditHealsRegardlessOfDueness` — carried: a placement happens for a non-due 4h trailing-ATR owner and for a non-due 1d static scalar owner.
- `TestRunHyperliquidLiquidationAuditReplacesScalarStop`, `...CancelThenPlaceIsANormalClamp`, `...FilledExternallyReportsFilled`, `...RestingClampEmitsNoClose` → `TestRunHyperliquidLiquidationAuditTightenOutcomes` — carried: `ImmediateFills 0` / no `CloseDetails`, cancel OID 4242, replace size 1.0, replace trigger `2340.5*(1+buffer)`, position oid 9001 at clamped trigger, oid 5150 + `hlLiquidationActionClamped`, external fill books 0 + `hlLiquidationActionFilledOnChain` (plus position stays 4242/2325).
- `TestRunHyperliquidLiquidationAuditFillAtSubmitReportsExited`, `...ImmediateCloseReasonsTheAudit`, `...ReportsBookedCloses` → `TestRunHyperliquidLiquidationAuditImmediateFillBooksClose` — carried on both rows: fills 1 / close details 1, `CloseDetails[0]` SC.ID `hl-eth` / Symbol `ETH` / non-empty Detail, position deleted, `CloseReason == "liquidation_clamp_sl_immediate"`, trade `Details` contains `Liquidation-clamp SL`, `hlLiquidationActionExited`.
- `TestAuditRetriesPlacementItStrippedSameCycle` (2 rows + first-try subtest), `TestAuditRetryAdoptsBookDiffResolvedOID`, `TestAuditRetryOutcomeUnknownKeepsStateAndReportsUnknown`, `TestAuditRetryCapRejectedStillProtectionLost` → `TestAuditRetriesPlacementItStrippedSameCycle` (4 rows + first-try subtest) — carried: exactly 2 placement calls, retry `cancelOID 0`, final oid 9002 / Clamped; book-diff oid 9002 adopted with `StateMutations >= 1`; outcome-unknown keeps 4242/2325 and reports `hlLiquidationActionOutcomeUnknown`; cap-rejected retry clears to 0/0 and reports `hlLiquidationActionProtectionLost`; first-try rest makes 1 call.
- `TestAuditRearmOutcomeUnknownRecordsTriggerAndStops`, `TestAuditRearmCapRejectedStillRetriedNextCycle`, `TestAuditRearmAdoptsBookDiffResolvedOID`, `TestAuditRearmProceedsWithMatchingSideAndUnknownGeometry` → `TestAuditRearmOutcomes` — carried: re-arm passes `cancelOID 0` (now on every row); outcome-unknown: 1 call across 2 passes, oid 0, recorded trigger `2340.5*1.005`, `hlLiquidationActionPlacementUnknown` after each pass; cap-rejected: 2 calls across 2 passes, `hlLiquidationActionRearmFailed`; book-diff: oid 9002, `hlLiquidationActionRearmed`; unknown geometry with matching side: trigger `2400*(1-0.03125)`.
- `TestLiquidationAuditIntervalSeconds`, `TestLiquidationAuditIntervalStrictlyShorterAndCoversManual` → `TestLiquidationAuditIntervalSeconds` — carried: 150, 0 without live HL perps/manual, 7200 for a 14400 fleet (strictly-shorter implied by the exact value), 1800 manual-only, floor `liquidationAuditMinIntervalSeconds` for a 30s fleet, 0 for paper manual.
- `TestHLLiquidationAlertMessageOmitsUnknownGeometry`, message half of `TestOutcomeUnknownIsNotProtectionLost`, `TestLiquidationAlertPlacementUnknownWording` → `TestHLLiquidationAlertMessage` — carried: no `$0.0000`, `NO exchange-side stop`, `**HL POSITION UNPROTECTED**` headline, unprotected true; armed refusal names `$2325.0000`/`$2340.5000`, not unprotected; outcome-unknown headline `OUTCOME UNKNOWN`, contains `could NOT be read`/`may be resting`/`KEPT`, excludes `no exchange-side stop right now`, not unprotected; tighten detail contains `CANCELLED`; fresh placement-unknown contains `Nothing was cancelled`, excludes `old trigger was CANCELLED`, not unprotected.
- `TestHLLiquidationArmClampActionOutcomeUnknown` → folded into `TestOutcomeUnknownIsNotProtectionLost` — carried: outcome-unknown → `hlLiquidationActionPlacementUnknown`; cap rejection → `hlLiquidationActionRearmFailed`; original predicate checks (`hlLiquidationActionUnprotected`, `hlLiquidationMayRetryReplace`) retained.

### hyperliquid_fills_test.go
- `TestLookupHyperliquidFillByOID_AggregatesPartialFills`, `_NoMatchReturnsFalse`, `_EmptyAddressShortCircuits`, `_AccumulatesFilledQty` → `TestLookupHyperliquidFillByOID` — carried: Count 2, Fee 0.80 (tol 1e-3, unchanged), ClosedPnLGross 150.00 (tol 1e-2, unchanged), FilledQty 0.311 (tol 1e-9, unchanged); ok=false on missing OID and on empty address. Every row now asserts Fee/PnL/FilledQty.
- `TestLookupHyperliquidFillByCoinSize_MatchesByCoinAndSize`, `_PicksNewestGroupNotSumOfWindow`, `_AggregatesPartialFillsByAnchorOID`, `_SetsFilledQtyNoOIDCase`, `_AccumulatesFilledQtyWithOID` → `TestLookupHyperliquidFillByCoinSize` — carried: Count 1 / Fee 0.40; OID 200 / Count 1 / Fee 0.50 / PnL 90.00; OID 555 / Count 2 / Fee 0.50 / PnL 70.00; FilledQty 0.211 with nil OID; FilledQty 0.211 summed across OID 77777. Every row now asserts OID, Count, Fee, PnL, FilledQty.

## Loosened / rewritten
1. `TestRunHyperliquidLiquidationAuditNoOpCases/"failed snapshot fetch ..."` — was (interrupted agent's merge): passed `tc.live` (true) as `snapshotFetched`, so the row no longer reached the `!snapshotFetched` early return at `scheduler/hyperliquid_liquidation_guard.go:859` that the original `TestRunHyperliquidLiquidationAuditSkipsWithoutSnapshot` exercised with `false` — now asserts: a `snapshot` table column, `false` for this row and `true` for the others, passed through to `runHyperliquidLiquidationAudit` — ground: the original test on origin/main and the signature (`snapshotFetched bool`, sixth parameter).
2. `assertFillField` (fills helper) — was: skipped the comparison when `want < 0`, a latent way for a table row to assert nothing — now asserts every Fee / ClosedPnLGross / FilledQty value on every row unconditionally — ground: no row uses a negative sentinel; the branch was dead.
3. `TestHLLiquidationAlertMessage` (from `...OmitsUnknownGeometry`) — headline check is `strings.Contains` on `**HL POSITION UNPROTECTED**` where the original used exact equality. Operator wording still pinned; left as the merge wrote it.

## Kept-but-flagged (audit findings, no edit)
- `TestRunHyperliquidLiquidationAuditNoOpCases/"failed snapshot fetch ..."` — even with `snapshotFetched=false` restored, the row cannot fail if the early return were deleted: its on-chain map is empty, so `collectHLLiquidationAuditCandidates` yields no candidate either way. The origin/main test had the same blind spot. A populated on-chain map with `snapshotFetched=false` would make it fail on that defect.
- `TestRunHyperliquidLiquidationAuditKeepsOriginalStopOnFailure` — its `{StopLossFilledExternally: true}` row duplicates the "filled externally" row of `...TightenOutcomes` (both assert 4242/2325 untouched). Kept; the loop is one table over failure classes.
- `TestAuditTightenOutcomeUnknownRecordsTriggerAndStops` — pins `pos.StopLossTriggerPx != 2340.5*1.005` with exact float equality and a literal `1.005` instead of `hlLiquidationStopBufferPct`. A buffer constant change breaks the literal. Left as-is (price contract).
- `TestHLLiquidationShouldNotifyThrottle`, `TestHLLiquidationAlertDueAndClear` — depend on the package-global `hlLiquidationAlerts` map and `effectiveAlertThrottleInterval()`; not `t.Parallel()`-safe. No test in these files calls `t.Parallel()`.
- `TestLookupHyperliquidFillByOID/"empty address short-circuits"` — now spins up a fills server that is never contacted (original did not). Harmless; assertion unchanged.
- `TestReconcileFillLookupSinceMs_BoundsTo24h` — restates `now - hlReconcileFillLookupWindow`. Kept; it pins the backward direction of the reconcile window.

## Code workarounds found (do NOT fix, report only)
None found.

## Removed
- `TestComputeTotalPortfolioValue_FallbackKeepsPeakFreezeSignal` — Obsolete-redundant — survivor `TestComputeTotalPortfolioValue/"fetch failure falls back..."` carries total==sum of member cash, usedFallback==true.
- `TestComputeTotalPortfolioValue_DelegatesCorrectly` — Obsolete-redundant — survivor `TestComputeTotalPortfolioValue/"mixed shared and non-shared..."` carries total==balance+spot PV, usedFallback==false.
- `TestReportSharedWalletDrift_WithinToleranceNoPanic` — Obsolete-redundant (asserted nothing) — survivor `TestReportSharedWalletDrift_ToleranceBoundary/"below tolerance records nothing"` carries same call + len(entries)==0.
## Merged
- 11 `TestWalletKeyFor_*` → `TestWalletKeyFor`; 8 `TestDetectSharedWallets_*` (+TopStep #1106 row restored) → `TestDetectSharedWallets`; 8 `TestComputeTotalPortfolioValue_*` → `TestComputeTotalPortfolioValue`; 2 `TestFetchSharedWalletBalances_*`; 6 `TestComputeInitialPortfolioPeak_*`; 5 `TestRebaselinePortfolioPeakAfterPrune_*`; 4 `TestComputeSubsetPortfolioValue_*`; 4 `TestPlanTradeLedgerForStrategy_NoOIDRepairSkips*`; 5 `TestReconcileSharedWallet_*` → `TestReconcileSharedWalletMemberValues`; 13 `TestSharedWalletDriftTracker_*` → `TestSharedWalletDriftTracker`; 4 `TestComputeSubsetDisplayValue_*`; 2 journal alert format tests → `TestFormatSharedWalletJournalAlerts`; 2 `TestReportSharedWalletDrift_*` → `_ToleranceBoundary`; 3 `TestHyperliquidKillSwitchFillShare_*` → one table. Every exact amount/flag/count carried.
## Loosened / rewritten
- None.
## Kept-but-flagged
- `TestHasSharedWalletBalanceFetcher_HLAndOKX`, `TestDisplayStrategyValue_PrefersSetValue` (small accessor pins, kept as gates).
## Code workarounds found
- None.

## Files: scheduler/db_test.go, scheduler/state_test.go   Before: 3319 lines / 80 Test funcs (db 2495/55, state 824/25)   After: 2962 / 54 (db 2246/40, state 716/14)

Counts measured with `wc -l` and `grep -c '^func Test'`. All assigned tests pass (`go test -count=1`), `go vet .` compiles, both files gofmt-clean. Removed nothing outright; every function that disappeared was folded into a table or subtest with its assertions carried.

## Removed
None. No pure duplicate found that another test fully covers.

## Merged
- `TestNewStrategyState`,`TestNewStrategyState_ConfigInitialCapital`,`TestNewStrategyState_NoConfigInitialCapital` → `TestNewStrategyState` (state_test.go) — assertions carried: ID/Type/Platform identity, Cash from Capital (1000, 600, 600), InitialCapital (1000, 505 from config, 600 fallback), non-nil Positions/OptionPositions/TradeHistory, RiskState.PeakValue = Capital, RiskState.MaxDrawdownPct. Peak and MaxDrawdown checks now run on all three cases (previously only the first).
- `TestValidatePerpsDirectionConfig`,`_NoConflicts`,`_OrphanState`,`_LegacyAllowShortsFallthrough`,`_RegimePolicyStampedTrendingDown`,`_RegimePolicyStampedTrendingUpConflict`,`_RegimePolicyUncertifiedWarnsForMigration`,`_WarningOrderDeterministic` → `TestValidatePerpsDirectionConfig` (state_test.go) — assertions carried: warning counts (2/0/0/1/0/1/1/2), joined text contains `hl-triple-ema-eth`, `hl-bear-sol`, `direction=`; `stamped regime="trending_up"` in the conflict warning; `DEFAULT-OFF` + `#1085` in the uncertified warning; ` BTC ` before ` ETH ` ordering. Subtest name keeps `#1085`.
- `TestSaveStateWithDB`,`TestSaveStateWithDB_Error` → `TestSaveStateWithDB` subtests (state_test.go) — CycleCount 3 persisted; closed DB returns error.
- `TestReconcileConfigInitialCapital`,`TestReconcileConfigInitialCapital_NoOpWhenAligned` → `TestReconcileConfigInitialCapital` subtests (state_test.go) — 1 info / 0 errs, in-memory 1000, untouched 200, persisted 1000 after SaveState; no infos/errs when aligned.
- `TestQueryTradeHistory_NoFilter`,`_ByStrategy`,`_BySymbol`,`_LimitClamped` → `TestQueryTradeHistory_Filters` (db_test.go) — total/len 2 with newest (`sell`) first; strategy filter matches every row and nonexistent → 0/0; symbol filter every row `BTC`; limit 9999 → 2 rows.
- `TestTradeExchangeFieldsRoundTrip`,`TestTradeExchangeFields_EmptyByDefault`,`TestQueryTradeHistory_ExchangeFields` → `TestTradeExchangeFieldsRoundTrip` (db_test.go) — ExchangeOrderID/ExchangeFee pairs (1234567890/1.75, 1234567891/1.79, 9876543210/2.50, ""/0) asserted via BOTH `LoadState` and `QueryTradeHistory` for every case (previously each path was checked on one case only).
- `TestSaveState_PreservesInitialCapital`,`_AllowsFirstInitialCapitalWrite`,`_AllowsNewStrategies`,`_AllowsBaselineWhenPrevZero` → `TestSaveState_InitialCapitalGuard` (db_test.go) — in-memory restore (initial 505, cash 632) and persisted (505/632); first write 1000; brand-new 2000 + old 1000; prev-zero baseline 1000. In-memory and persisted initial+cash now asserted for every case.
- `TestSetInitialCapital_ExplicitOverride`,`TestSetInitialCapital_RejectsInvalid` → `TestSetInitialCapital` subtests (db_test.go) — override 750 survives a stale 505 save; rejects 0, negative, unknown id.
- `TestSaveLoadState_LeaderboardSummaries`,`TestSaveLoadState_LastSummaryPost` → `TestSaveLoadState_TimestampMaps` (db_test.go) — 2 entries each, every key present, `time.Equal` per value.
- `TestSaveAndLoadDB_PendingCircuitCloseRoundTrip`,`TestSaveAndLoadDB_LegacyPendingHLJSON_MigratesOnLoad` → `TestSaveAndLoadDB_PendingCircuitClose` subtests (db_test.go) — `getPendingCircuitClose(hyperliquid)` non-nil, 1 symbol, ETH 0.2585, for the typed round trip and for injected legacy `{"coins":[...]}` JSON.
- `TestLifetimeTradeStatsAll_FreshInsert`,`_PartialClosesNetByPositionID`,`_PositionIDScopedByStrategy`,`_BreakevenPositionNeitherWinNorLoss`,`TestLifetimeTradeStats_SurvivesRiskStateReset` → `TestLifetimeTradeStatsAll` (db_test.go) — exact `LifetimeTradeStats{PositionsOpened,Wins,Losses}` per strategy for each case (s1 2/1/1 + s2 1/1/0; 1/1/0; s1 0/1/0 + s2 0/0/1; 0/0/0; 0/1/1), s3 absent from the map and zero from `LifetimeTradeStatsForStrategy`; `LifetimeTradeStatsForStrategy` now cross-checked for every expected strategy (previously only s1 in FreshInsert). `SurvivesRiskStateReset` had a no-op `_ = RiskState{}` line; its real assertion (stats derived from trades rows) is carried as the last subtest.

## Loosened / rewritten
- `TestSaveState_KillSwitchEventsCapped` → renamed `TestSaveState_KillSwitchEventsStoredAsIs` — was: name claimed a cap while the body asserts 60 rows stored — now asserts: unchanged (60 rows stored) — ground: db.go has no cap on `kill_switch_events` (delete-all + insert-all at db.go:1370-1374); the name was misleading, the assertion was correct.

## Kept-but-flagged (audit findings, no edit)
- `TestCorrelationSnapshotRoundTrip` — first half (non-nil, PortfolioGrossUSD 5000) duplicates `TestSaveAndLoadDBRoundTrip`; kept because the reload is the precondition for its unique nil-snapshot-clears assertion.
- `TestLoadState_TradeHistoryBoundedInSQL` — asserts `EXPLAIN QUERY PLAN` names `idx_trades_strategy_timestamp` and has no `SCAN TABLE`. That pins SQLite planner output text, but index use is the documented `LoadState` contract (CLAUDE.md), so left alone.
- `TestRecordClosedPosition_ExecuteSignal`, `TestRecordClosedOptionPosition_ExecuteClose` — wall-clock `DurationSeconds` windows (7100-7300, 10700-11000); tolerant, but a stalled CI could drift. Not changed.
- `TestSaveState_KillSwitchEventsStoredAsIs` — asserts absence of a cap; if a cap is ever intended (the old name suggests one was), this test and db.go both need the change.
- `TestOpenStateDB` — pins table list and `journal_mode=wal`; WAL is an operational contract (concurrent reader during save), kept.

## Code workarounds found (do NOT fix, report only)
None found. No non-test source was edited.

## Files: scheduler/config_test.go, scheduler/config_audit_test.go, scheduler/config_unknown_keys_test.go, scheduler/config_example_version_test.go, scheduler/config_regime_directional_warn_test.go   Before: 3472 lines / 112 Test funcs   After: 2763 lines / 78 Test funcs

Per file (before -> after, wc -l / grep -c '^func Test'): config_test.go 2813/86 -> 2158/58; config_audit_test.go 220/7 -> 220/7 (untouched); config_unknown_keys_test.go 348/16 -> 290/11; config_example_version_test.go 39/1 -> 39/1 (untouched); config_regime_directional_warn_test.go 52/2 -> 56/1.

Tests green before and after; `go vet .` clean. Seven-owner stop exclusivity tests, every LoadConfig load/migration test, and every unknown-key guard assertion are unchanged in effect.

## Removed
- `TestLoadConfigManualDefaultsStopLossATRMultTo2Point0` (config_test.go) — case: Obsolete-redundant — ground: byte-identical config JSON to the survivor (same manual strategy, no user_defaults, no default_stop_loss_atr_mult), same LoadConfig path (applyManualDefaults -> defaultManualStopLossATRMult) — survivor: `TestLoadConfigManualDefaultsAbsentPreservesHardcodedDefaults` carries: StopLossATRMult != nil and *StopLossATRMult == defaultManualStopLossATRMult, plus the tp_tiers fallback the removed test never asserted.
- `TestKnownStrategyConfigKeysIncludesCircuitBreaker` (config_unknown_keys_test.go) — case: Obsolete-redundant — ground: same knownStrategyConfigKeys() lookup — survivor: `TestKnownStrategyConfigKeysCoversCoreFields` carries: known["circuit_breaker"] (appended to mustHave).
- `TestLoadConfigHLPerpsPeersOnSameCoinMatching` (config_test.go) — case: Obsolete-redundant — ground: byte-identical two-peer JSON to `TestLoadConfigHLPerpsPeersOmittedStopLossDoesNotConflict`, same LoadConfig + EffectiveStopLossPct path — survivor: subtest "omitted stop_loss_* on both same-coin peers does not conflict" of merged `TestLoadConfigHLPerpsPeers` carries: len(cfg.Strategies)==2, EffectiveStopLossPct(sc)==0 for each peer.

## Merged
- `TestLoadConfigLeverageRejectsSpot`,`TestLoadConfigLeverageRejectsOutOfRange`,`TestLoadConfigSizingLeverageRejectsSpot`,`TestLoadConfigSizingLeverageRejectsOutOfRange`,`TestLoadConfigMarginPerTradeUSDRejectsSpot`,`TestLoadConfigMarginPerTradeUSDRejectsNonPositive`,`TestLoadConfigMarginModeRejectsInvalidValue`,`TestLoadConfigMarginModeRejectsSpot` -> `TestLoadConfigPerpsSizingFieldRejections` (config_test.go) — assertions carried: LoadConfig errors and each error contains the exact original substring (leverage is only supported for perps; leverage must be in; sizing_leverage is only supported for perps; sizing_leverage must be in; margin_per_trade_usd is only supported for perps; margin_per_trade_usd must be positive; margin_mode must be; margin_mode is only supported for HL perps) with the same field values (5, 150, 2, 200, 100, 0, "portfolio", "isolated") on the same spot / HL-perps strategy shapes.
- `TestLoadConfigPerpsLeverageDefault`,`TestLoadConfigPerpsLeverageExplicit`,`TestLoadConfigPerpsSizingLeverageExplicit`,`TestLoadConfigSizingLeverageAcceptsFractional`,`TestLoadConfigHLPerpsDefaultsToIsolatedMargin`,`TestLoadConfigHLPerpsExplicitCross` -> `TestLoadConfigPerpsSizingFieldsAccepted` (config_test.go) — assertions carried: Leverage==1 default; Leverage==10 and SizingLeverage==10; EffectiveExchangeLeverage==20 and EffectiveSizingLeverage==2; SizingLeverage==0.5; MarginMode=="isolated" default; MarginMode=="cross" explicit.
- `TestLoadConfigManualOptsOutWhenGlobalDefaultIsZero`,`TestLoadConfigManualExplicitATRMultPreserved`,`TestLoadConfigManualDefaultsStopLossATRMultOverride`,`TestLoadConfigManualDefaultsStopLossATRMultZeroOptsOut` -> `TestLoadConfigManualStopLossATRMultResolution` (config_test.go) — assertions carried: StopLossATRMult nil for default_stop_loss_atr_mult:0; ==2.5 explicit; ==2.25 from user_defaults.manual; nil for user_defaults.manual.stop_loss_atr_mult:0 (nil vs value distinguished in a switch as the originals' Fatal/Errorf pairs did).
- `TestLoadConfigHLPerpsPeersOmittedStopLossDoesNotConflict`,`TestLoadConfigHLPerpsPeersMismatchedMarginMode`,`TestLoadConfigHLPerpsPeersMismatchedLeverage`,`TestLoadConfigHLPerpsPeersMultipleStopLossAllowed`,`TestLoadConfigHLPerpsPeersSingleStopLossAllowed`,`TestLoadConfigHLPerpsPeersDifferentCoinsIndependent`,`TestLoadConfigHLPerpsPeersNoStopLossAllowed`,`TestLoadConfigHLPerpsPeersDefaultedMarginModeMatches` -> `TestLoadConfigHLPerpsPeers` (config_test.go) — assertions carried: both omitted peers EffectiveStopLossPct==0 and len==2; error contains "disagree on margin_mode" AND "ETH"; "disagree on leverage"; two stop_loss_pct owners load; single owner ==3 / omitted peer ==0; different coins with differing leverage+margin_mode load; two stop_loss_pct:0 peers load; defaulted margin_mode peer loads and both report "isolated". Same ids/args/capital/leverage/margin_mode values as the originals.
- `TestConfigValidationDMChannelsInvalidKey`,`TestConfigValidationDMChannelsEmptyValue`,`TestConfigValidationDMChannelsValidKeys`,`TestConfigValidationDMChannelsOrphanSuffix` -> `TestConfigValidationDMChannels` (config_test.go) — assertions carried: error contains "dm_channels key"; contains dm_channels["hyperliquid-paper"]; contains "platform prefix is empty"; valid keys load with Discord.DMChannels hyperliquid=111, hyperliquid-paper=222, Telegram.DMChannels okx-paper=444.
- `TestConfigValidationManualPerpsPeerLeverageMismatchRejected`,`TestConfigValidationManualPerpsPeerMarginModeMismatchRejected` -> `TestConfigValidationManualPerpsPeerMismatchRejected` (config_test.go) — assertions carried: validateConfig error contains "disagree on leverage" (manual 10 vs perps 5) and "disagree on margin_mode" (isolated vs cross, both leverage 5).
- `TestValidateStrategyJSONKeysFlagsInventedTPField`,`TestValidateStrategyJSONKeysHintsLegacyParams`,`TestValidateStrategyJSONKeysHintsStopLossTypos` -> `TestValidateStrategyJSONKeysUnknownFieldHints` (config_unknown_keys_test.go) — assertions carried: exactly 1 error each; contains strategy[hl-momentum-btc]: unknown field "take_profit_atr_mult" and "close_strategy"; contains "open_strategy: {name, params}"; contains "valid SL fields".
- `TestValidateUserDefaultsJSONKeysFlagsUnknownSibling`,`TestValidateUserDefaultsJSONKeysFlagsUnknownManualLeaf` -> `TestValidateUserDefaultsJSONKeysFlagsUnknownKeys` (config_unknown_keys_test.go) — assertions carried: exactly 1 error; contains user_defaults: unknown field "manaul"; contains user_defaults.manual: unknown field "margin".
- `TestLoadConfigRejectsUnknownUserDefaultsManualLeaf`,`TestLoadConfigRejectsUnknownUserDefaultsSibling` -> `TestLoadConfigRejectsUnknownUserDefaultsKeys` (config_unknown_keys_test.go) — assertions carried: LoadConfig (config_version 16, manual strategy) errors naming user_defaults.manual: unknown field "stop_loss_atr_mlt" / user_defaults: unknown field "manaul".
- `TestRegimeDirectionalPolicyWarningsEdgeCases` -> subtests of `TestRegimeDirectionalPolicyWarnings` (config_regime_directional_warn_test.go) — assertions carried: nil config returns nil; no-policy strategies return 0; single-label policy returns exactly 1; original 3-strategy case (exactly 1 warning, names dir-policy, cites #1076, [WARN] prefix) unchanged.

Shared fixtures added in config_test.go: `testSpotStrategyHead`, `testHLPerpsStrategyHead`, `testHLManualStrategyHead` (JSON fragments with the same field values as the inline originals) and a one-line `itoa` wrapper (strconv import added).

## Loosened / rewritten
None.

## Kept-but-flagged (audit findings, no edit)
- `TestLoadConfigDefaults` — pins IntervalSeconds/LogDir/DBFile/AutoUpdate constants; kept whole because the DefaultStopLossATRMult seeding in the same test is a stop-owner contract.
- `TestOrdinal`, `TestStrategyIntervalExceedsGlobalWarning` — pin exact [WARN] wording ("every 3rd portfolio cycle"); the ordinal suffix is the logic under test and the substring is its only observable, so left.
- `TestConfigValidationLeaderboardSummariesDuplicateKey` — pins the "[0]" index in the error text; incidental, left.
- `TestCanonicalizeRegimeBlockAliasPrecedenceDeterministic` (config_audit_test.go) — 50-iteration loop against map-order nondeterminism; legitimate for a Go map fold, kept.
- `TestConfigValidationCompoundAllScalarStopOwnersRejected` / `...ScalarPlusRegime...` — protected seven-owner contract, kept. The trailing_stop_atr_mult_regime vs trailing_stop_atr_mult pair is not asserted in these two tests; not checked whether another file covers it.
- `TestLoadConfigForProbeSkipsLiveCredentialChecks` — also depends on HYPERLIQUID_ACCOUNT_ADDRESS being cleared; correct, kept.

## Code workarounds found (do NOT fix, report only)
None found in the config load, validation, or unknown-key paths exercised by these tests.

## Removed
- `TestNewFieldsSinceFieldProperties` — Obsolete-redundant — survivor `TestNewFieldsSince` (superset sweep).
- `TestMigrateConfigAtFloorNeverStripsLegacyNames`, `TestVersionlessConfigWithInertPreFloorKeysMigrates`, `TestMigrateConfigV8PreservesFieldsAtCurrentVersion` — merged into `TestMigrateConfigRetainsInertLegacyDiscordKeys` subtests.
- `TestV19AtomicRenameSitesListsEveryMutationSite` — Incidental constant pin; `v19AtomicRenameSites` (config_migration_v19.go:122) has no non-test reference.
- `TestV19NoticeMentionsBothRenames` — Incidental notice wording; rename asserted by `TestMigrateV19AtrMultRegimeKey`, `TestLoadConfigV19Migrates*`.
- `TestFormatFloatPtr`, `TestFormatFloatPtrUSD`, `TestFormatFloatPtrPct` — Incidental display-formatter pins (config_reload.go change-list strings only).
- `TestValidateHotReloadCompatible_RegimeWindowOnlyChange`, `_SharedWalletPoolModeRequiresRestart`, `TestStrategyRestartShape_RegimeWindowOnlyChange` — Obsolete-redundant — rows in `TestValidateHotReloadCompatible` / `TestStrategyRestartShapeIgnoresHotReloadableFields` carry identical inputs and assertions.
## Merged
- 4 → `TestMigrateConfigWritesValues`; 2 → `TestMigrateConfigRejectsUnreadableInput`; 3 → `TestMigrateConfigV16UserDefaults`; 5 → `TestMigrateV15CloseKeys`; 2 → `TestMigrateV18TrailStopATRRegimeKey`; 2 → `TestNeedsV18TrailStopKeyMigration`; 3 → `TestMigrateV19AtrMultRegimeKey`; 2 → `TestNeedsV19AtrMultRegimeRename`; 2 → `TestV19RenameMap`; 20 `TestApplyHotReloadConfig*` guards → `TestApplyHotReloadConfigOpenPositionGuards` (every error substring and applied value carried; non-mutation now whole-object DeepEqual); 3 → `TestApplyHotReloadConfigTogglesWhileOpen`; 4 → `TestStrategyRestartShapeIgnoresHotReloadableFields`; 2 → `TestPortfolioRiskAccessorsNilSafe`.
## Loosened / rewritten
- `TestLoadConfigV19ReadOnlyLegacyKeyStillAttemptsRewrite` — now uses shared `readOnlyConfigDir` helper; same path and assertion.
## Kept-but-flagged
- Read-only-path tests skip as root / non-enforcing FS.
## Code workarounds found
- None. `v19AtomicRenameSites` is dead non-test code (follow-up candidate).

## Follow-on (unfiled)

- `v19AtomicRenameSites` (`scheduler/config_migration_v19.go:122`) is dead non-test code once its constant pin was removed; delete in a follow-up.
- `update_sh_feature_test.go` macOS bash 3.2 failures: Phase B.

---

### DIFF
--- a/scheduler/config_migration_test.go
+++ b/scheduler/config_migration_test.go
@@ -23,6 +23,9 @@ func TestNewFieldsSince(t *testing.T) {
 			if f.Version <= version {
 				t.Errorf("NewFieldsSince(%d) returned field %q with version %d", version, f.JSONPath, f.Version)
 			}
+			if f.JSONPath == "" || f.Description == "" || f.FieldType == "" || f.Version <= 0 {
+				t.Errorf("NewFieldsSince(%d) returned malformed field %+v", version, f)
+			}
 		}
 	}
 }
@@ -36,185 +39,86 @@ func TestMinSupportedConfigVersionFloor(t *testing.T) {
 	}
 }
 
-func TestNewFieldsSinceFieldProperties(t *testing.T) {
-	fields := NewFieldsSince(0)
-	for _, f := range fields {
-		if f.JSONPath == "" {
-			t.Error("field has empty JSONPath")
+func TestMigrateConfigWritesValues(t *testing.T) {
+	cases := []struct {
+		name   string
+		values map[string]string
+		check  func(t *testing.T, updated map[string]interface{})
+	}{
+		{
+			name:   "stamps_version_and_sets_value",
+			values: map[string]string{"discord.owner_id": "12345"},
+			check: func(t *testing.T, updated map[string]interface{}) {
+				discord, ok := updated["discord"].(map[string]interface{})
+				if !ok {
+					t.Fatal("discord section missing")
+				}
+				if discord["owner_id"] != "12345" {
+					t.Errorf("discord.owner_id = %v, want %q", discord["owner_id"], "12345")
+				}
+				if updated["interval_seconds"].(float64) != 300 {
+					t.Error("interval_seconds should be preserved")
+				}
+				if _, ok := updated["default_stop_loss_atr_mult"]; ok {
+					t.Error("default_stop_loss_atr_mult should no longer be backfilled on disk (#1285)")
+				}
+			},
+		},
+		{
+			name:   "creates_nested_paths",
+			values: map[string]string{"discord.dm_channels.hyperliquid": "999888777"},
+			check: func(t *testing.T, updated map[string]interface{}) {
+				discord := updated["discord"].(map[string]interface{})
+				dmCh := discord["dm_channels"].(map[string]interface{})
+				if dmCh["hyperliquid"] != "999888777" {
+					t.Errorf("discord.dm_channels.hyperliquid = %v, want %q", dmCh["hyperliquid"], "999888777")
+				}
+			},
+		},
+		{
+			name:   "nil_values_still_stamp_version",
+			values: nil,
+			check:  func(t *testing.T, updated map[string]interface{}) {},
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			dir := t.TempDir()
+			path := filepath.Join(dir, "config.json")
+			writeRawConfig(t, path, map[string]interface{}{
+				"interval_seconds": 300,
+				"strategies":       []interface{}{},
+			})
+			if err := MigrateConfig(path, tc.values, nil); err != nil {
+				t.Fatalf("MigrateConfig failed: %v", err)
+			}
+			if _, err := os.Stat(path + ".tmp"); !os.IsNotExist(err) {
+				t.Error("temp file should not exist after migration")
+			}
+			updated := readRawConfig(t, path)
+			if version := int(updated["config_version"].(float64)); version != CurrentConfigVersion {
+				t.Errorf("config_version = %d, want %d", version, CurrentConfigVersion)
+			}
+			tc.check(t, updated)
+		})
+	}
+}
+
+func TestMigrateConfigRejectsUnreadableInput(t *testing.T) {
+	t.Run("missing_file", func(t *testing.T) {
+		if err := MigrateConfig("/nonexistent/config.json", nil, nil); err == nil {
+			t.Error("expected error for missing file")
 		}
-		if f.Description == "" {
-			t.Errorf("field %q has empty Description", f.JSONPath)
+	})
+	t.Run("invalid_json", func(t *testing.T) {
+		path := filepath.Join(t.TempDir(), "config.json")
+		if err := os.WriteFile(path, []byte("not json"), 0600); err != nil {
+			t.Fatal(err)
 		}
-		if f.FieldType == "" {
-			t.Errorf("field %q has empty FieldType", f.JSONPath)
+		if err := MigrateConfig(path, nil, nil); err == nil {
+			t.Error("expected error for invalid JSON")
 		}
-		if f.Version <= 0 {
-			t.Errorf("field %q has invalid version %d", f.JSONPath, f.Version)
-		}
-	}
-}
-
-func TestMigrateConfigBasic(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-
-	original := map[string]interface{}{
-		"interval_seconds": 300,
-		"strategies":       []interface{}{},
-	}
-	data, err := json.MarshalIndent(original, "", "  ")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(path, data, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	values := map[string]string{
-		"discord.owner_id": "12345",
-	}
-	if err := MigrateConfig(path, values, nil); err != nil {
-		t.Fatalf("MigrateConfig failed: %v", err)
-	}
-
-	result, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(result, &updated); err != nil {
-		t.Fatal(err)
-	}
-
-	version := int(updated["config_version"].(float64))
-	if version != CurrentConfigVersion {
-		t.Errorf("config_version = %d, want %d", version, CurrentConfigVersion)
-	}
-
-	discord, ok := updated["discord"].(map[string]interface{})
-	if !ok {
-		t.Fatal("discord section missing")
-	}
-	if discord["owner_id"] != "12345" {
-		t.Errorf("discord.owner_id = %v, want %q", discord["owner_id"], "12345")
-	}
-
-	if updated["interval_seconds"].(float64) != 300 {
-		t.Error("interval_seconds should be preserved")
-	}
-	if _, ok := updated["default_stop_loss_atr_mult"]; ok {
-		t.Error("default_stop_loss_atr_mult should no longer be backfilled on disk (#1285)")
-	}
-}
-
-func TestMigrateConfigCreatesNestedPaths(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-
-	original := map[string]interface{}{"interval_seconds": 300}
-	data, err := json.MarshalIndent(original, "", "  ")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(path, data, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	values := map[string]string{
-		"discord.dm_channels.hyperliquid": "999888777",
-	}
-	if err := MigrateConfig(path, values, nil); err != nil {
-		t.Fatal(err)
-	}
-
-	result, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(result, &updated); err != nil {
-		t.Fatal(err)
-	}
-
-	discord := updated["discord"].(map[string]interface{})
-	dmCh := discord["dm_channels"].(map[string]interface{})
-	if dmCh["hyperliquid"] != "999888777" {
-		t.Errorf("discord.dm_channels.hyperliquid = %v, want %q", dmCh["hyperliquid"], "999888777")
-	}
-}
-
-func TestMigrateConfigNilValues(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-
-	original := map[string]interface{}{"interval_seconds": 300}
-	data, err := json.MarshalIndent(original, "", "  ")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(path, data, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	if err := MigrateConfig(path, nil, nil); err != nil {
-		t.Fatal(err)
-	}
-
-	result, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(result, &updated); err != nil {
-		t.Fatal(err)
-	}
-
-	version := int(updated["config_version"].(float64))
-	if version != CurrentConfigVersion {
-		t.Errorf("config_version = %d, want %d", version, CurrentConfigVersion)
-	}
-}
-
-func TestMigrateConfigAtomicWrite(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-
-	original := map[string]interface{}{"interval_seconds": 300}
-	data, err := json.MarshalIndent(original, "", "  ")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(path, data, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	if err := MigrateConfig(path, nil, nil); err != nil {
-		t.Fatal(err)
-	}
-
-	if _, err := os.Stat(path + ".tmp"); !os.IsNotExist(err) {
-		t.Error("temp file should not exist after migration")
-	}
-}
-
-func TestMigrateConfigMissingFile(t *testing.T) {
-	err := MigrateConfig("/nonexistent/config.json", nil, nil)
-	if err == nil {
-		t.Error("expected error for missing file")
-	}
-}
-
-func TestMigrateConfigInvalidJSON(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-	if err := os.WriteFile(path, []byte("not json"), 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	err := MigrateConfig(path, nil, nil)
-	if err == nil {
-		t.Error("expected error for invalid JSON")
-	}
+	})
 }
 
 func TestSetNestedField(t *testing.T) {
@@ -348,39 +252,6 @@ func TestMigrateConfigRejectsPreFloorVersions(t *testing.T) {
 	}
 }
 
-func TestMigrateConfigAtFloorNeverStripsLegacyNames(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-	writeRawConfig(t, path, map[string]interface{}{
-		"config_version": MinSupportedConfigVersion,
-		"discord": map[string]interface{}{
-			"channel_paper_trades": true,
-			"channel_live_trades":  true,
-			"spot_summary_freq":    "hourly",
-		},
-	})
-	if err := MigrateConfig(path, nil, nil); err != nil {
-		t.Fatalf("MigrateConfig failed: %v", err)
-	}
-	result, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(result, &updated); err != nil {
-		t.Fatal(err)
-	}
-	if v := int(updated["config_version"].(float64)); v != CurrentConfigVersion {
-		t.Errorf("config_version = %d, want %d", v, CurrentConfigVersion)
-	}
-	discord := updated["discord"].(map[string]interface{})
-	for _, key := range []string{"channel_paper_trades", "channel_live_trades", "spot_summary_freq"} {
-		if _, ok := discord[key]; !ok {
-			t.Errorf("discord.%s should NOT be removed for a supported config", key)
-		}
-	}
-}
-
 func assertVersionlessDMKeyError(t *testing.T, err error, key, path string, original []byte) {
 	t.Helper()
 	if err == nil {
@@ -492,71 +363,69 @@ func TestVersionlessConfigRemovedTranslationKeyAcceptance(t *testing.T) {
 	}
 }
 
-func TestVersionlessConfigWithInertPreFloorKeysMigrates(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-	writeRawConfig(t, path, map[string]interface{}{
-		"interval_seconds": 3600,
-		"strategies":       []interface{}{},
-		"discord": map[string]interface{}{
-			"owner_id":             "disc-owner",
-			"channel_paper_trades": true,
-			"channel_live_trades":  true,
-			"spot_summary_freq":    "hourly",
+func TestMigrateConfigRetainsInertLegacyDiscordKeys(t *testing.T) {
+	inert := map[string]interface{}{
+		"channel_paper_trades": true,
+		"channel_live_trades":  true,
+		"spot_summary_freq":    "hourly",
+	}
+	cases := []struct {
+		name string
+		obj  map[string]interface{}
+	}{
+		{
+			name: "at_floor_version",
+			obj: map[string]interface{}{
+				"config_version": MinSupportedConfigVersion,
+				"discord":        inert,
+			},
 		},
-	})
-	if err := MigrateConfig(path, nil, nil); err != nil {
-		t.Fatalf("MigrateConfig rejected a version-less config with inert pre-floor keys: %v", err)
-	}
-	after, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(after, &updated); err != nil {
-		t.Fatal(err)
-	}
-	if v := int(updated["config_version"].(float64)); v != CurrentConfigVersion {
-		t.Errorf("config_version = %d, want %d", v, CurrentConfigVersion)
-	}
-	discord := updated["discord"].(map[string]interface{})
-	for _, key := range []string{"channel_paper_trades", "channel_live_trades", "spot_summary_freq"} {
-		if _, ok := discord[key]; !ok {
-			t.Errorf("discord.%s should be retained (inert) for a version-less config", key)
-		}
-	}
-}
-
-func TestMigrateConfigV8PreservesFieldsAtCurrentVersion(t *testing.T) {
-	dir := t.TempDir()
-	path := filepath.Join(dir, "config.json")
-	original := map[string]interface{}{
-		"config_version": CurrentConfigVersion,
-		"discord": map[string]interface{}{
-			"spot_summary_freq": "hourly",
+		{
+			name: "versionless",
+			obj: map[string]interface{}{
+				"interval_seconds": 3600,
+				"strategies":       []interface{}{},
+				"discord": map[string]interface{}{
+					"owner_id":             "disc-owner",
+					"channel_paper_trades": true,
+					"channel_live_trades":  true,
+					"spot_summary_freq":    "hourly",
+				},
+			},
+		},
+		{
+			name: "versionless_without_strategies",
+			obj: map[string]interface{}{
+				"interval_seconds": 3600,
+				"discord":          inert,
+			},
+		},
+		{
+			name: "at_current_version",
+			obj: map[string]interface{}{
+				"config_version": CurrentConfigVersion,
+				"discord":        inert,
+			},
 		},
 	}
-	data, err := json.MarshalIndent(original, "", "  ")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(path, data, 0600); err != nil {
-		t.Fatal(err)
-	}
-	if err := MigrateConfig(path, nil, nil); err != nil {
-		t.Fatal(err)
-	}
-	result, err := os.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
-	var updated map[string]interface{}
-	if err := json.Unmarshal(result, &updated); err != nil {
-		t.Fatal(err)
-	}
-	discord := updated["discord"].(map[string]interface{})
-	if _, ok := discord["spot_summary_freq"]; !ok {
-		t.Error("discord.spot_summary_freq should NOT be removed when already at CurrentConfigVersion")
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "config.json")
+			writeRawConfig(t, path, tc.obj)
+			if err := MigrateConfig(path, nil, nil); err != nil {
+				t.Fatalf("MigrateConfig failed: %v", err)
+			}
+			updated := readRawConfig(t, path)
+			if v := int(updated["config_version"].(float64)); v != CurrentConfigVersion {
+				t.Errorf("config_version = %d, want %d", v, CurrentConfigVersion)
+			}
+			discord := updated["discord"].(map[string]interface{})
+			for _, key := range []string{"channel_paper_trades", "channel_live_trades", "spot_summary_freq"} {
+				if _, ok := discord[key]; !ok {
+					t.Errorf("discord.%s should NOT be removed (inert) for a supported config", key)
+				}
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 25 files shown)


richkuo:

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-30T10:28:13Z) in 3dc543a78c7a8721bfd6bcd3881cf9a78adfc4ed.

Verification: targeted tests, Go vet, and the scheduler build pass. The full local suite has the existing macOS Bash 3.2 failures in untouched updater tests. Pull request CI checks are green.

### Fixed
1. **The liquidation-audit "failed snapshot fetch" no-op row cannot detect removal of the snapshot gate** — populated the row with known liquidation, net-side, and on-chain data so the failed snapshot is the only no-op condition (`scheduler/hyperliquid_liquidation_guard_test.go:1271`).
2. **Aliased tier fixtures make the rejected-reload whole-object DeepEqual partially vacuous on the `sl_after` rows** — changed the tier fixtures to fresh builders for each reload closure, so the live and expected configurations have independent nested maps (`scheduler/config_reload_test.go:230`).

### Deferred to follow-up
1. **Delete the now-dead `v19AtomicRenameSites` list in a follow-up** — out of scope under the reviewer's `### Create Follow-up Issue` routing; filed as #1489.

### Test edits
1. **TestRunHyperliquidLiquidationAuditNoOpCases** (`scheduler/hyperliquid_liquidation_guard_test.go:1271`) — Wrong; ground: the review finding and issue #1485 rule 1b require the kept test to reach a real audit candidate and detect its claimed snapshot-gate behavior; now asserts that known exchange data still produces no action when the snapshot fetch fails.
2. **TestApplyHotReloadConfigOpenPositionGuards** (`scheduler/config_reload_test.go:230`) — Wrong; ground: the review finding and issue #1485 rule 1b require the rejected-reload comparison to use independent mutable data; now builds fresh nested tier maps for each configuration closure.

---
Created with LLM: GPT-5 | high | Harness: Codex

richkuo:
@claude sonnet review

PR #1502, opened 2026-09-01
fix(#1501): reject unconfirmed hyperliquid fills before booking [C97, GPT-5.6 Luna, max]

### LINKED ISSUE #1501: [C97] Reject unconfirmed Hyperliquid fills before recording
**Complexity: 97/100** — Capability 3 (Risk 4: money and trade-state integrity); Volume 22 (Scope 4 + Coupling 3 + Verification 4) — Opus 5, xhigh · fableplan: yes

## Problem

The operator reported that `hl-mr-hype-60` wrote 22 open and external-close trade pairs between 2026-08-30 22:05 UTC and 2026-08-30 22:57 UTC. The reported opens had empty `exchange_order_id` values, no corresponding Hyperliquid fills, and live trade alerts. Repository data cannot verify the incident count or wallet state.

The code path can produce that result. `shared_scripts/check_hyperliquid.py:run_execute` treats a Hyperliquid response without a usable `filled` status as a successful execution with an empty or zero-valued fill. `scheduler/main.go:runHyperliquidExecuteOrder` checks subprocess and top-level errors, then accepts that result and clears the live-execution failure throttle.

The automatic trade path then uses the cycle price and local sizing when the returned fill price or size is zero. It can create a position, write an open trade, and send a live trade alert without confirmed exchange execution. The account reconciler later sees that the modeled position is absent on-chain and records an external close, using average cost when no close-price source exists.

This failure is conditional. It occurs when the exchange or Software Development Kit (SDK) returns a non-filling response without an exception or top-level error. Existing exception and top-level-error paths already stop execution.

The current live consumers have different validation:

- Automatic open, close, and flip execution checks transport and top-level errors only.
- Automatic scale-in execution checks transport and top-level errors only, then substitutes the requested size and cycle price when fill data is absent.
- Manual open, add, and close commands require a fill object but do not require both positive average price and positive filled size. Manual open and add can substitute requested size.
- The daemon manual-close path accepts any non-nil fill and can queue a zero-price action.
- Hedge open and add already require positive average price and positive filled size through `hedgeExecuteConfirmed`.
- A positive fill can omit its order identifier. Existing Python tests define this as valid behavior.

## Goal

Accept a live Hyperliquid execution as confirmed only when it contains a positive average fill price and positive filled size. Keep model state, trade history, pending actions, and executed-trade alerts unchanged after an unconfirmed result. Surface the failure through the correct operator channel and preserve protection-cancellation state for reconciliation.

## Approach

1. In `shared_scripts/check_hyperliquid.py`, classify the market-order response explicitly. Return a precise top-level error for a rejected status, an empty status list, a missing `filled` object, malformed fill values, a non-positive average price, or a non-positive filled size. Keep a positive fill valid when its order identifier or fee is absent.
2. Add one shared Go validation contract for `RunHyperliquidExecute`. After transport parsing, require a non-nil result, no top-level error, a non-nil execution and fill, positive `AvgPx`, and positive `TotalSz`. Return the parsed result with the validation error so cancellation metadata remains available.
3. Apply that contract before mutation in every live consumer: automatic open, close, and flip; automatic scale-in; manual open, add, and close; daemon manual close; and hedge open or add. Reuse or replace `hedgeExecuteConfirmed` so the fill rule has one owner.
4. Route automatic failures through `notifyLiveExecFailure` with the strategy, direction, symbol, and a stable “exchange returned no confirmed fill” error. Clear the failure throttle only after fill validation succeeds. Manual commands must return an execution error and must not queue a pending action.
5. When the execute attempt cancelled stop-loss or take-profit orders before fill validation failed, retain the cancellation result and update or reconcile the recorded protection state. The model must not report cancelled protection as resting.
6. Keep paper Hyperliquid execution, replay mirror booking, OKX execution, and generic perps portfolio simulation unchanged. These paths can use modeled fills and do not share the live Hyperliquid confirmation contract.

## Acceptance criteria

- A live Hyperliquid result is confirmed only when `Execution.Fill.AvgPx > 0` and `Execution.Fill.TotalSz > 0`.
- A positive fill without `OID` succeeds and records an empty exchange order identifier.
- A missing result, missing execution, missing fill, rejected status, empty status list, malformed fill, zero price, or zero size returns an execution failure.
- An unconfirmed automatic execution does not change the strategy position, cash, trade history, risk state, replay log, hedge exposure, or protection snapshot.
- An unconfirmed automatic execution does not send a trade-executed alert. It uses the existing throttled live-execution failure notification.
- An unconfirmed manual execution does not insert a pending manual action and returns a clear command error.
- Automatic scale-in, bidirectional flip, manual open, manual add, manual close, daemon manual close, and hedge open or add follow the same confirmation rule.
- A failure after protection cancellation preserves the cancellation metadata and leaves recorded protection state consistent with the on-chain result.
- Paper Hyperliquid, replay mirror, OKX, and generic perps simulation behavior remain unchanged.
- Python tests cover a rejected status, empty statuses, a missing `filled` object, malformed values, zero price, zero size, and a valid fill without an order identifier.
- Go tests cover the shared validator and each live consumer family. They assert zero state mutation, zero trade rows, zero pending actions, correct failure notification, and protection-cancellation handling.
- The focused Python and Go tests pass, followed by the full scheduler and shared-script test suites.

## Plain simple English

The scheduler can record a live trade when Hyperliquid did not confirm a fill. This creates false positions, false trade history, and false alerts. The scheduler must require a real fill price and size before it changes trading state.

---
LLM: GPT-5 | high | Harness: validate-issue

richkuo:
## Plain simple English

The scheduler now accepts a live Hyperliquid order only when the exchange reports a positive filled size and average price. Failed responses produce an operator error and leave positions, trade records, pending actions, and trade alerts unchanged. Confirmed protection cancellations remain recorded for reconciliation.

## Summary

- Added one Go fill confirmation contract for all live Hyperliquid execute consumers.
- Rejected invalid Python order responses with exit code 1.
- Removed scale-in fallback values and cleared only confirmed-cancelled protection records.
- Added consumer, parser, protection, and Python regression tests.

## Verification

- `uv run --no-sync python -m py_compile shared_scripts/check_hyperliquid.py` passed.
- `uv run --no-sync python -m pytest shared_scripts/test_check_hyperliquid.py -q` passed: 95 tests.
- `go -C scheduler test . -count=1 -timeout 180s` passed.
- `uv run --no-sync python -m pytest shared_scripts/test_close_hyperliquid_position.py shared_scripts/test_check_execute_fees.py -q` passed: 35 tests.

Test edit: `test_fill_empty_statuses` is Outdated. Ground: issue #1501 acceptance criteria require a top-level error and exit code 1. It now asserts both. The helper catches `SystemExit`, and positive-fill tests assert exit code 0.

Closes #1501

---
LLM: GPT-5.6 Luna | max | Harness: Codex CLI

### DIFF
--- a/scheduler/executor.go
+++ b/scheduler/executor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 	"sort"
@@ -70,14 +71,16 @@ type HyperliquidExecution struct {
 }
 
 type HyperliquidExecuteResult struct {
-	Execution                 *HyperliquidExecution `json:"execution"`
-	Platform                  string                `json:"platform"`
-	Timestamp                 string                `json:"timestamp"`
-	Error                     string                `json:"error,omitempty"`
-	CancelStopLossError       string                `json:"cancel_stop_loss_error,omitempty"`
-	CancelStopLossSucceeded   bool                  `json:"cancel_stop_loss_succeeded,omitempty"`
-	StopLossError             string                `json:"stop_loss_error,omitempty"`
-	StopLossFilledImmediately bool                  `json:"stop_loss_filled_immediately,omitempty"`
+	Execution                   *HyperliquidExecution `json:"execution"`
+	Platform                    string                `json:"platform"`
+	Timestamp                   string                `json:"timestamp"`
+	Error                       string                `json:"error,omitempty"`
+	CancelStopLossError         string                `json:"cancel_stop_loss_error,omitempty"`
+	CancelStopLossSucceeded     bool                  `json:"cancel_stop_loss_succeeded,omitempty"`
+	CancelStopLossSucceededOIDs []int64               `json:"cancel_stop_loss_succeeded_oids,omitempty"`
+	CancelStopLossFailedOIDs    []int64               `json:"cancel_stop_loss_failed_oids,omitempty"`
+	StopLossError               string                `json:"stop_loss_error,omitempty"`
+	StopLossFilledImmediately   bool                  `json:"stop_loss_filled_immediately,omitempty"`
 }
 
 type HyperliquidStopLossUpdateResult struct {
@@ -328,6 +331,8 @@ func RunHyperliquidExecute(script, symbol, side string, size, stopLossPct float6
 	return parseHyperliquidExecuteOutput(stdout, string(stderr), err)
 }
 
+var runHyperliquidExecuteFn = RunHyperliquidExecute
+
 func RunHyperliquidUpdateStopLoss(script, symbol, side string, size, triggerPx float64, cancelStopLossOID int64) (*HyperliquidStopLossUpdateResult, string, error) {
 	args := []string{
 		"--update-stop-loss",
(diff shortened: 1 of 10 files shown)


richkuo:
@claude review

Issue #1500, opened 2026-08-31
[C37] The volume-feature GMM regime candidate changes label too often to pass the stability gate


richkuo:
**Complexity: 37/100** — Capability 1 (Uncertainty 2 — the smoothing mechanism is known and the reachable setting needs discovery); Volume 12 — Opus 5, high · fableplan: no

## Problem

BTC `volume:gmm:k=5` is the only candidate in the #1218 re-run that clears every promotion arm except one. Its stored record shows:

- Held-out Kruskal H 107.83 against the incumbent hand-rule's 85.69.
- Permutation p at 0.000556, the minimum achievable value at the run's resolved `n_perm=1799`, so it passes the family-wise Bonferroni alpha of 0.001111.
- Five latent states mapping to five distinct names, and `non_degenerate_all=true` across all five evaluation windows.
- `stability_ok=false`. Its label transition rate is 0.2539 against the incumbent's 0.1374, so `stability_gain` is −0.1165 and `ship=false`.

The candidate changes label about twice as often as the hand-rule. Nothing else blocks it.

The churn has a plausible mechanical source. `empirical_transition` builds the transition matrix from the fitter's hard assignments, and the GMM assignment is a per-bar argmax with no temporal term, so the resulting matrix carries less diagonal mass than a Hidden Markov Model fit. `forward_filter_labels` (`backtest/regime_hmm.py:58`) then runs a forward pass bounded by `filter_window` and takes a per-bar argmax over the filtered posterior, with no dwell constraint. A weakly sticky transition matrix and a short filter window together produce frequent label switches.

Two floors bound any fix. `min_transition_rate` for this run is 0.0687, so over-smoothing fails non-degeneracy from the other side. `min_active_labels` is 5, and this candidate sits exactly at 5, so any smoothing that merges one state fails as well. The reachable band is narrow, and this issue may measure that no setting reaches it.

## Goal

Find whether decoder smoothing brings this candidate's transition rate to or below the incumbent's while it keeps five active labels and stays above the churn floor. Record the answer either way.

## Approach

1. Sweep `filter_window` for this candidate across the five evaluation windows, and record transition rate, active labels, maximum occupancy, and held-out Kruskal H at each setting. `filter_window` is an existing model field, so this arm needs no decoder change.
[...]

PR #1503, opened 2026-09-02
fix(#1499): give each latent regime state a distinct label [C64, Fable 5.1, high]

### LINKED ISSUE #1499: [C64] The regime model gate rejects candidates because the label mapping merges separate states
**Complexity: 64/100** — Capability 2 (Uncertainty 3 — the correct naming rule is open design judgment; Coupling 3 sets the same floor); Volume 14 — Opus 5, xhigh · fableplan: no

## Problem

The #1218 re-run recorded non-degeneracy as the dominant failure arm and read it as unsupervised fits collapsing to about four effective states. The stored artifacts show a different cause. Every candidate that passed both `verdict.ship` and the family-wise Bonferroni correction occupies its full latent state count. Each one loses states at the naming step:

| Run | Candidate | Latent states | Distinct names | Floor |
|---|---|---|---|---|
| BTC/USDT 1h | `htf:hmm:k=6` | 6 | 4 | 5 |
| ETH/USDT 1h | `canonical:kmeans:k=7` | 7 | 4 | 6 |
| ETH/USDT 1h | `funding:kmeans:k=6` | 6 | 3 | 6 |
| ETH/USDT 1h | `all_enriched:kmeans:k=6` | 6 | 4 | 6 |

`map_latent_to_names` (`backtest/regime_vol_model.py:209`) names each latent state by running the hand-rule threshold function `map_composite_label` on that state's raw centroid. Two centroids that fall in the same hand-rule cell receive the same name. `forward_filter_labels` (`backtest/regime_hmm.py:58`) then emits `states[j]`, so both latent states produce one label string. `non_degeneracy` (`backtest/regime_vol_model.py:283`) counts distinct names, so a model with six separated volatility states reports four active labels and fails `min_active_labels`.

The condition is widespread across the swept grid: 48 of 90 BTC candidates and 60 of 90 ETH candidates carry at least one duplicate name in `states`.

BTC `htf:hmm:k=6` is the clearest case. It records held-out Kruskal H 137.87 against the incumbent's 80.47, a permutation p at the minimum achievable value for the run, a stability gain of +0.065, and `ship=true` with Bonferroni passed. Its six latent states map onto four names, because `trending_up_choppy` and `trending_down_choppy` each name two states.

Two consequences follow. The gate scores the naming step while it reports on the model. The mapping also discards information that live use needs, because two states with different forward volatility share one label and therefore one per-regime ATR multiplier downstream.

## Goal

Give each separated decoded state its own label, so the promotion gate measures the model. Then re-run the bake-offs under the unchanged gate and record what the corrected mapping produces.

## Approach

1. Pin the mechanism with a test on the stored artifacts: assert that a fitted model whose centroids collide under `map_composite_label` produces a label stream with fewer distinct labels than latent states.
2. Design the naming rule. `map_latent_to_names` already computes `volatility_rank` per state, so rank-ordered assignment is one option. Disambiguation of colliding names by rank is another. The rule must keep the seven named labels as the output vocabulary, because `allowed_regimes` and the per-regime stop dictionaries key on those names.
3. Keep the label semantics defensible. A name must still describe the state it marks, so a rank-based rule needs evidence that the assigned name matches the state's measured behavior.
4. Hold the gate fixed. No change to `derive_thresholds`, `non_degeneracy`, the candidate-self-v2 `gate_verdict`, alpha, or the permutation resolution belongs in this issue. Any gate change belongs to the #1211 lineage.
5. Re-run the #1080 and #1095 harnesses on BTC and ETH with the corrected mapping and write the result to `docs/research/`, whatever the verdict.

## Acceptance criteria

- A regression test fails on the current mapping and passes on the corrected one, using a fitted model with colliding centroids.
- No fitted model emits a `states` list with duplicate names, or the harness records the duplicate explicitly and the gate reads distinct decoded states.
- The gate code and thresholds are byte-identical to their current form.
- The hand-rule incumbent path is unchanged, and its held-out separation numbers reproduce the values in the #1218 artifacts.
- A research document in `docs/research/` records the re-run for BTC and ETH, names every arm that each near-miss candidate passes or fails, and states plainly whether any candidate now clears the full bar.
- The document states whether the corrected mapping changes what each label means, because #1074 live wiring depends on that answer.

## Plain simple English

Our test of new market-condition models says the models are too simple. The test is wrong. The models do find six or seven different market states. The step that gives each state a name merges some of them, because it reuses the old rule's names. The test then counts the names and rejects a good model. This also loses real information, because two different states would share one setting.

---
LLM: Opus 5 | high | Harness: Claude Code

richkuo:
Closes #1499

## Summary

- `backtest/regime_vol_model.py`: new naming rule `distinct_nearest_cell` (`MODEL_VERSION` 2). A state whose hand-rule name is uncontested keeps it. Colliding states are assigned distinct names by a minimum-cost assignment over the nine hand-rule cells (rectangular Hungarian). Cost is the z-scaled distance from the centroid to the nearest point of each cell plus a fixed penalty per relabel, so the assignment minimises the relabel count first and the total displacement second, with a deterministic tiebreak. k above the vocabulary size keeps hand-rule names and records the duplicates in `naming.duplicate_names`.
- Every `mapping[i]` now records `handrule_name`, `relabeled` and `displacement_z`; the fitted model carries a `naming` summary and the 1080, 1083 and 1095 harness records copy it into their artifacts.
- `non_degeneracy`, `derive_thresholds`, `gate_verdict`, the alpha resolution and the permutation-count resolution are unchanged (`git diff` touches none of them).
- `docs/research/1499-distinct-latent-state-labels.md` plus three artifacts in `docs/research/1499-artifacts/` record the BTC and ETH re-run of the 1080 and 1095 harnesses on a cache frozen at the last bar the 1218 run saw.

## Results of the re-run

- Duplicate-name candidates: 9 of 18 (1080), 48 of 90 (1095 BTC), 60 of 90 (1095 ETH) in 1218. Zero in every run now.
- Five candidates clear the full bar (raw ship, Bonferroni, non-degenerate on all five windows): BTC `htf:hmm:k=6` and `htf:kmeans:k=7`; ETH `canonical:kmeans:k=7`, `funding:kmeans:k=7` and `all_enriched:hmm:k=7`. The BTC pair and ETH `all_enriched:hmm:k=7` pass Bonferroni at exactly the corrected alpha (knife edge, recorded).
- Every 1218 near-miss has a per-arm pass/fail entry in the document. BTC `volume:gmm:k=5` (no collision) is unchanged and still fails stability only (that is #1500).
- Label meaning: uncontested states keep their meaning. A relabeled state's name is the nearest free hand-rule cell in z units and its centroid does not satisfy that cell's rule. `mapping[i].handrule_name` and `relabeled` carry the truth for #1074 live wiring; the document states the consequence.

## Verification

- Incumbent hand-rule held-out p-values, permutation steps to alpha and transition rates reproduce the 1218 artifacts exactly for all three runs. The incumbent Kruskal-Wallis H differs in the fourth digit; the 1218 run saw the final cached candle in progress, and substituting that in-progress close (BTC 62719.11, ETH 1777.81) reproduces the 1218 H to the last digit. Unchanged-name `canonical`, `funding` and `htf` candidates reproduce their p-value and stability gain exactly; the `volume` and `all_enriched` tails move by the amounts listed in the document.
- Regression proof: with the old `regime_vol_model.py` swapped in, the new tests in `backtest/tests/test_regime_vol_model.py` give 7 failed, 1 passed; with the fix, the file passes 70 tests.
- Full suite: `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` gives 2858 passed. `py_compile` passes on every changed Python file.

## Test edits

None of the pre-existing tests were edited. All test changes are additions.

## Adopted plan

None. The issue carried no implementation plan comment.

## Plain simple English

The regime model gave two different market states the same name when their centres fell in the same hand-rule box. The gate then counted them as one state and rejected good models. Now each state gets its own name, chosen from the nearest free box. Five models pass the full gate for the first time. The gate rules did not change.

---

https://claude.ai/code/session_013E8gonXiGvgzZKEHYZKuHi

### DIFF
--- a/backtest/regime_vol_model.py
+++ b/backtest/regime_vol_model.py
@@ -7,13 +7,16 @@ for _p in (_THIS_DIR, os.path.abspath(os.path.join(_THIS_DIR, "..")),
         sys.path.insert(0, _p)
 
 from dataclasses import dataclass
+import math
 
 import numpy as np
 from regime_hmm import stationary_distribution
 
 FEATURES = ["return_eff", "range_eff", "efficiency", "adx"]
 MODEL_TYPE = "unsupervised_vol_regime"
-MODEL_VERSION = 1
+MODEL_VERSION = 2
+NAMING_RULE = "distinct_nearest_cell"
+_RELABEL_PENALTY = 1e6
 
 
 def standardize(features):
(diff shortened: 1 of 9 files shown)


richkuo:
@claude review

Issue #1218, opened 2026-07-03
[C20] Re-run the 1080/1095 vol-regime bake-offs under candidate-self-v2 gate semantics


richkuo:
**Complexity: 20/100** — scope: low (run two existing merged harnesses, record an evidence doc); risk: low (read-only offline research, no live or Go path touched); uncertainty: medium only in outcome — the tooling, gate, and correction machinery are already merged.

## Problem
The #1080 (canonical four-feature) and #1095 (enriched-feature, PR #1168) unsupervised vol-regime bake-offs were last run under the pre-#1211 gate, whose `incumbent_trustworthy` hard-gate abstained **every** candidate verdict regardless of candidate quality. #1211 (PR #1215) revised `gate_verdict` to candidate-self-v2 (`backtest/regime_calibrate.py`): `ship = separation_ok and stability_ok`, judged on the candidate's **own** block-shuffle permutation significance + non-inferiority to the incumbent's KW-H + stability gain. Both harness docstrings already record that re-running under v2 "can now surface a gate-passing winner instead of a blanket abstain" (`backtest/research/regime_1095_enriched_vol_model.py`, `regime_1080_unsupervised_vol_model.py`). No re-run has happened, so blocker 2 of #1074 ("a candidate model clears the forward-volatility separation/stability gate") is currently **unmeasured, not failed**.

## Goal
Determine whether any existing candidate (k-means / GMM / HMM × K grid × feature subsets) now passes the promotion gate under candidate-self-v2, producing the evidence #1074 blocker 2 needs — in either direction.

## Approach / acceptance criteria
- Re-run both harnesses **unchanged** (no gate or threshold edits — semantics changes belong to #1211's lineage, not here):
  - `uv run --no-sync python backtest/research/regime_1080_unsupervised_vol_model.py`
  - `uv run --no-sync python backtest/research/regime_1095_enriched_vol_model.py` (BTC, plus ETH per its docstring; needs the OHLCV cache; funding-bearing subsets that can't be built are reported "unavailable", never fit on nothing)
- Run the **full candidate family in one invocation each** — the Bonferroni correction divides alpha across the combined structurally-eligible grid, so partial/cherry-picked sweeps under-correct and are invalid as promotion evidence.
- Record a reproducible evidence doc under `docs/research/` with command lines, JSON artifacts, per-candidate `verdict.ship` / `any_ships`, and the resolved `n_perm`.
[...]

PR #1508, opened 2026-09-02
feat(#1500): sweep decoder smoothing for the volume gmm regime candidate [C37, Fable 5.1, low]

### LINKED ISSUE #1500: [C37] The volume-feature GMM regime candidate changes label too often to pass the stability gate

richkuo:
Closes #1500

## Summary

- Adds two off-by-default decoder options to `forward_filter_labels` in `backtest/regime_hmm.py`: `decode_min_dwell` (label hysteresis) and `decode_stickiness` (diagonal boost for decoding only). Absent, `0`, `None`, or `1` gives a byte-identical decode. Both live on the model dict, so they survive the #1074 move to `shared_tools/`.
- Adds the one-shot sweep script `backtest/research/regime_1500_gmm_stability_sweep.py` (registered in `docs/backtesting-registry.md`), its BTC artifact, and the research record `docs/research/1500-gmm-regime-stability-smoothing.md`.
- Result: `filter_window` is inert for this candidate (identical decode from 8 to 512). Dwell 3 is the only in-band dwell. At dwell 3 with stickiness 2 the model passes the raw `gate_verdict` (H 87.58, stability gain +0.0517) but its p of 0.01667 misses the Bonferroni alpha of 0.001111. **The candidate does not ship. Bonferroni significance is the binding arm.**
- Context: `htf:hmm:k=7`, `all_enriched:kmeans:k=5` and `all_enriched:kmeans:k=7` clear the full bar under stickiness 20. They are recorded, and no promotion is proposed. A follow-up bake-off re-run with stickiness is unfiled.
- Gate code and thresholds in `regime_calibrate.py` and `regime_diagnostics.py` are byte-identical to `origin/main`.

## Verification

- `uv run --no-sync python -m pytest backtest/tests/test_regime_hmm.py backtest/tests/test_regime_vol_model.py backtest/tests/test_regime_calibrate.py`: 100 passed.
- `py_compile` on the two touched Python files passes.
- Full sweep run on the frozen #1218 OHLCV cache with `n_perm=1799`; artifact committed.
- Test edits: none. Five tests added in `backtest/tests/test_regime_hmm.py`, including a byte-identical decode check against a frozen copy of the old decoder, look-ahead safety under dwell, and negative-value rejection.
- Plan: none.

## Plain simple English

We tried to make the volume-based regime model change its label less often. A three-bar hold does that and keeps all five labels. But the smoothed model no longer passes the strict significance test, so it does not ship. The gate is unchanged. Three other models look better with smoothing and are recorded for later.

https://claude.ai/code/session_01EhWRwr4TWSf7J9oByhhL2k

---

### DIFF
--- a/backtest/regime_hmm.py
+++ b/backtest/regime_hmm.py
@@ -4,6 +4,8 @@ import numpy as np
 MODEL_TYPE = "label_anchored_hmm"
 MODEL_VERSION = 1
 FEATURES = ["return_eff", "range_eff", "efficiency", "adx"]
+DECODE_MIN_DWELL_KEY = "decode_min_dwell"
+DECODE_STICKINESS_KEY = "decode_stickiness"
 
 
 def stationary_distribution(transition: np.ndarray) -> np.ndarray:
@@ -65,12 +67,23 @@ def forward_filter_labels(features: np.ndarray, model: dict):
     em_mean = np.array([e["mean"] for e in model["emissions"]], dtype=float)
     em_var = np.array([e["var"] for e in model["emissions"]], dtype=float)
     log_init = np.log(np.asarray(model["init"], dtype=float) + 1e-300)
-    log_A = np.log(np.asarray(model["transition"], dtype=float) + 1e-300)
+    A = np.asarray(model["transition"], dtype=float)
+    stickiness = float(model.get(DECODE_STICKINESS_KEY) or 0.0)
+    if stickiness < 0:
+        raise ValueError(f"{DECODE_STICKINESS_KEY} must be >= 0")
+    if stickiness > 0:
+        A = A + stickiness * np.eye(k)
+        A = A / A.sum(1, keepdims=True)
+    log_A = np.log(A + 1e-300)
     w = int(model["filter_window"])
+    min_dwell = int(model.get(DECODE_MIN_DWELL_KEY) or 0)
+    if min_dwell < 0:
+        raise ValueError(f"{DECODE_MIN_DWELL_KEY} must be >= 0")
     default_label = states[int(np.argmax(model["init"]))]
 
     labels = np.array([default_label] * n, dtype=object)
     conf = np.zeros(n, dtype=float)
+    held, pending, pending_run = None, None, 0
     for i in range(n):
         lo = max(0, i - w + 1)
         alpha = log_init.copy()
@@ -88,6 +101,17 @@ def forward_filter_labels(features: np.ndarray, model: dict):
             seen = True
         if seen:
             j = int(np.argmax(alpha))
+            if min_dwell > 1:
+                if held is None or j == held:
+                    pending, pending_run = None, 0
+                else:
+                    pending_run = pending_run + 1 if pending == j else 1
+                    pending = j
+                    if pending_run >= min_dwell:
+                        pending, pending_run = None, 0
+                    else:
+                        j = held
+                held = j
             labels[i] = states[j]
             conf[i] = float(np.exp(alpha[j] - _logsumexp(alpha)))
     return labels, conf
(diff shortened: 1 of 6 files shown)


richkuo:
@claude review

Issue #1509, opened 2026-09-03
[C97] Live and paper strategies in one scheduler share every portfolio-wide risk gate and latch


richkuo:
**Complexity: 97/100** — Capability 3 (Risk 4 — portfolio kill switch, daily loss, and exposure gates are auto-protective money paths; Coupling 4 — one new scope dimension threads risk, state, DB, Discord, and UI); Volume 22 — Opus 5, xhigh · fableplan: yes

## Problem

Every portfolio-wide risk mechanism aggregates over all strategies in the process with no mode filter. Live and paper Hyperliquid strategies can already share one config (mode is only checked for credential presence at `scheduler/config.go:1480-1527`) and already share one batch subprocess (`scheduler/hl_batch.go:29-35` has no mode in the key). Running both modes in one process today makes paper affect live and live affect paper:

- Drawdown latch. `AggregatePerpsMarginInputs` (`scheduler/risk.go:507-513`) and `PortfolioNotional` (`scheduler/risk.go:697-699`) iterate all strategy states. `computeSubsetPortfolioValue` (`scheduler/shared_wallet.go:333-349`) and `computeInitialPortfolioPeak` (`scheduler/shared_wallet.go:402-420`) add paper capital to the equity peak. One `PortfolioRiskState` (`scheduler/risk.go:332`) and one SQLite row `portfolio_risk … CHECK (id = 1)` (`scheduler/db.go:202-218`) hold the latch.
- Kill switch blast radius. `main.go:1306` `if !killSwitchFired {` wraps the whole strategy loop, so a live latch freezes paper books and a paper drawdown can latch and stop live trading.
- Daily loss. `evaluateDailyLossLimit` (`scheduler/daily_loss.go:22-39`) sums `DailyPnL` and `InitialCapital` across all states; only pooled IDs are excluded.
- Exposure cap and correlation. `computeAssetDeltas` (`scheduler/correlation.go:32-47`) nets a live long against a paper long in one `AssetExposure`. `evaluateExposureCap` (`scheduler/exposure_cap.go:33-44`) and `evaluateNotionalCapHold` (`scheduler/notional_cap.go:15`) consume that total.
[...]

PR #1512, opened 2026-09-03
feat(#1510): let a paper strategy name its live replay source [C64, Opus 5, xhigh]

### LINKED ISSUE #1510: [C64] Replay mirror cannot pair live and paper twins that run in one process
**Complexity: 64/100** — Capability 2 (Coupling 3 — replay source and mirror share one decision log and strategy identity across process boundaries; Risk 2 — paper-only writes with a data-integrity contract on the log); Volume 14 — Opus 5, xhigh · fableplan: no

## Problem

The live-to-paper replay mirror assumes the live source and the paper mirror run in different processes under the same strategy id. The source writes `dec.StrategyID = s.ID` (`scheduler/replay_log.go:239`) and the mirror reads `decisionLog.PendingDecisions(sc.ID)` under its own id (`main.go:2170`). Docs state the pairing rule as "paper with the same strategy id" (`SKILL.md:663`, `docs/POST_UPDATE_HISTORY.md:238`), and the shared DB lives in `StateDirectory=go-trader/shared` so two systemd instances can reach it.

One config rejects duplicate ids (`scheduler/config.go:1391`). When live and paper twins move into one process they need distinct ids, and the mirror finds zero rows under its own id, so mirroring silently stops. There is no `replay_source_id` field anywhere. The drift script `scripts/check-live-paper-config-drift.sh` also pairs twins by id across deployments (`classify()` line 75, pair output near line 190).

## Goal

A paper strategy can name the live strategy it mirrors, so the replay mirror works when both run in one process with distinct ids, and keeps working across processes as today.

## Approach

Correctness and safety outrank diff size.

1. Add per-strategy `replay_source_id` (string). Validation at `scheduler/config.go:1451-1459`: allowed only with `replay_sharing="live_mirror"` on a paper HL perps strategy; the referenced id, when present in the same config, must be a live HL perps strategy on the same symbol and timeframe. Absent, the mirror keys on its own id as today.
2. `replayMirrorPaperActive` and the `PendingDecisions` read at `main.go:2169-2197` resolve the source id through a helper `replayMirrorSourceID(sc)`; the watermark in `StrategyState.ReplayMirrorWatermark` stays keyed on the paper strategy and records the source id, so a source change resets the watermark and logs a WARN.
3. In-process ordering: when the source and mirror share a process, the live strategy's decision is written in the same cycle before the paper strategy's apply step. Sort due strategies so live replay sources evaluate before paper mirrors of them, or apply on the next cycle with a documented one-cycle lag. Pick the same-cycle order and test it.
4. Hot reload: `replay_source_id` follows the `replay_sharing` rule (flat-only).
5. Update `scripts/check-live-paper-config-drift.sh` to pair by `replay_source_id` when set, else by id, and to accept twins inside one deployment.
6. Tests: mirror with distinct ids in one config, source id validation errors, watermark reset on source change, same-cycle ordering, drift script pairing.

## Acceptance criteria

- A config with `hl-x-live` (live, `live_mirror`) and `hl-x-paper` (paper, `live_mirror`, `replay_source_id: hl-x-live`) mirrors live decisions into the paper book in the same cycle.
- Cross-process mirroring with the same id and no `replay_source_id` behaves exactly as before.
- A `replay_source_id` that names a missing, paper, or different-symbol strategy fails config load with a clear error.
- The drift script pairs twins by `replay_source_id` and reports zero false drift for an in-process pair.

## Plain simple English

Paper strategies can copy the decisions of their live twin. Today this only works when the twin has the same name and runs in a separate program. If both run in one program they must have different names, and the copy stops without a message. This change lets a paper strategy name its live twin directly.

Part of umbrella #1511.

---
LLM: Fable 5.1 | low | Harness: new-issue

richkuo:
## Plain simple English

A paper strategy can copy the trades of its live twin. Before this change the two had to have the same name and run as two separate programs. If they run in one program their names must differ, and the copy stopped with no message. A paper strategy can now name its live twin directly, and the copy happens in the same cycle as the live trade.

## Summary

Closes #1510.

- **New per-strategy `replay_source_id`** (paper HL perps, opt-in). `replayMirrorSourceID(sc)` resolves the source id (`replay_source_id`, else the strategy's own id) and now backs `replayMirrorPaperActive`; the decision-log read in the cycle keys on it instead of `sc.ID`.
- **Load validation** (`scheduler/config.go`): allowed only with `replay_sharing="live_mirror"` on a paper HL perps strategy. The named strategy must exist in the same config, run live, be HL perps, set `replay_sharing="live_mirror"`, and trade the same symbol and timeframe. Self-reference is refused with a message that says to omit the field. One live source takes one mirror, so two mirrors cannot consume each other's rows.
- **Same-cycle ordering**: `orderReplaySourcesBeforeMirrors` reorders the due list so a live source evaluates before the paper mirror that names it. The move is minimal — only a mirror that sits ahead of its source moves, to the slot right after it. With no in-process pair the slice is returned unchanged, so existing deployments see the same order as today.
- **Watermark source tracking**: the watermark stays keyed on the paper strategy and now records its source in `StrategyState.ReplayMirrorWatermarkSource`, persisted in the new `strategies.replay_mirror_watermark_source` column (idempotent `ALTER TABLE` migration). A source change resets the watermark and the in-memory progress and logs a WARN. A blank stored source reads as the strategy's own id, so a pre-upgrade mirror with no `replay_source_id` never resets.
- **Hot reload**: `replay_source_id` follows the `replay_sharing` rule — hot-reloadable while flat, refused with open positions, and masked out of the restart-shape comparison.
- **Drift audit**: `scripts/check-live-paper-config-drift.sh` pairs twins by `replay_source_id` when set, else by id, so twins inside one deployment pair. `id` and `replay_source_id` are excluded from the field comparison, so an in-process pair with differing ids reports no drift; the pair lines annotate the real ids.
- **Docs**: `CLAUDE.md`/`AGENTS.md` subsystem line, `SKILL.md` (new Adjustable Settings row, the `replay_sharing` row's "same strategy id" wording, the blocked-while-open list, the hot-reload paragraph), and a `#1510` entry in `docs/POST_UPDATE_HISTORY.md` with a pointer added to the `#1431` entry.

### Behavior kept as it is

Cross-process mirroring under the same id with no `replay_source_id` is untouched: the source resolves to the strategy's own id, the watermark does not reset, and the same rows book the same way. `TestReplayMirrorCrossProcessSameIDUnchanged` proves it end to end against a real decision-log database.

### Deviation from the issue

The issue's Approach says the referenced id is checked only "when present in the same config", while the acceptance criteria require a **missing** id to fail the load. This build fails the load, because a typo would otherwise disable mirroring in silence — the exact failure the issue exists to remove — and the Goal keeps cross-process pairing working "as today", which means the same id and no `replay_source_id`.

## Verification

- `gofmt -l scheduler` — clean.
- `go -C scheduler build .` — ok.
- `go -C scheduler test ./... -count=1` — ok (the module lives in `scheduler/`; `go test ./...` from the repo root has no module to select).
- Drift-audit section of `scripts/test_update_helpers.sh` — passes, including the two new in-process fixtures. The earlier systemd-fleet section of that file fails the same way on `origin/main` on macOS (bash 3.2, `out[@]: unbound variable`), so the drift block was run on its own.
- No Python source changed, so the Python suite was not run.

### Tests

New `scheduler/replay_source_id_test.go`: source-resolution table, cross-process parity, applying a named source's rows into the paper book (and that the paper id itself has no rows), the validation table (missing, paper, wrong symbol, wrong timeframe, source not sharing, mirror without `replay_sharing`, source id on a live strategy, spot, self-reference, two mirrors of one source), the cycle-ordering table, watermark reset with its WARN, the watermark-source state round trip, flat-only hot reload, the unknown-key guard, and `LoadConfig` on a full in-process pair. New drift-audit cases in `scripts/test_update_helpers.sh` for an in-sync in-process pair and one that drifts on cadence.

**No existing test was edited or removed.**

---
LLM: Opus 5 | xhigh | Harness: work-on-issue

https://claude.ai/code/session_011texkRC3noVQpS9MGCSNuD


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-09-03T04:55:09Z) in b35eadde.

The review was a bare LGTM with no findings. This push only merges `origin/main` (the #1509 portfolio-scope partition) into the branch to clear the merge conflict.

### Resolved merge conflicts
1. `scheduler/main.go` — took the `origin/main` body and re-applied this PR's two hunks by hand: the `orderReplaySourcesBeforeMirrors` due-set ordering (`scheduler/main.go:675`) and the `replayMirrorSourceID` / `syncReplayMirrorWatermarkSource` replay block (`scheduler/main.go:2274`). No other lines differ from `origin/main`.
2. `SKILL.md` — kept both sides of the hot-reload sentence: the `portfolio_risk.paper` override wording from main and the `replay_source_id` flat-only clause from this PR.

Verification after the merge: `go build`, `go vet`, `go test ./...` (pass), `pytest -n auto` (2886 passed), `pytest shared_scripts/` (268 passed).

---

richkuo:
@claude sonnet review

Issue #1511, opened 2026-09-03
[C12] Live and paper strategies run in separate processes and see different market snapshots


richkuo:
**Complexity: 12/100** — Capability 0 (Risk 1 — deployment docs, scripts, and example config only); Volume 12 — Opus 5, high · fableplan: no

## Problem

Live and paper Hyperliquid strategies run as separate systemd instances (`systemd/go-trader@.service`: one config and one state directory per `%i`; `SKILL.md:275` port map; `README.md:58-63` install recipe). Each process pulls candles, marks, and regime bundles on its own clock, so paper and live evaluate different market snapshots and their decisions diverge (the 2026-08-25 `hl-vwap-eth-60` discussion). Inside one process the batch runner already shares one candle fetch across modes (`scheduler/hl_batch.go:29-35` has no mode in the key, `hl_batch.go:195` carries mode per slot) and marks are fetched once per cycle (`main.go:729`).

Two blockers stop a merge into one process. Every portfolio-wide risk gate aggregates across modes (child issue 1), and the replay mirror pairs twins by identical id across processes (child issue 2). Once both land, the fleet still needs a merge path: the drift script `scripts/check-live-paper-config-drift.sh` pairs twins across deployments, `config.example.json:79,88` ships paper-only examples, and docs present separation as the convention (`SKILL.md:141, 188, 663`, `docs/POST_UPDATE_HISTORY.md:227-238`).

## Goal

Live and paper Hyperliquid strategies run in one scheduler process, evaluate the same market snapshot each cycle, and stay fully isolated at the portfolio-risk level. The paper-testing instance is retired with a documented migration.

## Approach

Correctness and safety outrank diff size. Order of work:

- [ ] #1509 Partition every portfolio-wide gate, latch, display, and reset by mode.
- [ ] #1510 Let a paper strategy name its live replay source so twins can share one process.
- [ ] Merge procedure (this issue): a `scripts/merge-paper-into-live.sh` that takes the paper instance's config and state directory, renames each paper strategy id with a `-paper` suffix, sets `replay_source_id` where `replay_sharing="live_mirror"` is set, appends the strategies to the live config, imports the paper strategy rows, trades, and positions into the live SQLite state, and refuses to run while either daemon is live. Idempotent and dry-run by default.
- [ ] Update `check-live-paper-config-drift.sh` to detect in-process twins and compare them with the same strip-mode rule.
[...]

PR #1513, opened 2026-09-03
feat(#1509): partition portfolio risk gates and latch by live and paper scope [C97, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1509: [C97] Live and paper strategies in one scheduler share every portfolio-wide risk gate and latch

richkuo:
## Plain simple English

One safety system watched every strategy together, so a paper loss could stop live trading and paper positions could hide real live exposure. Live and paper strategies now each get their own safety system: their own drawdown peak, kill switch, daily loss limit and exposure totals. One mode can no longer stop the other.

## Summary

Closes #1509. Part of #1511.

One scheduler process now holds two portfolio scopes, `live` and `paper`, derived only from `isLiveArgs` (`--mode=live`). Each scope keeps its own drawdown peak, kill-switch latch, event log, daily-loss ledger, notional total, exposure total and correlation model, evaluated once per cycle over that scope's strategies.

**Scope model.** New `scheduler/portfolio_scope.go` holds `PortfolioScope`, `portfolioScopeFor`, `activeScopes` (only scopes with a configured strategy), the subset helpers, `dueStrategiesNotLatched`, and the testable per-cycle pair `measureScopeCycleRisk` / `applyScopeCycleRisk`. `strategyMode` is deleted and `hyperliquidModeFromArgs` becomes a wrapper over `isLiveArgs`, so one classifier decides mode everywhere. The old parser accepted a bare `live` positional that `isLiveArgs` rejects, which could have made a strategy paper for risk and live for the candle fetch.

**State and persistence.** `AppState.PortfolioRisk` and `AppState.CorrelationSnapshot` become maps keyed by scope. `portfolio_risk` is rebuilt with `scope` as its primary key, `correlation_snapshot` likewise, and `kill_switch_events` gains a `scope` column; the rebuild introspects `PRAGMA table_info` and is idempotent. Legacy unscoped rows load under a sentinel and `assignLegacyPortfolioScope` places them in `LoadStateWithDB` (which owns the roster): `live` when `HasLiveStrategy`, else `paper`, saved immediately so a second boot finds nothing to move. A conflicting pre-existing scoped row keeps its values, ORs the latch, and records a `migration_conflict` event.

**Actions and alerts.** A live latch keeps the exchange close plan, restricted to the live roster. A paper latch calls `forceClosePaperScopePositions` (virtual close at mark, no exchange call), guarded once by the persisted `KillSwitchCloseApplied`. `AutoResetConfirmedFlatKillSwitch` and `ClearLatchedKillSwitchSharedWallet` act on the live scope only, so paper never auto-resets without an owner. The reset DM names its scope, uses a per-scope single-flight flag, and accepts `reset` for one latched scope or `reset live` / `reset paper` when both are latched; a bare `reset` with two latches is refused. Paper equity is always available and trusted, so `DrawdownReadingSubstituted` and `UntrustedOverLimitSince` can never be set on a paper scope.

**Surfaces.** `/status` gains `portfolio_risk_by_scope`, `total_value_by_scope`, `total_notional_by_scope` and `correlation_by_scope`; the legacy fields keep today's values from the live scope, or paper when no live strategy exists. Circuit-breaker and correlation Discord responses and the dashboard risk and correlation panels render one block per scope. `resolveChannelKey` gains `isLive` and prefers a non-empty `<platform>-paper` key, so summaries, leaderboards and Sharpe groups split by mode only when that channel is configured.

**Config.** New optional `portfolio_risk.paper` override block with the parent's fields. `scopeRiskConfig` merges non-zero override fields over a clone of the parent (zero means inherit); a nested `paper.paper` is rejected. Override fields hot-reload with change lines, and `paper.max_notional_usd` is restart-required like the parent.

Existing single-mode deployments keep today's behavior: one scope is evaluated, so the cycle inputs, the reason strings and the on-disk values are unchanged apart from the scope key on rows and the scope label on displays.

## Verification

- `gofmt -l scheduler` empty; `go -C scheduler vet ./...` clean; `go -C scheduler build ./...` clean.
- `go -C scheduler test ./... -count=1` passes, including the untouched `portfolio_risk_review_1449_*`, `risk_test.go` and `kill_switch_*` intent.
- `go -C scheduler test ./... -race` on the new scope suites is clean.
- Mixed live and paper config, one cycle: prints `[config] portfolio scopes: live=1 paper=1 strategies` and one `Portfolio peak initialized [<scope>]` line per scope; `/status` returns `portfolio_risk_by_scope` keys `["live","paper"]`.
- Single-mode config: one peak-init line, one scope key, and the legacy `portfolio_risk` object mirrors that scope.
- Seeded legacy unscoped row plus a roster holding a live strategy: logs `Legacy unscoped portfolio risk row placed in the live scope`, the latch and its event land under `live`, and the second boot logs no placement.
- Two boots on the same DB leave identical `SELECT scope, kill_switch_active, peak_value FROM portfolio_risk` rows, and `PRAGMA table_info(portfolio_risk)` shows `scope` as the primary key.

New tests cover: `TestPortfolioScopeFor`, `TestActiveScopes_OnlyConfiguredScopes`, `TestStrategiesInScope_AndStateFilters`, `TestMigrateSchema_PortfolioRiskScope_Idempotent`, `TestMigrateSchema_PortfolioRiskScope_LegacyRowBecomesUnassigned`, `TestAssignLegacyPortfolioScope`, `TestLoadStateWithDB_PlacesLegacyRowAndPersists`, `TestSaveLoadState_PerScope`, `TestCycleScopeRisk_LiveLatchLeavesPaperFree`, `TestCycleScopeRisk_PaperLatchLeavesLiveFree`, `TestCycleScopeRisk_PaperEquityAlwaysTrusted`, `TestCycleScopeRisk_SingleModeMatchesLegacy`, `TestPaperKillSwitch_ForceClosesPaperOnly`, `TestLiveKillSwitch_LeavesPaperBooks`, `TestAutoResetConfirmedFlat_LiveScopeOnly`, `TestParseKillSwitchResetReply`, `TestManualResetOneScope_OtherStaysLatched`, `TestKillSwitchResetPrompt_NamesScope`, `TestEvaluateDailyLossLimit_Scope`, `TestEvaluateExposureCap_Scope`, `TestComputeAssetDeltas_Scope`, `TestPortfolioNotional_Scope`, `TestAggregatePerpsMarginInputs_Scope`, `TestComputeCorrelation_PerScope`, `TestManualStateView_ScopeLatch`, `TestManualStateView_ScopeHoldsDoNotCross`, `TestResolveChannelKey_PaperPrefersPaperKey`, `TestResolveChannelKey_NoPaperKeyFallsBack`, `TestSendToScopeChannels`, `TestFormatCircuitBreakersResponse_PerScope`, `TestHandleStatus_PerScopeFields`, `TestScopeRiskConfig_Merge`, `TestValidateConfig_PaperOverrideRanges`, `TestConfigExample_PaperOverrideLoads`, `TestHotReload_PortfolioRiskPaperOverride`, `TestHotReload_PaperMaxNotionalRestartRequired`, `TestHotReload_WhileLatched_KeepsLatch`, `TestRebaselinePeakAfterPrune_PerScope`.

## Adopted plan

Built to the Fable 5.1 implementation plan on the issue: https://github.com/richkuo/go-trader/issues/1509#issuecomment-5520210401

Deviations:

1. Plan step 8 says the whole-portfolio `totalPV` used by `logMgr.LogSummary` becomes "the strategy's own scope total". A per-cycle summary line has no strategy, so `LogSummary` keeps a process-wide total, now computed as the sum of the per-scope totals. That equals the previous value in a single-mode deployment. `notifyPerStrategyCircuitBreakerWithSnapshot` does take the strategy's own scope total, as the plan asks.
2. Plan steps 8 and 11 leave the per-scope cycle evaluation inline in `main()`, which no test can reach. The body was extracted unchanged into `measureScopeCycleRisk` and `applyScopeCycleRisk` so the plan's own `TestCycleScopeRisk_*` table tests the shipped code rather than a copy of it. `checkPortfolioRiskWithEquityAvailability` is untouched, as the plan requires.
3. Plan step 16's verify line runs `./go-trader --config scheduler/config.example.json --probe-only`. The binary has no `--probe-only` flag (it belongs to the check scripts). The example config is verified instead by `TestConfigExample_PaperOverrideLoads`, which loads it through `LoadConfig` and asserts the override merge, plus the mixed and single-mode `--once` smoke runs above.
4. Plan step 14 says the paper key is preferred when "any backend has that key". A key present with an empty channel id would capture the grouping and silently drop the summary, so the preference requires a non-empty id. `TestResolveChannelKey_NoPaperKeyFallsBack` covers the empty case.
5. Plan step 6 keeps `dailyLossStatusNote` and `exposureCapStatusNote` taking a `*PortfolioRiskConfig`. They now take the `*Config`, because emitting one block per active scope needs the roster and `scopeRiskConfig` for each scope. The single-scope output is unchanged.

## Test edits

All are case **Outdated**. Ground: issue #1509 Approach step 1, "Replace the single `AppState.PortfolioRisk` with a map keyed by scope", and the adopted plan steps 3, 6, 11, 13 and 14, which name each changed signature. No expectation was weakened.

1. `db_test.go`, `risk_test.go`, `shared_wallet_test.go`, `kill_switch_limit_orders_test.go`, `ui_trade_actions_test.go`, `portfolio_risk_review_1449_test.go`, `portfolio_risk_review_1449_deferral_test.go` read and write the latch through `scopeRisk(ScopeLive)` and the scoped maps. Every assertion is unchanged and now binds to the live scope.
2. `daily_loss_test.go` and `exposure_cap_test.go`: `dailyLossStatusNote` and `exposureCapStatusNote` take a `*Config` and `manualExposureCapStatus` takes a scope. The fixtures are wrapped in a config carrying the same strategies; the expected note text is unchanged.
3. `discord_commands_test.go` and `ui_ops_test.go`: `formatCorrelationResponse` and the stored correlation take the per-scope map. The same snapshot is asserted under the live scope.
4. `notifier_test.go`: `resolveChannelKey` takes `isLive`. The existing cases pass `true` and assert the same keys.
5. `kill_switch_close_test.go`: `formatKillSwitchResetPrompt` takes the scope and the latched list. The single-latch case asserts the same content.
6. `notional_cap_test.go`: the source guard now matches `notionalCapSkipsStrategyCycle(sr.NotionalBlocked)`, and its AST check accepts the `NotionalBlocked` selector as well as the old identifier. It still forbids a raw notional-conditioned `continue`, which is the behavior it guards.

---
LLM: Opus 5 | xhigh | Harness: work-on-issue


richkuo:

richkuo:
Addressed review feedback (github-actions[bot] · 2026-09-03T05:33:48Z) in e3052b2.

### Fixed
1. **The paper kill switch has no reset path of any kind when no DM owner is configured, so an ownerless deployment's paper scope latches permanently.** — Confirmed against origin/main, where the ownerless path force-closed and auto-reset via `CanAutoResetWithoutOwner`. New `applyPaperKillSwitchCycle` runs the virtual close once and, with no owner, calls `AutoResetConfirmedFlatKillSwitch` on the paper scope with the paper equity and re-baseline flag; it also fires on a restart that loads `KillSwitchCloseApplied=true`, and an owner added later takes over because the reset clears `KillSwitchCloseApplied` and the DM loop keys on `HasOwner` (`scheduler/risk.go:1464`). Covered by `TestPaperKillSwitch_AutoResetsWithoutOwner` (first cycle, restart mid-latch, owner configured; live latch and live books untouched in every case).
2. **The legacy `/status` fields `total_value` and `total_notional` stay process-wide while the shipped docs say all legacy fields mirror the live scope, and the legacy surface is internally inconsistent in mixed mode.** — Both legacy totals now take the `statusLegacyScope` value; with no active scope they keep the process-wide reading (`scheduler/server.go:333`). `TestHandleStatus_PerScopeFields` asserts mixed mirrors live and paper-only mirrors paper.
3. **Every paper reset prompt after the first close cycle carries an empty close-plan message, so the operator sees no drawdown reason.** — The prompt now builds `formatPaperKillSwitchPromptMessage(sr.Reason)` on every issue (`scheduler/main.go:1359`); the one-shot close message is only the channel broadcast. Covered by `TestPaperKillSwitchPromptMessage_CarriesReason`.

Docs: `docs/POST_UPDATE_HISTORY.md` and `SKILL.md` state the paper auto-reset rule and that the prompt carries the reason on every issue.

Verification: `gofmt -l scheduler` empty, `go -C scheduler vet ./...` clean, `go -C scheduler test ./... -count=1` passes.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-09-03T05:54:02Z) in 22d0cc63.

### Fixed
1. **The upgrade puts the whole-portfolio peak into the `live` scope while the live total drops to live-only equity, so a mixed live+paper deployment can latch and market-close its real book on the first cycle after the upgrade.** — Confirmed. `assignLegacyPortfolioScope` now re-bases the moved live peak whenever the roster also holds paper strategies, via `liveScopeRebasedPeak` (`scheduler/state.go:501`, `scheduler/shared_wallet.go:493`): the sum of live per-strategy `RiskState.PeakValue`, configured capital only for a strategy with no peak yet, and wallet equity for a shared-wallet pool. The latch, timestamp and events move untouched, a `scope_basis_rebaseline` event is recorded, and the placement is saved in the same `LoadStateWithDB` write. A whole-roster capital floor was deliberately left out because live books already below configured capital would otherwise read as a drawdown. `TestAssignLegacyPortfolioScope_MixedRosterRebasesLivePeak` covers the reviewer's three cases: paper share removed with no first-cycle latch, latch intact through the re-base, and a genuine live loss still latching.
2. **The `paper` scope's first peak is seeded from configured `capital`, not from the paper book's current value, so a paper book already below its starting capital is force-closed on the first cycle after the upgrade.** — Confirmed. The startup seed loop skips a zero-peak scope whose strategies already carry persisted state (`scheduler/main.go:290`, `scopeHasPersistedState`); the first cycle then seeds the peak from the scope's current total through the existing peak ratchet in `checkPortfolioRiskWithEquityAvailability`, which yields zero drawdown on that cycle. A scope with only new strategies keeps the configured-capital seed. `TestCycleScopeRisk_NewScopeSeedsPeakFromCurrentValue` covers 70% of capital (no latch), a later loss (latches), above capital (seeds at the higher value), and `TestScopeHasPersistedState` covers the fresh-scope fallback.
3. **When both scopes latch at once, two reset prompts wait on the same owner DM and one consumes the other's reply, so the operator must answer twice.** — Confirmed. One single-flight waiter again; when both scopes latch, one prompt built by `formatKillSwitchResetPromptForScopes` carries both sections and accepts `reset live` or `reset paper` (`scheduler/main.go:1366`). The reply is parsed against the scopes latched at reply time, so a bare `reset` clears a lone survivor after the other auto-resets, and the acknowledgement names any scope still latched, which is re-prompted next cycle. `TestKillSwitchResetPromptForScopes` covers the prompt and the reply routing.

### Test edits
1. **TestAssignLegacyPortfolioScope** (`scheduler/portfolio_scope_persistence_test.go:139`) — Outdated; ground: finding 1 of this review, the mixed-roster placement now re-bases the peak; the case is narrowed to a live-only roster where the intact-peak assertion still holds, and the mixed roster now asserts the re-based peak, intact latch and event in `TestAssignLegacyPortfolioScope_MixedRosterRebasesLivePeak`.

Docs: `docs/POST_UPDATE_HISTORY.md` and `SKILL.md` describe the re-base, the deferred seed, and the single prompt.

Verification: `gofmt -l scheduler` empty, `go -C scheduler vet ./...` clean, `go -C scheduler test ./... -count=1` passes.

---

richkuo:
@claude review

PR #1527, opened 2026-09-05
feat(#1523): separate live and paper state files behind one storage owner [C99, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1523: [C99] One scheduler process writes live and paper state into one SQLite file
**Complexity: 99/100** - Capability 3 (Risk 4, Uncertainty 2: routing controls persisted money records; the full caller set needs discovery); Volume 24 (Scope 4: storage, config, commands, scripts, and a shared storage contract; Coupling 4: locks, reload, background workers, and two databases; Verification 4: concurrent ownership and crash recovery) - Opus 5, xhigh · fableplan: yes

## Problem

At baseline `38d78534`, the scheduler opens one state database. Portfolio risk is partitioned by scope after #1509, but strategy records still share that database. The maintainer requires separate live and paper database files.

A wrapper with the unchanged method signatures cannot safely provide this separation. `UpdateTradeDiagnosticsMetrics` receives only a database-local row identifier. `DeletePendingManualActionsThrough` receives only a database-local high-water mark. Queries such as `RecentTrades` and `LifetimeTradeStatsAll` have no strategy selector. None of these calls can choose a file from a strategy identifier alone.

Stored strategy identifiers also differ from process identity requirements. Existing files can contain the same strategy identifier, but the in-memory strategy map requires unique keys. Renaming only the config identifier can initialize an empty strategy or cause startup to prune an existing book. The stored strategy rows do not contain an authoritative live or paper mode field.

## Goal

Run both modes in one process with separate state files, stable stored identities, explicit ownership for every operation, and recovery that cannot lose or duplicate a recorded effect across a partial save.

## Approach

Part of #1526. This issue supplies the storage and inspection contracts used by #1525. Existing single-file deployments remain supported.

1. Add optional root `paper_db_file`. When absent, retain the existing single-file layout. When present, resolve both paths and verify distinct physical files, including relative paths, symbolic links, and hard links. Validate parent-directory resolution for new files. Database paths and storage identity mappings require restart; a scope change cannot move an existing book between files.
2. Keep process strategy identifiers unique. Add optional per-strategy `storage_strategy_id`, defaulting to the configured identifier. Storage identity is `(scope, storage_strategy_id)`. Require an unambiguous mapping within each file. Translate identifiers at the storage boundary in both directions, including position owners, hedge references, queued actions, diagnostics, and historical queries. Keep replay source references in the process identifier namespace. A renamed process identifier must load and save the existing stored book without rewriting stored identifiers.
3. Introduce an explicit storage owner that holds the physical database handles and the immutable identity map. Classify every table and every reader and writer. Strategy records, option records, diagnostics, and manual actions belong to their strategy scope. Portfolio risk, kill-switch events, and correlation records belong to their explicit scope. Live wallet ledgers, transfers, cashflow records, and live pending limit orders belong to the primary file, with live-only mutation guards. Process metadata and shared regime history, transitions, and alert deduplication belong to the primary file for new writes. Regime history uses market and classifier keys. Preserve existing paper-file history without copying it, and expose its source when read.
4. Change interfaces where ownership is absent. Background diagnostic work carries the owning scope with the row identifier. Backfill plans bind to their source file and strategy. Drain and acknowledge manual actions per file; acknowledge only successfully applied, durably recorded actions. A high-water mark from one file must never delete rows in another file or discard a failed action. Unknown ownership must produce an error before mutation.
5. Define combined reads explicitly. Query each required file, translate identities, combine results, then apply global ordering and pagination. Include source scope and a stable tie-breaker for colliding row identifiers and timestamps. Counts, statistics, dashboard, exports, and diagnostics must include both files where requested. A failed file read must produce a failure or an explicit incomplete result, never a successful report with missing scope data. Historical rows without a current strategy retain source-file provenance and must not be assigned to a scope by an identifier suffix.
6. Split state loading and saving by physical owner. Loading scope data must work without a process metadata row in the paper file; the current `LoadState` returns early when that row is absent. Apply legacy unscoped risk placement using the owning file's validated scope before combining state. Keep each file internally transactional. Return save outcomes per scope, retain failed state for retry, and prevent new exposure in a scope with an unacknowledged persistence failure. Retrying a failed scope must not duplicate effects already committed in the other scope. Preserve protective management subject to durable recording and the existing confirmed-fill rules; report any action that cannot run safely.
7. Add read-only ownership inspection for deployment checks. For existing files, validate strategy mappings and existing scoped risk rows before normal startup writes or pruning. Reject cross-scope or ambiguous current books with the identifier and source file named. Current config alone cannot prove the historic mode of an arbitrary unknown row. Require explicit resolution of ambiguous legacy ownership; do not move or delete rows automatically. Normal schema migrations remain per file and separate from read-only inspection.
8. Extend the existing singleton lock mechanism to both files in deterministic order, before migrations, state mutations, or trading. Release all acquired locks if either acquisition fails. Include `--once` in scheduler ownership checks. Migration preflight must use the same ownership contract and coordinate with the existing manual-action locks. Preserve sanctioned manual command behavior while preventing a migration from racing a scheduler or manual write.
9. Audit all direct database openers and callbacks, including manual commands, diagnostics workers, fee and trade-ledger backfills, agent information, server routes, trade recording, and shutdown. Replace the previous fixed caller count with a complete inventory from the implementation baseline. Update config validation and reload, read-only inspection, both-file backups and restores, command output, documentation, and the single-file compatibility tests.

There is no transaction across the files. A process crash can stop both modes. Separate files provide record separation, but do not isolate faults in shared process memory.

## Acceptance criteria

- Database fixtures prove live and paper trade, position, option, risk, and manual-action writes in both directions, including identical stored strategy identifiers and identical row identifiers in separate files.
- A process alias resumes existing cash, open positions, risk latches, pending actions, diagnostics, and trade history. No stored identifier is renamed, no book is reset, and no unmatched book is pruned during failed startup validation.
- Background updates and backfills change only the selected file. A failed manual action survives acknowledgement of another action, including when the other action belongs to the other file.
- Combined history, pagination, counts, and exports produce deterministic complete results, including equal timestamps and row identifiers. A missing or unreadable file is reported explicitly.
- Startup rejects aliases of the same physical file, invalid identity maps, conflicting ownership, and a lock held on either file before any database mutation. Restart and reload tests cover paths, identities, and scope changes. A second scheduler cannot use either owned file, including with `--once`.
- Fault injection covers failure before either commit, failure after one file commits, repeated save failure, retry, and restart. Confirmed records survive without duplicate booking, risk reset, or cross-scope acknowledgement.
- Read-only inspection changes no database or journal. Normal migrations work independently, and paper scope loading does not depend on primary process metadata.
- Backups and restores cover both files. Existing single-file behavior passes its contract tests. Go tests use database fixtures and process or lock stubs without spawning Python.

## Plain simple English

Live and paper records must stay in separate files. Some database operations do not identify which file owns a row, and renaming a strategy can disconnect it from its history. These gaps can send a write to the wrong file or hide an existing position.

---

richkuo:
## Plain simple English

Live and paper records can now live in two separate database files. One owner holds both files, sends every write to the correct file, and joins reads that cross both. A strategy can be renamed without losing its money, positions or history. Every file is locked before any change, and a failed save stops new positions in that mode only.

## Summary

Closes #1523

Adds optional root `paper_db_file` and per-strategy `storage_strategy_id`. Omit `paper_db_file` and the single-file layout is unchanged. Set it and the paper scope's books, risk row, kill-switch events and correlation snapshot move to a second file, while `db_file` keeps the live scope, process metadata, the live-only wallet and cash-flow tables, and shared regime history.

A new `StateStore` (`scheduler/state_store*.go`) owns the physical handles and an immutable identity map. It routes every write to the file that owns the row and combines every read that spans scopes. Storage identity is `(scope, storage_strategy_id)`, unique per file; the same value in both files is the supported alias, so a strategy `id` can be renamed with no stored rewrite and no book reset.

### Behavior changes that apply to every deployment

| Change | Before | After |
| --- | --- | --- |
| File ownership | lock taken after `OpenStateDB`, `LoadState` and the startup writes; skipped for `--once` | every state file locked in role order **before** any migration or startup write; `--once` takes it too, so it refuses while the daemon runs (exit 79 names the file and holder pid) |
| Layout check | none | read-only ownership check at startup; a rejected layout exits **80**. Units now set `RestartPreventExitStatus=78 79 80` |
| Persistence hold | new exposure allowed until three failed saves, then the whole cycle skipped | any unacknowledged failed save holds position-increasing signals for **that scope** at the six regime-gated dispatch sites and refuses `manual-open`/`manual-add`; closes, trailing stops, ratchet, protection and hedge keep running. Three consecutive failures still skip that scope's strategies |
| Manual acknowledgement | `DELETE ... WHERE id <= maxDrained`, so a failed lower-id action was removed by a later success | deleted **by row id inside the transaction that persists the effect**, and only for the strategy that transaction wrote |
| `--summary` / `--leaderboard` | migrating open, startup writes | read-only open, no migration, no lock |
| `backfill --apply` | `pgrep -x go-trader` | file ownership plus each file's manual-action lock |

New read-only command: `go-trader storage-inspect [--config <path>] [--json] [--require-idle]`. It reports, per file, the canonical path, owned scopes, any held lock and pid, mapped strategies with position counts, orphans (flagged when they hold positions), the pending-action count and risk rows with their latch. It writes nothing and tolerates a pre-#1509 schema.

### Deviation from the adopted plan

One, taken under the plan-deviation policy's correctness override, and it is the only one.

The plan's decision 1 has the store translate strategy identifiers "at that boundary" — that is, at every store call site — with `StateDB` methods keeping stored identifiers in their signatures. That would require translating at every one of the ~100 sites that hold a `*StateDB`, including the deep executor, protection-sync and stamping paths, where a single missed translation writes a process identifier into a file that expects a stored one, silently.

Instead the per-file `StateDB` carries the file's half of the identity map and translates inside itself, at the SQL boundary (`toStorageID` / `fromStorageID` in `scheduler/state_db_identity.go`). Every existing method keeps its process-namespace signature, so the deep call paths are unchanged and it is structurally impossible to skip a translation. The plan's structure is otherwise implemented as written: the store still partitions by scope (`saveScopeBooks` / `loadScopeBooks`), still owns routing, combining, acknowledgement and the persistence hold, and `StateDB` still returns orphans separately rather than keying them into the roster. `CLAUDE.md` records the rule as "ids translate INSIDE `StateDB`, never at a call site".

Two smaller points inside plan steps, neither changing an observable contract:

- `storage_strategy_id` is trimmed by the `effectiveStorageStrategyID` accessor rather than materialised into the config struct at load, so a config rewrite never gains the key for every strategy.
- Read-only inspection asserts the database file's bytes and content are unchanged and that no rollback journal appears. It cannot assert the absence of `-wal` / `-shm`: SQLite's read-only WAL path must be able to create the shared-memory index and cannot delete it on close. The plan's own risk list anticipates this. The test comment states it.

Plan: https://github.com/richkuo/go-trader/issues/1523#issuecomment (the `## Implementation plan (Fable 5.1)` comment on the issue).

## Test edits

Every one is disclosed with its case and its checkable ground. Nothing was weakened.

| Test | Case | Ground | Replacement asserts |
| --- | --- | --- | --- |
| `TestDrainPendingManualActionsAlerts` (`manual_test.go`) | Outdated | Issue #1523 acceptance criterion: "A failed manual action survives acknowledgement of another action, including when the other action belongs to the other file." | The failed DOGE close keeps its queue row while the three applied actions are acknowledged |
| Four `flushOffCycleLiquidationAuditState` sub-tests (`hyperliquid_liquidation_guard_test.go`) | Outdated | Issue #1523 Approach item 6 ("Return save outcomes per scope") and the adopted plan's decision 8 ("`saveFailures` becomes per scope") | The store's per-scope failure counter and the persistence hold, in place of the returned int |
| `TestKnownSubcommandsMatchDispatch` (`main_test.go`) | Outdated | The adopted plan's step 3 registers `storage-inspect` in `knownSubcommands`; the test's own failure message instructs the update | The list including `storage-inspect`, still guarding dispatch drift |
| `TestMirrorSaveBeforeMarkWiring` (`replay_mirror_test.go`) | Outdated | The per-strategy handle in the cycle is now `stratDB` and the fleet save is `SaveStateWithStore`, both from this change | Unchanged invariant: the strategy-book save still runs before `MarkDecisionsApplied`, and the block still avoids a fleet-wide save |

Every other test file change is a mechanical call-site adaptation to a changed signature (`*StateDB` → `*StateStore`, `DeletePendingManualActionsThrough` → the queue-clearing helper). Their assertions are unchanged.

## New tests

`scheduler/storage_split_test.go` and `scheduler/storage_inspect_test.go` cover the contracts the issue names: layout resolution against relative paths, symlinked parents, symlinked files and hard links; identity mapping both ways including the two-file alias and an unmapped identifier; ownership acquisition with release-all on a held second file; by-id acknowledgement; a split round trip with the **same stored identifier and the same row identifiers in both files**; a process alias resuming cash, positions, latch and history with no stored rewrite; orphan reporting instead of adoption; fault injection through `storeCommitHook` proving the committed scope is not re-booked on retry and the failed one is not marked persisted; paper books loading with no primary metadata row; per-file legacy row placement and the refusal when it is ambiguous; deterministic combined pagination across equal timestamps; an unreadable file failing the whole read; live-only writes refusing a paper strategy; per-file drain and acknowledgement; per-scope save failures and the hold; plan-to-file binding; the combined TradingView export; and the ownership inspection cases including a pre-scope schema and `--require-idle`.

## Verification

- `gofmt -l scheduler/` — clean
- `go -C scheduler build .`, `go -C scheduler vet .` — clean
- `go -C scheduler test ./...` — ok, 44s
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2929 passed
- `uv run --no-sync python scripts/check_test_budget.py` — wording-only tests 170 (baseline 170)
- `bash -n scripts/update.sh scripts/update_helpers.sh` — clean; the new `update_resolve_db_exclude` assertions in `scripts/test_update_helpers.sh` pass (that script's later `systemd`-dependent check fails identically on `main` on macOS)
- `wc -c CLAUDE.md` — 14,998 bytes
- Manual smoke on a two-file config: `--once` created both files, traded the paper scope and saved; `storage-inspect` showed `hl-paper-eth` and a paper-scope risk row in the paper file and nothing in the primary; a second `--once` exited 79 naming the held file and pid; an aliased layout was rejected before any file was opened; `--require-idle` passed on an idle layout.

---

https://claude.ai/code/session_01HkUFBcasS7FYpgH6xYykuB


richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Fable 5.1 · 2026-09-05T03:47:04Z) in db9b69ab.

### Fixed
1. **The root `go-trader.service` unit still lacks exit code 80 in `RestartPreventExitStatus`, so a rejected storage layout crash-loops instead of staying down.** — `go-trader.service:14` now reads `78 79 80`, in lockstep with `systemd/go-trader@.service:14`; the docs that stated it are now true.
2. **The persistence hold does not block option open actions, unlike every sibling gate at the same dispatch site.** — the options branch drops open actions through `pausedOptionsActions` while the scope is held, closes and theta-harvest untouched (`scheduler/main.go:1976`). The hold still sits below the three-strike skip, so the single-failure case is covered.
3. **`CountPendingLimitOrders` errors for a paper-scope strategy, which makes `manual-open`/`manual-add` unusable for paper `type=manual` strategies in a split layout.** — the read resolves through a new `liveOnlyReadFileForStrategy` that reports 0 for a paper strategy and still errors for an unmapped id (`scheduler/state_store_reads.go:563`, `:580`); `InsertPendingLimitOrder` keeps refusing paper. Asserted in `TestStateStoreRoutesLiveOnlyWritesAndRefusesPaperStrategies` (`scheduler/storage_split_test.go`).
4. **`probeStateDBLockHolder` briefly acquires the exclusive ownership lock, so an inspection racing a daemon start can make the daemon exit 79 and stay down.** — `acquireStateDBLock` retries a contended `flock` up to 10 times at 10 ms (`scheduler/singleton_lock.go:48`, `:75`); the probe holds for microseconds, so the window is closed. A genuinely held lock still returns the same `stateDBLockedError` with the holder pid after the retries, so the two-inspection and self-probe cases are unchanged. Asserted in `TestAcquireStateDBLock_SurvivesProbeWindow` (`scheduler/singleton_lock_test.go`).

No test edits.

Verification: `gofmt -l scheduler/` clean; `go -C scheduler build .`, `go -C scheduler vet .` clean; `go -C scheduler test ./...` ok (43s); `scripts/check_test_budget.py` 170 (baseline 170).

---

richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Opus 5 · 2026-09-05T04:38:56Z) in 9b05fdd4.

### Fixed
1. **The eager diagnostics insert writes the process identifier into `trade_diagnostics.strategy_id`, so an aliased strategy's rows are keyed under a name every read and every correction filters against the other one.** — `StateDB.InsertTradeDiagnostics` now translates with `toStorageID` inside `StateDB` and inserts through `insertTradeDiagnosticsRowAs` (`scheduler/trade_diagnostics_db.go:16`); the single-file layout is unchanged because a nil identity map returns the id as-is. Asserted end to end in `TestStateStoreEagerDiagnosticsInsertUsesStoredIdentifier` (`scheduler/storage_split_test.go`): stored key equals the storage id, the process-id read returns the row, and a correction keyed on the stored id hits it.
2. **The singleton and storage-layout refusals now run before the notifier exists, so a daemon that refuses to start sends no owner DM and stays down silently.** — the notifier is built before the ownership block (`scheduler/main.go:168`); exit 79 and both exit 80 paths DM the owner through `sendStartupRefusalDM` and call `cleanupNotifier()` before `os.Exit` (`scheduler/main.go:176`, `:189`, `:196`; `scheduler/state_store.go:585`). The DM names the file and holder pid, or each layout rejection (`storageRefusalDMBody`, `scheduler/state_store.go:594`). A config with no backend is a nil-safe no-op. On `main`, the notifier was already built before the read-only report modes, so their behaviour is unchanged.
3. **Diagnostics rows flushed by the replay-mirror book save carry no scope, so the split-storage worker drops them and their trade-quality metrics are never computed.** — both replay-mirror saves route through `store.SaveStrategyBook(stratState)`, which resolves the scope and the pending acks (`scheduler/main.go:2372`, `:2387`).
4. **`LoadStateWithStore` dereferences the primary handle without checking it, so a read-only command panics when only the paper file exists on disk.** — nil guard returning a missing-primary-file error (`scheduler/state_store_load.go:40`). Asserted in `TestLoadStateWithStoreReportsMissingPrimaryFile` (`scheduler/storage_split_test.go`); the both-absent case still returns the existing "no state file exists yet" error from `openToolStateStoreReadOnly`.

### Test edits
1. **`TestMirrorSaveBeforeMarkWiring`** (`scheduler/replay_mirror_test.go:333`) — Outdated; ground: confirmed finding 3 above, which replaces the `SaveStrategyBookWithDB(stratState, stratDB)` call the test pinned; now asserts the same save-before-mark ordering and the same no-fleet-save guard against `store.SaveStrategyBook(stratState)`.
2. **`scripts/test_budget_baseline.json`** — one entry added for `TestStorageRefusalDMNamesEveryRejection`, an operator-decision wording test (the DM must name each rejection so the owner can act without the journal).

Verification: `gofmt -l scheduler/` clean; `go -C scheduler build .`, `go -C scheduler vet .` clean; `go -C scheduler test ./...` ok (44s); `scripts/check_test_budget.py` 171 (baseline 171).

---

richkuo:
@claude review

PR #1528, opened 2026-09-05
feat(#1524): shared Hyperliquid market feed with synchronized sealed snapshots [C99, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1524: [C99] Hyperliquid checks can read different market inputs and run on different schedules
**Complexity: 99/100** - Capability 3 (Risk 4, Uncertainty 3: market data and outage gates control live entries and protection inputs; the history bootstrap medium and the non-Hyperliquid consumer scope were unsettled and change the edit list); Volume 24 (Scope 4: Go feed and scheduling plus Python payload consumers, a shared contract, and docs; Coupling 4: subprocess parity, concurrent socket updates, and reload; Verification 4: timing, reconnect, and outage reproduction) - Opus 5, xhigh · fableplan: yes

## Problem

At baseline `38d78534`, eligible Hyperliquid batch members already share one candle fetch across modes. Singleton strategies, unsupported batch arguments, changed inputs, and batch failures can use individual checks. The adapter also has a candle cache. Thus every strategy check does not necessarily issue its own exchange request.

Separate systemd instances have private temporary directories and separate polling clocks. One process alone also does not align checks: eligibility uses per-strategy last-run times, those times are assigned after each strategy finishes, and drawdown can shorten a strategy's effective interval.

A shared socket alone cannot guarantee identical evaluation inputs. Signal checks also use higher-timeframe candles and funding data. Regime classification can use a different configured timeframe. Current candle conversion prefers the exchange close timestamp, and signal code reads the latest frame row. Account data cannot replace missing candle-derived inputs for every close or trailing decision.

The repository has no websocket client today. The Go binary already calls the Hyperliquid `/info` endpoint directly for mid prices, and the Python adapter calls the same endpoint for candle snapshots. Root config decoding ignores unknown keys; the unknown-key guard covers only strategy and `user_defaults` keys.

## Goal

Give compatible live and paper evaluations the same identified, immutable market snapshot on a common schedule, while maintaining explicit freshness, recovery, and protective behavior when data is unavailable.

## Approach

Part of #1526. Ship after #1523 in the umbrella sequence. Preserve single-instance deployment support, and keep the backtester and tuning page outside the live feed.

This issue covers Hyperliquid perps and `manual` strategies only. Strategies on other platforms keep legacy polling under every `market_feed` value. Six REST candle and mid consumers stay outside the feed and outside the zero-call criterion: manual-open ATR fetch (`--fetch-atr`), LLM entry review candles, the UI candle chart, the status server mid fetch, the liquidation guard mid fetch, and OKX perps checks that fetch their own candles. Name each in the implementation plan so a reviewer can confirm the exclusion.

1. Add one Go feed owner with one websocket connection to the Hyperliquid host. Own subscriptions, reconnect backoff, liveness, and bounded candle history there. Build the subscription set from all in-scope scheduler consumers: strategy candles, higher-timeframe filters, regime timeframe overrides, and required history lengths. Key raw data by exchange host, market namespace, symbol, and timeframe. Store the maximum required lookback per key and slice it for consumers. Preserve classifier settings in derived-result keys. Derive subscriptions from validated config generations on reload; publish a new generation only after its required data is ready. The feed owner has its own lock and never takes the scheduler state lock; snapshot reads, the reload generation swap, and feed alerts run outside that lock. The feed goroutine stops on the shutdown contexts with the rest of the process.
2. Synchronize compatible signal checks using common deadlines that do not depend on individual completion times. Give twins with the same effective cadence the same evaluation identifier and snapshot. Define restart, overrun, missed-deadline, and reload behavior without replaying missed trading opportunities. Preserve faster risk and protection checks, including the off-cycle liquidation audit. If effective cadences differ, report the difference and compare only matching evaluation identifiers; do not claim universal decision parity.
3. Bootstrap sufficient history before declaring a key ready through a Go client on the existing Hyperliquid `/info` candle snapshot endpoint, beside the existing Go mid-price client. Reuse the Python adapter's candle-row conversion contract, including its preference for the exchange close timestamp, and prove parity with a test that feeds one exchange response to both converters. Buffer socket updates during bootstrap, combine them by bar identity, and prevent an older response from replacing a newer update. On reconnect, repair missing history before declaring recovery. Validate timestamps, ordering, duplicate updates, finite positive prices, candle geometry, and required lookback. Handle missing bars and markets with sparse trading explicitly; do not manufacture candles to satisfy readiness.
4. Define a versioned snapshot envelope with evaluation identifier, config generation, source key, history coverage, bar open and close times, receive time, capture time, per-key readiness, and data provenance. State how closed and forming bars are represented. Preserve existing consumer timestamp and forming-bar behavior unless a separate behavior change is authorized. Resolve needed REST recovery before sealing the snapshot, then freeze it for all consumers. Later socket updates belong to a later snapshot.
5. Inject the snapshot into batch checks, individual checks, same-cycle batch retries, and Hyperliquid regime bundle checks. A retry reuses the original evaluation data. Include higher-timeframe frames; when regime uses a different timeframe, provide its own keyed frame from the same snapshot. Fetch funding inputs used by an evaluation once through the feed owner or a cycle-owned dependency provider, and include their provenance. Payload absence means legacy polling only when the feed is disabled or the caller is outside the feed contract. With the feed enabled, malformed, stale, incomplete, or mismatched payloads must never silently trigger a private worker fetch.
6. Use standard input for the potentially large market payload, extending the existing batch request envelope. Define bounded payload sizes, serialization errors, and the individual and regime check input contract. Keep command-line flags small and add protocol support to both startup probe argument sets. Script compatibility checks remain distinct from tests that prove payload use. Go tests use pure helpers and subprocess stubs; Python contract tests verify actual parsing and skipped adapter fetches.
7. Define separate freshness limits for candle coverage, recently received market updates, and mid prices. Socket connectivity alone does not prove freshness. Specify grace and timeout values, supported intervals, sparse-market behavior, and decision-age limits in the implementation plan. Recheck entry eligibility before dispatch if a cycle runs long. A stale key triggers at most one cycle-owned REST recovery per key, with a shared result or shared failure for all consumers. Cache use must preserve source age; a REST response is not automatically fresh.
8. Feed fresh mids into the existing cycle price map and checks from the sealed snapshot. Protective account reconciliation remains separately sourced and must not mutate that evaluation snapshot. If both feed and REST inputs fail, hold position-increasing signals at every affected dispatch path without setting the operator pause state. Separate protective dispatch from successful signal evaluation. Keep confirmed account reconciliation, existing on-chain orders, and reduce-only actions that have sufficient verified inputs. A candle-dependent close, ratchet, or trailing calculation with missing inputs remains explicitly degraded; preserve existing protection and alert. Never fabricate average true range, erase a stop, loosen protection, or claim all closes can run from account data alone.
9. Add a root config field `market_feed` with the values `websocket` and `rest`. An omitted field means `rest`, which keeps the existing polling and scheduling behavior, so no deployed config changes behavior until an operator opts in. `websocket` enables the feed owner. Validate the value explicitly in `loadConfig` after decoding and fail startup on any other value; the existing unknown-key guard does not cover root keys, so a misspelled root key keeps `rest`. Log the effective feed mode at startup so an operator can confirm the opt-in took effect. Cover the field in `scheduler/config.example.json`, and make a change of the value restart-required in `validateHotReloadCompatible`, following the existing storage identity reload check. Do not add an environment variable switch. Measure zero candle REST calls only in steady state after all required histories are ready and current; bootstrap, reconnect repair, and stale recovery are permitted calls. Record request counts and snapshot identities, without claiming reduced Python import cost. The subprocess model remains in place.
10. Update the operator and agent docs that describe the REST-per-batch flow: the SKILL.md batched signal check section and its subsystem mechanism entry, the `hl_batch.go` guardrail line in CLAUDE.md, and the README architecture diagram node for the check scripts.

## Acceptance criteria

- A controlled-clock test gives compatible twins the same evaluation identifier and snapshot despite different check durations. Cover restart, overdue work, drawdown-driven cadence changes, and reload without duplicate signal dispatch.
- Socket and REST stubs cover empty startup history, short history, gaps, duplicate or out-of-order updates, reconnect during bootstrap, silent stale keys on a connected socket, missing mids, and invalid values. Unready inputs cannot permit entries.
- A parity test feeds one exchange candle response to the Go bootstrap converter and the Python adapter converter and asserts identical rows, including the close-timestamp convention.
- Batch, singleton, batch-disabled, changed-input retry, and shared-failure retry paths consume the same sealed data. Tests include a higher-timeframe filter and a regime timeframe that differs from the signal timeframe.
- Healthy steady-state evaluations make zero candle REST calls for covered consumers. Bootstrap and repairs are counted separately. A stale key has one shared recovery attempt per cycle, with the same outcome visible to each consumer. The six excluded consumers and non-Hyperliquid strategies are outside this count and keep their existing behavior under both `market_feed` values.
- Feed and REST failure tests cover flat and open positions, long and short positions, close-only signals, scale-in, stale marks, and candle-dependent protection. Entries remain held. Verified protective actions still run, existing protection remains intact, and unsupported calculations report degradation.
- Legacy polling and injected data produce equivalent results for the same frozen inputs, including timestamp and forming-bar conventions. Snapshot comparison checks content as well as identifiers. Different account state or fills remain valid reasons for different resulting trades.
- `market_feed` omitted or `rest` passes the existing batch and scheduling tests unchanged; `websocket` enables the feed; an unknown value fails `loadConfig`; a SIGHUP that changes the value is refused with a restart-required message; startup logs the effective mode.
- A race-detector test proves the feed lock never nests inside the scheduler state lock, and a shutdown test proves the feed goroutine exits on the shutdown contexts.
- Payload compatibility passes both startup probes. Feed-off behavior passes existing contracts. Go tests do not launch Python; Python tests exercise the payload readers and adapter fallback policy.
- The docs named in step 10 describe the feed and the excluded consumers.

## Plain simple English

Sharing a connection does not ensure that live and paper strategies read the same data or run together. Missing history and failed data reads can also stop protective calculations. The design needs common evaluation times, complete shared inputs, and explicit rules for each failure.

---
LLM: Fable 5.1 | high | Harness: validate-issue

richkuo:
## Plain simple English

Live and paper Hyperliquid strategies can now read the same market data at the same moment. One websocket keeps the candles and prices in the scheduler, and every check gets the same frozen copy. If the data is missing or old, the strategy makes no new trade, but stops, take-profits and account repair still run. The old behaviour stays the default.

## Summary

Adds a root `market_feed` config field. `rest` (the default, and what an omitted field means) keeps today's polling and scheduling untouched. `websocket` opens one Hyperliquid socket, keeps the candle history and mid prices in the Go process, and hands every covered check one sealed, identified snapshot on standard input.

Scope is Hyperliquid `perps` and `manual` strategies. Every other platform keeps legacy polling under both values. Seven REST consumers stay outside the feed and outside the zero-call criterion: the manual-open ATR fetch (`--fetch-atr`), the manual-open regime-trail label check, the LLM entry-review candles, the UI candle chart, the status-server mid fetch, the liquidation-guard off-cycle mid fetch, and OKX perps checks that fetch their own candles.

What lands:

- **Feed owner** (`market_feed.go`): one socket, its own `feedMu`, a bar ring per `marketFeedKey{Host, Namespace, Symbol, Timeframe}` sized `required + 50`, bar validation, socket/REST merge rules and readiness. It never takes the scheduler state lock; alerts leave through a channel the cycle drains outside `mu`.
- **Transport** (`market_feed_ws.go`): keepalive, read deadline, jittered backoff, resubscribe, and reconnect repair before a key serves again. The goroutine exits on the shutdown contexts.
- **Requirements** (`market_feed_requirements.go`): the key set comes from every in-scope consumer — signal frames, higher-timeframe filter frames, regime timeframe overrides, mid coins, funding needs. Each key stores the highest lookback any consumer asks for and `frameFor(key, required)` slices it per consumer, so legacy row counts are preserved. A generation publishes only once each required key is ready or has failed explicitly.
- **Scheduling** (`market_feed_schedule.go`): epoch-aligned deadlines, so twins on one cadence share an evaluation identifier and a snapshot no matter how long each check takes. A missed deadline is never replayed. Out-of-scope strategies keep last-run scheduling.
- **Snapshot** (`market_feed_snapshot.go`): at most one REST recovery per stale key per cycle, then a frozen deep copy. Later socket updates belong to a later snapshot. The payload is capped at 8 MiB.
- **Dispatch** (`market_feed_dispatch.go`): batch (`v:2` with a `market` object), singleton, changed-input retry, shared-failure retry and Hyperliquid regime-bundle checks all read the same sealed data through `--market-stdin`. A malformed, stale, incomplete or mismatched payload is an explicit error, never a private worker fetch.
- **Parity** (`hyperliquid_candles.go`): the Go bootstrap mirrors the Python adapter's widening loop and row conversion, including the exchange close-timestamp preference. One golden fixture feeds both converters.
- **Outage**: no ready frame yields signal 0, close fraction 0, a price from the freshest verified mid and a `degraded` reason plus one throttled owner alert per key. Entries are held at both in-scope dispatch sites, and a sealed snapshot older than `max(60s, interval/2)` at dispatch also holds them. Trailing stops, ratchets, protection sync, hedge sync and reconciliation keep running on verified inputs. With no verified mark either, the strategy is skipped with a CRITICAL line. `sc.Paused` is never written.
- **Operator surface**: startup logs the effective mode, each cycle logs `[feed] snapshot=<id>/<gen> keys=… rest{bootstrap,repair,recovery}=… steady_candle_rest=0`, and `/status` carries a `market_feed` block.

## Verification

All run on this branch:

- `go -C scheduler build ./...`, `go -C scheduler vet ./...`, `gofmt -l scheduler` (empty)
- `go -C scheduler test ./...` — pass
- `go -C scheduler test -race -run MarketFeed .` — pass (also added as a CI step)
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/ -q -n auto` — 2931 passed
- `uv run --no-sync python -m pytest shared_scripts/ -q` — 282 passed
- `uv run --no-sync python3 scripts/check_test_budget.py` — wording-only tests 171 (baseline 171)
- `./scheduler/trading-scheduler probe --config <websocket config>` — exit 0
- `--once` smoke against the live venue, daemon stopped, both modes:
  - `rest`: `Market feed: rest (legacy polling)`, `Signal: HOLD | BTC @ $79656.50`
  - `websocket`: `Market feed: websocket (Hyperliquid perps + manual)`, `[feed] BTC|1h: bootstrap complete`, `[feed] snapshot=300s/1788592500/1 keys=1 ready=1 stale=0 rest{bootstrap,repair,recovery}=1,0,0 steady_candle_rest=0`, `Signal: HOLD | BTC @ $79656.50`, exit 0. The check argv carried `--market-stdin` and the Python side logged no `Fetching … from Hyperliquid` line.
- `python3 -c "print(len(open('CLAUDE.md','rb').read()))"` — 14995 (under the 15,000-byte cap)

Acceptance criteria covered by tests: `TestDeadlineScheduleContract` (twins, restart, overrun, drawdown cadence change, reload, `rest` parity), `TestMarketFeedHistoryContract` and `TestMarketFeedSocketContract` (empty/short history, gaps, duplicate and out-of-order updates, reconnect during bootstrap, silent stale key on a connected socket, missing mids, invalid values), `TestCandleSnapshotConversionParity` plus `test_candle_snapshot_conversion_parity`, `TestEveryCheckPathConsumesTheSealedSnapshot`, `TestSealedSnapshotRecoversStaleKeysOncePerCycle`, `TestFeedOutageHoldsEntriesAndKeepsProtection`, `test_market_payload_matches_legacy_polling_for_the_same_frozen_rows`, `TestMarketFeedConfigModes` and `TestMarketFeedModeChangeIsRestartRequired`, `TestMarketFeedNeverNestsInsideSchedulerLock` with `TestMarketFeedSourceNeverTouchesTheSchedulerLock`, `TestMarketFeedStopsOnShutdown`, and the probe tests.

## Test edits

- `scheduler/probe_cmd_test.go` `TestRunProbeHappyPath` and `scheduler/version_probe_test.go` `TestProbeRunsExtraArgvForHL` — **Outdated**. Checkable ground: the adopted implementation plan step 10 and the issue's acceptance criterion "Payload compatibility passes both startup probes" authorize the added probe argvs. The replacements keep every previous probe assertion and add the two Hyperliquid market-signal probes and the regime market probe.
- `scheduler/regime_store_test.go` `stubRegimeBundleCheck`, `scheduler/hl_batch_test.go`, `scheduler/manual_test.go`, `scheduler/regime_gate_on_failure_test.go` — mechanical call-site and stub signature updates for new parameters. No assertion changed.

## Adopted plan and deviations

Built to the implementation plan on the issue: https://github.com/richkuo/go-trader/issues/1524#issuecomment-5549968103 (Fable 5.1, xhigh).

Deviations:

1. `check_regime.py` moves its `--probe-only` exit after `parse_args()`. The plan adds a `--market-stdin` regime probe, but the old early exit returned 0 before argparse ever saw the flag, so the probe could not catch a Go/Python argv mismatch — the only thing it exists for.
2. Funding rides new Go clients (`hyperliquid_funding.go`) fetched at seal time with a 5-minute TTL, and Python filters payload records to its own bar-0 start so `funding_skew` sees exactly the record set legacy polling gave it. The plan named the payload fields but not the fetcher.
3. Single-check and regime-bundle payloads ride the same `{"v":2,"market":…}` envelope as the batch rather than a bare market object, so one Python parser covers all three paths.
4. `lastEvaluated` stores a typed mark (identifier, deadline, cadence) instead of a bare identifier string. The "never dispatch a deadline at or before the last consumed one" rule needs the deadline itself, and parsing it back out of the identifier would be fragile.
5. `resolveHyperliquidCheckOutcome` was not extracted. The outage contract is tested at the `runHyperliquidCheck` boundary plus a `pausedBlocksSignal` state matrix (flat and open, long and short, close-only, scale-in), which covers the same states without restructuring the 4,400-line cycle loop.
6. CLAUDE.md sat 2 bytes under its 15,000-byte cap, so wording was compressed across existing bullets (no guardrail dropped) to fit the two new guardrails. Final size 14,995 bytes.

Closes #1524

---
Updated with LLM: Fable 5.1 | high | Harness: Claude Code

https://claude.ai/code/session_01VmhZ6kZoAo9rBgiPZSbYd2


richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Fable 5.1 · 2026-09-05T07:36:52Z) in 3b5e91e9428845f3c8132d5fb0c816674c84b324.

### Fixed
1. **The singleton sealed payload caps the signal frame at 200 rows while the same check's argv asks for `regimeRequiredOhlcvLimit` rows.** `frameSpecs` now reads `entry.SignalLookback`, the same `max(200, regime limit)` the ring stores and the argv names (`scheduler/market_feed_dispatch.go:39`). `TestSingletonPayloadRowCountMatchesTheArgvLimit` covers regime off, period 14 and period 120 (`scheduler/market_feed_dispatch_test.go:280`); batch and regime-bundle payloads are unchanged.
2. **The feed's bar validation rejects the missing-close-timestamp candle shape that the PR's own parity fixture asserts is real.** `feedBarFromRaw` takes the key interval and synthesizes `CloseMs = OpenMs + intervalMs - 1` when `HasClose` is false, so ordering and staleness still work; `feedBar.row()` falls back to the open time exactly as `hlCandleRowFromRaw` does, so the payload row matches legacy polling (`scheduler/market_feed.go:80`, `scheduler/market_feed.go:108`). A close before open and a bar spanning two intervals are still rejected, and a later full update replaces the synthesized bar (`scheduler/market_feed_test.go:168`).
3. **`publishGeneration` has no staleness guard, so an outlived startup `ApplyGeneration` goroutine can clobber a later reload generation.** `ApplyGeneration` takes an `applySeq` token under `feedMu`; `publishGeneration` no-ops and logs when a newer request has already published (`scheduler/market_feed_requirements.go:246`, `scheduler/market_feed_requirements.go:287`). `TestMarketFeedGenerationPublishIsNeverClobberedByAnOlderRequest` blocks the first bootstrap, publishes a second request, then releases the first and asserts the generation counter, published set and key set all stay the newer request's (`scheduler/market_feed_test.go:236`).
4. **A degraded feed evaluation spuriously clears the script-failure streak.** `finishHyperliquidCheck` skips `clearScriptFailure` when the result carries a `Degraded` reason (`scheduler/main.go:3486`). The outage test seeds a streak before the degraded evaluation and asserts it survives (`scheduler/market_feed_dispatch_test.go:329`). A later real successful run still clears it, and the feed-degraded tracker is untouched.
5. **The "six REST consumers" disclosure omits a seventh: the manual-open regime-trail label check.** SKILL.md and the PR body now list `resolveManualRatchetRegimeLabel` as the seventh consumer (`SKILL.md:904`). Documented rather than routed through the feed: the CLI open path (`scheduler/manual.go:153`) runs in a separate process with no feed owner, and the UI path shares the same off-cycle shape as the already-disclosed `--fetch-atr`.

### Test edits
1. **TestEveryCheckPathConsumesTheSealedSnapshot, singleton subtest** (`scheduler/market_feed_dispatch_test.go:200`) — Outdated; ground: Needs Fixing item 1 and the issue's acceptance criterion that legacy row counts are preserved; now asserts the singleton frame's `required` equals the strategy's `SignalLookback`.
2. **TestMarketFeedHistoryContract, "invalid bars are rejected and never stored"** (`scheduler/market_feed_test.go:194`) — Outdated for the "no close timestamp" row only; ground: Recommended Optional item 1's invariant that the feed accepts every candle shape the Python adapter accepts, backed by `hlCandleRowFromRaw` and the shared parity fixture; the row moves to the new subtest "a bar without a close timestamp is accepted like the Python adapter accepts it", which asserts the synthesized close, the open-time row timestamp, REST merge acceptance, and replacement by a later full update.
3. Every other `feedBarFromRaw` call in `scheduler/market_feed_test.go` — mechanical signature update (interval argument added), no assertion changed.

Verification on 3b5e91e9: `go -C scheduler build ./...`, `go -C scheduler vet ./...`, `gofmt -l scheduler` empty, `go -C scheduler test ./...` pass, `go -C scheduler test -race -run 'MarketFeed|Feed|Singleton' .` pass, pytest main suite 2931 passed, `shared_scripts/` 282 passed, `check_test_budget.py` 171 (baseline 171), CLAUDE.md 14995 bytes.

---

richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Claude Opus 5 · 2026-09-05T08:06:39Z) in b168f251f1a4075cf44088e4cead1fb2e47368ca.

### Fixed
1. **The sealed payload drops funding whenever the coin carries no mid, so a funding strategy's whole cycle fails instead of degrading.** `marketPayloadFor` attaches funding in its own loop over the requested coins, independent of `mids` (`scheduler/market_feed_snapshot.go:342`). A missing, failed or incomplete funding entry is now a feed hold: `holdFor` reports it like a missing frame (`scheduler/market_feed_dispatch.go:151`, `scheduler/market_feed_snapshot.go:401`), so the strategy degrades with entries held and protection running, and `batchPayload` refuses a group whose funding is unavailable so members fall back and degrade (`scheduler/market_feed_dispatch.go:96`). Python's hard-fail on a missing funding block stays as the guard against a Go bug; it is no longer reachable from a healthy seal. Covered by `TestSealedPayloadCarriesFundingIndependentlyOfMids` (`scheduler/market_feed_snapshot_test.go:227`) and `TestFundingStrategyDegradesInsteadOfSkippingWhenFundingIsUnavailable` (`scheduler/market_feed_dispatch_test.go:480`).
2. **A key that is repairing after a reconnect is still reported ready, so a frame with a hole in it is served as fresh.** `keyReadiness` returns not-ready with status `repairing` until a repair or recovery fetch succeeds (`scheduler/market_feed.go:366`); the cycle seal then runs its once-per-cycle recovery on it, a failed recovery leaves the frame absent so the strategy degrades, and a successful one serves the key in the same cycle. `TestSealedSnapshotRepairsAReconnectingKeyBeforeServingIt` covers the failed repair, the same-cycle recovery and the reconnect repair with no degraded cycle (`scheduler/market_feed_snapshot_test.go:278`).
3. **The batch path never clears the feed-degraded tracker, so recovery is silent and the next outage is throttled out of alerting.** The `batch.lookup` branch clears the strategy's signal and higher-timeframe keys on a healthy, non-degraded batched result (`scheduler/main.go:3394`); the singleton path uses the same `clearDegradedFor` (`scheduler/main.go:3424`). Degraded alerts are now raised on the key that actually failed (`scheduler/main.go:3415`), so a twin degraded on its higher-timeframe key is not cleared by a twin that is healthy on the signal key. Covered by `TestFeedDegradedAlertsCountOncePerKeyPerCycleAndClearOnEveryPath` (`scheduler/market_feed_dispatch_test.go:434`).
4. **The degraded alert counts dispatches, not cycles, so twins double every count.** `notifyMarketFeedDegraded` records once per key per sealed snapshot id (`scheduler/market_feed_dispatch.go:242`); the same test asserts twins on one key in one snapshot count one cycle.

Verification on b168f251: `go -C scheduler build ./...`, `go -C scheduler vet ./...`, `gofmt -l scheduler` empty, `go -C scheduler test ./...` pass, `go -C scheduler test -race -run 'MarketFeed|Feed|Singleton|Sealed|Funding' .` pass, pytest main suite 2931 passed, `shared_scripts/` 282 passed, `check_test_budget.py` 171 (baseline 171). No test edits.

---

richkuo:
@claude review

PR #1530, opened 2026-09-06
fix(#1529): gate sub-lot partial closes before they reach the venue [C87, Fable 5.1, medium, fableplan]

richkuo:
Closes #1529

## Plain simple English

After a partial take-profit, the exchange fills a rounded quantity. The bot then saw a tiny remainder and sent an order too small for the exchange every cycle, with a failure alert each time. The bot now floors the remainder to the venue lot size and treats a zero or below-minimum order as done. A full close is never blocked.

## Summary

- `platforms/hyperliquid/adapter.py`: `MIN_ORDER_NOTIONAL_USD = 10.0`; `sz_decimals_from_meta_cache(symbol)` reads the meta cache file with no TTL and no fetch; `lot_size_decimals(symbol)` returns the resolved lot size or `None`; `_sz_decimals` keeps its default of 3 and its one-shot refresh.
- `shared_strategies/close/_helpers.py`: `current_close_fraction` returns `0.0` when the remainder is at or below `1e-9` of the initial quantity. No absolute floor; other platform check scripts are untouched.
- `shared_scripts/check_hyperliquid.py`: `apply_venue_close_gate` floors `current_quantity * close_fraction` by lot, and when the floored quantity is zero or its value at the resolved price is below the venue minimum, rewrites the decision to `close_fraction 0.0`, recomputes `signal`, and stamps `close_gate: below_venue_minimum` plus `close_gate_detail`. A `close_fraction >= 1` is never rewritten. `resolve_venue_lot_decimals` uses the adapter when present, else the meta cache on the sealed path; any failure yields `None` and no gate. One stderr line reports a rewrite.
- `scheduler/`: `CloseGate` on `StrategyDecisionFields`; an Info log in `runHyperliquidCheck` when the gate fired; `runHyperliquidExecuteOrder` returns `(nil, false)` on `Signal == 0` before any sizing, so a gated close cannot reach the execute subprocess or the failure-alert throttle.
- Docs: SKILL.md trailing-ratchet paragraph describes the gate; CLAUDE.md `shared_scripts/` bullet carries the guardrail (file at 14998 bytes).

## Verification

- `gofmt -l scheduler` empty; `go -C scheduler build .` passes.
- `go -C scheduler test ./... -count=1`: every test passes except `TestUpdateHelpersEnvfileParsing790`, which fails identically on `origin/main` in this environment (`scripts/test_update_helpers.sh`, untouched by this PR).
[...]

PR #1531, opened 2026-09-06
feat(#1525): verified deployment handoff for merging a paper instance into the live process [C99, Fable 5.1, medium, fableplan]

### LINKED ISSUE #1525: [C99] Moving paper trading into the live process needs a verified deployment handoff
**Complexity: 99/100** - Capability 3 (Risk 4, Uncertainty 3: a wrong storage mapping turns the paper book into an orphan that the next save drops, and live risk config is rewritten in place; the effective-setting source is now named but the fields it must gain need discovery); Volume 24 (Scope 4: deployment script, shell tests, service override, drift script, inspect JSON, startup line, CI job, example config and four docs across bash, Go and Python; Coupling 4: two stopped units, file locks, service sandbox and live/paper parity; Verification 4: restart races, interrupted apply, and a root systemd sandbox fixture) - Opus 5, xhigh · fableplan: yes

## Problem

After #1523 and #1524, a paper instance can share the live scheduler while retaining its existing database. The migration procedure must preserve both storage identity and effective behavior.

At baseline `fff25950`, startup looks up stored state by the configured strategy identifier, initializes absent strategies, and prunes unmatched state. A stored book that maps to no configured strategy in its file is an orphan, and the next full save of that file drops it. Appending a suffix only in config therefore cannot preserve a colliding paper book. Root config also controls inherited risk, cadence, regime, close defaults, and notification behavior. Copying the strategy array alone can change these settings.

The service template uses `ProtectSystem=strict` and grants state-directory access for its own instance and the shared directory. The ownership lock, the manual-action lock and the SQLite write-ahead log sit beside the database file. Pointing the live config at another instance's database therefore requires explicit write access to that database directory.

Existing facilities to extend, never duplicate: the drift audit already pairs through `replay_source_id`; startup already prints the scope counts and both resolved database paths; `go-trader inspect --all --json` already prints each strategy's effective post-migration cadence, leverage, stops, take-profit ladder, regime windows and direction; `go-trader storage-inspect --json` already reports mappings, orphans, positions, pending actions, risk latches and lock holders without a write. The inspect output does not yet carry scope risk, notification routing, replay fields or the ATR method. A config drift check does not prove matching runtime snapshots or matching trades. No shipped example config runs live and paper together, and no continuous-integration job runs shell or systemd tests.

## Goal

Provide a repeatable, recoverable deployment handoff that preserves both databases and effective strategy behavior, grants the required service access, and verifies that only the combined scheduler owns the files.

## Approach

Part of #1526. Depends on the storage identity, read-only inspection, and lock contracts in #1523, and the evaluation contract in #1524.

1. Add `merge-paper-instance.sh` with live and paper instance arguments. Default to dry run; require `--apply` for deployment changes. Validate instance names with the existing update helper, deployment roots, binary compatibility, source configs, paper mode, and resolved database identity. Both deployments must use a compatible release with the early ownership-lock contract from #1523 before cutover. Refuse if either unit is active. Take both database ownership locks and both manual-action locks with the same lock files the scheduler uses, and hold them through validation and apply so a restart or manual write cannot race the handoff. A scheduler that starts while the script holds a lock exits 79, stays down under the unit's restart guard, and pages the owner; the operator docs state this.
2. Give colliding paper process identifiers a deterministic suffix only when the resulting identifier is free. Set `storage_strategy_id` to the original stored identifier using #1523's mapping. Preserve existing explicit mappings. Resolve replay mirrors through `replay_source_id`, keeping it in the process identifier namespace. Reject missing, ambiguous, or duplicate source mappings. Leave stored identifiers and records unchanged.
3. Compare normalized effective configs before and after the proposed move by running each deployment's own binary with `inspect --all --json`, so the comparison uses the scheduler's configuration semantics and never a second implementation of config loading. Extend that JSON with the fields it lacks: the scope risk limits after paper override inheritance, notification routing including the `-paper` channel keys, replay sharing and source, and the ATR method. Preserve each strategy's cadence, sizing, stop and close defaults, average true range settings, regime inputs, direction policy, scope risk, replay behavior, and notification routing. Populate supported paper overrides and explicit per-strategy values as needed. If a shared root setting cannot express both original configurations, refuse with the exact conflict. In particular, zero-valued paper risk overrides currently inherit live values; never claim that they disable a nonzero live limit. Preserve live settings as well as paper settings.
4. Reference the existing paper database through `paper_db_file`. After the script holds the locks, run `storage-inspect --json` without `--require-idle`, because that flag rejects any held lock including the script's own; refuse when a reported lock holder differs from the script's process. Use that report to verify storage mappings, ownership, open positions, pending actions, and risk latches. Read-only database inspection is allowed. The script must never open through the normal migrating writer, copy records, rename stored identifiers, create a replacement database, or invoke a trading cycle as a validation step. `--probe-only` proves script compatibility; it does not prove database ownership or service write access.
5. Include an instance-specific systemd drop-in override that grants only the required paper database directory, using the existing writable-directive helper so a state directory and a deploy-tree path each get the correct directive. The apply operation installs the reviewed config and override; it does not start either service or execute trades. Record prior config, override contents, and relevant unit state for rollback. Print the required service reload, disable, and startup commands in safe order. Validate the resulting unit configuration and access under the service account and sandbox with an isolated fixture.
6. Make apply repeatable and recoverable. Write and validate staged files first, retain the prior deployment files, and replace each file atomically. Track progress across the config and service-override replacements because they do not share one transaction. On failure, restore completed deployment-file changes or report the exact recovery state and keep services stopped. A repeated apply must not append duplicate strategies or add another suffix. Recheck source-config fingerprints and ownership before mutation. Document rollback with both units stopped and both database locks held; rollback restores deployment files while preserving database records written after cutover.
7. Extend the existing drift audit to use explicit replay source mappings and verified twin pairing in one config or across deployments. A suffix alone cannot prove that two strategies are twins. Compare effective cadence and sizing, and report incompatible or ambiguous pairs. Keep static config drift separate from #1524's runtime snapshot comparison. Startup already prints both resolved database paths; add the storage alias mapping needed for inspection to a successful start, which today prints it only on a rejected layout.
8. Add a shipped example config that runs live and paper together with `paper_db_file`, and add a continuous-integration job that runs the shell tests and the Linux service fixture. Update installation and deployment instructions, the steps that retire the paper instance's status port and any tunnel mapping, drift documentation, project guardrails, and post-update documentation. Explain separate database backups and restores, unchanged stored identities, service access, rollout, and rollback. State that a process crash or stuck cycle can stop both modes and that there is no transaction across the two state files.

## Acceptance criteria

- A fixture pair with colliding identifiers retains its existing paper cash, positions, risk latches, pending actions, and history through process aliases. Neither stored identifier changes, no empty replacement book is initialized, and no orphan appears in the inspection report.
- Effective-setting comparison reads the binary's inspect output and preserves both scopes, including inherited defaults and paper risk limits. A zero-inheritance conflict, incompatible regime setting, ambiguous twin, second suffix collision, or invalid storage mapping causes failure before writes.
- Dry run and apply leave both database files and their write-ahead logs byte-identical. The shared-memory index file is excluded, because a read-only open updates its read marks and can create it. These checks end before normal scheduler startup; byte identity is not required after a real trading cycle or an authorized schema migration.
- Active services, ownership-lock contention, manual-write contention, changed source configs, invalid units, failed probes, and failed read-only inspections cause no partial active deployment. A controlled restart race cannot produce two schedulers for the paper file.
- Tests cover interruption between deployment-file replacements, failed override installation, retry, repeated apply, and rollback. Recovery preserves both databases and leaves a clearly identified valid config and service override.
- An isolated Linux service fixture, run in continuous integration, proves write and journal access under the combined unit's sandbox and proves that a second scheduler cannot own either database. Startup and `--once` smoke tests use disposable databases and exchange stubs; they cannot submit live orders.
- Drift checks cover in-process and cross-instance twins, explicit replay mappings, ambiguous pairing, and effective-setting differences. A successful startup prints the scope counts, both resolved database paths, and the storage alias mapping.
- Operator docs cover the service reload and cutover sequence, backups, rollback, the exit-79 page during a handoff, and accepted process and transaction failure boundaries.

## Plain simple English

Moving paper strategies by copying config can hide their existing positions, change their risk settings, or leave the live program unable to write the paper file. The move needs identity checks, setting checks read from the program itself, correct service access, and a recovery procedure that preserves both databases.

---
LLM: Fable 5.1 | high | Harness: validate-issue

richkuo:
Closes #1525

## Plain simple English

A paper instance can now move into the live scheduler with one script. The script checks both deployments, holds both database locks, builds the merged config, proves it with the live program, installs it with a journal, and can roll back. Both state files stay where they are. Nothing trades and nothing writes a database.

## Summary

- `scripts/merge-paper-instance.sh` (dry run by default, `--apply`, `--rollback`). Preflight with distinct exit codes (deployments, binary version, `storage-inspect` compatibility, configs, stopped units, paper-only config, database identity). A python lock holder takes both databases' ownership and manual-action locks for the run and writes its pid the way the scheduler does; contention exits 3 naming the holder pid. Read-only inspection through each deployment's own binary refuses foreign lock holders, rejections and paper orphans. Compose: `-paper`/`-paper2` suffix only when free, `storage_strategy_id` set to the stored id, `replay_source_id` in the process namespace with missing, non-live and already-claimed sources refused, per-strategy `interval_seconds`/`atr_method` stamps, `portfolio_risk.paper` with the zero-inherits refusal, `-paper` channel keys, `paper_db_file`, dropped keys listed, every other differing root key refused. Proof: the live binary's `storage-inspect --json` must map every paper book with no orphan, and its `inspect --all --json` must match each strategy's pre-move effective view. Override drop-in via `update_paper_override_directive`, optional `systemd-analyze verify`, cutover commands printed in order. Apply journals `config` and `override` steps with retained copies and atomic moves; interruption restores and exits 4; a re-run on an interrupted journal restores first; a complete journal with matching files is a no-op; a repeat merge skips already-merged strategies.
- `inspect --all --json` gains `scope`, `storage_strategy_id`, `capital*`, `margin_per_trade_usd`, `risk_per_trade_pct`, `atr_method`, `htf_filter`, `allowed_regimes`, `regime_gate_on_failure`, `hurst_gate_enabled`, `replay`, `notification`, `scope_risk` (with `zero_override_inherits`).
- Startup prints `[storage]   alias <role>: <process id> -> <storage id>` on a successful start (`formatStorageAliasLines`).
- `update_helpers.sh`: `update_canonical_db_path`, `update_state_lock_paths`, `update_start_state_lock_holder`/`update_stop_state_lock_holder`, `update_file_fingerprint`, `update_db_fingerprint`, `update_resolve_config_db_path`, `update_unit_dropin_path`, `update_paper_override_directive`.
- Drift audit: explicit `replay_source_id`, equal id, or `-paper` suffix backed by `storage_strategy_id`; a bare suffix is `AMBIGUOUS` and gates; effective cadence and sizing from a binary beside the config (`RAW` marker otherwise); `INCOMPATIBLE` twins reported and left alone.
- `scheduler/config.live-paper.example.json`; Linux service fixture `scripts/test_merge_paper_service_fixture.sh`; CI jobs `shell` and `service-fixture`; docs in README, SKILL.md (§ Storage Ownership handoff subsection, drift verdicts, port retirement), CLAUDE.md guardrail (14,997 bytes), POST_UPDATE_HISTORY.

## Verification

- `go -C scheduler test ./...`: every test passes except `TestUpdateHelpersEnvfileParsing790`, which fails on this macOS host on the base commit too (the fleet-shim case compares a `/var/folders` path against its `/private/var` canonical form). It runs on Linux CI.
- `TestMergePaperInstance` (72 s): preflight exit codes 10 to 17, ownership and manual-action lock contention (exit 3), dry run with byte-identical databases and `VERDICT: READY`, deploy-tree `ReadWritePaths`, zero-override and regime and channel refusals (exit 21), second suffix, invalid explicit storage mapping (exit 20), apply, repeated apply no-op, repeat dry run without duplicates, rollback with and without a prior drop-in, interruption after each step, unwritable unit dir, resume of an interrupted journal.
- `scripts/test_update_helpers.sh` with `GO_TRADER_BIN`: helper cases, storage-alias twins, ambiguity, incompatible marker, effective cadence and `RAW` fallback, all existing drift verdicts unchanged.
- `gofmt -l` clean; `scripts/check_test_budget.py` at baseline; `wc -c CLAUDE.md` 14997.
- Not verified locally: `scripts/test_merge_paper_service_fixture.sh` needs Linux, root and systemd; the `service-fixture` CI job runs it.

## Test edits

- `TestConfigExampleCopyDoesNotTriggerTheUpgradePath`: Outdated. Ground: plan step 13 extends it to load the new example. Now loops over both example files and asserts the split layout and the paper alias.
- `TestUpdateShellScriptSyntax`: Outdated. Ground: plan step 10 adds the new scripts. The list now includes the drift, merge, merge-test and service-fixture scripts.

## Plan

Adopted the maintainer's `## Implementation plan` comment on #1525 (Fable 5.1 authored, 2026-09-06). Deviations:

1. Step 1: `inspect --all` was unreachable as documented (the flag parser rejected `--all`) and `--json` mixed loader and state log lines into stdout. Both fixed; the plan's verify command needed them.
2. Step 6: paper channel routing is preserved by synthesizing the `<platform>-paper` key from the paper deployment's effective channel when it only had a bare platform key, so a paper strategy never inherits the live channel. Also added `platforms`, `risk_free_rate`, `alert_throttle_interval`, `kill_switch_reset_dm_timeout` and `telegram` to the refuse-on-difference set, and refuse any unlisted differing root key.
3. Step 7: the diff drops `notification.*_key` fields (channel ids still compare) and `scope_risk.zero_override_inherits` (the loader's default warn threshold shows as inherited in the merged view while every effective value compares equal).
4. Step 9 and acceptance: a zero-length `-wal` created by a read-only open fingerprints like an absent one, since it holds no frames.
5. Step 11: `INCOMPATIBLE` is reported and the pair is left alone as `SKIP`; it does not gate, so the existing `paper2` drift case keeps its verdict as the plan requires.
6. Step 12: the shell CI job runs `test_update_helpers.sh` with a built binary and the merge tests through `go test`, because the fixture databases are written with the Go state writer. The fixture asserts both `.db` files are written; SQLite removes the `-wal` on a clean close.
7. Preflight sources each deployment's `.env` before running its binary, because `LoadConfig` requires the Hyperliquid secret for a live strategy.

---
Created with LLM: Fable 5.1 | medium | Harness: Claude Code

https://claude.ai/code/session_01C4kbNLS9QGfJgp5Nx6zKUB


richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Fable 5.1 review · 2026-09-06T09:16:57Z) in 30bf1673ac603dc66bb9fad47c2287fa5ea1b456.

### Fixed
1. **The lock holder creates root-owned lock files that later block the merged service's start and its manual actions.** — The holder records whether each lock file existed before `os.open`; a file it created is `fchown`ed to the database's owner, or to its directory's owner when the database is absent, and a pre-existing file keeps its ownership (`scripts/update_helpers.sh:275-298`). The service fixture no longer pre-creates and chowns the paper locks; it deletes them, starts the holder on both databases and asserts all four lock files are owned by the service user before the exit-79 start (`scripts/test_merge_paper_service_fixture.sh:103-112`).
2. **`update_canonical_db_path` resolves symlinks the scheduler does not, so the handoff's locks can miss the scheduler's lock files.** — The helper resolves symlinks only when `os.stat` succeeds on the absolute path, so a missing component or a dangling symlink keeps the unresolved absolute path exactly as `canonicalDBPath` does (`scripts/update_helpers.sh:207-218`). `TestUpdateCanonicalDBPathMatchesScheduler` runs the shell helper against `canonicalDBPath` for an absent file behind a symlinked parent, a present file and a dangling symlink (`scheduler/merge_paper_instance_test.go:85-129`).
3. **The preflight probe reports a rejected storage layout as "binary cannot inspect its storage layout" with exit 12.** — The probe ignores the exit code and requires only a report carrying `layout`; a report with rejections passes through to the later `storage-check`. A config the binary refuses to load exits 20 with the loader errors shown; exit 12 is left for no report or no `layout` (`scripts/merge-paper-instance.sh:668-679`). Test: a split-layout paper config that `storage-inspect` rejects with a full report passes the probe and reaches the classify refusal (`scripts/test_merge_paper_instance.sh`, "preflight passes a rejected layout through"). No single-file layout can produce a rejection through config alone (identity collisions are loader validation errors), so the rejection-under-exit-20 path is covered by the existing `badmap` and `storage rejection:` handling at `scripts/merge-paper-instance.sh:251`, `:864-870`.
4. **A dry run ignores an interrupted apply journal and reports READY against the half-applied config.** — The journal gate runs in every mode. A dry run refuses an interrupted journal with exit 24 and never restores; a complete journal whose files match is reported and the dry run continues; a complete journal whose files changed refuses with exit 24 as apply does; a rolled-back journal is reported and left in place (`scripts/merge-paper-instance.sh:823-855`). Tests cover the three journal states (`scripts/test_merge_paper_instance.sh`, "dry run reports the journal state").
5. **Rollback and resume overwrite the current live config without checking it against the journal's recorded result, keeping no copy.** — The journal now records `staged_config`, `staged_override` and `override_prior_fp` fingerprints beside `result_*`; every restore path compares the live config and the drop-in against the original, staged and result fingerprints and copies a file that matches none to `<file>.merge-edited.<timestamp>` before overwriting it, naming the copy; an archive failure is CRITICAL and stops the restore (`scripts/merge-paper-instance.sh:735-760`, `:958-966`). Archive was chosen over refuse so the failure-path restore inside `--apply` can never strand the deployment half-applied. Tests: rollback after a hand edit to both files keeps both copies and still restores; a clean rollback archives nothing (`scripts/test_merge_paper_instance.sh`, "rollback keeps a copy of a hand-edited merged config").

### Corrected scope (partial)
1. **Composing a paper risk block into a live config that has no `portfolio_risk` always fails the script's own proof step.** — The proof failure is real, but the cited mechanism is not: `LoadConfig` replaces a nil `PortfolioRisk` with `{MaxDrawdownPct: 25}` and a zero `WarnThresholdPct` with 60 (`scheduler/config.go:1153-1158`), so the pre-move live view is `{25, 60, ...}`, never `null`. The staged `portfolio_risk: {"paper": …}` therefore loaded with `max_drawdown_pct: 0`, which failed the proof and, had it passed, would have removed the live scope's default drawdown limit. Fixed: compose reads both deployments' effective `scope_risk` from their own `inspect --all --json` output, writes the effective live limits explicitly into the root when the live block has no nonzero root field, and derives the paper override from every field whose effective paper value differs from the live one; the zero-inherits refusal and the existing-override comparison use effective values (`scripts/merge-paper-instance.sh:283-302`, `:409-452`). The reviewer's shape (2), `portfolio_risk: {}`, is rejected by config validation (`portfolio_risk.max_drawdown_pct must be in (0, 100]`) before any inspection, so it exits 20 with the loader error. Tests: live without a block (root gains 25 and 60, paper override kept), live with `{}` (exit 20), paper without a block (effective paper `daily_max_loss_usd` of zero refuses against a live 500; equal effective limits need no override) (`scripts/test_merge_paper_instance.sh`, three "portfolio_risk block" cases).

### Test edits
1. **canonical db path case** (`scripts/test_update_helpers.sh:508-520`) — Outdated; ground: review finding 3 on this PR and `canonicalDBPath` (`scheduler/singleton_lock.go:36-45`), which keeps the unresolved absolute path when a component is missing; now asserts the unresolved path and lock paths for an absent file behind a symlinked parent, then the resolved path once the file exists.
2. **service fixture second start** (`scripts/test_merge_paper_service_fixture.sh:103-112`) — Outdated; ground: review finding 1 on this PR; the pre-create and chown workaround is removed and the fixture now asserts the holder creates all four lock files owned by the service user.

Verification: `go -C scheduler test ./...` (every test passes except `TestUpdateHelpersEnvfileParsing790`, failing on this macOS host on the base commit too; it runs on Linux CI); `TestMergePaperInstance` with the new cases; the merge-helper section of `scripts/test_update_helpers.sh` in isolation (the full script stops earlier on the same macOS-only fleet-shim case); `gofmt -l` clean; `scripts/check_test_budget.py` at baseline. Not run locally: `scripts/test_merge_paper_service_fixture.sh` needs Linux, root and systemd; the `service-fixture` CI job runs it.

---

richkuo:

richkuo:
Addressed review feedback (github-actions[bot], Opus 5 review · 2026-09-06T09:57:27Z) in c1bd96a32bce27f3dd67bc71e89bac96e238a78b.

### Fixed
1. **The preflight runs each deployment's binary against the deployment's own `config.json`, so `LoadConfig` migrates that file in place — as root, leaving it root-owned `0600` and unreadable by the service user.** — Root cause fixed in Go: `MigrateConfig` is split into `migrateConfigData` (pure) plus the write (`scheduler/config_migration.go:154-170`), `loadConfig` takes a `readOnly` flag that applies a pending migration in memory (`scheduler/config.go:867-893`), and `loadConfigQuietForJSON` uses `LoadConfigReadOnly` for `inspect` and `storage-inspect` (`scheduler/storage_inspect_cmd.go:62-70`). `TestLoadConfigReadOnlyNeverRewritesTheFile` covers a v18 file and a v19 file with a legacy key behind a symlink: bytes, symlink and mtime unchanged, no `.tmp`, migrated field present in memory (`scheduler/config_load_readonly_test.go:9`). Defense in the script: `require_units_stopped` runs before any binary (`scripts/merge-paper-instance.sh:678`); both binaries run against copies under the work directory (`:679-682`); `config_copy_intact` fingerprints each copy after every run and exits 18 if an older binary rewrote it (`:686-692`, `:912-913`); the staged config is fingerprinted across the proof (`:923`); and two configs with different `config_version`s refuse with exit 18 (`:719`). Rule 1: the probe call site is this PR's. Tests: v15 live and paper configs reach READY byte-identical with no `.tmp` and the staged config keeps the live version; mixed versions exit 18 (`scripts/test_merge_paper_instance.sh:564`).
2. **The drift audit now executes each deployment's binary, so auditing a running fleet rewrites deployment configs and can replace the transition symlink with a regular file.** — Same Go fix: `inspect` loads read-only, so the audit never writes. The effective-view fixture now points the paper deployment's `scheduler/config.json` symlink at a v15 file and asserts the audit still yields an effective view (no `RAW`), the symlink survives, the bytes are unchanged and no `.tmp` appears (`scripts/test_update_helpers.sh:868-887`). Rule 1: `effective_view` is this PR's code.
3. **Compose only compares root keys that exist in the paper config, so a root key set only in the live config silently applies to the moved paper strategies and the proof step does not see it.** — Both sweeps compare in both directions: the `REFUSE_ON_DIFFERENCE` loop triggers when the key is in either config, and the catch-all iterates the union; a `DROPPED` key present only in the live config is simply kept. An absent `market_feed` compares as `rest`, matching `marketFeedMode` (`scheduler/config.go:145-153`), so an explicit `rest` beside an absent key does not refuse (`scripts/merge-paper-instance.sh:495-513`). Tests: live-only `market_feed: websocket` and a live-only `telegram` block both exit 21 naming the key (`scripts/test_merge_paper_instance.sh:592`).
4. **A paper strategy that has never persisted a book makes the handoff refuse with a message that reads like a storage fault.** — The count check refuses only when the file maps more books than the config carries; fewer is reported as `N configured strategies without a stored book yet` and the merge proceeds; orphans still refuse (`scripts/merge-paper-instance.sh:262-264`, `:281-283`). Test: a second paper strategy with no row reaches READY, is reported and is moved (`scripts/test_merge_paper_instance.sh:616`).

### Resolved merge conflicts
1. `CLAUDE.md` — origin/main (#1530) rewrote the same guardrail lines this PR had compressed; main's wording was kept throughout, the `merge-paper-instance.sh` guardrail line was re-added and now names `LoadConfigReadOnly`, and other lines were trimmed for size only (14,996 bytes). Hand-resolved set: `CLAUDE.md`; `SKILL.md` and `scheduler/main.go` auto-merged. Findings were addressed in this cycle, so the routing below is the normal step-down.

### Test edits
1. **effective-view drift fixture** (`scripts/test_update_helpers.sh:868-887`) — extended, no assertion removed; ground: review finding 2 on this PR; now also asserts the audit is read-only for a v15 config behind the transition symlink with a binary beside it.

Verification: `go -C scheduler test ./...` (every test passes except `TestUpdateHelpersEnvfileParsing790`, which fails on this macOS host on the base commit too and runs on Linux CI); `TestMergePaperInstance` with the six new cases; the merge-helper and drift section of `scripts/test_update_helpers.sh` in isolation with a built binary; `TestLoadConfigReadOnlyNeverRewritesTheFile`; `gofmt -l` clean; `scripts/check_test_budget.py` at baseline; `wc -c CLAUDE.md` 14996. Not run locally: the service fixture (Linux, root, systemd; the `service-fixture` CI job runs it).

---

richkuo:
@claude review

Issue #1532, opened 2026-09-06
[C72] A final full close below the venue minimum on a shared coin is rejected every cycle


richkuo:
**Complexity: 72/100** — Capability 3 (Risk 4, Uncertainty 2 — live reduce-only full close on a shared coin; the venue rule for full-close exemption is not documented); Volume 12 (Scope 2, Coupling 2, Verification 2) — Opus 5, xhigh · fableplan: yes

## Problem

The venue close gate from #1530 never rewrites a full close (`close_fraction >= 1`), as issue #1529 acceptance criterion 3 requires. When another live Hyperliquid strategy shares the coin, `shouldCloseFullPosition` (`scheduler/main.go:3569-3574`) downgrades the full close to a sized reduce-only order for the whole remaining quantity, sent through `market_open` in `shared_scripts/check_hyperliquid.py` `run_execute`. When that remaining quantity is worth less than the Hyperliquid minimum order value of $10 (after venue-floored tier fills, or after an adverse price move), the venue rejects the sized order every cycle. The execute failure alert repeats under the throttle in `scheduler/failure_alerts.go`, and the strategy can never exit the position.

The single-owner path is different: `market_close(symbol, sz=None)` closes the whole on-chain position. Hyperliquid documents the $10 minimum in its API error responses ("Order must have minimum value of $10.") with no documented exemption for reduce-only or full-close orders, so whether a whole-position `market_close` below $10 is accepted is not confirmed from documentation and must be checked against the venue.

A noop rewrite is not a valid remedy: it would strand the position with no exit.

## Goal

A below-minimum position whose coin is shared with a peer strategy still reaches a flat state or a clear operator decision, with no repeating reject-and-alert loop.

## Approach

1. Confirm on the venue (testnet or a documented probe) whether a reduce-only order that closes the whole on-chain position below $10 is accepted, and whether a sized reduce-only order below $10 is rejected. Record the result in `SKILL.md`.
2. In the shared-coin sized-close branch, detect a remaining quantity valued below the venue minimum (reuse the gate threshold from `apply_venue_close_gate` and `resolve_venue_lot_decimals`).
[...]

PR #1536, opened 2026-09-07
fix(#1534): escalate or refuse an operator shared-coin close below the venue minimum [C87, Opus 5, high, fableplan]

### LINKED ISSUE #1534: [C87] A manual or force close of a small shared-coin position is rejected by the venue with no hold or escalation
**Complexity: 87/100** — Capability 3 (Risk 4, Uncertainty 3 — the change decides which live reduce-only order reaches the venue on an operator command; a wrong escalation takes a peer's exposure, and the escalate-or-refuse policy below is settled by validation, not yet by a prior spec); Volume 12 (Scope 2, Coupling 2, Verification 2) — Fable 5.1, high · fableplan: yes

## Problem

The shared-coin full-close floor from issue 1532 is wired only into the automatic strategy cycle. The manual close dispatch and the force-close path still send the sized reduce-only close for a shared-coin position whose remainder is worth less than the venue minimum plus the 3% gate margin. The venue rejects that order, and the operator sees the rejection with no hold, no escalation to a whole-position close, and no stranded-remainder alert.

Both paths compute the close mode with the same peer count helper the automatic cycle uses, so any position on a coin with two or more live Hyperliquid perps strategies takes the sized order. The manual command-line path also runs with no notifier, so the only output is the process error text.

The stranded-remainder alert text currently tells the operator that the bot's force-close sends the same sized order the venue rejects and cites this issue. That copy becomes stale when this issue lands.

## Goal

An operator full close of a small shared-coin position gets one deliberate outcome: a whole-position close when every peer is provably flat on-chain and in its own book, or a loud refusal that names the peer exposure. It never sends the sized order the venue rejects.

## Approach

Policy, decided: a hand close is never held. When the remainder is below the gate on a shared coin, the command escalates to the whole-position close under the same peer-flat proof the automatic path uses (peer virtual quantity zero in book, and the refetched on-chain quantity on the coin no larger than the operator's own position on the same side). In every other case the command refuses with an explicit message and sends no order: a peer holds quantity, the on-chain account cannot be read, or the refetch fails. A hand command cannot defer to the next cycle, so an unreadable account refuses rather than defers.

A stored hold reason on the position (peer busy or venue rejected, stamped by the automatic cycle) does not block the hand close. The hand close re-proves the peers from fresh state and decides on that proof alone.

Reuse: extract the pure floor decision so the automatic apply step and both operator cores call the same function, and share the pre-escalation account refetch. The operator cores need three inputs the manual dependency set does not carry today: a positions fetcher for the configured account address (the same fetcher and environment variable the scheduler cycle uses), the peer virtual quantity on the coin taken from the full state of every strategy (the same snapshot helper the cycle uses), and a mark price for the remainder from the existing mid-price fetcher. Add these to the shared manual dependency set and populate them in both constructors: the command-line one and the daemon one behind the local status server. Test fakes cover them.

Apply the decision in both cores before the venue call: the manual close core switches its execute call to the whole-position mode on escalate, and the force-close core drops the partial size on escalate. On refusal, both return the operator error and, when a notifier is present, send the stranded-remainder alert with the refusal reason.

Rewrite the stranded-remainder alert recovery text so it no longer says that force-close sends the same sized order, and no longer cites this issue.

Keep the kill-switch and circuit-breaker gates on manual open, add, and stop-loss edits unchanged; close paths carry none today and gain none.

Future: a partial force-close with an explicit quantity below the venue minimum has no gate on either operator path. The check-script gate covers the automatic cycle only. Out of scope here.

## Acceptance criteria

1. A manual or force close of a shared-coin position below the gate, with every peer flat on-chain (refetched) and in book, escalates to the whole-position close and never sends the sized order.
2. The same close with a peer holding quantity, in book or on-chain, refuses with a message that names the peer's book quantity and the on-chain quantity, and sends no order.
3. The same close when the account refetch fails or returns no readable positions refuses with a message and sends no order.
4. A stored hold reason on the position does not change the outcome of items 1 to 3.
5. A manual close of a single-owner position, or of a shared-coin position at or above the gate, sends the same order it sends today.
6. When a notifier is present, a refusal sends the stranded-remainder alert once with the refusal reason; the command-line path with no notifier still prints the full reason.
7. The stranded-remainder alert recovery text no longer states that force-close sends the sized order and no longer cites this issue.
8. Tests cover items 1 to 5 for both cores with fake execute, fake closer, and fake positions fetcher, without spawning Python.

## Plain simple English

When an operator closes a small position by hand on a coin that two strategies share, the exchange refuses the order because it is too small. The bot does not help: no larger close, no clear message. The operator must get one clear result each time: a full close when it is safe, or a clear refusal.

---
Validated with LLM: Fable 5.1 | low | Harness: Claude Code

richkuo:
## Plain simple English

When an operator closes a small position by hand on a coin that two live strategies share, the exchange used to refuse the order because it was too small, and the bot said nothing useful. Now the command gives one clear result each time: a full close when every other strategy is proved empty, or a clear refusal that sends no order.

## Summary

Closes #1534

The shared-coin full-close floor from #1532 was wired only into the automatic strategy cycle. `manualCloseCore` and `forceCloseCore` still sent the sized reduce-only close for a shared-coin position whose closing value sits below the venue minimum plus the 3% margin. The venue rejected that order and the operator got the raw rejection: no escalation, no refusal, no stranded-remainder alert. The command-line path runs with no notifier, so its only output was the process error text.

Both operator cores now take one deliberate outcome before the venue call:

- **Escalate** to the whole-position close only when every peer is flat in its own book and on the refetched on-chain account, on the same side and no larger than the operator's own position: the same two-input proof `applySharedCoinFullCloseFloor` requires. `manualCloseCore` switches `closeFullPosition` to true; `forceCloseCore` leaves `partialSz` nil.
- **Refuse** in every other case, sending no order: a peer holds quantity in book or on-chain, the mark is unreadable, the account is unreadable, or the pre-escalation refetch fails. A hand command cannot defer to the next cycle, so an unreadable account fails closed.

A hand close never holds and never defers, and a stored `SharedCloseHoldReason` (`peer_busy` or `venue_rejected`, stamped by the cycle) is never an input to the decision: the operator core passes `heldReason=""` and re-proves the peers from fresh state. A refusal writes no hold to the position, so the cycle stays the sole owner of the stored hold reason.

**Reuse.** The refetch step of the automatic floor is extracted into `resolveSharedCloseFloorEscalation`, so the cycle and both operator paths share one resolution and one refetch discipline. `applySharedCoinFullCloseFloor` calls it and keeps its existing log lines and result mutations unchanged. `decideOperatorSharedCloseFloor` is the pure operator decision and reads the on-chain account only once the coin is shared and the value is below the gate, so an above-gate or single-owner close makes no venue call at all.

**Plumbing.** `manualCoreDeps` gains `fetchPositions` (a wrapper over `fetchHyperliquidStateFn`, the same fetcher and `HYPERLIQUID_ACCOUNT_ADDRESS` the cycle uses), and `manualStateView` gains `PeerVirtualQty` from `hlPeerVirtualQtyOnCoin(snapshotHyperliquidVirtualQuantities(...))`. Both operator constructors already route through `manualStateViewFromStateWithStore`, so no per-constructor edit was needed for the peer quantity. The fetch runs outside the state lock, as the existing `fetchMids` call does.

**Alerts.** A refusal with a notifier present sends the stranded-remainder alert once under a new `operator_refused` hold label whose recovery text says the hand close was refused, no order was sent, and the scheduler's own hold is unchanged. The command-line path with no notifier still returns the full reason. The two existing recovery strings no longer state that force-close sends the same sized order the venue rejects, and no longer cite this issue.

Force-close `--dry-run` previews the same decision without alerting: it reports `full market_close (escalated: peers flat)` on an escalation, or the refusal and its reason, so the preview matches what the real command would do.

Out of scope, as the issue states: a partial force-close with an explicit quantity below the venue minimum stays ungated on both operator paths.

## Verification

- `go -C scheduler build .`
- `go -C scheduler vet ./...`
- `gofmt -l scheduler` (empty)
- `go -C scheduler test ./...` — one failure, `TestUpdateHelpersEnvfileParsing790`. It fails identically on an untouched `main` checkout at `6b7612a8` and exercises `scripts/test_update_helpers.sh` against `scripts/update.sh`, which this change does not touch. Every other Go test passes.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2943 passed
- `uv run --no-sync python scripts/check_test_budget.py`

The new operator-core regression test was run against the unfixed cores first, before the wiring landed, and failed on every escalate and refuse case (`whole-position close = false, want true`; `err = <nil>, want a refusal`).

## Tests

- `TestDecideOperatorSharedCloseFloor` covers escalate, peer in book, peer on-chain, a peer going busy between the read and the refetch, refetch failure, first-read failure, unreadable account, unreadable mark, no account reader, single owner, and value at or above the gate. It counts on-chain reads, so the above-gate and single-owner cases assert that no venue call is made.
- `TestOperatorSharedCloseFloorGatesBothCores` drives `manualCloseCore` and `forceCloseCore` with a fake execute, a fake closer and a fake positions fetcher, asserting the whole-position close on escalate, zero venue calls on every refusal, the unchanged order for a single-owner coin and for a shared coin above the gate, and the same escalation under a stored `venue_rejected` hold. No Python is spawned.
- `TestSharedCloseStrandedAlertRecoveryTextDropsForceCloseClaim` guards the rewritten recovery text for all three hold reasons.

**Test edits: none.** No existing test was rewritten, weakened or removed. The new recovery-text test asserts operator-decision wording, so it is listed in `scripts/test_budget_baseline.json`; that is the only entry added.

## Plan

Built to the implementation plan on the issue: https://github.com/richkuo/go-trader/issues/1534#issuecomment-5563792249 (Fable 5.1, low, `issueplan`).

Deviations:

1. **The pure decision owns the on-chain read instead of taking a pre-fetched view.** The plan passed `onChain` and `refetch` separately into `decideOperatorSharedCloseFloor`. Taking one `fetchOnChain` closure lets the decision check the shared-coin test, the mark and the gate before it reads the account, so an above-gate or single-owner close makes no network call and existing tests with no positions fetcher keep their current behavior. Reason: correctness of the unchanged paths, plus no needless venue reads.
2. **Force-close `--dry-run` reports the decision.** The plan named only the escalated mode string. A dry run that previewed a sized order the real command would refuse would mislead the operator, so the refusal is previewed too, without an alert and without an error.
3. **The alert hold constant is named `hlSharedCloseHoldOperatorRef`** rather than `hlSharedCloseHoldOperatorRefused`, to keep the const block aligned. Its value is the `operator_refused` string the plan specifies.
4. **Plan step 7 found nothing to update.** No test pinned the old recovery text, matching the plan's own note.

---

https://claude.ai/code/session_0139cnaqbwT7DypSgVJZ3kvz


richkuo:
@cladue fable review effort:high

richkuo:

richkuo:
Addressed review feedback (Claude/Fable 5.1 · 2026-09-07T02:19:45Z) in 95ea5e8cf38d02a8caf8d805207aca06de79a6bd.

### Fixed

1. **`manual-close --dry-run` previews a sized order the real command would refuse or escalate.** — The dry-run branch now computes `intentFullClose`, `shouldCloseFullPosition` and the floor decision before it prints, and reports the same three outcomes force-close previews: `full market_close (escalated: peers flat)`, `REFUSED — <reason>`, or the sized close, plus the unreadable-mark warning (`scheduler/manual_core.go:983-1024`). A single-owner coin, a value above the gate and a partial close print the sized preview unchanged. Scope rule 1: this PR added the floor block to the real path only, creating the divergence.

2. **The refusal alert states a $0.00 value and a gate breach that were never measured.** — Fixed at its source and structurally. The only path that reached the alert with an unmeasured remainder was the mark-unreadable refusal, which no longer refuses (see the judgment call below), so `RemainderUSD` is now measured on every refuse path. `formatSharedCloseStrandedAlert` also no longer renders a value it was not given: a non-positive remainder prints `the closing value of the final full close could not be measured against the $10.30 gate` in place of `worth $0.00, under the $10.30 gate` (`scheduler/hyperliquid_shared_close_floor.go:211-219`), so the CRITICAL alert cannot assert an unmeasured value whatever the caller passes. Scope rule 1: this PR added the operator refusal alert path.

1. **An unreadable mark refuses a shared-coin full hand close at any position size.** — Implemented the reviewer's Option B, and confirmed the finding first: `decideOperatorSharedCloseFloor` refused before the gate compare and `operatorSharedCloseFloorDecision` turns any mids failure into price 0, so a mids outage blocked every shared-coin hand close at any size. The decision now sets a new `MarkUnreadable` flag, withholds only the escalation, and returns neither refuse nor escalate (`scheduler/hyperliquid_shared_close_floor.go:151-180`); both cores send the sized reduce-only close and print the reason as a warning (`scheduler/manual_core.go:1097-1098`, `scheduler/manual_core.go:1257-1262`). No on-chain account read happens on this path.

   Grounds. A sized reduce-only close is peer-safe at any size: it can only reduce this strategy's own quantity, so the refusal bought no protection against the hazard issue #1534 names, while costing availability on an operator emergency command. The venue's own minimum still rejects a truly small one, which is the pre-PR behavior. Issue #1534's Approach lists the validated refusals as a peer holding quantity, an unreadable on-chain account, and a failed refetch; an unreadable mark is not among them, and acceptance criterion 5 keeps an above-gate close unchanged, which an unmeasurable value cannot rule out.

   Alternatives rejected: keep the refusal (fails criterion 5 whenever the position is in fact above the gate, and blocks kill-switch flattening by hand during a mids outage, for no peer-safety gain); refuse only below some assumed value (there is no measured value to compare, so the assumption is the same fabrication finding 2 reports).

### Test edits

1. **`TestDecideOperatorSharedCloseFloor` / "unreadable mark refuses without reading the account"** (`scheduler/hyperliquid_shared_close_floor_test.go:426-427`) — Outdated; ground: issue #1534's Approach names the validated refusals and does not include an unreadable mark, and its acceptance criterion 5 keeps an above-gate close unchanged; the review's Requires Human Review item 1 put the decision. Now asserts `MarkUnreadable` with no refusal and no escalation, zero on-chain reads, and a reason naming the withheld escalation and the unchanged sized close. A second case covers a large position under a NaN mark.

No other test was rewritten, weakened or removed.

### New coverage

- `TestOperatorSharedCloseFloorGatesBothCores` gains an unreadable-mark case: both cores still send the sized order, and the on-chain account is never read.
- `TestManualCloseDryRunPreviewsSharedCloseDecision` covers the four manual-close dry-run previews (escalate, refuse, above the gate, unreadable mark) and asserts zero venue calls.
- `TestSharedCloseStrandedAlertRecoveryTextDropsForceCloseClaim` now asserts the measured rendering keeps the value and the unmeasured rendering states no value.

### Verification

`go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler` (empty), `go -C scheduler test ./...` — only `TestUpdateHelpersEnvfileParsing790` fails, identically on an untouched `main` at `6b7612a8`; it exercises `scripts/test_update_helpers.sh`, which this PR does not touch. `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2943 passed. `scripts/check_test_budget.py` — 172 of 172, unchanged.

## Plain simple English

The bot refused a hand close when it could not read the price. This stopped large closes also. Now it only stops the full-position close and sends the usual order with a warning. The test mode of manual-close shows the same result as the real command. The critical alert no longer gives a value that was not measured.

---

https://claude.ai/code/session_016Yia7pXCtgopDEbcEqUPj9

richkuo:

richkuo:
Addressed review feedback (Claude/Opus 5 · 2026-09-07T03:28:33Z) in 07d7d70b87538e88c836616a73971098db40f43c.

### Fixed

1. **An escalated `manual-close` books the full book quantity as realized, even when the whole-position close fills less.** — Confirmed and fixed. `manualCloseCore` now reads `Execution.Fill.TotalSz` and books that quantity: it attributes only the virtual quantity when the fill exceeds the book and pro-rates the fee, warns when the fill is under the requested close, and derives `IsFullClose` from the filled quantity instead of the intent (`scheduler/manual_core.go:1143-1156`, `scheduler/manual_core.go:1163-1186`). The zero-fill and already-flat cases were already an error before the accounting, because `confirmHyperliquidExecuteFill` rejects a non-positive `TotalSz` (`scheduler/executor.go:490-492`) and `_extract_execute_fill` returns an error for a `market_close` with no filled status (`shared_scripts/check_hyperliquid.py:836-850`); the new test asserts no trade is queued on that path. A position that survives a short fill now also clears the protection order ids the close already cancelled, which the error path already did and the success path never needed while it always deleted the position (`scheduler/manual_core.go:1157-1159`). Scope rule 1: this PR added the escalation that sets `closeFullPosition = true` on a shared coin (`scheduler/manual_core.go:1112-1115`), routing shared-coin closes into the unclamped accounting; `hlPeerFlatOnCoin` proves peers flat by accepting an on-chain quantity at or below the book quantity (`scheduler/hyperliquid_shared_close_floor.go:78-88`), so the escalation gate is exactly the state where the two can differ.

2. **A refused `manual-close` has already cancelled the strategy's resting limit orders, but the refusal and its CRITICAL alert both say "no order sent".** — Confirmed and fixed. Both operator refusal reasons and the `operator_refused` recovery text now end in "no close order sent", which is accurate on every refusal (`scheduler/hyperliquid_shared_close_floor.go:164`, `scheduler/hyperliquid_shared_close_floor.go:206`, `scheduler/hyperliquid_shared_close_floor.go:230`). `clearRestingLimitRemainderForPositionAction` returns the order ids it cancelled on the venue (`scheduler/manual_core.go:368-440`), and the manual-close refusal appends them with the statement that they are not restored, into both the error and the CRITICAL alert body (`scheduler/manual_core.go:1106-1109`, `scheduler/hyperliquid_shared_close_floor.go:213-223`). The decision stays after the cancel on purpose: the cancel step can reconcile a stale position snapshot to the true size (`scheduler/manual_core.go:1039-1049`), so deciding before it would gate on a quantity the venue has already moved past. A refusal with no resting order reads exactly as it did. Scope rule 1: this PR added the operator refusal that follows the cancel step.

### Deferred to follow-up

1. **A rejected operator close leaves the position with no exchange-side stop, and no operator path re-arms it.** — Reviewer-routed to `### Create Follow-up Issue`; filed as #1538. Confirmed against the code first: `run_execute` cancels the trigger orders before it places the order (`shared_scripts/check_hyperliquid.py:1351-1378`), `reconcileManualExecuteProtection` only records the succeeded cancels (`scheduler/manual_core.go:265-276`), and nothing re-arms, unlike `rearmProtectionAfterFailedClose` (`scheduler/hyperliquid_shared_close_floor.go:311-330`). No existing issue covers it.

### Test edits

1. **`TestDecideOperatorSharedCloseFloor`, the four `"no order sent"` reason assertions** (`scheduler/hyperliquid_shared_close_floor_test.go:411-420`) — Outdated; ground: the review's `### Recommended Optional` item 1 states the invariant that a refusal claiming no venue action must leave the venue untouched or name the action it took, which required the wording change. Each case now asserts the replacement wording `"no close order sent"`; every other assertion in those cases, including the peer quantities and the fetch-error text, is unchanged.

2. **`TestManualCloseDryRunPreviewsSharedCloseDecision` / "a peer holding quantity previews the refusal"** (`scheduler/manual_test.go:2082`) and **`TestOperatorSharedCloseFloorGatesBothCores` / "a peer holding on-chain refuses"** (`scheduler/manual_test.go:2185`) — Outdated; same ground. Both now assert `"no close order sent"`; the `REFUSED` marker, the venue-call counts and every other assertion are unchanged.

No other test was rewritten, weakened, skipped or removed.

### New coverage

- `TestManualCloseBooksTheVenueFillNotTheBookQuantity` drives an escalated shared-coin close through four fills — under the book, equal to the book, above the book, and none — and asserts the queued quantity, fee, realized profit or loss and `IsFullClose`, plus that a zero fill queues no trade at all.
- `TestManualCloseRefusalNamesCancelledRestingLimitOrders` asserts a refusal names the resting order it cancelled and that a refusal with no resting order names none, with no close order reaching the venue in either case.

### Verification

`go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler` (empty), `go -C scheduler test ./...` — only `TestUpdateHelpersEnvfileParsing790` fails, identically on an untouched `main` at `9a35c5d8`; it exercises `scripts/test_update_helpers.sh`, which this PR does not touch. `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2943 passed. `scripts/check_test_budget.py` — 172 of 172, unchanged. All four GitHub Actions checks passed on the prior head.

## Plain simple English

A hand close on a shared coin can now close the whole position. The exchange closes only the quantity it really holds, which can be less than the bot's own book. The bot recorded the book quantity, so the profit and the trade size could be wrong and the position was deleted too early. It now records what the exchange filled. A refused hand close also says which resting orders it already cancelled.

---

https://claude.ai/code/session_01GhmMRpRYGhMTdydWZ7ZLu1

richkuo:
@claude review

PR #1537, opened 2026-09-07
feat(#1535): install the shipped systemd unit and reload systemd before restart [C58, Opus 5, high]

### LINKED ISSUE #1535: [C58] update.sh installs the shipped systemd unit and reloads systemd before restart
**Complexity: 58/100** — Capability 60 (Risk 55, Uncertainty 40 — a wrong unit install can leave a live deployment unable to restart, though rollback and the existing `.prev` pattern bound it); Volume 45 (Scope 40, Coupling 45, Verification 50) — Opus 5 / high · fableplan: no

## Problem

`scripts/update.sh --restart` swaps the binary and Python tree, then restarts the unit with `sudo systemctl restart` (`scripts/update.sh:863`, and again in `do_rollback` at `scripts/update.sh:466`). It never compares the repo's unit file against the installed one and never runs `systemctl daemon-reload`. The only path that installs a unit and reloads systemd is `scripts/install-service.sh:63-80`, which an operator runs by hand.

PR #1527 changed both shipped units to `RestartPreventExitStatus=78 79 80` (`go-trader.service:14`, `systemd/go-trader@.service:14`). A deployment updated with `update.sh` alone still runs under the old unit, so a storage-ownership refusal (exit 80) is restarted by systemd in a loop instead of staying down. The same gap applies to every future unit-file change: `ProtectSystem`, `ReadWritePaths`, `EnvironmentFile`, `ExecStart`.

## Goal

After `scripts/update.sh --restart` in systemd mode, the installed unit file matches the unit shipped at the deployed SHA and systemd has reloaded it, so the restart runs under the new unit. A deployment whose unit already matches sees no change. The update never touches operator drop-ins.

## Approach

Correctness and safety outrank diff size.

1. Pure decision helpers in `scripts/update_helpers.sh`, tested in `scripts/test_update_helpers.sh` with no systemd:
   - `update_unit_source_path <repo-root> <unit-name>`: `go-trader@<i>.service` maps to `systemd/go-trader@.service`, a plain unit maps to `go-trader.service`; any other name yields empty (skip with a note, never guess).
   - `update_unit_sync_decision <installed-path> <source-path> <needs-reload>`: returns `install`, `reload`, or `none`. `install` when the source exists and `cmp -s` differs or the installed file is missing; `reload` when the files match but `NeedDaemonReload=yes`; `none` otherwise.
2. In `scripts/update.sh`, a `begin_phase unit` step between `swap` and `restart`, systemd mode only:
   - Resolve the installed path with `systemctl show -p FragmentPath --value "$service_unit"` and `NeedDaemonReload`. An empty `FragmentPath` (unit not found) skips the step; the existing signal fallback then applies.
   - On `install`: `sudo cp` the installed unit to `<path>.prev` (same retention pattern as `go-trader.prev`), `sudo install -m 0644` the source, then `sudo systemctl daemon-reload`. Log it (`[update] unit: installed <path> from <source>`).
   - On `reload`: `sudo systemctl daemon-reload` only.
   - `do_rollback` restores `<path>.prev` when the unit phase installed it and runs `daemon-reload` before its own restart, so a rollback runs the old binary under the old unit.
   - Never write to `<unit>.d/` drop-ins; the merge handoff (`scripts/merge-paper-instance.sh:95`) owns those.
3. `--all` runs the step per instance with that instance's resolved unit.
4. Refuse, before the swap, when `FragmentPath` resolves outside `/etc/systemd/system` (a vendor or generator unit is not ours to overwrite).
5. Docs: `SKILL.md` deploy section and `docs/POST_UPDATE_HISTORY.md` entry state that `update.sh --restart` now installs the shipped unit and reloads systemd; `install-service.sh` stays the first-install path.

## Acceptance criteria

- `scripts/test_update_helpers.sh` covers the two helpers: template and plain name mapping, unknown name empty, `install` on differing or missing installed file, `reload` on match plus `NeedDaemonReload=yes`, `none` on match.
- On a host whose installed unit lacks exit 80, `update.sh --restart` installs the unit, runs `daemon-reload`, and `systemctl show -p RestartPreventExitStatus` reports `78 79 80` after the restart.
- On a host whose unit already matches and needs no reload, the update log shows no unit install and no `daemon-reload`.
- A rollback triggered after a unit install restores the previous unit file and reloads before restarting.
- Drop-in files under `<unit>.d/` are byte-identical before and after an update.
- A `FragmentPath` outside `/etc/systemd/system` stops the update before the swap with a clear message.
- `bash -n scripts/update.sh` and the CI `shell` job pass.

## Plain simple English

The update script replaces the program but does not replace the service file that tells systemd how to run it. So a new rule in that file, such as "do not restart after a storage error", does not apply until someone reloads systemd by hand. This change makes the update script install the new service file and reload systemd itself.

---
Created with LLM: Fable 5.1 | low | Harness: new-issue

richkuo:
## Plain simple English

The update script replaced the program but not the service file that tells systemd how to run it. So the rule "do not restart after a storage error" never reached a machine that only ran the update script. The update script now installs the new service file and reloads systemd itself, keeps the old file, and puts it back if the update fails.

## Summary

Closes #1535

`scripts/update.sh --restart` swapped the binary and restarted the unit, but never compared the loaded unit file against the one this repo ships and never ran `daemon-reload`. A deployment updated only with `update.sh` therefore kept the unit installed at first install, so the `RestartPreventExitStatus=78 79 80` guard added by #1527 never reached it: a storage-ownership refusal (exit 80) was crash-looped by systemd instead of staying down. The same gap applied to every future unit change (`ProtectSystem`, `ReadWritePaths`, `EnvironmentFile`, `ExecStart`).

A new `unit` phase runs in systemd mode between the binary swap and the restart:

- It resolves the loaded fragment with `systemctl show -p FragmentPath` and maps the unit name to the shipped source: `go-trader.service` for a plain unit, `systemd/go-trader@.service` for `go-trader@<instance>.service`.
- On a difference it keeps the loaded file as `<unit>.prev`, installs the shipped file at mode 0644, and runs `daemon-reload` before the restart.
- On a match that systemd has not reloaded (`NeedDaemonReload=yes`) it runs `daemon-reload` only. On a match with no reload pending it does nothing.
- `do_rollback` restores `<unit>.prev` and reloads before it restarts the previous binary, so a failed update lands on the old binary under the old unit. The restore runs before the `go-trader.prev` check, so a unit rollback still happens when there is no binary to restore.
- A `daemon-reload` that fails right after an install restores the previous unit before the run fails, so disk and systemd never disagree.

Guards:

- Drop-ins under `<unit>.d/` are never written. `scripts/merge-paper-instance.sh` still owns those.
- A unit name this repo does not ship (a hand-written `go-trader-live.service`, say) is logged and skipped, never guessed.
- A fragment that is a symlink whose target differs from the shipped unit is left to the operator instead of being flattened into a regular file.
- A fragment outside `/etc/systemd/system` (vendor or generator) stops the update before the swap, with the path in the message.
- `--all` inherits the phase once per instance, because each child already resolves its own unit.
- An empty `FragmentPath` (unit not loaded) skips the phase, so the existing signal fallback is unchanged and `update.sh` never creates a unit that was not already there. `scripts/install-service.sh` stays the first-install path.

New pure helpers in `scripts/update_helpers.sh`: `update_unit_source_path`, `update_unit_fragment_scope`, `update_unit_sync_decision`, `update_unit_install_with_backup`, `update_unit_restore_backup`. The install and restore helpers run their commands through `update_unit_sudo`, so the tests exercise the real file moves without root.

### Verification

- `bash -n scripts/update.sh scripts/update_helpers.sh scripts/test_update_helpers.sh` — pass.
- `GO_TRADER_BIN=<built binary> bash scripts/test_update_helpers.sh` — `OK: update_helpers tests passed`. Note: the pre-existing `--all` fleet fixture compares `mktemp -d` paths against a child's `pwd`, which differ on macOS through the `/var` symlink; that case fails on this machine before and after this change, and passes once the fixture path is physical. CI runs it on Linux, where it already passes.
- `go build ./...` from `scheduler/` — pass.
- `go test -count=1 -run 'TestMergePaperInstance|TestUpdateShellScriptSyntax' .` — `ok`.
- `python3 scripts/check_test_budget.py` — `wording-only tests: 171 (baseline 171)`.
- Red-then-green on the symlink guard: with the symlink branch removed from `update_unit_sync_decision` the new assertion fails with `got=install want=skip`; with the guard it passes.

### Tests

`scripts/test_update_helpers.sh` gains, with no test edited or removed:

- Name mapping: template and plain units, with and without the `.service` suffix, an unshipped name, a traversal instance name, a non-service unit, empty input, and a trailing slash on the repo root. Two assertions pin that the two mapped paths are files this repo actually ships.
- Fragment scope: `/etc/systemd/system`, a vendor path, a generator path, a drop-in path, a relative path, and empty.
- The decision matrix: `install` on a differing or missing installed unit, `reload` on a match plus `NeedDaemonReload=yes`, `none` on a match, `none` with no shipped source or no fragment path, `skip` on a differing symlink and on a dangling symlink, and `reload` on a symlink already carrying the shipped bytes.
- A real install-then-restore round trip: the shipped bytes land at mode 0644, the replaced unit is kept as `.prev`, a rollback puts it back and consumes the backup, a missing backup refuses instead of clobbering the unit, no `.prev` is invented when there was no installed unit, a missing shipped source refuses and leaves the unit as it was, and an operator drop-in beside the unit is byte-identical throughout.

### Notes on the issue's approach

- The issue's step 2 puts the `FragmentPath` scope refusal "before the swap"; it runs in preflight, before the pull, which is earlier and cheaper. The refusal applies only when the unit name maps to a shipped source: for a unit this repo does not ship we already skip and never write, so a custom-unit deployment keeps updating.
- The symlink case is not in the issue. A fragment symlinked by the operator would have been replaced by a regular file, so `update_unit_sync_decision` gained a `skip` outcome for it.
- `CLAUDE.md` is unchanged. It sits 4 bytes under its documented 15,000-byte ceiling, and fitting the new deploy guardrail there means rewording unrelated guardrail lines. The mechanism is documented in `SKILL.md` § Auto-Update (Unit sync) and § Build, deploy, and test mechanics, plus a `docs/POST_UPDATE_HISTORY.md` runtime-default entry, per `CLAUDE.md`'s own pointer to SKILL.md for mechanism. Say the word and I will spend the byte budget.

### Operator note

The first `update.sh --restart` after this lands installs the shipped unit on any deployment whose unit drifted. An operator who edited a shipped unit in place should move those edits into a `<unit>.d/` drop-in first; otherwise the update replaces the file and keeps the old one as `<unit>.prev`.

---

https://claude.ai/code/session_014zZY4Ta2THBegnQe7LFNt7


richkuo:
@claude review

PR #1539, opened 2026-09-07
fix(#1538): re-arm the exchange-side stop after a rejected hand close [C91, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1538: [C91] A rejected hand close leaves the position with no stop-loss on the exchange
**Complexity: 91/100** — Capability 3 (Risk 4, Uncertainty 2 — the exchange-side stop-loss is an auto-protective mechanism and the operator path has no re-arm); Volume 16 (Scope 2, Coupling 3, Verification 3) — Opus 5, xhigh · fableplan: yes

## Problem

`manual-close` cancels the exchange-side stop-loss and take-profit trigger orders **before** it sends the close order. `run_execute` in `shared_scripts/check_hyperliquid.py` runs the `cancel_trigger_order` loop first, then calls `market_close` or `market_open`. When the venue then rejects the close order — a value under the Hyperliquid minimum, an insufficient-margin rejection, a transient venue error — `run_execute` exits 1 with the protection already gone from the exchange.

`manualCloseCore` handles that error by calling `reconcileManualExecuteProtection`, which only records the succeeded cancels in the strategy book and returns the error. Nothing re-arms the stop-loss. The position stays open with no exchange-side protection until the next scheduler cycle runs its own protection sync, and an operator running the command by hand on a stopped scheduler gets no cycle at all.

The cycle path already solves this: `rearmProtectionAfterFailedClose` re-arms the protection sync, the trailing stop and the scalar stop after a failed close. `force-close` is unaffected because it cancels after the fill. Only the operator `manual-close` path has the gap.

This was found while reviewing the operator shared-coin close floor work in PR #1536. It is pre-existing for every rejected manual close, and the fix needs an operator-side re-arm path (state handle, lock, liquidation maps) that PR #1536 does not carry, so it is tracked here.

## Goal

A rejected `manual-close` order leaves the position with the same exchange-side protection it had before the command ran, or tells the operator plainly that the protection is gone and how to restore it.

## Approach

Give the operator path the same re-arm the cycle path has, without widening the manual command surface further than it needs.

1. Extend `manualCoreDeps` with a re-arm dependency alongside `reconcileCanceledProtection`, wired in both construction sites: the command-line dependencies and the dashboard trade-action dependencies. The dashboard site already holds the state pointer and the lock, so it can call `rearmProtectionAfterFailedClose` directly; the command-line site loads the state through the store and must supply the liquidation maps and the notifier the re-arm needs.
2. Call it from `manualCloseCore` on the close-order error path, after the existing `reconcileManualExecuteProtection` call, and only when the cancel actually succeeded for at least one order id — a cancel that failed left the trigger live and re-arming would duplicate it.
3. Respect the documented lock order and never re-arm while holding a lock a subprocess call needs. The re-arm reads the on-chain account, so it belongs outside the state write lock.
4. When the re-arm itself fails or no re-arm path is configured, print a CRITICAL warning naming the position, the cancelled order ids and the fact that the exchange-side stop is gone, and send it to the notifier. A silent failure is worse than the current behavior.
5. Leave `manual-add`, `force-close` and the cycle path unchanged.

## Acceptance criteria

1. A `manual-close` whose close order is rejected by the venue, with the stop-loss cancel confirmed succeeded, re-arms the exchange-side stop-loss before the command returns, and the command still exits with an error.
2. A `manual-close` whose close order is rejected and whose stop-loss cancel failed does not re-arm, and says so, because the original trigger is still live on the exchange.
3. A `manual-close` that fills normally re-arms nothing and books the fill exactly as it does today.
4. A re-arm that itself fails produces a CRITICAL alert naming the symbol and the cancelled order ids, and the command exits with an error.
5. `force-close` and the scheduler cycle path behave exactly as they do today.
6. A Go test covers criteria 1, 2 and 4 without spawning Python.

## Plain simple English

Manual-close first removes the stop-loss from the exchange, then sends the close order. If the exchange refuses the close order, the position stays open with no stop-loss on the exchange. Only the next scheduler cycle puts it back, and an operator with a stopped scheduler gets no cycle. Put the stop-loss back immediately, or warn the operator clearly.

---

richkuo:
## Plain simple English

A hand close first removes the stop-loss from the exchange, then sends the close order. If the exchange refused the close order, the position stayed open with no stop-loss on the exchange until the next scheduler cycle. Now the command puts the recorded stop back at once, or tells the operator loudly that it could not.

## Summary

Closes #1538

`manual-close` cancels the on-chain stop-loss and take-profit triggers before it sends the close order (`run_execute` in `shared_scripts/check_hyperliquid.py`). A venue rejection therefore left the position open with no exchange-side stop. The cycle path already re-arms after a failed close; the operator path did not, and an operator running the command against a stopped scheduler got no cycle at all.

The operator path now restores the recorded trigger itself:

- `manualCloseCore` snapshots the stop OID, trigger, side and quantity from the refreshed position before `d.execute`, then calls `restoreManualStopLossAfterFailedClose` on the close-order error path, after the existing `reconcileManualExecuteProtection` call. The command still returns the same failure, so the exit code stays non-zero (criterion 1).
- `decideManualCloseRearm` (new, pure, table-tested) is a four-way decision: **no cancel requested** (a partial close, or no stop armed), **cancel not confirmed** (the parsed result names no succeeded OID, which in `run_execute` means either the cancel loop never ran or it explicitly failed, so the original trigger is still live and re-arming would duplicate it), **stop cancelled**, and **outcome unknown** (the subprocess returned no result at all). Only the last two re-arm (criteria 1 and 2).
- The restore reads the account outside every lock, caps the size with `hlSLEffectiveQty`, one-way tightens through `clampStopInsideLiquidation`, and places through `d.updateSL` with the previous OID as the cancel target. `check_hyperliquid.py --update-stop-loss` verifies that OID on-chain before it cancels, so a stop that is still resting is replaced rather than duplicated, and one that already filled places nothing.
- Any outcome that leaves the position unprotected prints a CRITICAL line naming the symbol, the cancelled order ids and the `manual-update-sl` recovery command, and sends the same text to all channels and the owner DM (criterion 4). `runManualClose` now builds a notifier, as `manual-open` already does, so the alert reaches Discord with the scheduler stopped.
- `manualCoreDeps.recordRearmedStopLoss` is wired at both construction sites. The command-line site writes the restored OID and trigger into the book (so `inspect` is truthful with the daemon stopped) and queues an `update-sl` action so a running daemon adopts it at the next drain; an immediate fill queues a `close` action instead, which drains ahead of the on-chain reconcile so nothing double-books. The dashboard site applies the result to `ss.state` under `ss.mu` via `applyTrailingStopUpdateResult`, matching the cycle path.

`manual-add`, `force-close` and the scheduler cycle path are unchanged (criteria 3 and 5). Every new test is Go-only and spawns no Python (criterion 6).

`CLAUDE.md` gained the `manual*.go` guardrail. Two wording-only compressions kept it under the 15,000-byte limit and dropped no guardrail: `exact-match certs` on the regime line is still stated on the later regime line, and the test-budget clause `none asserting only wording, a constant or a round-trip (operator-decision wording aside)` is still stated by the preceding Testing line plus the `check_test_budget.py` clause that follows it.

## Verification

- `go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler/` all clean.
- `go -C scheduler test ./...` passes except `TestUpdateHelpersEnvfileParsing790`, which fails identically on `origin/main` (`scripts/test_update_helpers.sh`, unrelated to this diff).
- Regression proof: with the re-arm body short-circuited, every new row and the rewritten daemon row fail; with it in place they all pass.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` 2943 passed; `shared_scripts/test_*.py` 292 passed; `scripts/ .github/scripts/` 51 passed.
- `python3 scripts/check_test_budget.py`: wording-only tests 172, baseline 172.
- `wc -c CLAUDE.md` = 14992.

## Test edits

- `TestDaemonManualCloseRejectsUnconfirmedWithoutQueueing` rewritten as `TestDaemonManualCloseRejectsUnconfirmedAndRestoresTheStop` (`scheduler/hyperliquid_execute_confirmation_test.go`). **Case: Outdated. Ground:** issue #1538 acceptance criterion 1 replaces the behaviour it asserted (a rejected daemon close left `StopLossOID == 0`). The replacement asserts the rejected close still returns 409 and queues no action, that `updateSL` is called with the cancelled OID 111, the recorded $1900 trigger and the 0.4 size, and that the daemon book keeps the quantity and holds the restored OID 222 at $1900.
- `stubTradeDeps` gained a `fetchPositions` stub that fails by default, so no dashboard test can reach the network.

## New tests

- `TestManualCloseRearmDecision`: the pure decision helper over all four outcomes, including a parsed result that names no cancel outcome.
- `TestManualCloseRestoresTheStopAfterAVenueRejection`: `manualCloseCore` end to end through `newCLIManualCoreDeps` with stubbed `execute`, `updateSL`, `fetchPositions`, `fetchMids` and a mock notifier backend. Rows: confirmed cancel restores the trigger at the on-chain-capped size and queues `update-sl`; an unconfirmed cancel calls no placement and says the trigger is still live; a failed re-arm alerts with the symbol and the cancelled order ids; an immediate fill queues a `close` at the trigger; a liquidation price inside the trigger tightens the re-arm; an unreadable execute outcome still re-arms against the previous order id.

## Plan

Adopted plan: https://github.com/richkuo/go-trader/issues/1538#issuecomment-5565929748 (Fable 5.1, 2026-09-07).

Deviations, both grounded in the code:

1. **Operator-visible text goes on `res.outf`, not `res.errf`.** The plan put the CRITICAL and the hold message on stderr. `ui_trade_actions.go` builds the dashboard error body from `res.uiMessage()`, which keeps only non-stderr lines, so a stderr CRITICAL would never reach a dashboard operator. The issue's Goal requires telling the operator plainly, so these lines use `outf` and reach both surfaces. Diagnostic warnings (account read failure, size cap, clamp, subprocess stderr) stay on `errf`.
2. **A failed `recordRearmedStopLoss` raises its own CRITICAL alert** instead of the plan's plain warning. A successful on-chain placement whose book write fails leaves the book and the venue disagreeing about the stop-loss, which the operator has to see.

---
https://claude.ai/code/session_019VZzXHnSTkNJakzqR9nbHT


richkuo:

richkuo:
Addressed review feedback (claude-fable-5 · 2026-09-07T08:26:07Z) in 0498def7e440f9b5ba45f4a61aa31de752dbb101.

### Fixed

1. **The cancel-not-confirmed branch asserts the stop is still resting without verifying it, and re-arms nothing when the cancel actually took effect on-chain.** — Confirmed at the cited grounding: `run_execute`'s per-OID `except Exception` and its explicit-rejection path both append to `cancel_failed_oids` (`shared_scripts/check_hyperliquid.py:1364-1368`), so an unconfirmed cancel is never proof the trigger survived. The branch no longer returns: it prints that the previous stop is being verified rather than assumed live, then falls through to the same `d.updateSL` restore the outcome-unknown case uses (`scheduler/manual_close_rearm.go:83-88`). `run_update_stop_loss` reads the open orders first, so a stop still resting is cancelled and replaced with no duplicate, one already gone is placed fresh, and one that filled places nothing — the reviewer's three must-survive cases. The alert is now outcome-specific instead of always claiming the position is unprotected: a venue that reports the previous stop still resting says so (`scheduler/manual_close_rearm.go:164`), an unreadable open-order check says the state is unverified (`:167`), and only a cancelled stop with no replacement says UNPROTECTED (`:169-171`). CLAUDE.md line 35 reworded to match.

2. **The daemon-path re-arm is not serialized with cycle-side trailing stop replacements.** — Scope rule 1: this PR adds the new in-daemon `--update-stop-loss` call site, so the race it opens is PR-caused. The placement now runs under `lockHyperliquidTrailingUpdate(snap.Symbol)` (`scheduler/manual_close_rearm.go:129-133`), the same per-symbol lock taken at `scheduler/hyperliquid_trailing_stop.go:550`, `scheduler/hyperliquid_liquidation_guard.go:637` and `:737`. The lock is released before `d.recordRearmedStopLoss` takes the state lock, matching the order the cycle sites use (`rearmScalarStopAfterFailedClose` releases the trailing lock inside `hlLiquidationClampReplace` before its own `mu.Lock()`), so no inversion is introduced.

3. **A successful CLI re-arm blocks the operator's retry of the close until a scheduler drain runs.** — Scope rule 1: only this PR queues an `update-sl` whose values the book already holds. `pendingSLActionExists` now skips such an action (`scheduler/manual_sl.go:55-61`, `:63-80`) and takes the position at all three call sites (`scheduler/manual_core.go:1091`, `:1267`, `:1567`), so both the full-close gate and the SL-edit gate open once the book matches. The reviewer's three must-survive cases hold by construction: a genuine `manual-update-sl` never writes the book, so its queued action never matches; a `cancel-sl` is never treated as adopted; and an unreadable placement queues OID 0, which `slActionAlreadyInBook` rejects, so it still gates.

### Deferred to follow-up

1. **Cancelled take-profit triggers are never restored after a rejected close, and the success message does not say so.** — Reviewer-routed to `### Create Follow-up Issue`; filed as #1540, which covers both the restore and the interim one-line notice on the success path.

### Test edits

1. **TestManualCloseRestoresTheStopAfterAVenueRejection / "an unconfirmed cancel leaves the live trigger alone"** (`scheduler/manual_close_rearm_test.go:120`) — Outdated; ground: review finding 1 on this PR (`Needs Fixing`) replaces the behaviour it asserted; renamed "an unconfirmed cancel is verified on-chain and restored, never assumed live" and now asserts `updateSL` is called with the previous OID and the recorded trigger, the book holds the new OID, an `update-sl` action is queued, and the operator output says the stop was verified rather than assumed live.
2. **TestManualCloseRestoresTheStopAfterAVenueRejection / "a failed re-arm alerts with the symbol and the cancelled order ids"** (`scheduler/manual_close_rearm_test.go:169`) — Outdated; ground: the same finding requires the alert to state only verified on-chain state; the expected phrase moved from "UNPROTECTED on-chain" to "UNVERIFIED", with the cancelled-order-ids assertion unchanged.
3. **TestManualCloseRearmDecision — the `rearms()` assertion** (`scheduler/manual_close_rearm_test.go:59-62`) — Obsolete; ground: finding 1 deletes the covered behaviour, since the not-confirmed decision now re-arms and the predicate has no remaining caller; the surviving rows still assert the decision value the helper returns for every case, including two renamed for accuracy.
4. **TestPendingSLActionExists** (`scheduler/manual_sl_test.go:145`) — Outdated; ground: finding 3's fix changes the function signature; the three existing calls pass `nil`, which preserves every prior assertion verbatim, and new rows assert that an adopted `update-sl` does not gate while a drifted OID, a drifted trigger, an empty book and a `cancel-sl` all still do.

New tests: a row for a venue that still reports the old stop resting (no duplicate placed, the alert says the position is most likely still protected), and `TestManualCloseRearmWaitsForThePerSymbolStopLock`, which holds the per-symbol lock and asserts the re-arm places nothing until it is released.

Verification: `go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler/` clean; `go -C scheduler test ./...` passes except `TestUpdateHelpersEnvfileParsing790`, which fails identically on `origin/main` and is unrelated to this diff; `pytest` over `shared_scripts/test_*.py`, `scripts/` and `.github/scripts/` 343 passed; `check_test_budget.py` reports 172 of a 172 baseline; `wc -c CLAUDE.md` = 14993.

---

richkuo:

richkuo:
Addressed review feedback (Claude Opus 5 · 2026-09-07T09:03:36Z) in 2b895251585c3c06bb78fba639787f920aae7050.

### Fixed

1. **The re-arm places a stop without confirming the position still exists on-chain, in the one state where it most likely does not.** — Confirmed end to end: `hlSLEffectiveQty` returns the recorded quantity uncapped for an absent coin (`scheduler/hyperliquid_trailing_stop.go:54-59`), `hlLiquidationPxForSide` returns 0 so no clamp applies, and `run_update_stop_loss` leaves `should_place` true for a cancel OID that is gone and unfilled. The read is now used as evidence, not only as a size cap: when it succeeds and `hlPositionGoneForSide` finds the coin absent or net on the other side, nothing is placed and the operator is told the close most likely filled and the reconciler will book it (`scheduler/manual_close_rearm.go:114-121`, `:180-188`). This is the same test the liquidation guard applies with the same two maps (`scheduler/hyperliquid_liquidation_guard.go:586-589`). Must-survive (3) holds: a failed read still places at the recorded size with a warning and no clamp, since a rejected close usually means the position is open.

2. **A re-armed stop that fills immediately books the close at gross PnL with no fee, so cash is permanently overstated.** — Confirmed: the drain sets `RealizedPnL = a.RealizedPnL + a.FillFee`, `PnLGross`, `FeeSource=userfills` and `ss.Cash += a.RealizedPnL` (`scheduler/manual.go:519-542`) while the daemon path deducts a modeled fee (`scheduler/portfolio.go:282-290`), and the empty `ExchangeOrderID` makes the row `missing_oid` to the backfill (`scheduler/backfill_trade_ledger.go:249-256`). Of the two offered remedies the deferral is taken: the CLI path queues no close and leaves the booking to the on-chain reconciler, which uses the venue's real fill and fee rather than the trigger price and a model (`scheduler/manual_close_rearm.go:154-156`, `:196-206`). It matches the sibling `StopLossFilledExternally` case in the same function and needs no change to the shared drain path, which hardcodes `FeeSource=userfills` for every operator close — setting a modeled fee there would have mislabelled it, failing the finding's own must-survive (2). With no queued close, must-survive (1) becomes "cash is untouched until the reconciler books the venue's numbers" and (3) holds because no row is written for the backfill to double-count.

3. **After an immediate fill on the CLI path, a leftover position keeps a stop-loss order id that no longer exists.** — Confirmed, including the consequence: `armed := pos.StopLossTriggerPx > 0` makes a residue with a stale trigger skip the guard's re-arm branch (`scheduler/hyperliquid_liquidation_guard.go:557-562`). `rearmedStopLossBookValues` now mirrors `applyTrailingStopUpdateResult`'s case set (`scheduler/manual_close_rearm.go:196-212`): an immediate fill, an external fill, or a succeeded cancel with nothing resting clears the book's stop fields and queues a `cancel-sl` so a running daemon clears its in-memory copy too; a resting replacement and an unreadable outcome keep their existing writes; a cancel that failed with the old order still resting changes nothing. The clear is guarded on the book still pointing at the order the command saw removed (`:223-225`), which satisfies must-survive (3). `slActionAlreadyInBook` now treats a `cancel-sl` as adopted when the book is already clear (`scheduler/manual_sl.go:55-67`), so the clear does not reintroduce the retry lockout fixed last cycle.

4. **The new CLAUDE.md guardrail forbids the direct book write this same PR adds.** — Confirmed. `CLAUDE.md:35` now reads "SL edits queue `PendingManualAction`, never a bare book write; only the re-arm also writes it", so `manual-update-sl` and `manual-cancel-sl` stay forbidden while the write-then-queue pair is named as the exception. Three wording-only compressions kept the file at 14,996 bytes and dropped no guardrail: the Go build command and "from repo root" are both already stated on the Patterns line that owns them, and "Review format SSoT" loses only the redundant "SSoT".

5. **The daemon re-arm writes its stop-loss log lines to stdout instead of the strategy log.** — Confirmed at `scheduler/ui_trade_actions.go:99` and `scheduler/logger.go:38-46`. It now uses `cfg.LogDir` through a new `strategyLoggerOrStdout` helper (`scheduler/ui_trade_actions.go:99-100`, `:341-348`) that falls back to a stdout logger on an empty directory or an open failure, so must-survive (1) and (2) hold — a logging problem never aborts the book write — and the file handle is closed on return.

### Test edits

1. **TestManualCloseRestoresTheStopAfterAVenueRejection / "a re-arm that fills immediately queues the close at the trigger"** (`scheduler/manual_close_rearm_test.go:144`) — Outdated; ground: review finding 2 on this PR (`Needs Fixing`) replaces the behaviour it asserted; renamed "a re-arm that fills immediately books no close and leaves it to the reconciler" and now asserts no `close` action is queued, a `cancel-sl` is queued instead, and the operator is told the reconciler books the close with its venue fill and fee.

New rows in the same test: an immediate fill after an unconfirmed cancel clears the stale order id and queues `cancel-sl` (finding 3, must-survive 1); a venue reporting no open position places nothing (finding 1, must-survive 1); a venue reporting the coin net short places nothing for a long book (finding 1, must-survive 2).

Verification: `go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler/` clean; `go -C scheduler test ./...` passes except `TestUpdateHelpersEnvfileParsing790`, which fails identically on `origin/main` and is unrelated to this diff; `pytest` over `shared_scripts/test_*.py`, `scripts/` and `.github/scripts/` 343 passed; `check_test_budget.py` reports 172 of a 172 baseline; `wc -c CLAUDE.md` = 14996.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Claude Opus 4.8 (1M) · 2026-09-07T09:38:33Z) in 24a723f7bea8e147a292aac98216a1d124bc8093.

Growth check: diff 798 lines vs 637 at first push (1.25x); cycle 4.

### Fixed
1. **The `hyperliquid_protection.go` guardrail dropped the `strategyUsesTieredTPATRClose` gate on on-chain TP, undisclosed among the PR's listed compressions.** — Confirmed. `origin/main` carried "on-chain TP only when `strategyUsesTieredTPATRClose` AND live (paper never)" and this PR's compression left only the live gate, while `hyperliquidPlacesOnChainTPs` still returns `strategyUsesTieredTPATRClose(sc)` after the live check (`scheduler/hyperliquid_protection.go:587-595`) and `strategyTPTiersForRegime` returns nil for a non-tiered strategy (`scheduler/hyperliquid_protection.go:88-91`). `CLAUDE.md:37` now reads "on-chain TP needs `strategyUsesTieredTPATRClose` AND live (paper never)". The byte room came from `CLAUDE.md:16`, which repeated the same symbol: it keeps its distinct guardrail as "On-chain TP gate is NOT `len(tiers)>0`" and no longer restates the symbol. All three must-survive conditions hold: the file is 14,991 bytes (under 15,000), the `live` / `paper never` gate is still stated, and the tiered gate is named on the line that owns the mechanism.

No test edits this cycle. Verification: `go -C scheduler build ./...`, `go -C scheduler vet ./...` and `gofmt -l scheduler` clean; `go -C scheduler test ./...` passes except `TestUpdateHelpersEnvfileParsing790`, which fails identically on `origin/main` on this machine and passes in the CI `go` job; test budget 172/172. The Python suites were not re-run: this push changes only `CLAUDE.md`.

---

richkuo:
@claude sonnet review

PR #1541, opened 2026-09-07
fix(#1540): restore cancelled take-profit tiers after a rejected hand close [C97, Opus 5, xhigh, fableplan]

### LINKED ISSUE #1540: [C97] A rejected manual close leaves take-profit orders missing without immediate recovery
**Complexity: 97/100** - Capability 3 (Risk 4, Uncertainty 3 - live exit orders and persistent recovery state; the recovery and adoption contract needs a plan); Volume 22 (Scope 3, Coupling 4, Verification 4 - operator, dashboard, queue and protection changes with their tests; process coordination; concurrent recovery proof) - Opus 5, xhigh · fableplan: yes

## Problem

Validated against main at `cdd7f6a7`. A full `manual-close` passes the recorded stop-loss and take-profit order IDs to the venue for cancellation before submitting the close. If the close fails, `reconcileManualExecuteProtection` clears confirmed cancellations and `restoreManualStopLossAfterFailedClose` restores only the stop-loss. The successful stop-loss message does not report the missing take-profit protection.

The manual cancellation helper, `clearHyperliquidProtectionOIDsMatching`, clears each matching take-profit order ID and sets its existing armed flag to false. A later scheduler protection sync can therefore restore the cancelled tiers. Recovery still requires a running scheduler, a valid protection plan and successful venue operations. The command must not promise that the next cycle will restore them. With the scheduler stopped, no cycle provides this recovery.

This take-profit gap existed before the stop-loss restore merged in pull request (PR) #1539. That PR supplies the stop-loss recovery pattern, but its snapshot and recovery action do not implement take-profit restoration. The existing manual position view already copies take-profit IDs and armed flags. The shared protection planner resolves regime-dependent tiers and uses the position's risk anchor price, which can differ from average cost after a scale-in.

The command-line recovery path saves the book and queues a stop-loss action. The dashboard instead updates scheduler memory under its state lock and saves the book. The existing `update-sl` action does not adopt take-profit IDs. A take-profit restore must cover both paths and coordinate with scheduler reconciliation.

## Goal

After a rejected full manual close, restore eligible take-profit protection before the command returns. If recovery is incomplete or uncertain, report the affected orders and required operator action. Keep the close result as an error and preserve correct position and fill records.

## Approach

Keep one issue with this implementation checklist. Obtain a Fable plan before building.

- [ ] Capture the position identity, side, quantity, risk anchor, entry average true range, resolved tiers, order IDs and armed flags before execution. Reuse the existing protection planner and reduce-only placement mechanism. Limit recovery to live Hyperliquid strategies that pass `hyperliquidPlacesOnChainTPs`.
- [ ] Restore take-profits independently of whether a stop-loss existed or its restore succeeded. Preserve the existing stop-loss recovery result. A take-profit-only recovery must not replace a restored stop-loss or change completed tiers.
- [ ] Classify each requested cancellation and verify uncertain outcomes before placement. Preserve orders that still rest; do not recreate filled tiers. Recheck position identity, side and exposure after stop-loss recovery. Place nothing for a closed or reversed position. Bound restored quantities to the remaining owned exposure and preserve shared-coin ownership.
- [ ] Distinguish confirmed cancellations from filled tiers when constructing recovery input. Preserve the existing manual cancellation helper's armed-flag behavior. Do not reset every armed flag or change ordinary scheduler tier semantics.
- [ ] Define the command-line persistence and scheduler-adoption contract before implementation. Reuse `StateStore` and the pending manual-action mechanism where suitable; implement take-profit adoption explicitly. Cover the dashboard's memory-and-book path. Specify ordering, position identity checks, repeated adoption, restart recovery and coordination with concurrent protection sync. A stale action must not attach orders to a replacement position.
- [ ] Handle partial placement, immediate fills, unreadable replies and storage or queue failures. Preserve confirmed outcomes and enough recovery state to avoid duplicate placement. Leave fill accounting to the existing confirmed-fill reconciliation path. Shared sync success alone is insufficient proof that every required tier rests.
- [ ] Report restored, missing and unverified protection through both operator surfaces. On incomplete recovery, send the existing critical operator notification, identify the symbol and affected order IDs, and require exchange verification and restoration or closure. Do not promise recovery on the next scheduler cycle.
- [ ] Preserve forced-close and partial-close behavior, normal scheduler protection policy, and the recent stop-loss recovery fixes. Shared plumbing changes must carry regression coverage. Add no separate placement implementation.

## Acceptance criteria

1. With the scheduler stopped, a rejected full live manual close restores eligible, confirmed-cancelled take-profit tiers before return when the position remains open and venue operations succeed. The command still returns the original close error.
2. Recovery also works when no stop-loss was recorded. A restored stop-loss remains intact. A stop-loss that closes the position prevents subsequent take-profit placement.
3. Paper and non-tiered strategies place no take-profit orders. Completed tiers remain completed. Regime selection and risk-anchor pricing match the shared planner.
4. Resting, cancelled, filled and unknown order outcomes are handled separately. Unknown reads or lost placement replies never cause blind duplicate placement. Reduced exposure, reversed positions and shared-coin peers remain safe.
5. Successful recovery reaches the stored book and the running scheduler through the appropriate command-line or dashboard path. Repeated adoption and restart are safe. Concurrent scheduler work cannot duplicate recovery orders or overwrite them with stale state.
6. Partial recovery, immediate fills, unavailable state, placement failures and persistence failures produce accurate operator output. Missing or unverified protection raises a critical notification with the symbol and affected order IDs. Recovery must not invent fill prices, fees or closed quantities.
7. Add Go tests for operator decisions and both adoption paths without spawning Python, database-backed restart and failure tests, controlled concurrency tests, and Python venue-stub contract tests for reused protection behavior. Cover the safety cases above and retain forced-close, partial-close and stop-loss regression coverage.

## Plain simple English

A rejected manual close can leave take-profit orders missing while the stop-loss is restored. A later scheduler cycle can restore them, but recovery depends on the scheduler and the exchange. The command needs immediate recovery and a clear warning when protection remains missing or its state is unknown.

---
Created with LLM: Opus 5 | xhigh | Harness: fix-pr-review

richkuo:
## Plain simple English

A rejected manual close now restores eligible take-profit orders and reports missing or uncertain protection. The scheduler skips protection placement while an operator command or queued recovery is pending. Recovery retains unverified order IDs, and immediate fills complete their tiers so the scheduler does not place them again.

## Summary

Closes #1540.

- Restore cancelled take-profit tiers after a rejected full live manual close through the existing protection planner and reduce-only placement path. Recheck exposure and position identity after stop-loss recovery; preserve completed and unaffected tiers.
- Persist command-line recovery in the book and a `restore-tp` action carrying previous IDs, replacement IDs, armed flags, and position identity. Dashboard recovery updates scheduler memory. Adoption is idempotent and reports conflicting IDs. Rows for closed or replaced positions are acknowledged without attaching orders to another position.
- The cycle takes the strategy file's manual-action lock without waiting and checks for pending position or protection changes before it places orders. Stop-loss and take-profit recovery each hold the symbol lock through placement and bookkeeping.
- Whole-recovery errors retain order IDs whose cancellation was unconfirmed. Dust tiers carry a size-skipped result and produce an unverified recovery outcome. Ordinary cycle sync records immediate take-profit fills as completed tiers; confirmed-fill reconciliation retains responsibility for fill accounting.
- Incomplete recovery reports the affected symbol and order IDs through the existing critical notification path. The close returns its original error.

## Verification

- Go build and vet pass; gofmt and git diff checks are clean.
- The full Go suite has one failing test: `TestUpdateHelpersEnvfileParsing790`. The same failure reproduces on the unchanged base checkout at `cdd7f6a7`: the shell fixture reports `a record missing x.service or --restart`. The affected helper, script, and test have no changes in this pull request. Other Go tests pass.
- The protection lock tests pass with the Go race detector.
- Python regression suite: 2,943 passed. Hyperliquid script tests: 100 passed. Python syntax checks pass.
- Test budget passes: 172 wording-only tests against a baseline of 172.

### Test edits

- `TestInterpretManualCloseTPRestore`: **Outdated**. Ground: blocking review finding 2 requires preservation of order IDs whose cancellation was not confirmed. The subprocess-error and result-error cases now expect Unverified with the previous ID; confirmed-cancelled tiers still expect Missing. Added nil-result and size-skipped cases for blocking finding 2 and optional finding 1.
- New tests: `TestRunHyperliquidProtectionSyncManualActionGuard`, `TestApplyHyperliquidProtectionSyncImmediateTiers`, `TestDrainRestoreTPForClosedPosition`, `TestManualCloseRearmSerialisesWithProtectionSync`, and `TestSyncProtection.test_size_skipped_tiers_report_unverified`. These cover file-lock contention, persisted queue state, read failure, both symbol-lock orderings, immediate fills, stale-row acknowledgement, and dust reporting.
- The initial implementation added defaulted arguments to `TestSyncProtection._run_sync` and a failure stub to `stubTradeDeps`; those helper changes retained existing assertions and defaults.

## Plan

The initial implementation used the [Fable plan](https://github.com/richkuo/go-trader/issues/1540#issuecomment-5574162137).

Recovery calls `RunHyperliquidSyncProtection` through its dependency to retain the raw per-tier result and operator-specific notifications. It queues recovery only when the recorded tier values change. The ordinary-cycle immediate-fill gap identified by the plan is now fixed in this pull request.

---


richkuo:

richkuo:
Addressed the review from github-actions[bot] (2026-09-07T19:07:28Z) in 65416242a21543fd53229da80387c099ba41251c.

### Fixed

1. **The per-symbol lock cannot serialise the command-line restore against the daemon's cycle sync, so two full sets of take-profit orders can rest on-chain.** The cycle takes the existing manual-action file lock without waiting, reads the queue through StateStore, and skips placement while a position or protection action awaits adoption; the lock covers placement and bookkeeping (`scheduler/hyperliquid_protection.go:27`). Tests cover active lock contention, persisted queue rows after restart, unrelated rows, and read failure.
2. **A whole-recovery failure clears unconfirmed take-profit order ids from the book, so the next cycle places a duplicate tier order.** Whole-recovery failures keep unconfirmed IDs as Unverified; confirmed-cancelled tiers remain Missing (`scheduler/manual_close_tp_restore.go:147`). Subprocess errors, nil results, and result-level errors are covered.
3. **The stop-loss re-arm leg runs outside the new per-symbol lock, so a cycle sync can place a second stop concurrently.** Stop-loss recovery holds the symbol protection lock through the account read, placement, and bookkeeping (`scheduler/manual_close_rearm.go:118`). Tests exercise both operation orderings and retain the immediate-fill flat result.
4. **`tp_filled_immediately` is now parsed but the ordinary cycle path still ignores it, leaving a repeat-placement loop.** Ordinary sync clears the filled tier ID and marks the tier armed/completed (`scheduler/hyperliquid_protection.go:472`). Tests cover a resting peer tier, mixed immediate and external fills, legacy two-tier fields, and the next protection plan. This existing gap needs no new mechanism and is in scope under rule 3.
5. **A dust tier is reported "verified still resting" without any verification.** The script reports skipped tier sizes, including a whole position below lot precision; recovery retains unconfirmed IDs and reports Unverified (`shared_scripts/check_hyperliquid.py:1162`, `scheduler/manual_close_tp_restore.go:176`). Python tests cover a skipped middle tier and a fully skipped position.
6. **A `restore-tp` row whose position left the book retries with an error every cycle forever.** A missing position makes the restore action stale; the existing drain transaction acknowledges its row without a book mutation (`scheduler/manual.go:635`). Database tests cover the first drain and restart.

### Test edits

1. **TestInterpretManualCloseTPRestore** (`scheduler/manual_close_tp_restore_test.go:121`): Outdated. Ground: blocking finding 2 requires retention of IDs whose cancellation was never confirmed. The subprocess-error and result-error cases now expect Unverified with the previous ID; confirmed-cancelled tiers still expect Missing. Added nil-result and skipped-size cases for blocking finding 2 and optional finding 1.

### Added coverage

- `TestRunHyperliquidProtectionSyncManualActionGuard` (`scheduler/hyperliquid_protection_test.go:1021`): lock contention, queue persistence, unrelated actions, and read failure.
- `TestApplyHyperliquidProtectionSyncImmediateTiers` (`scheduler/hyperliquid_protection_test.go:1103`): completed-tier state and the next cycle plan.
- `TestDrainRestoreTPForClosedPosition` (`scheduler/manual_close_tp_restore_test.go:882`): durable acknowledgement without position or trade changes.
- `TestManualCloseRearmSerialisesWithProtectionSync` (`scheduler/manual_close_tp_restore_test.go:920`): both operation orderings and immediate-fill propagation.
- `TestSyncProtection.test_size_skipped_tiers_report_unverified` (`shared_scripts/test_check_hyperliquid.py:802`): dust-tier result reporting with real planner execution.

### Verification

- Go build and vet pass; gofmt and git diff checks are clean.
- The full Go suite has one failing test: `TestUpdateHelpersEnvfileParsing790`. The same failure reproduces on the unchanged base checkout at `cdd7f6a7`: the shell fixture reports `a record missing x.service or --restart`. The affected helper, script, and test have no changes in this pull request. Other Go tests pass.
- The protection lock tests pass with the Go race detector.
- Python regression suite: 2,943 passed. Hyperliquid script tests: 100 passed. Python syntax checks pass.
- Test budget passes: 172 wording-only tests against a baseline of 172.

## Plain simple English

The scheduler waits for operator recovery to be adopted before it places protection. Recovery keeps order IDs when their state is uncertain. Immediate fills complete their tiers, small skipped tiers report uncertainty, and recovery rows for closed positions are removed through the normal saved acknowledgement.

---
Created with LLM: GPT-6 | high | Harness: fix-pr-review

richkuo:

richkuo:
Addressed the review from github-actions[bot] (2026-09-07T20:46:45Z) in 0d236a7e3bbbf821142d91648f8dc5bdcaea1f32.

### Fixed

1. **The new manual-action guard silently suppresses the same-cycle protection re-arm after a failed close, so a position that just lost its exchange stop keeps none for the rest of the cycle.** `runHyperliquidProtectionSync` now takes an `hlProtectionGuardMode` (`scheduler/hyperliquid_protection.go:686`). `rearmProtectionAfterFailedClose` passes `hlProtectionGuardStopLegAfterFailedClose` (`scheduler/hyperliquid_shared_close_floor.go:317`), which proceeds with `stopLegOnlyProtectionPlan` stripping `Tiers`, `TPOIDs`, `TPArmedTiers`, `ForceTPReplace` and `CancelTPOIDs` (`scheduler/hyperliquid_protection.go:757`), so the queued `restore-tp` row still owns every tier id while the stop goes back in the same cycle. A strategy whose stop belongs to another owner returns without placing, since the trailing and scalar arms cover it. Skipping is no longer only an `Info` line: three consecutive blocked syncs for one strategy and symbol raise the CRITICAL through the notifier once, and any sync that runs clears the count (`scheduler/hyperliquid_protection.go:26`), so an operator command's one-cycle hand-off stays quiet while a backfill or a stuck row is reported. Verified: the stop-leg mode places the stop and leaves `TPOIDs` untouched, the non-owner case places nothing, and the counter alerts once and re-arms after a clear.

2. **A queued row the drain cannot apply now disables all protection placement for that strategy and symbol for as long as the row exists, with no terminal state and no alert.** Every retained row now reaches a terminal state. A `restore-tp` row whose position is owned by a peer is acknowledged without mutating the book and raises a CRITICAL naming the restored ids, because adopting them would attach the orders to another strategy's position (`scheduler/manual.go:659`). Any action whose apply keeps failing past `staleManualActionMaxAge` (one hour from `created_at`) is acknowledged through the same persisting transaction with a CRITICAL naming the action and the error (`scheduler/manual.go:381`). A row with no readable insert time never expires, so a parse failure keeps the conservative retry. A row that fails once and succeeds later still applies; the strategy-not-in-config branch is untouched, and a second queued row for the same symbol keeps the guard set. The repeat-block alert from finding 1 covers the window before a row expires.

3. **The command-line restore writes the book and queues its recovery row as two separate writes, so a failed book save leaves restored orders on the venue that the next cycle duplicates.** `StateStore.SaveStrategyBookQueueingManualAction` threads the queue insert into `saveStrategyBookWithAcks`, so the book change and the row commit or roll back together (`scheduler/state_store.go:296`, `scheduler/db.go:1807`, `scheduler/db.go:2016`). The take-profit restore uses it (`scheduler/manual_close_tp_restore.go:418`) and so does the stop-loss re-arm's `update-sl` / `cancel-sl` row (`scheduler/manual_close_rearm.go:255`), which carried the same two-write hazard. A restored order id is therefore never resting on the venue with neither the book nor a queued row naming it.

4. **A take-profit placement whose reply is lost leaves an untracked resting order, because the tier path has no book-diff resolver like the stop-loss path.** Fixed in this PR rather than filed, at the author's explicit instruction; scope rule 1 also applies, because this PR routes the new manual-close recovery through the same placement code and maps its error to `manualCloseTPMissing`, which clears the tier and lets the next cycle re-place it. `_place_tp` snapshots the open-order ids before each placement and, on an unreadable status or an exception, re-reads them through `_resolve_placement_by_book_diff`, the generalised `_resolve_sl_placement_by_book_diff` (`shared_scripts/check_hyperliquid.py:773`, `shared_scripts/check_hyperliquid.py:1200`). One fresh id takes the tier; no fresh id keeps the error and leaves the tier unplaced; a failed snapshot, a failed re-read, or two or more fresh ids set `tp_outcome_unknown[idx]`. An unknown tier is recorded at id 0 and **armed** on both the cycle path (`applyUnknownTPPlacementOutcome`, `scheduler/hyperliquid_protection.go:508`) and the recovery path (`manualCloseTPOutcomeUnknown`, `scheduler/manual_close_tp_restore.go:167`), so the next cycle places no second order at the same price, and both raise the CRITICAL that names the tier. An explicit venue rejection stays a plain error and never triggers a book diff.

### Test edits

No existing assertion, fixture or expectation was changed, so no test needed an Outdated, Wrong or Obsolete ground. Disclosed in full:

1. **TestInterpretManualCloseTPRestore** and **TestApplyRestoredTakeProfitTiers** (`scheduler/manual_close_tp_restore_test.go:177`, `scheduler/manual_close_tp_restore_test.go:242`): one new table case each for the unresolved-placement outcome. Every existing case is unchanged.
2. Every `runHyperliquidProtectionSync` call in `scheduler/*_test.go` gains the `hlProtectionGuardFull` argument. Compile-only, following the new parameter.
3. **TestRunHyperliquidProtectionSyncManualActionGuard** (`scheduler/hyperliquid_protection_test.go:1078`): `clearHLProtectionGuardBlocks` calls added so the new consecutive-block counter does not leak between subtests. No assertion changed.
4. **TestSyncProtection._run_sync** (`shared_scripts/test_check_hyperliquid.py:701`): defaulted `open_oids_sequence` and `tp_place_error` arguments. Existing defaults and every existing assertion are unchanged.

### Added coverage

- `TestRunHyperliquidProtectionSyncStopLegAfterFailedClose` (`scheduler/hyperliquid_protection_test.go:1181`): the stop leg re-arms under a queued row, the tiers are stripped from the plan and left untouched in the book, and a non-owner strategy places nothing.
- `TestHLProtectionGuardBlockAlertsOnRepeat` (`scheduler/hyperliquid_protection_test.go:1157`): one alert at the threshold, none after, per-symbol counting, and re-alerting after a clear.
- `TestApplyUnknownTPPlacementOutcome` (`scheduler/hyperliquid_protection_test.go:1240`): the tier is armed at id 0, the next protection plan places nothing, and the alert names the tier.
- `TestDrainRestoreTPTerminalStates` and `TestDrainAcknowledgesExpiredFailingAction` (`scheduler/manual_close_tp_restore_test.go:1004`, `scheduler/manual_close_tp_restore_test.go:1053`): peer ownership acknowledged with a critical and no book mutation, a fresh failure retried, an expired failure acknowledged, and an owned row still adopting.
- `TestSaveStrategyBookQueueingManualActionIsAtomic` (`scheduler/manual_close_tp_restore_test.go:1095`): both writes commit together, a rejected queue write rolls the book back, and a row naming another strategy is refused.
- `TestSyncProtection.test_unresolved_tp_placement_resolves_by_book_diff` (`shared_scripts/test_check_hyperliquid.py:836`): unreadable reply and exception both resolved to the fresh id, no fresh id kept as an error, and failed snapshot, failed re-read and ambiguous diff each reported unknown.

### Verification

- Go build, `go vet` and `gofmt` are clean. The protection-lock and guard tests pass with the race detector.
- Full Go suite: one failing test, `TestUpdateHelpersEnvfileParsing790`. It reproduces on the unchanged base checkout at `cdd7f6a7` with the same message (`a record missing x.service or --restart`), it is local to this machine, and CI reports all four checks green on this branch. No file it covers is in this diff.
- Python regression suite: 2,943 passed. Hyperliquid script tests: 106 passed.
- Test budget: 172 wording-only tests against a baseline of 172.
- `CLAUDE.md` is 14,994 bytes, inside the 15,000-byte cap; the new guardrails were added and equivalent wording trimmed elsewhere, dropping no guardrail. `SKILL.md` records the guard modes, the three bounds on the skip, the single-transaction write and the tier book-diff resolver.

## Plain simple English

The daemon can now put the stop back in the same cycle after a rejected close, even while an operator recovery waits to be read. It tells the operator when it must skip protection again and again. Old queued rows now end instead of blocking forever. The book and its recovery row are saved together. A take-profit order with a lost reply is found by comparing the exchange order list, and is never placed twice.

---

richkuo:
@claude review

richkuo:
CI follow-up on 0d236a7e: the `go` check failed. Fixed in 17aa0437.

**Correction to my previous verification note:** I reported the full Go suite as having one failure (`TestUpdateHelpersEnvfileParsing790`). It had two. I read only the tail of the run output and missed the second, so the claim was wrong.

### Fixed

1. **CI Failure — `go`** — `TestApplyManualActionRestoreTP/a_position_owned_by_another_strategy_errors` still asserted the pre-fix behaviour (`scheduler/manual_close_tp_restore_test.go:342`). Blocking finding 2 required the `restore-tp` ownership mismatch at `scheduler/manual.go:659` to stop erroring on every cycle and reach a terminal state, so the error the case expected no longer exists.

### Test edits

1. **TestApplyManualActionRestoreTP**, case `a position owned by another strategy errors` (`scheduler/manual_close_tp_restore_test.go:342`) — **Outdated**. Ground: blocking review finding 2 on this PR, which requires every retained `pending_manual_actions` row to reach a terminal state and names `scheduler/manual.go:641` (the `restore-tp` ownership mismatch) as the sibling branch that must stop erroring. The case now asserts what that fix produces: no error, the peer's book left untouched (`TPOIDs` and armed flags unchanged), and exactly one critical naming the restored order id and the recorded owner. The `wantErr` and `wantConflicts` fields become a per-case `wantCritical` list so each case checks its own message instead of one hardcoded id pair; every other case keeps its assertions unchanged.

### Verification

- `go vet` and `gofmt` clean. Full Go suite passes except `TestUpdateHelpersEnvfileParsing790`, which reproduces unchanged on the base checkout at `cdd7f6a7` and whose `shell` and `service-fixture` CI checks are green on this branch.
- The `shell`, `python` and `service-fixture` checks all passed on 0d236a7e; only `go` failed.

## Plain simple English

One older test still expected an error that the review fix removed on purpose, so the Go check failed. The test now checks the new result: the row is marked done, the other strategy's records are not changed, and one alert names the order and the owner.

---

PR #1547, opened 2026-09-10
fix(#1546): alias every merged paper strategy as <base>-paper [C68, Opus 5, xhigh]

### LINKED ISSUE #1546: [C68] merge-paper-instance.sh: a paper strategy with a free bare id merges with no -paper alias
**Complexity: 68/100** — Capability 2 (Risk 3 — rewrites the live config the scheduler loads and the in-process id a strategy runs under, and recovery needs `--rollback`; Uncertainty 3 — two viable designs with different edit lists: alias only, or alias plus a drift-audit report for a paper book with no live twin); Volume 18 (Scope 3: merge script, shell test, drift audit, example config, `SKILL.md`, `docs/POST_UPDATE_HISTORY.md`; Coupling 3: the alias contract is mirrored in the script, the audit and three docs; Verification 3: the shell test runs real binaries against SQLite fixtures) — Opus 5, xhigh · fableplan: no

## Problem

`merge-paper-instance.sh` gives a paper strategy a `-paper` / `-paper2` alias only when its bare id is already taken by a strategy in the merged config (`scripts/merge-paper-instance.sh:544-556`). A paper strategy whose bare id is free merges under that bare id with no alias.

That leaves the merged live config without the live/paper twin naming the rest of the merged-architecture work assumes:

- Operators grepping for `hl-rmc-eth-60-paper` to find the paper twin of `hl-rmc-eth-60` get nothing when the paper deployment had no live counterpart at merge time and was committed under its bare id.
- The drift audit pairs a paper strategy with a live one through `replay_source_id`, an equal id, or a `-paper` suffix backed by `storage_strategy_id` (`scripts/check-live-paper-config-drift.sh:213-234`). A bare-id paper book holds the base name itself, so a live counterpart cannot be added under that name later without renaming the paper book first. An alias plus the bare `storage_strategy_id` leaves the base name free and makes that later pairing work with no second migration.
- The shipped example (`scheduler/config.live-paper.example.json`) models a paper twin only as `<base>-paper` beside `<base>`, so a reader has to infer that bare ids are accepted too.

Concrete example from the move being prepared now: `go-trader-paper-testing` is being merged into `go-trader-live`. The paper-testing book has 20 ETH-only strategies across 5m/15m/30m/60m (`hl-rmc-eth-5/15/30/60`, `hl-temac-eth-5/15/30/1h`, `hl-bps-eth-5m/15m/30m/1h`, `hl-vrs-eth-5m/15m/30m/1h`, `hl-abr-eth-5/15/30/1h`). None share a base name with anything currently in `go-trader-live`, so under the current rule they all merge under their bare ids.

Two earlier claims in this issue were traced and do not hold:

- A bare-id paper book with no live counterpart does **not** trip the drift audit's `AMBIGUOUS` gate. That gate fires only when the stripped base is itself a live id (`check-live-paper-config-drift.sh:228-231`). With no live entry in the group the audit skips the strategy with no output (`:251`).
- Renaming alone does not give the drift audit a twin to compare. The audit needs a live strategy carrying the base id; an alias cannot create one. For the 20 paper-testing strategies the audit stays silent before and after this change. Reporting that silence is a separate decision, taken in the Approach below.

## Goal

Make the `-paper` alias unconditional. Every paper strategy merged into the live process gets an alias of the form `<base>-paper`, `<base>-paper2`, `<base>-paper3`, ..., picked deterministically by scanning the merged config, and the alias becomes the new in-process identifier. `storage_strategy_id` keeps pointing at the bare stored identifier from the paper database.

The behavior must remain:

- A paper strategy whose bare id collides with an existing live strategy still ends up aliased (it gets `<base>-paper` when free, otherwise `<base>-paper2`, ...).
- A paper strategy whose bare id is itself free ends up aliased too. No more bare-id paper books.
- Re-running `--apply` after a successful merge is a no-op. The skip keys on `storage_strategy_id` (`merge-paper-instance.sh:372-375`), so it survives the wider rename.
- `--rollback` restores the retained `config.json.pre-merge-<paper>` file and never touches a database (`restore_from_retained`, `merge-paper-instance.sh:1032-1043`). The paper source config is never rewritten, so the bare ids come back with the restored live config.

## Approach

Part of the merged live/paper architecture follow-ups to #1525 / #1531. The rename conflict case needs no new refusal: the resolver increments the numeric suffix until the candidate is free of both `taken` and `remaining` (`merge-paper-instance.sh:546-550`), so it always terminates.

1. In the compose pass (`merge-paper-instance.sh`, the `MERGE_PY` heredoc, `cmd_compose`), run every paper strategy through the alias resolver instead of only the colliding ones. The resolver scans the merged config for `<base>-paper`, then `<base>-paper2`, `<base>-paper3`, ... and takes the first free suffix. The result becomes the strategy's new `id`; `storage_strategy_id` continues to name the bare stored id.
2. Teach `--diff` to preview the alias plan. `cmd_root_diff` today prints differing root keys plus the compose refusals it can see without inspect (`merge-paper-instance.sh:267-274`) and prints no rename at all; the rename lines come from the dry run's compose report. Add a config-only alias preview (`hl-rmc-eth-5 -> hl-rmc-eth-5-paper`) to the same static path, so the operator can review every rename before any lock or binary runs.
3. Decide how the drift audit treats a paper strategy with no live twin. Today it is skipped with no output (`check-live-paper-config-drift.sh:251`). Report it as `UNPAIRED` instead, with the same non-gating shape as the existing unset-mode `UNPAIRED` line, so a merged paper book is visible in the audit whether or not a live counterpart exists. This is the part that actually makes the audit see these 20 strategies; the alias alone does not.
4. Update `scripts/test_merge_paper_instance.sh`. The base fixture's live and paper configs both hold `hl-x` (`:60`, `:82`), so every existing case already exercises the collision path and keeps its current expectations. Add a case where the paper strategy has no live counterpart and the bare-id to alias rename must still fire, and re-point the "second collision suffix" case at a paper strategy that does not collide with a live strategy. Add an assertion for the new `--diff` alias preview.
5. Update the documentation that states the current rule:
   - `SKILL.md`, the Storage Ownership handoff subsection, which reads "a colliding paper id gets the first free `-paper`/`-paper2` suffix".
   - `docs/POST_UPDATE_HISTORY.md:21`, which reads "`-paper` suffix plus `storage_strategy_id` on a colliding id".
   - The script's `--help` text, which describes `--diff` output and must name the alias preview.
6. Record the operator-facing consequences in the same documentation pass:
   - Every merged paper strategy changes its in-process handle, so Discord manual commands, the dashboard and `#T` labels for those strategies use the alias after cutover.
   - `/go-trader-paper-to-live` changes only the strategy args and keeps the id (`scheduler/ui_structural.go:308-321`), so a promoted strategy runs live under a `-paper` name. State the operator step that renames it, or strip the suffix on promotion.
   - A paper config that already holds both `hl-x` and `hl-x-paper` produces `hl-x-paper2` and `hl-x-paper-paper`. Either accept and document that, or refuse a paper id that already carries a `-paper` suffix.

`scheduler/config.live-paper.example.json` already carries the target shape: its paper strategy has `id: hl-vwap-eth-60-paper`, `storage_strategy_id: hl-vwap-eth-60` and `replay_source_id: hl-vwap-eth-60`. Extend it only if a second pair showing a paper book with no live counterpart helps the reader.

## Acceptance criteria

- A paper instance with no live counterparts merges under `<base>-paper` aliases for every paper strategy, with `storage_strategy_id` pointing at the bare stored id. Verified with `storage-inspect --json` on the merged live deployment and with `inspect --all --json` showing the new `id`s.
- A paper instance whose every paper strategy collides with a live strategy merges under `<base>-paper2` or higher without overwriting the existing `<base>-paper` twin. Verified by a regression test in `scripts/test_merge_paper_instance.sh`.
- A re-applied merge is a no-op: the second run reports every paper strategy as already merged and adds no second copy.
- `--diff` lists every alias the apply step would create, from the two config files alone, with no lock, binary or unit stop.
- The drift audit reports a merged paper alias with no live twin as `UNPAIRED` and keeps exit 0 for that case. A paper alias whose live twin exists still pairs and compares as it does today, and an unbacked `-paper` suffix over an existing live base still gates as `AMBIGUOUS`.
- `--rollback` restores the retained `config.json.pre-merge-<paper>` file, so the paper strategies are back under their bare ids and no database is touched.
- `SKILL.md`, `docs/POST_UPDATE_HISTORY.md` and the script's `--help` text state the unconditional rule, the preserved `storage_strategy_id`, and the changed operator handles.
- `go test ./...` passes, including `TestConfigExampleCopyDoesNotTriggerTheUpgradePath`, and `scripts/test_merge_paper_instance.sh` passes with its fixtures.

## Plain simple English

Today a paper strategy keeps its plain name when that name is free in the merged live config. We want every merged paper strategy to get a `-paper` name. Then live and paper entries read as a pair, and an operator can find the paper copy by name. The stored identity does not change.

---
Validated with LLM: Opus 5 | xhigh | Harness: validate-issue

richkuo:
Closes #1546

## Plain simple English

A paper strategy kept its plain name when that name was free in the merged live config. Now every merged paper strategy gets a `-paper` name, and the stored identity does not change. The live and paper entries read as a pair, and an operator can find the paper copy by name.

## Summary

- **Unconditional alias.** `scripts/merge-paper-instance.sh` aliased a paper strategy only when its bare id was already taken in the merged config, so a paper deployment with no live counterpart merged under bare ids. `resolve_paper_alias` now takes the first free `<base>-paper` / `-paper2` / `-paper3` name for **every** paper strategy, and `storage_strategy_id` carries the bare stored id, so no stored book moves.
- **Idempotent resolver.** A paper id that already carries the alias keeps it when the name is free, so a re-alias never yields `-paper-paper`, and a colliding already-aliased id resolves from its base (`hl-x-paper` -> `hl-x-paper2`).
- **`--diff` preview.** The resolver is factored into `compose_alias_plan`, which `cmd_root_diff` reuses, so `--diff` prints the alias every paper strategy would take from the two config files alone, with no lock, binary or unit stop.
- **Drift audit.** `scripts/check-live-paper-config-drift.sh` gains a non-gating `UNPAIRED (no live twin)` line for a `-paper` alias whose base names no live strategy in the audited deployments. A suffix over an existing live base with no backing `storage_strategy_id` still prints `AMBIGUOUS` and still gates.
- **Docs and example.** `SKILL.md`, `docs/POST_UPDATE_HISTORY.md` and the script's `--help` state the unconditional rule, the preserved `storage_strategy_id`, the `--diff` preview and the operator consequences. `scheduler/config.live-paper.example.json` gains a paper strategy with no live counterpart, so the example shows the alias in the case the old rule left bare.
- **One suffix rule.** `scripts/paper_alias.py` is the single definition of the `<base>-paper[N]` parser that the merge script and the drift audit both read, so the audit pairs on the exact names the merge composes. `scripts/test_paper_alias.py` asserts the parser's outcomes, that `scripts/` defines it in exactly one file, and that both tools import it.

Unchanged on purpose: `--rollback` still restores the retained `config.json.pre-merge-<paper>` file (which brings the bare ids back) and never touches a database; the re-apply no-op still keys on `storage_strategy_id`, so it survives the wider rename.

## Verification

| Check | Result |
|---|---|
| `go test ./...` | Passes except `TestUpdateHelpersEnvfileParsing790` (see below) |
| `go test -run 'TestMergePaperInstance\|TestUpdateShellScriptSyntax'` | Pass (the full `test_merge_paper_instance.sh` suite) |
| `go test -run TestConfigExample` | Pass |
| `uv run --no-sync python -m pytest -n auto` | 2968 passed |
| `gofmt -l scheduler/` | Clean |
| `scripts/check_test_budget.py` | 172 wording-only tests, baseline 172 |
| `bash -n` on all three edited shell scripts | Clean |
| Drift audit by hand, paper alias with no live twin | `UNPAIRED (no live twin) hl-z-paper`, exit 0 |
| Drift audit by hand, paper alias whose live twin exists | `PAIR hl-y` / `IN SYNC`, exit 0, no UNPAIRED line |
| Drift audit by hand, `-paper` suffix with no backing `storage_strategy_id` | `AMBIGUOUS hl-y-paper`, exit 1 |
| Both scripts with `scripts/paper_alias.py` removed | Exit 2, the missing path named |

**Pre-existing failure, not caused by this change.** `TestUpdateHelpersEnvfileParsing790` fails on this macOS host at `FAIL: a record missing x.service or --restart`, inside the `update.sh` service-record assertions of `scripts/test_update_helpers.sh`. The same failure reproduces on unchanged `origin/main` (verified by running the `origin/main` copy of that script against the same binary), and `scripts/update.sh` and `scripts/update_helpers.sh` are untouched by this diff. CI runs that job on `ubuntu-latest`.

**Left to CI.** Because that assertion aborts the script before the drift section, the two new drift-audit assertions in `scripts/test_update_helpers.sh` did not execute locally. The behavior they assert was verified by hand against both fixtures (the two drift rows in the table above); the assertions themselves first run in the CI `shell` job on Linux.

## Deviations from the issue

The issue's Approach is followed, with these judgment calls named:

1. **`CLAUDE.md` is unchanged.** It sits at its documented 15000-byte cap and forbids growth ("Shorten wording, drop no guardrail"). A clause naming the unconditional alias needed 78 bytes that could only come from compressing already-dense guardrail text. `CLAUDE.md` delegates mechanism to `SKILL.md` and `docs/POST_UPDATE_HISTORY.md`, both of which now state the rule, so no guardrail is lost. Say the word and I will free the bytes and add the clause.
2. **`docs/POST_UPDATE_HISTORY.md` gets a new entry** rather than a rewrite of the #1525 entry, which is archived history; the stale clause in that entry now points at this issue.
3. **The suffix rule now has one owner** (follow-up from the first review). `scripts/paper_alias.py` holds the single definition of `paper_alias_base`; both shell scripts pass their own `SCRIPT_DIR` to their embedded Python block as `GO_TRADER_SCRIPT_DIR` and import it, and each refuses with exit 2 before any work when that file is absent. The drift audit's `strip_paper_suffix` is removed and its two call sites take the shared name. `scripts/test_paper_alias.py` pins the contract, the single definition, and both importers.

Against the issue as validated, two of its original claims were already corrected in the issue body and are reflected here: a bare-id paper book with no live counterpart was never caught by the `AMBIGUOUS` gate (it was skipped silently), and the alias by itself cannot give the drift audit a twin to compare, which is why the `UNPAIRED` report carries that part of the goal.

## Test edits

`scripts/test_merge_paper_instance.sh`, case "a paper strategy with no stored book passes": it asserted `strategies.2.id == "hl-new"`, that a paper strategy with no live counterpart merges under its bare id.

- **Case:** Outdated.
- **Ground:** issue #1546's Goal ("A paper strategy whose bare id is itself free ends up aliased too. No more bare-id paper books.") and its acceptance criterion that a paper instance with no live counterparts merges under `<base>-paper` aliases.
- **Replacement asserts:** the `rename hl-new -> hl-new-paper (storage_strategy_id=hl-new)` report line, the staged id `hl-new-paper`, and `storage_strategy_id` `hl-new`.

The shared-module follow-up needed no test edit: it keeps every output byte the same, and it adds `scripts/test_paper_alias.py`. No other test was weakened, skipped or removed. Added coverage: the "second collision suffix" case is widened to cover a paper id with no live collision alongside the colliding one (renamed to "second suffix, with and without a live id collision"); a new case covers an already aliased paper id keeping its name; the `--diff` static-preview case asserts the alias preview line; `scripts/test_update_helpers.sh` adds the `UNPAIRED (no live twin)` case and asserts that a paired alias is not flagged; `scheduler/config_example_version_test.go` gains one assertion for the new example strategy's alias mapping.

---
Updated with LLM: Opus 5 | xhigh | Harness: Claude Code
https://claude.ai/code/session_01EWsUr8UgX4VHo9wXVsApQP


richkuo:
@claude review

richkuo:
The review's follow-up item is done in this PR instead of a new issue, at `2e5f95e`.

`scripts/paper_alias.py` now holds the single definition of `paper_alias_base`. Both shell scripts pass their own `SCRIPT_DIR` to their embedded Python block as `GO_TRADER_SCRIPT_DIR` and import the rule from it; each refuses with exit 2 before any work when that file is absent. The drift audit's `strip_paper_suffix` is removed and its two call sites take the shared name, so the suffix contract has one owner.

`scripts/test_paper_alias.py` pins it three ways: a table over bare, plain-alias, numeric-suffix, non-numeric-suffix and double-alias ids; an assertion that `scripts/` defines the parser in exactly one file; and an assertion that both paper tools import it. `scripts/` is already in the pytest testpaths, so the CI python job runs them.

The refactor keeps every output byte the same, so no test needed an edit. Verified: the paper handoff suite passes, pytest passes (2968), the test budget is unchanged at 172, `bash -n` is clean on both scripts, and the three pairing outcomes were run by hand against fixtures (PAIR/IN SYNC at exit 0, AMBIGUOUS at exit 1, UNPAIRED at exit 0). With the module removed both scripts exit 2 and name the missing path.

## Plain simple English

Two scripts each kept a copy of the same "-paper" name rule. The rule is now in one file that both scripts read, and a new test makes sure it stays in one place. Nothing an operator sees has changed.

---
Updated with LLM: Opus 5 | xhigh | Harness: Claude Code
https://claude.ai/code/session_01EWsUr8UgX4VHo9wXVsApQP

richkuo:
@claude sonnet review

PR #1550, opened 2026-09-10
fix(#1549): preview the merge channel plan in --diff and drop redundant -paper keys [C66, Opus 5, xhigh]

### LINKED ISSUE #1549: [C66] merge-paper-instance.sh: --diff hides the paper channel keys the merge adds, and a key that repeats the live value is still added
**Complexity: 66/100** — Capability 2 (Risk 2 — the composed config decides which Discord channel each moved strategy routes through, so a wrong result mixes paper alerts into a live channel; it is reversible from the retained pre-merge config and no persisted trade state is touched; Uncertainty 2 — the merge helper and the runtime resolver are both established, and the remaining choice is where the report threads through the two callers); Volume 16 (Scope 3: the merge script, its test script, and `SKILL.md`; Coupling 2: `apply_paper_discord_maps` is called from both the preview path and the compose path, so a report-shape change moves both; Verification 3: the test script needs a built binary and a live/paper database fixture directory) — Opus 5, high · fableplan: no

## Problem

`merge-paper-instance.sh` already merges the paper deployment's Discord channel maps. `apply_paper_discord_maps` (`scripts/merge-paper-instance.sh:355`) walks `CHANNEL_MAPS` (`channels`, `trade_alert_channels`, `dm_channels`) under the `discord` block, adds each paper value at `<platform>-paper`, keeps the live value at the bare key, and refuses when a `-paper` key already exists with a different value. It runs from the compose path (`:711`) and from the `--diff` preview path (`:459`). The runtime then routes a paper strategy through that key: `MultiNotifier.resolveChannelKey` prefers `<platform>-paper` for a non-live strategy (`scheduler/notifier.go:246`), and `scheduler/discord.go:156` and `scheduler/inspect_cmd.go:750` mirror the same rule. `scripts/test_merge_paper_instance.sh:252` already asserts that a live `hyperliquid` and a paper `hyperliquid` compose into `hyperliquid` plus `hyperliquid-paper`.

Two gaps remain in that already-working merge.

**1. `--diff` never names the channel keys it will add.** `collect_compose_refuse_previews` calls `apply_paper_discord_maps` and discards the report (`_report, conflicts = ...` at `scripts/merge-paper-instance.sh:459`). It surfaces only the conflict previews. The compose path keeps its report and prints it as `compose: discord.channels.hyperliquid-paper=<id>` lines, but compose runs inside `--apply` and the dry run, both of which need both units stopped and both locks held. So the one mode an operator can run against a live pair, `--diff`, is the one mode that does not show which channel keys the merge would create. The operator sees a refusal preview or nothing, and learns the added keys only after the merge is composed.

**2. A paper value equal to the live value is duplicated.** When live and paper carry the same value at the same key, the loop at `scripts/merge-paper-instance.sh:371-380` still writes `mm["<platform>-paper"] = val` because the target key is absent. The merged config then carries `hyperliquid` and `hyperliquid-paper` with one identical channel id. Routing is unaffected, because both modes resolve to the same channel. But the merged config gains a key that carries no routing decision, and a later operator reading it cannot tell a deliberate split from a merge artifact.

Neither gap loses per-trader routing. The routing itself is correct today.

**Refuted against `origin/main`.** These were part of an earlier reading of this issue and do not hold:

- **There is no root-level `channels`, `trade_alert_channels`, or `dm_channels` config key.** They exist only inside `DiscordConfig` and `TelegramConfig` (`scheduler/config.go:18-20`, `:49-51`). `CHANNEL_MAPS` names the keys nested under `discord`, which is what the code uses.
- **A merge does not silently take live's value for an unrecognized root key.** `collect_root_diffs` classifies any differing non-dropped root key as `unknown`, and `refuse_root_conflicts` exits 1 (`scripts/merge-paper-instance.sh:718` → `:260`). `scripts/test_merge_paper_instance.sh:680` asserts that refusal.
- **There is no per-strategy `trade_channel` field.** The string appears nowhere in the repository. Channel selection is by mode inside the resolver, so there is no per-strategy override to rewire and nothing to journal for it.
- **`telegram` channel maps are deliberately not merged.** `telegram` is in `REFUSE_ON_DIFFERENCE` (`scripts/merge-paper-instance.sh:174`), so any Telegram difference refuses the whole merge. That is the fail-closed choice; `--align-to-live` is the operator path. This issue keeps it.
- **A numeric channel suffix has no reader.** The resolvers look up `<platform>-paper` only, and `validateDMChannelsMap` rejects a `dm_channels` key that contains `-paper` without ending in it (`scheduler/config.go:2130`). A `-paper2` channel key would be dead or invalid config.
- **The example config and the documentation already show the convention.** `scheduler/config.live-paper.example.json` carries `discord.channels` with `hyperliquid` plus `hyperliquid-paper` and `trade_alert_channels` with `hyperliquid-live` plus `hyperliquid-paper`; `SKILL.md:825` states that `-paper` channel keys are added, and `SKILL.md:1101` documents the resolver.
- **Rollback already reverses the channel merge.** `--rollback` restores the whole retained `config.json.pre-merge-<paper>` (`scripts/merge-paper-instance.sh:132`), so a per-key channel journal entry adds no coverage.

## Goal

`--diff` names every channel key the merge would add, so an operator can check the routing before either unit is stopped. The composed config carries a `-paper` channel key only when it changes where a paper strategy routes.

- `--diff` prints one line per channel key the compose step would add, for each of the three maps under `discord`, in the deterministic order the merge helper already uses, alongside the existing conflict previews and alias previews.
- A paper value that differs from the live value at the same key still composes to `<platform>-paper`, exactly as today.
- A paper value equal to the live value at the same key adds no `-paper` key. The merged map keeps the single bare key.
- A paper key that already ends in `-paper` composes unchanged, exactly as today.
- A `-paper` key that already exists in live with a different value still refuses, exactly as today.
- A re-applied merge stays a no-op.

## Approach

Part of the merged live/paper deployment follow-ups to #1525 / #1531 / #1544 / #1546.

1. In `apply_paper_discord_maps` (`scripts/merge-paper-instance.sh:355`), skip the `-paper` write when the resolved paper value equals the value already at the bare key the paper side resolved from. Report the skip so both callers can name it. The conflict branch and the `-paper`-suffixed passthrough branch keep their current behavior.
2. Thread the report out of `collect_compose_refuse_previews` (`:459`) instead of discarding it, and print it from `cmd_root_diff` (`:285`) as a channel-plan section next to the existing `diff: compose-refuse` and `diff: alias` lines. The compose path (`:711`) keeps printing the same report under its `compose:` prefix, so the two modes agree on wording for the same plan.
3. Update `scripts/test_merge_paper_instance.sh`:
   - `--diff` names the `-paper` channel key it would add for a paper strategy whose platform map differs from live.
   - `--diff` names the added key for each of `channels`, `trade_alert_channels`, and `dm_channels`.
   - live and paper carry the same value at `discord.channels["hyperliquid"]` → the staged config has `hyperliquid` and no `hyperliquid-paper`.
   - live and paper carry different values → the staged config has both, unchanged from today.
   - the existing conflicting `-paper` refusal case (`:296`) still refuses.
4. Update `SKILL.md:825` so the `-paper` channel sentence states that `--diff` previews the added keys and that a paper value equal to live's adds no key, and update the `--help` text in `scripts/merge-paper-instance.sh` where it describes what `--diff` prints.

## Acceptance criteria

- `--diff` on a live/paper pair whose `discord.channels["hyperliquid"]` values differ prints a line naming `discord.channels.hyperliquid-paper` and its value, with no binary run, no lock held, and neither unit stopped.
- `--diff` names the added key for each of the three maps under `discord` when the paper deployment sets all three.
- A live and paper pair carrying the identical value at `discord.channels["hyperliquid"]` composes to a merged `discord.channels` holding `hyperliquid` once and no `hyperliquid-paper`. Verified by reading the staged config.
- A live and paper pair carrying different values at that key composes to both `hyperliquid` (live value) and `hyperliquid-paper` (paper value), unchanged from the behavior asserted today at `scripts/test_merge_paper_instance.sh:252`.
- A live config that already holds `discord.channels["hyperliquid-paper"]` with a different value than the paper deployment resolves to still refuses with exit 21 and names the key.
- A paper key that already ends in `-paper` still composes unchanged.
- A second `--apply` on a complete journal stays a no-op.
- `--diff` and the compose path print the same channel-plan content for the same pair, so the preview matches what apply composes.
- No new empty channel map is written: a paper deployment that sets only `discord.channels` leaves `trade_alert_channels` and `dm_channels` absent in the merged config.

## Plain simple English

The merge script already sends paper trades to the paper channel. But it does not show which channel keys it will add before the merge runs, so the operator cannot check the routing first. It also adds a key that repeats a channel the live program already uses.

---
Validated with LLM: Opus 5 | xhigh | Harness: validate-issue

richkuo:
Closes #1549

## Plain simple English

Before this change the merge script did not tell the operator which Discord channel keys it would add, and the only mode that runs while both programs are up was the mode that showed nothing. Now `--diff` lists every channel key the merge would add. The script also stops adding a paper key that repeats a channel the live program already uses.

## Summary

Two changes to `scripts/merge-paper-instance.sh`.

**1. `--diff` prints the channel plan.** `collect_compose_refuse_previews` returned the discord conflict previews and threw the merge helper's report away. It now returns the report too, and `cmd_root_diff` prints one `diff: channel-plan <line>` per entry beside the existing `diff: compose-refuse` and `diff: alias` lines, or `diff: channel-plan no discord channel keys to add` when there is nothing to add. The compose path prints the same content under its own `compose:` prefix, so the preview and the apply agree. `--diff` still reads only the two config files: no binary, no lock, no unit stop.

**2. A paper channel value that changes no routing adds no key.** `apply_paper_discord_maps` wrote `<platform>-paper` whenever the key was absent, even when the value repeated what the live config already carried at the bare key. The merged config then held two keys with one channel id and no way to tell a deliberate split from a merge artifact. The helper now prunes such a key and reports it as `discord.<map>.<platform>-paper not added (paper value <id> already routes through discord.<map>.<key>)`.

The prune runs after every write, so the two existing refusals are untouched: a live `-paper` key holding a different value still refuses with exit 21, and a same-platform type-keyed self-conflict still refuses. A key is pruned only when **every** `(platform, type)` pair that resolved to it would fall back to the same value through the merged map, mirroring `resolveChannel` (`scheduler/discord.go:144`), which prefers the `platform` key over the strategy-type key. A key the paper config set explicitly as `<platform>-paper` is never pruned.

`SKILL.md` and the `--help` text now state both behaviors.

### Deviation from the issue's Approach

Approach step 1 says to compare the paper value with "the value already at the bare key the paper side resolved from". That comparison changes routing when the paper side resolves its value from the strategy-type key while the merged map also holds a `platform` key with a different value: the runtime resolver prefers `platform`, so dropping the `-paper` key would send that paper strategy to a different channel. The skip therefore compares against the runtime fallback for every pair that shares the target key. Everything else follows the Approach.

## Verification

- `go -C scheduler test -run TestMergePaperInstance -count=1 .` — pass. This is the harness that runs `scripts/test_merge_paper_instance.sh` end to end against a built binary and a live/paper database fixture pair.
- **Red then green.** With the fix reverted and the new tests in place, the run fails at `--diff names the -paper channel key the compose step would add (missing 'diff: channel-plan discord.channels.hyperliquid-paper=C-paper')`. Driving the embedded merge helper directly against the pre-change and post-change versions, the equal-value cases (`equal skips`, `stype value equal to platform value skips`) fail before and pass after, while `conflict refuses`, `type self-conflict`, `explicit -paper kept`, `mixed fallback keeps key`, `platform fallback wins` and `idempotent` pass in both.
- `bash -n` on both scripts — pass.
- `python3 scripts/check_test_budget.py` — `wording-only tests: 172 (baseline 172)`.
- `go -C scheduler build .` — pass.
- `go -C scheduler test ./...` — one failure, `TestUpdateHelpersEnvfileParsing790`. It is pre-existing on the base commit `5bdf6489`: it exercises `scripts/test_update_helpers.sh`, which sources `scripts/update_helpers.sh` and runs `scripts/update.sh`, and this change touches none of those three files.
- Not run: the pytest suites. This change edits shell and Markdown only, and no Python package changed.

### Test edits

None. `scripts/test_merge_paper_instance.sh` gains cases only; no existing assertion was weakened, narrowed, or removed.

New cases:

1. `--diff` names `discord.channels.hyperliquid-paper=C-paper` for a pair whose values differ, inside the existing static-preview block.
2. `--diff previews every discord channel key the merge would add`: the paper deployment sets all three maps under `discord` and `--diff` names the added key for each, writing no staged config.
3. `a paper channel value equal to live adds no -paper key`: `--diff` and the dry run both name the same skip, and the staged config holds `discord.channels.hyperliquid` once with no `-paper` key and no empty `trade_alert_channels` or `dm_channels` map.

The pre-existing cases that pin the unchanged behavior stay as they are: the differing-value compose at `scripts/test_merge_paper_instance.sh:252`, the exit-21 refusal on a conflicting live `-paper` key, and the type-keyed self-conflict previews.

---

https://claude.ai/code/session_01TcGCvzqxXsjzJwkCHZaxRo


richkuo:
@claude review

richkuo:
## Review disposition

Addressed the review on `c6a15eb`. Head is now `75e61a0`. One finding, one follow-up issue.

### Recommended Optional

**1. Pruning the last `channels.<platform>-paper` key lets paper-scope kill-switch alerts leak to live-only channels — ✅ Confirmed, fixed here.**

The mechanism holds exactly as described. `SendToScopeChannels` scopes a paper broadcast by scanning `b.channels` for keys ending in `-paper` and falls through to `SendToAllChannels` when it finds none (`scheduler/notifier.go:267-289`); `SendToAllChannels` sends to every distinct non-empty value in that map (`scheduler/notifier.go:204-215`). `b.channels` is `cfg.Discord.Channels` or `cfg.Telegram.Channels` (`scheduler/notifier.go:140,145`), so `discord.channels` is the only one of the three maps with a scope-broadcast reader. The prune could remove the last `-paper` key there, which is a defect this PR introduces.

Scope: PR-caused, so in scope for this PR however much mechanism the fix needs. It is also a protection-surface path (a kill-switch alarm), so it gets fixed rather than deferred.

Fix (`scripts/merge-paper-instance.sh`): `apply_paper_discord_maps` now models the notifier's own choice. `paper_scope_broadcast_values(mm, dropped)` returns the `-paper` values when any survive and every value otherwise, matching `SendToScopeChannels` and its fall-through. For `map_key == "channels"` only, the prune is held back whenever the post-prune set differs from the pre-prune set. The kept key reports as `discord.channels.<k>=<v> (kept: dropping it would change which channels paper-scope alerts reach)`, so `--diff` names it as kept rather than silently adding it.

The guard blocks a **narrowing** prune too, which the finding did not raise: with `channels: {hyperliquid: C-hl, okx: C-okx}` and a paper side setting `hyperliquid: C-hl` plus `okx: C-okx-paper`, pruning `hyperliquid-paper` while `okx-paper` survives would stop paper alerts reaching `C-hl`, a channel they reach today. Losing a protection alert is worse than over-sending one, so the same equality test covers it.

The prune pass moved after the explicit `-paper` passthrough loop so the broadcast set is measured on the complete merged map, including keys the paper config set explicitly. Report line order is unchanged.

`trade_alert_channels` and `dm_channels` keep the unrestricted prune: no code reads either map by suffix scan, and both per-strategy resolvers fall back.

Your three "must survive" states, each asserted:

1. First merge where `channels.<platform>` matches live and a distinct live-only platform channel exists → `hyperliquid-paper` is kept; the paper broadcast reach is identical before and after. New integration case in `scripts/test_merge_paper_instance.sh` (`an identical paper channel keeps its -paper key when a live-only channel exists`), which also asserts the redundant `dm_channels` key is still skipped in the same run.
2. Paper config with an explicit `channels.<platform>-paper` key → still kept, via `pinned`, unchanged.
3. Merged config already holding a live `channels.<platform>-paper` key → a pre-existing key is never in `added`, so it is never a prune candidate; isolation holds.

Verified with a direct harness over the embedded merge helper: 16 cases, including the three above plus the narrowing case, all green after the fix; the three isolation cases and the new integration assertion are red against `c6a15eb`.

### Created follow-up issue

**#1551 — `[C58] Paper-scope alerts go to live channels when no channel key ends in -paper`.** The `-paper` suffix scan is a proxy for "this channel serves the paper scope", and the proxy is wrong for any hand-written config whose paper strategies route through a bare `platform` or strategy-type key, with or without a merge. Fixing it needs a paper-scope channel set that `MultiNotifier` does not have today (scope rule 2: new mechanism the yardstick never asked for), so it is filed rather than built here. The issue names the merge-script guard as the workaround to remove once the notifier resolves paper channels itself.

### Test edits

None. `scripts/test_merge_paper_instance.sh` gains one case; no existing assertion was weakened, narrowed, or removed.

### Verification

- `go -C scheduler test -run TestMergePaperInstance -count=1 .` — pass. Red against `c6a15eb` at `--diff says the channel key is kept for paper-scope alerts`.
- `go -C scheduler build .` — pass.
- `python3 scripts/check_test_budget.py` — `wording-only tests: 172 (baseline 172)`.
- `go -C scheduler test ./...` — one failure, `TestUpdateHelpersEnvfileParsing790`, pre-existing and local-only: it reads `scripts/update.sh` and `scripts/update_helpers.sh`, which this branch never touches, and the `go` CI job passed on `c6a15eb`.
- `bash -n` on both scripts — pass.

Docs: `SKILL.md` and the `--help` text now state the guard and name `scheduler/notifier.go:267` as the reason.

---
Updated with LLM: Opus 5 | xhigh | Harness: fix-pr-review

richkuo:
@claude sonnet review

richkuo:
## Review disposition

Addressed the `Needs Updates` review on `75e61a0`. Head is now `39f1fba`. One blocking finding, confirmed and fixed; one follow-up issue filed.

### Needs Fixing

**1. Pruning a redundant `dm_channels` `-paper` key permanently silences paper trade-alert DMs — ✅ Confirmed, fixed.**

Every step of the claim traces. `tradeAlertRoutes` sets `dmKey = platform + "-paper"` for a paper strategy and reads `b.dmChannels[dmKey]` as a plain map index (`scheduler/notifier.go:315-321`). There is no fallback to the bare `platform` key, none to the strategy-type key. With the key gone, `dmDest` is `""` and both DM sends are skipped by their own `route.dmDest != ""` guards: the per-trade alert (`scheduler/main.go:3404`, reached from `sendTradeAlerts` at `scheduler/main.go:2781`) and the LLM entry-analysis digest (`scheduler/main.go:447`).

The contrast with the other two maps holds as you described. `resolveChannel` falls back `platform` then `stratType` (`scheduler/discord.go:144-152`); `resolveTradeAlertChannel` falls back `<platform>-paper` → `platform` → `stratType` → `resolveTradeChannel` over the base `channels` map (`scheduler/discord.go:163-181`). `merged_channel_route_key` mirrors those two and mirrors nothing in the DM path.

The blindness of the proof step traces too: `inspect` resolves `dm_key`/`dm_channel` through `resolveChannelKeyOverMaps` (`scheduler/inspect_cmd.go:739`, `:750`), which tries `<platform>-paper` and then falls back to the bare keys, so the pre-move and merged views compare equal while the real send path has lost its route.

Scope: PR-caused (the prune is this PR's), so in scope under rule 1, and a silent notification loss on a trade path is a safety-carve-out fix regardless.

Fix: the skip is now limited to maps whose runtime resolver falls back off the `-paper` key, named in `CHANNEL_MAPS_WITH_BARE_KEY_FALLBACK` = `channels`, `trade_alert_channels`. `dm_channels` keys are always written. `channels` keeps the paper-scope broadcast guard from the previous cycle on top of that.

Your three "must survive" states:

1. Paper `dm_channels` sets the bare `<platform>` key to the value the merged bare key already holds → the `-paper` key is written. Asserted in the `scopeisolation` case, which now requires `discord.dm_channels.hyperliquid-paper` to equal `D-live` in the staged config and to appear in the `--diff` channel plan.
2. Paper `dm_channels` sets the bare `<type>` key instead → new case `a type-keyed dm map still gets its -paper key`: live and paper both carry `dm_channels: {perps: D-perps}`, the merged config gets `dm_channels.hyperliquid-paper = D-perps`, and the `channels` map still skips its redundant key in the same run, so the two maps are shown diverging by rule and not by accident.
3. Paper `dm_channels` sets `<platform>-paper` explicitly → still pinned, untouched by the fix.

### Created follow-up issue

**#1552 — `[C54] inspect reports a paper DM channel the send path never uses`.** The reporting gap you identified is real and pre-existing: `inspect`'s DM view falls back off `<platform>-paper` while `tradeAlertRoutes` does not, so a paper strategy with only a bare `dm_channels` key shows a DM route it does not have, and the merge proof cannot see a DM route being lost. Fixing it changes what the proof compares for every strategy in both merge directions, a cross-cutting invariant this PR's yardstick (#1549) never asked for, so it is filed rather than built here. This PR no longer depends on it: the merge never drops a `dm_channels` key now, so the hazard is out of the merge path.

### Test edits

Two assertions in `scripts/test_merge_paper_instance.sh`, both added by this PR's own previous commit, **case Wrong**: they asserted the redundant `dm_channels` `-paper` key is skipped, which was never correct behavior. **Ground:** `scheduler/notifier.go:315-321` resolves a paper strategy's DM destination on that literal key with no fallback, so the key is load-bearing and dropping it silences the DM. They now assert the key is written and named in the `--diff` channel plan. No other test changed; one new case was added.

### Verification

- `go -C scheduler test -run TestMergePaperInstance -count=1 .` — pass. Red against `907c0326` at `the dm map has no bare-key fallback, so its key is always added`.
- `go -C scheduler build .` — pass.
- `python3 scripts/check_test_budget.py` — `wording-only tests: 172 (baseline 172)`.
- `bash -n` on both scripts — pass.
- `go -C scheduler test ./...` — one failure, `TestUpdateHelpersEnvfileParsing790`, which fails only on this macOS checkout; CI's `shell` job runs that same script and passes.
- CI was green on `907c0326` (all four jobs) after a separate commit fixed `TestModelOnlyClose_HedgeLegAdjustsDailyPnLNeverStreak`, which failed in the first minute of a UTC day and is unrelated to this review.

Docs: `SKILL.md` and the `--help` text now state that `dm_channels` keys are always added, and name `scheduler/notifier.go:313` and `scheduler/inspect_cmd.go:750` as the reason.

---
Updated with LLM: Opus 5 | xhigh | Harness: fix-pr-review

richkuo:
@claude review

PR #1554, opened 2026-09-11
fix(#1552): inspect reports the DM route the send path uses [C66, Grok 4.6, xhigh]

### LINKED ISSUE #1552: [C66] inspect reports a paper DM channel the send path never uses
**Complexity: 66/100** — Capability 2 (Risk 3, Uncertainty 3 — compose still promotes a bare or type DM key into `-paper`, and exact-key inspect makes that dry-run a proof refuse; this issue keeps that write and treats the refuse as required); Volume 16 (Scope 3, Coupling 2, Verification 3) — Opus 5, xhigh · fableplan: no

## Problem

Two resolvers answer "which direct message channel does this strategy use", and they disagree.

The send path is `MultiNotifier.tradeAlertRoutes` (`scheduler/notifier.go:343-352`). For a paper strategy it sets `dmKey = platform + "-paper"`; for a live strategy it sets `dmKey = platform`. It reads `b.dmChannels[dmKey]` directly. There is no fallback: no bare `platform` key on paper, and no strategy-type key in either mode. With no matching key in `dm_channels`, `dmDest` is `""` and the DM is skipped, for the per-trade alert (`scheduler/main.go:3404`, reached from `sendTradeAlerts` at `scheduler/main.go:2781`) and for the LLM entry-analysis digest (`scheduler/main.go:447`).

The reporting path is `inspect`, which resolves `dm_key` and `dm_channel` with `resolveChannelKeyOverMaps` (`scheduler/inspect_cmd.go:739`, `:750`). That helper tries `<platform>-paper` first and then falls back to the bare `platform` key and the strategy-type key, the same shape the *channel* resolvers use (`scheduler/discord.go:144-161`). The fallback is right for `channels`; it is wrong for `dm_channels`. `TestBuildStrategyInspectionJSONScopeFields` pins that fallback today: `dm_channels` holds only `{perps: D-all}`, and both a live strategy and a paper strategy expect `dm_channel` `D-all` (`scheduler/inspect_scope_fields_test.go:17-59`). The send path would send no DM in either case.

So a paper strategy whose `dm_channels` holds only a bare `hyperliquid` key sends no DM at all, while `inspect --all --json` reports `dm_channel` as that bare key's value and `dm_key` as `hyperliquid`. An operator reading the inspect output sees a DM route that does not exist.

This matters most in `scripts/merge-paper-instance.sh`. Its proof step compares each moved strategy's pre-move `inspect` view against the merged view and refuses on a difference, and notification channels are part of that view. The proof strips keys that end in `_key`, so only `dm_channel` is compared (`scripts/merge-paper-instance.sh:793-794`). Because both sides read the forgiving resolver, a change that removes a paper strategy's real DM route compares equal and passes the proof. PR #1550 hit exactly this: a prune of a redundant `dm_channels` `-paper` key would have silenced the paper DM, and the merge's own proof could not see it. That PR now never skips a `dm_channels` key (`CHANNEL_MAPS_WITH_BARE_KEY_FALLBACK` is only `channels` and `trade_alert_channels`, `scripts/merge-paper-instance.sh:200-201`, `422-423`), which removes the drop hazard from compose and leaves the reporting gap in place.

Compose is a third resolver. `apply_paper_discord_maps` still reads each paper map with fallback (`<platform>-paper`, then bare `platform`, then strategy type) and writes `<platform>-paper` (`scripts/merge-paper-instance.sh:385-402`). A paper config whose `dm_channels` holds only `hyperliquid` or only `perps` therefore stages a new `-paper` key. Today's inspect fallback makes `dm_channel` match on both sides, so those dry-runs pass (`scripts/test_merge_paper_instance.sh:795`, `:816`). After inspect matches the send path, paper inspect is empty and staged inspect is not, so the proof must refuse.

## Goal

`inspect` reports the DM route the send path would actually take, so an operator and the merge proof both read the truth. Compose keeps writing every `dm_channels` `-paper` key it already writes.

- A paper strategy whose `dm_channels` holds no `<platform>-paper` key reports an empty `dm_key` and `dm_channel`.
- A paper strategy whose `dm_channels` holds `<platform>-paper` reports that key and value, unchanged.
- A live strategy whose `dm_channels` holds a bare `<platform>` key reports that key and value, unchanged.
- A live or paper strategy whose `dm_channels` holds only a type key reports an empty `dm_key` and `dm_channel`.
- The channel and trade-alert views keep their current fallback, which matches their own resolvers.
- The merge proof refuses when a merge would drop a paper strategy's real DM route.
- The merge proof refuses when paper inspect reports no DM and the staged inspect would send one.

## Approach

1. Split the DM resolution in `scheduler/inspect_cmd.go` out of `resolveChannelKeyOverMaps` into a helper that mirrors `tradeAlertRoutes`: `<platform>-paper` for a paper strategy and bare `<platform>` for a live one, with no type-key fallback on either.
2. Leave `resolveChannelKeyOverMaps` as it is for `channel_key`/`channel`, and leave `resolveTradeAlertKeyOverMaps` alone; both mirror resolvers that do fall back.
3. Add a Go test over a paper strategy whose `dm_channels` carries only a bare `platform` key: `inspect` reports an empty DM view, and the same config with `<platform>-paper` reports that key. Update `scheduler/inspect_scope_fields_test.go` so the `{perps: D-all}` fixture expects an empty DM view for both the live strategy and the paper strategy.
4. Do not change compose: `dm_channels` keys are still always added. Retarget `scripts/test_merge_paper_instance.sh`: the "never prune" dry-run must put an explicit `<platform>-paper` key on the paper map so inspect agrees and the proof still shows the key is kept; the promote-from-bare and type-key dry-runs (`:795`, `:816`) must expect a proof refuse, because paper inspect is empty and staged inspect is not.

## Acceptance criteria

- A paper strategy with `dm_channels` holding only `hyperliquid` reports `dm_key` and `dm_channel` as empty in `inspect --all --json`.
- The same strategy with `dm_channels` holding `hyperliquid-paper` reports that key and its value.
- A live strategy with `dm_channels` holding `hyperliquid` reports that key and value, unchanged.
- A live or paper strategy with `dm_channels` holding only `perps` reports `dm_key` and `dm_channel` as empty.
- The `channel_key`, `channel`, `trade_alert_key` and `trade_alert_channel` fields are unchanged for every strategy in every mode.
- A merge that would leave a paper strategy with no `<platform>-paper` DM key, where it had one before, refuses at the proof step instead of passing.
- A merge whose paper inspect DM is empty and whose staged inspect DM is set refuses at the proof step.
- `go test ./...` and the paper handoff tests pass under those retargeted expectations.

## Plain simple English

The program that sends a trade direct message looks up one exact name. The program that reports the setup also accepts other names. So a strategy can look as if it has a direct-message channel when none is used, and the merge check can accept a change that starts sending those messages.

---
Created with LLM: Opus 5 | xhigh | Harness: fix-pr-review

richkuo:
Closes #1552

## Summary

- `inspect` now resolves `dm_key`/`dm_channel` with `resolveDMKeyOverMaps`, which mirrors `tradeAlertRoutes`: `<platform>-paper` for paper, bare `<platform>` for live, no type-key fallback.
- Channel and trade-alert inspect fields still use the fallback helpers.
- Compose still always writes `dm_channels` `-paper` keys. The merge proof now refuses when paper inspect reports no DM and the staged inspect would send one.
- Plan: none.

### Test edits

- `TestBuildStrategyInspectionJSONScopeFields` (`scheduler/inspect_scope_fields_test.go`) — Outdated. Ground: issue #1552 Goal that inspect reports the send-path DM route; `tradeAlertRoutes` (`scheduler/notifier.go:345-352`) looks up only that exact key. `wantDM` is empty for the `{perps: D-all}` fixture, and the assertion now includes `dm_key`.
- `scripts/test_merge_paper_instance.sh` never-prune dry-run — Outdated. Ground: the same Goal. Paper now carries an explicit `hyperliquid-paper` key so inspect agrees and the proof still shows the key is kept.
- `scripts/test_merge_paper_instance.sh` type-keyed dm dry-run — Outdated. Ground: issue #1552 Approach step 4. It now expects exit 22 and a `notification.dm_channel` proof refuse; compose still writes the `-paper` key first.

### Verification

- `TestNotificationRoutingJSONDMExactKey` failed on the unfixed helper (paper bare/type and live type reported a fallback DM) and passes after the exact-key split.
- `go -C scheduler test -run 'TestNotificationRoutingJSONDMExactKey|TestBuildStrategyInspectionJSONScopeFields' .` passes.
- `go -C scheduler test -count=1 -run 'TestMergePaperInstance$' .` passes.
- `gofmt -l` is clean on the edited Go files.
- `go -C scheduler test ./...` is otherwise green except `TestUpdateHelpersEnvfileParsing790`, which fails on the unchanged base on this macOS host and is deferred to CI.

## Plain simple English

Inspect now reports the same direct-message channel the send path uses. A paper strategy with no name that ends in -paper shows none. The merge check then refuses a merged setup that would start sending those messages.

---
Created with LLM: Grok 4.6 | xhigh | Harness: Cursor


richkuo:
@claude revidew

richkuo:
@claude review

PR #1560, opened 2026-09-14
test(update.sh): canonicalize temp dir for macOS in update helper test [C15, Fable 5.1, low]

richkuo:
## Plain simple English

One shell test always failed on a Mac. The test made a temporary folder, but the Mac reported that folder with a different path name. The test now uses the real path name, so it passes on a Mac and on Linux.

## Summary

- `scripts/test_update_helpers.sh`: resolve the fleet temp dir with `pwd -P` after `mktemp -d` so the child `cwd` record matches the assertion prefix. macOS returns `/var/folders/...` from `mktemp` but `pwd` resolves to `/private/var/...`.

## Verification

- `go -C scheduler test -run TestUpdateHelpersEnvfileParsing790 .` passes on macOS. It failed on `main` before this change.

---

https://claude.ai/code/session_01FWE685Xjt1rpYXc32nUqq3


richkuo:
@claude review

Issue #1561, opened 2026-09-15
[C99] Support several paper sources in one production scheduler and dashboard


richkuo:
**Complexity: 99/100** - Capability 3 (Risk 4, Uncertainty 4: production cutover and stored-book integrity; the storage-group and risk-partition contracts need a plan); Volume 24 (Scope 4: scheduler, storage, configuration, dashboard, deployment tools, tests, and operator documents with new ownership contracts; Coupling 4: process ownership, database locks, restart, and Go/Python parity; Verification 4: interrupted cutover, restart recovery, and live-service proof) - Opus 5, xhigh · fableplan: yes

## Problem

The operator wants one production scheduler and one dashboard for live trading and several paper sources. The reported deployment inventory has one live unit, five paper coin units, and a separate paper-testing unit. This inventory and the claim that cutover has never run are operator reports; repository validation did not inspect the host.

Validation against origin/main at `623566b6` found these limits:

- `PortfolioScope` classifies strategies as Live or Paper from their execution arguments. It does not identify each source deployment or coin. `activeScopes` returns only those two modes. The existing risk and correlation maps therefore cannot provide independent paper-source views. See the [scope classifier and membership helpers](https://github.com/richkuo/go-trader/blob/623566b6/scheduler/portfolio_scope.go#L9).
- The storage layout supports a primary database and one optional paper database. `StateStore` routes through those roles. A second merge from a different paper database is explicitly refused. Repeated single-source merges cannot achieve the proposed topology. See the [storage layout](https://github.com/richkuo/go-trader/blob/623566b6/scheduler/storage_layout.go#L180) and [merge refusal](https://github.com/richkuo/go-trader/blob/623566b6/scripts/merge-paper-instance.sh#L1011).
- The handoff requires both live and paper services to stop, then holds database ownership locks. Storage changes also require a restart. Uninterrupted live execution is not provided by this mechanism. See the [service check](https://github.com/richkuo/go-trader/blob/623566b6/scripts/merge-paper-instance.sh#L883) and [reload checks](https://github.com/richkuo/go-trader/blob/623566b6/scheduler/config_reload.go#L428).
- The main dashboard table already filters All, Paper, and Live. It has no source-deployment selector. The overview feed is a separate application programming interface (API) endpoint and must be included in the change. See the [table filter and overview request](https://github.com/richkuo/go-trader/blob/623566b6/scheduler/static/ui/app.js#L2030).
[...]

PR #1562, opened 2026-09-15
feat(#1561): runtime ownership, storage and risk partitions for several paper sources [C99, Opus 5, xhigh, fableplan]

richkuo:
Part of #1561.

This is the first of the four pull requests the adopted [implementation plan](https://github.com/richkuo/go-trader/issues/1561#issuecomment-5675685971) defines. It delivers plan steps 1 to 8: the runtime ownership, storage and risk partitions. The plan reserves `Closes #1561` for its fourth pull request, after the handoff script and the dashboard land.

## Summary

The scheduler can now own several independent paper deployments in one process.

- **Ownership model.** `RiskPartition{Scope, Source}` is the sole risk and storage owner, with the text form `live`, `paper` and `paper:<id>`. Execution mode stays `PortfolioScope`, still classified only by `isLiveArgs`. A new root list `paper_sources` (`id`, `label`, `db_file`, optional `portfolio_risk`) plus the per-strategy key `paper_source` give each source a stable identity that survives restart and never depends on an alias or a coin name.
- **Storage.** One file per partition, resolved in the stable order primary, paper, then sources by id. That order is the single lock and save order every process and tool takes. Every pair must be a distinct physical file. Stored books are untouched: a source database keeps writing scope `'paper'` exactly as it did standalone, so the existing `CHECK` constraint stays valid and no migration runs. Loading maps `(file, scope)` onto the partition that file owns, so equal stored identifiers in different source databases stay distinct.
- **Failure isolation.** Save failures, persistence holds and the three-strike cycle skip are per partition. A diagnostics update routes by the row's `SourceRole`, because with several paper files a scope alone cannot choose a file.
- **Risk.** Cycle risk, peaks, latches, daily loss, exposure caps, correlation, portfolio warnings and the paper kill-switch close each run once per partition, on that partition's own roster. Limits layer root, then `portfolio_risk.paper`, then the source override, with zero inheriting the layer above. The kill-switch reset DM accepts `reset paper:<id>`, lists the latched partitions, and refuses a bare `reset` when more than one is latched.
- **Operator surfaces.** Channel routing tries `<platform>-paper:<id>` before `-paper` before the bare key; the DM key stays an exact match with no fallback, so a moved book never DMs another source's route. `/status` keys every by-scope map by partition text, adds `paper_sources`, and stamps each strategy with its partition. `storage-inspect` prints the partitions each file owns, and `inspect --all --json` carries `partition` and `paper_source`. SIGHUP refuses adding, removing or re-pointing a source, moving a strategy between sources, and a per-source notional-cap change.

Refusals owned by `validatePaperSourcesConfig`: a blank, reserved (`live`/`paper`/`primary`), malformed or duplicate id; a blank `db_file`; an unknown `paper_source` reference; a source named by a live strategy; a nested `paper` override; and, in the layout, two partitions resolving to one physical file.

A config with no `paper_sources` behaves exactly as before, in both the single-file and the split layout.

### Verification

Run locally on macOS at base `623566b6`:

- `go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler/` — clean.
- `go -C scheduler test ./...` — pass, including `TestMergePaperInstance`, which drives the `merge-paper-instance.sh` shell fixtures.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2943 passed.
- `uv run --no-sync python -m pytest shared_scripts/ scripts/` — 327 passed.
- `bash scripts/test_update_helpers.sh` — pass.
- `uv run --no-sync python scripts/check_test_budget.py` — 172 wording-only tests, baseline 172.

New behaviour is covered by `scheduler/paper_source_test.go`: partition text round trip and refusals, partition order, the config refusal table, limit layering, the layout's four same-file refusals and lock order, equal stored identifiers across three source files, a three-source save and restart round trip, risk isolation (one source's latch, force-close and reset leave the others untouched), per-partition save failures and the due-set drop, diagnostics routing by role, and the channel and DM routing ladder.

Nothing is deferred to CI beyond the Linux-only checks CI always runs; there is no release check left outstanding for this pull request. Cutover itself (plan steps 17 and 18) runs on the host after the fourth pull request merges.

### Test edits

Disclosed per CLAUDE.md. Every other test change in this diff is a mechanical rename of a helper this change replaced, with no assertion changed.

- `TestPaperScopeChannelValues` (`scheduler/portfolio_scope_persistence_test.go`) — **Outdated**. Ground: plan step 7 replaces the single paper destination list with a per-partition set. It now indexes the returned map by partition and adds three source cases.
- `TestParseKillSwitchResetReply` and `TestKillSwitchResetPromptForScopes` (`scheduler/portfolio_scope_test.go`) — **Outdated**. Ground: plan step 6 requires the reply options to name each latched partition. The refusal wording moved from a fixed "reset live or reset paper" to the generated option list, and five source cases were added.
- `TestConfigExampleCopyDoesNotTriggerTheUpgradePath` (`scheduler/config_example_version_test.go`) — **Outdated**. Ground: plan step 1 adds a named source to `scheduler/config.live-paper.example.json`, so the example's third strategy now stores in that source's file.

### Deviations from the adopted plan

- **`update_resolve_db_exclude` is pulled forward** from plan step 9 (its second pull request). A configured source database that `scripts/update.sh` does not exclude is a data-integrity risk from the moment `paper_sources` loads, so it ships with the runtime that makes it reachable. The rest of step 9 stays where the plan put it.
- **CLAUDE.md grows by about 509 bytes**, against plan step 15's "does not grow" constraint. The new ownership subsystem needs guardrails that wording compression alone could not fund, and the file was already over its own 15k target before this change. Plan step 15 owns the rebalance.

No other step was overridden; the traced code matched the plan everywhere else.

## Plain simple English

The scheduler can now hold several paper deployments at the same time. Each one keeps its own database file, its own risk limits, its own kill switch and its own alert channels. A problem in one paper source cannot close, reset or overwrite another one, or the live books. Nothing changes for a setup that has no paper sources.

---


richkuo:

richkuo:
Addressed review feedback (claude-fable-5 · 2026-09-15T08:50:31Z) in 7c8576e7.

### Fixed

1. **`validateDMChannelsMap` refuses the `-paper:<id>` DM keys the new source routing reads** — new `parseDMChannelKey` (`scheduler/config.go:2194`) accepts exactly the three shapes `paperChannelKeys` builds (`<platform>`, `<platform>-paper`, `<platform>-paper:<source id>`), with the id checked against `paperSourceIDPattern`, so `hyperliquid-paper:btc` loads while `hyperliquid-paperx`, `hyperliquid-paper:` and `hyperliquid-paper:BTC!` stay refused and the platform-typo warning still derives `hyperliquid` (`scheduler/config.go:2208`); a key naming an undeclared source now warns that the route can never fire.

2. **The paper kill-switch channel message instructs a fixed `reset paper` reply that can clear the wrong partition's latch** — `paperKillSwitchHeader` and `paperKillSwitchManualResetLine` are now functions of the partition (`scheduler/risk.go:1435`, `scheduler/risk.go:1443`), so each broadcast names the partition that fired and prints the exact reply that clears it (`reset paper:btc`), which `parseKillSwitchResetReply` resolves to that same partition whether one or several are latched; `formatPaperKillSwitchMessage`, `formatPaperKillSwitchPromptMessage` and `formatPaperKillSwitchAutoResetMessage` take the partition and the auto-reset substitution replaces that partition's own manual line. Every default-paper string is byte-identical to before.

3. **CLAUDE.md cites `scheduler/notifier.go:345` for a rule whose code this diff moved** — the brittle line anchor is dropped in favour of the `tradeAlertRoutes` function name already in the sentence (`CLAUDE.md:20`), so the citation survives the next move; `SKILL.md:825` carried the same stale anchor plus a `<platform>-paper` key description this diff made incomplete, and both are corrected there (scope rule 3: a stale citation in executable agent-instruction surface, no new mechanism).

### Test edits

1. **`TestPaperKillSwitch_ClosesOnlyPaperBooks`, `TestPaperKillSwitch_AutoResetsWithoutOwner`, `TestPaperKillSwitchPromptMessage_CarriesReason`, `TestKillSwitchResetPromptForScopes`** (`scheduler/portfolio_scope_test.go:459`, `:757`, `:760`, `:772`, `:823`) — Outdated; ground: review finding `### Needs Fixing` item 2, which authorizes the partition-specific message API. Each call site now passes the partition; the assertions are unchanged in substance, since every default-paper string the tests match is byte-identical, so each test still fails on the same defect it guarded before.

New coverage for the two invariants: `TestConfigValidationDMChannels` (`scheduler/config_test.go`) gains three malformed-key refusals and one case where a valid `hyperliquid-paper:btc` key both passes `LoadConfig` and is read back by `tradeAlertRoutes`, crossing config validation and the send path in one case — the gap the review named. `TestPaperKillSwitchMessageNamesItsPartition` (`scheduler/paper_source_test.go`) pins, for the default paper partition and for `paper:btc`, that the reply the broadcast prints is the reply `parseKillSwitchResetReply` resolves to that same partition while both are latched.

Verification: `go -C scheduler build .`, `go vet ./...` and `go test ./...` all pass; `gofmt -l scheduler/` clean; `scripts/check_test_budget.py` at baseline 172.

---

richkuo:

richkuo:
Addressed review feedback (Claude Opus 5 · 2026-09-15T09:15:06Z) in 9f0604a3.

### Not changed (refuted)

1. **A `paper_source` whose file does not already hold that strategy's books starts it empty, then deletes the stored book on the first save** — the traced chain omits `inspectStorageOwnership`, which `scheduler/main.go:187` runs **before** `LoadStateWithStore` at `scheduler/main.go:247` and before any write. For every stored id a file does not own, `inspectOneStorageFile` calls `foreignRoleForStorageID` (`scheduler/storage_inspect.go:154`, `scheduler/storage_inspect.go:262`); when that id is mapped to another file in the layout it appends a rejection, `si.OK()` (`scheduler/storage_inspect.go:52`) turns false, and `scheduler/main.go:198` exits `ExitStorageOwnership` (80) with an owner DM. Running the finding's own scenario — `hlp` stored in the primary (or in `paper_db_file`), then `paper_source: "btc"` added — inspection returns `primary state file: stored strategy "hlp" is mapped to the paper:btc state file; move or re-map it by hand` and refuses. The prune WARN at `scheduler/main.go:334` is reachable only for a stored id mapped to **no** file in the layout, which is the retired-strategy case the branch exists for; a `readOnlyReport` run skips the inspection but opens every file through `OpenStateStoreReadOnly` (`scheduler/main.go:216`) and can never save. The stated `Reachability` is therefore refuted at its first step, and the invariant the finding asks for is the one `CLAUDE.md` already records as "unknown ownership errs BEFORE mutation". Both must-survive cases are likewise already satisfied: the documented fold starts clean and a retired strategy still prunes.

### Fixed

1. **A declared paper source with no `dm_channels` key loses its trade DMs silently, while the mirror-image mistake warns** — new `warnPaperSourceDMGaps` (`scheduler/config.go:2194`), called for both backends beside `validateDMChannelsMap`, warns once per missing `<platform>-paper:<id>` key when the plain `<platform>-paper` key is present, naming both the key to add and the key that no longer routes; it stays silent when the source carries its own key, when `dm_channels` is absent, and for an unrelated platform. Scope rule 1: the sourced DM key that supersedes the plain one is added by this PR, so the silent drop is a hazard this PR creates.

### Test edits

None. Both changes are additions; no existing test was edited, weakened, or removed.

New coverage: `TestPaperSourceRelocationRefusedBeforeAnyWrite` (`scheduler/paper_source_test.go`) pins the refuted finding's own invariant, which had no test before — nothing in the suite exercised `inspectStorageOwnership`. It asserts that books left in the primary and books left in `paper_db_file` are both refused before any save, with the refusal naming the moved book and the file that now owns it, while the documented fold still starts clean and a retired strategy still prunes. `TestPaperSourceDMGapWarning` asserts the new warning names `dm_channels["hyperliquid-paper:btc"]` and the plain key, and stays silent in the three cases that must not warn.

Verification: `go -C scheduler build .`, `go vet ./...` and `go test ./...` all pass; `gofmt -l scheduler/` clean; `scripts/check_test_budget.py` at baseline 172.

---

richkuo:
@claude review

PR #1563, opened 2026-09-15
feat(#1561): pair folded paper-source aliases in the drift audit [C99, Opus 5, xhigh, fableplan]

richkuo:
Part of #1561. Follows #1562, which is merged.

This is the second of the four pull requests the adopted [implementation plan](https://github.com/richkuo/go-trader/issues/1561#issuecomment-5675685971) defines, covering plan steps 9 to 11. The plan reserves `Closes #1561` for its fourth pull request.

## Summary

The operator tooling now reads a folded paper source's strategy alias.

- **Shared alias rule (step 10).** `scripts/paper_alias.py` gains `paper_alias_suffix(source)`: the bare `-paper` suffix with no source, `-paper-<id>` for a named one. It refuses a reserved id (`live`, `paper`, `primary`) and any id outside the scheduler's `^[a-z0-9][a-z0-9_-]{0,31}$` pattern, so the shell tools and the scheduler agree on which ids can name a source. `paper_alias_base(sid, source=None)` reads that suffix; the legacy no-source path is unchanged, including the `-paper<n>` collision form. A suffix-only id now reports no base instead of an empty one.
- **Drift audit.** `check-live-paper-config-drift.sh` passes each strategy block's own `paper_source`, so a `-paper-<id>` suffix counts as an alias only when the block names that source. An id that merely looks like one (`hl-paper-rsi`) is still not an alias, and pairing still requires `storage_strategy_id` to equal the base, exactly as the bare `-paper` path does. The UNPAIRED and AMBIGUOUS lines name the suffix they read.
- **Pairing keys are not drift.** `paper_source` joins `id`, `replay_source_id` and `storage_strategy_id` in the audit's ignored-key set. A live twin can never carry a `paper_source`, so without this every folded pair would report an OTHER difference and land on SKIP, leaving a folded fleet uncertifiable.
- **Steps 9 and 11 already shipped** with the runtime in #1562: `update_resolve_db_exclude` already prints every `paper_sources[].db_file`, and `readStateFilesReadOnly` already labels each open position with its file's partition. `update_paper_override_directive` is already generic over a directory, and its per-source reuse belongs to the handoff script in plan step 12. This pull request adds step 11's missing verify point.

### Verification

Run locally on macOS at base `19e6efb9`:

- `go -C scheduler build .`, `go -C scheduler vet ./...`, `gofmt -l scheduler/` — clean.
- `go -C scheduler test ./...` — pass, including `TestMergePaperInstance`, which drives the `merge-paper-instance.sh` shell fixtures.
- `bash scripts/test_update_helpers.sh` — pass with and without `GO_TRADER_BIN`, so the new folded-source case is proven against the binary's own `inspect --all --json` and not only against raw config keys.
- `uv run --no-sync python -m pytest shared_scripts/ scripts/` — 329 passed.
- `uv run --no-sync python -m pytest shared_strategies/ shared_tools/ platforms/ backtest/` — 2943 passed.
- `uv run --no-sync python scripts/check_test_budget.py` — 172 wording-only tests, baseline 172.

New behaviour is covered by:

- `scripts/test_paper_alias.py`: the suffix table (no source, blank source, a valid id, each reserved id, an uppercase id, a leading hyphen, a separator character, an over-long id) and the source-aware base table (match, collision form, wrong source, bare `-paper` under a named source, a named-source id read with no source, a reserved source, a suffix-only id).
- `scripts/test_update_helpers.sh`: a folded source that pairs through `storage_strategy_id`, the same alias without it reporting AMBIGUOUS and gating, a `-paper-<id>` id with no `paper_source` that is not read as an alias, and an effective-view case where the deployment's own binary supplies the inspect output.
- `scheduler/agent_info_test.go`: `TestReadStateFilesReadOnlyLabelsPaperSourcePartitions`, a five-file fixture asserting `live`, `paper` and each `paper:<id>` label on the snapshot and in its note, plus the role fallback when one file owns two partitions.

Nothing is deferred to CI beyond the Linux-only checks CI always runs.

### Test edits

Disclosed per CLAUDE.md.

- `test_the_suffix_rule_is_defined_once` (`scripts/test_paper_alias.py`) — **Outdated**. Ground: plan step 10 moves the suffix construction into the shared module, so the single-definition guard must name the new function too. The regex now also matches `paper_alias_suffix`, and the assertion still requires exactly `paper_alias.py` to define the rule.
- No other test was changed. Every other test difference in this diff adds a case or a file.

### Deviations from the adopted plan

- **Plan steps 9 and 11 were already delivered by #1562**, so this pull request carries step 10 plus step 11's verify point rather than three steps of new code. Step 9's `update_resolve_db_exclude` was pulled forward with disclosure in that pull request; nothing in step 9 is left outstanding.
- **The audit's ignored-key set gains `paper_source`**, which the plan does not name. Without it the audit reports a false difference on every folded pair, which would defeat the pairing the same step adds.

No other step was overridden; the traced code matched the plan everywhere else.

## Plain simple English

A folded paper deployment renames its strategies with a source suffix. The drift audit can now read that suffix and match each paper strategy to its live twin, but only when the strategy says which source it belongs to. This keeps the audit able to certify a merged fleet, and it still refuses to guess.

---


richkuo:

richkuo:
Addressed review feedback (claude-fable-5 · 2026-09-15T13:50:10Z) in 0dafb3d8d54858680253e383e2a6216432006746.

### Fixed

1. **The audit silently certifies a folded twin whose id keeps the bare `-paper` alias while `paper_source` is set** — Confirmed, and caused by this PR. `alias_read` (`scripts/check-live-paper-config-drift.sh:137-147`) now tries the named-source suffix first and falls back to the bare `-paper` rule, returning the suffix that matched so the UNPAIRED and AMBIGUOUS lines name the rule they read (`scripts/check-live-paper-config-drift.sh:275-282`). Reproduced against the pre-fix script on a live `hl-y` plus `{"id": "hl-y-paper", "paper_source": "btc", "storage_strategy_id": "hl-y"}` at 3600s against 300s: exit 0 with no PAIR line before, `DRIFT interval_seconds live=300 paper=3600` and exit 1 after. All three must-survive cases are fixtures now: the `-paper<n>` collision form under a named source pairs and gates, the bare alias under a named source with no `storage_strategy_id` reports AMBIGUOUS naming `-paper`, and an id whose bare base matches no live id still does not gate.

2. **Pairing evidence carried by `storage_strategy_id` alone is never read or reported** — Confirmed, in scope under scope rule 3 (no new mechanism; the same pairing branch). A paper block whose `storage_strategy_id` names a live strategy now pairs on it when no alias rule names a live base (`scripts/check-live-paper-config-drift.sh:265-268`). Pairing rather than only surfacing: the stored identity is the same proof the alias branch already demands, it makes the AMBIGUOUS line's own instruction ("set replay_source_id or storage_strategy_id to pair it") true, and it gives the operator the drift report instead of a note to go look. Must-survive cases are fixtures: a stored id naming a live id under an unrelated block id pairs and reports its drift, a `-paper-<id>` id with no `paper_source` key pairs through its stored id, and a stored id matching no live id stays silent at exit 0.

### Test edits

1. **The `source-unset` drift fixture** (`scripts/test_update_helpers.sh:1004-1015`) — Outdated; ground: review finding `Recommended Optional 1` above, which authorizes pairing a block whose `storage_strategy_id` names a live id even when no alias suffix fires. The paper block drops that key, so the case still asserts exactly what it names: a `-paper-<id>` id is not read as an alias without `paper_source`. Storage-id pairing moved to its own fixture pair.

Verification: `bash scripts/test_update_helpers.sh` passes with and without `GO_TRADER_BIN`; `uv run --no-sync python -m pytest shared_scripts/ scripts/` 329 passed; `go -C scheduler test ./... -run 'UpdateSh|MergePaper|PaperSource|AgentInfo|ReadState'` passes (no Go source changed in this push); `bash -n` clean on both changed shell scripts; `scripts/check_test_budget.py` 172, baseline 172.

---

richkuo:

richkuo:
Addressed review feedback (Claude Opus 5 · 2026-09-15T20:58:33Z) in cb363263bec6d00f1186c93685651563d12b2f26.

### Fixed

1. **A `-paper-<id>` id whose source is declared in the same config, but which carries neither `paper_source` nor `storage_strategy_id`, is silently certified** — Confirmed, in scope under scope rule 3 (no new mechanism; the same pairing branch, plus the declared ids carried onto each entry at `scripts/check-live-paper-config-drift.sh:230`). Reproduced on a live `hl-y` plus `{"id": "hl-y-paper-btc"}` at 3600s against 300s with `paper_sources` declaring `btc`: exit 0 and no line before, `AMBIGUOUS hl-y-paper-btc … the -paper-btc suffix alone does not prove a twin of hl-y` and exit 1 after. `declared_alias_read` (`scripts/check-live-paper-config-drift.sh:165-174`) is the last resort after the block's own `paper_source`, the bare `-paper` rule and the stored identity, and it reports only when a declared source's suffix yields a base that names a live strategy (`scripts/check-live-paper-config-drift.sh:296-302`). All three must-survive cases are fixtures: `hl-paper-rsi` in a deployment declaring a source `rsi` stays silent because `hl` names no live strategy; a `-paper-<id>` name for an undeclared source stays silent; a paper-only deployment with an empty live set prints nothing.

### Resolved judgment calls

1. **A `-paper-<id>` id whose source is declared in the same config, but which carries neither `paper_source` nor `storage_strategy_id`, is silently certified** — the finding suggested printing a line without pairing, leaving whether it gates open. It gates, through the existing AMBIGUOUS path (`scripts/check-live-paper-config-drift.sh:296-302`), because the finding's own invariant is that adopting the convention never removes a gate the bare `-paper` rule applied, and that rule gates at exit 1. Alternatives rejected: pairing on the name alone, which is the same evidence the AMBIGUOUS branch already refuses and which would certify a block the scheduler places in the default paper partition, not in `btc`; a new non-gating MISNAMED line, which would add a fourth report shape and still let a drifted fleet exit 0.

Verification: `bash scripts/test_update_helpers.sh` passes with and without `GO_TRADER_BIN`; `uv run --no-sync python -m pytest shared_scripts/ scripts/` 329 passed; `go -C scheduler test ./... -run 'UpdateSh|MergePaper|PaperSource|AgentInfo|ReadState'` passes (no Go source changed in this push); `bash -n` clean on both changed shell scripts; `scripts/check_test_budget.py` 172, baseline 172. No test was edited; every test difference adds a case.

---

richkuo:
@claude sonnet review

PR #1564, opened 2026-09-15
feat(#1561): hand off several paper deployments in one merge [C99, Opus 5, xhigh, fableplan]

richkuo:
Part of #1561. Follows #1563, which is merged. Third of the four pull requests the adopted [implementation plan](https://github.com/richkuo/go-trader/issues/1561#issuecomment-5675685971) defines, covering step 12 only: the handoff script folds several paper deployments into one live scheduler.

## Summary

`scripts/merge-paper-instance.sh` gains `--source <id>=<instance>`, repeatable. Each one folds that deployment into the partition `paper:<id>` that #1562 built: the merged config gains one `paper_sources` entry with the deployment's canonical database, every moved strategy is aliased `<base>-paper-<id>` (numeric suffix when the name is taken) and stamped `paper_source=<id>`, and the merged unit gains one drop-in `50-merge-paper-<id>.conf` that keeps that database's directory writable.

`--paper <instance>` is unchanged and may be given with any number of sources. A run with only `--paper` produces the same staged config, the same journal line names and the same drop-in name as before.

- **One fold table.** The shell builds one row per folded deployment in the runtime's partition order (default paper first, then sources by id). Preflight, the ownership locks, the fingerprint checks, compose, the proof, apply and rollback all walk it, and the Python helper reads the same table from one plan file.
- **Preflight** resolves every deployment (directory, binary on one release, config, unit stopped, `config_version` equal to live's, every strategy in paper mode, no nested `portfolio_risk.paper`). Exit 16 covers a deployment that already splits its own state or already folded sources of its own, two deployments on one canonical database, and a folded database that is the live one. Exit 17 covers a source the live config declares on another file, and a database the live config already owns under `paper_db_file` or another source's entry. A source declared on the same file is a repeat run.
- **Risk layers**: root, then `portfolio_risk.paper`, then `paper_sources[<id>].portfolio_risk`, written only where the source's effective view differs from the default paper partition's. A zero where the layer above sets a limit still refuses.
- **Channel keys follow the resolver.** A moved strategy's route is written to `<platform>-paper:<id>`, and the key is not added when the first merged key below it already carries that value. `dm_channels` keys are always added. A `-paper` key in a source deployment's own map is carried over as `-paper:<id>`, so folding a source never re-routes the default paper partition.
- **Replay mirrors, root-key conflicts and alias collisions** are checked across every folded deployment at once.
- **The proof** compares each moved strategy's effective view against its own deployment's and asserts the staged partition and `paper_source` are exactly what the fold table names; the staged storage layout is checked per role, so each source's book must map under its own alias in its own file.
- **One journal per run** names every folded deployment and carries per-deployment fingerprints. Apply installs the config, then one drop-in per deployment with its own begin and done marker; an interruption at any step restores the config and every drop-in it had installed. Rollback restores the same set and never rewrites a stored book, including after the merged process has written state.

Operator documents (SKILL.md, README.md, the cutover runbook) stay with plan steps 15 and 16 in the fourth pull request. No deviation from the plan.

## Verification

- `go -C scheduler test ./... -run TestMergePaperInstance` passes; the script prints `OK: merge-paper-instance tests passed`.
- `go -C scheduler test ./...` passes (308s). `go -C scheduler build .` and `gofmt -l scheduler` clean. No Go source changed.
- `bash scripts/test_update_helpers.sh` passes.
- `uv run --no-sync python -m pytest shared_scripts/ scripts/` — 329 passed, including the structural test that both paper tools load the shared alias rule.
- `bash -n` clean on all three changed shell scripts; `scripts/check_test_budget.py` at 172, baseline 172.
- `scripts/test_merge_paper_service_fixture.sh` now declares a `paper_sources` entry with its own state directory, so the sandbox proof covers a source directory, the lock holder covers its two lock files and the second start still exits 79. That fixture runs only on Linux as root, so CI verifies it and this machine skips it.

New fixture cases in `scripts/test_merge_paper_instance.sh`: two sources folding into their own partitions with their own aliases, channel keys, risk overrides, drop-ins and proof; the ownership locks taken in the runtime's order (primary, then each source by id); the repeat dry run and the repeat apply; rollback; rollback after the merged process wrote state; an interrupted apply at the config step and at the override step; `--paper` and `--source` in one run; two sources on one database; a source the live config declares on another database; a source deployment carrying `paper_sources` or `paper_db_file`; a root-key difference in a source; two sources mirroring one live strategy; the `--diff` preview of a source alias, its stamp, the channel key it would add and a redundant one; a taken source alias taking the numeric suffix; and the argument refusals.

### Test edits

- `scripts/test_merge_paper_instance.sh`, the complete-journal assertion — **Outdated**. Ground: plan step 12 gives the merged unit one drop-in per folded deployment, so a run installs more than two deployment files and the message can no longer say "both deployment files". The assertion now reads "every deployment file matches its result" and still proves the dry run reports a complete journal.
- No other test was changed. Every other test difference adds a case or a fixture.

## Plain simple English

The handoff tool could move one paper deployment into the live scheduler. Now it can move several in one run, and it keeps each one's trades, money limits and chat channels separate from the others. If a step fails, or an operator asks for a rollback, it puts every file back and never changes a stored trade book.

---


richkuo:

richkuo:
Addressed review feedback (Claude Fable 5.1 · 2026-09-15T23:24:35Z) in 5e56baf75a9b121ca609cad3e7d56d2847f03baa.

All six findings confirmed against the code and all in scope. No test was edited; every test difference adds a case.

### Fixed

1. **A `--source` id equal to a previously folded `--paper` instance name (or vice versa) silently replaces that partition's drop-in on a later run.** — Confirmed and reproduced. Each apply now records `fold <key> <partition> <instance>` per folded deployment (`scripts/merge-paper-instance.sh:1865`), and `scan_other_journals` (`scripts/merge-paper-instance.sh:1615-1631`) refuses at exit 24 when another complete journal already folded one of this run's keys as a different partition, naming that partition and the drop-in both runs would write. Scope rule 1: the source-id namespace this PR adds is what makes the second run legal. Must-survive cases are fixtures: a repeat run with the same arguments still re-applies as a no-op, the reverse direction (`--source a=x` first, then `--paper a`) is refused the same way, and a later run that keeps every fold key on its own partition still reaches READY.

2. **An interrupted apply is invisible to every run with a different fold set, which then certifies and applies over the half-applied state.** — Confirmed. The same scan (`scripts/merge-paper-instance.sh:1615-1621`), called before the own-journal gate (`scripts/merge-paper-instance.sh:1667`), refuses at exit 24 when any journal other than this run's holds neither `complete` nor `rolled-back`, and names the journal and the two ways out. Must-survive cases are fixtures: complete and rolled-back journals do not refuse, and the same-key resume path is untouched because the scan skips this run's own journal.

3. **Rolling back an older run after a newer run applied reverts the newer merge and strands its journal.** — Confirmed. `journal_that_installed_live_config` (`scripts/merge-paper-instance.sh:1635-1646`) refuses the rollback at exit 24 when the live config is the result another complete journal recorded, and names that journal. A config that matches no journal still takes the `archive_if_edited` path, so a hand-edited merged config rolls back exactly as before, and rolling back the newest run stays one command.

4. **Preflight and compose disagree on what counts as a repeat run.** — Confirmed. Preflight computes the verdict once (`fold_merged_flag`, `scripts/merge-paper-instance.sh:1274-1290`) and carries it in the plan file; `fold_already_merged` now reads `merged` (`scripts/merge-paper-instance.sh:608`) instead of comparing the live config's raw `db_file` string against a canonical path. A genuine database move still refuses at exit 17 in preflight, and a byte-identical repeat run still skips every already-merged book.

5. **The staged storage proof counts every book in a partition file against only this run's fold, so repeat runs falsely refuse once the merged process has run.** — Confirmed. The staged check now expects every strategy the staged config places in that partition, via the new `partition-count` helper (`scripts/merge-paper-instance.sh:676-686`, used at `scripts/merge-paper-instance.sh:1773`), instead of this run's `ids + skipped`. A book with no configured owner is still refused by the orphan check, which the existing invalid-storage-mapping fixture holds at exit 20.

6. **Rollback demands full preflight over every folded deployment it does not need.** — Confirmed. The deployment, binary, version, config-copy and storage-inspect preflight now runs for a merge only (`scripts/merge-paper-instance.sh:1333`, `:1356`, `:1389`); a rollback resolves each folded database from its config when that config is still there and names it when it is gone (`scripts/merge-paper-instance.sh:1399`). Fixtures cover rollback with the deployment deleted, rollback with the config deleted too, and lock contention still refusing at exit 3.

### Resolved judgment calls

1. **A `--source` id equal to a previously folded `--paper` instance name (or vice versa) silently replaces that partition's drop-in on a later run.** — The finding's suggested remedy was to refuse when the existing drop-in's content differs from the staged override. That rule would refuse a case the script supports and a fixture holds: a drop-in that existed before any merge (`[Service]\nNice=5`) is legitimately replaced and retained for the rollback, and its content differs by definition. I refused on journal ownership instead (`scripts/merge-paper-instance.sh:1622-1630`), which is evidence about the hazard itself — another applied merge owns that drop-in name for another partition — and leaves every legitimate replacement working. Alternatives rejected: content comparison, which breaks the pre-existing drop-in contract; refusing every reused fold key, which would also refuse the legitimate run that adds a source to an applied merge.

2. **Rolling back an older run after a newer run applied reverts the newer merge and strands its journal.** — The finding asked for a refusal "behind an explicit override". I refused without adding an override flag, because the evidence identifies the newer run by name and the remedy is one command (roll that run back first), so an override would only offer a way to reach the outcome the finding calls wrong. The discrimination is the newer journal's `result_config` fingerprint rather than "matches none of this journal's fingerprints", which keeps the hand-edited-config archive path working with no flag and no test edit. Alternatives rejected: a `--force-rollback` flag, which adds a surface whose only use is the damaging path; refusing on any fingerprint mismatch, which would break the hand-edit case the finding requires to survive.

Verification: `go -C scheduler test ./... -run TestMergePaperInstance` passes; `go -C scheduler test ./...` passes; `bash scripts/test_update_helpers.sh` passes; `uv run --no-sync python -m pytest shared_scripts/ scripts/` 329 passed; `bash -n` clean on both changed shell scripts; `scripts/check_test_budget.py` at 172, baseline 172. No Go source changed in this push.

---

richkuo:

richkuo:
Addressed review feedback (Claude Opus 5 · 2026-09-16T00:04:26Z) in bc0eae31b0112e6896e9ec4fb3c9bb0599558544.

All five findings confirmed against the code and all in scope under rule 1: every one of them lives in code this PR adds. No follow-up issue was filed.

### Fixed
1. **Stacked merges share one retained drop-in copy, so the second rollback in the required order is permanently refused.** — the retained drop-in path now carries the whole key set (`scripts/merge-paper-instance.sh:233-240`), and `restore_from_retained` falls back to the old `.pre-merge` name only for a journal that records no folds (`scripts/merge-paper-instance.sh:1599-1601`). Rule 1: this PR turns `MERGE_KEY` into a joined key set, so two runs can share a fold key and own different journals; on main the journal name and the fold key were the same string.
2. **A merge applied by an earlier release leaves no fold record, so the drop-in collision the new scan exists to refuse still happens.** — `journal_fold_partition` now derives the key from the file name and the partition `paper` for a journal that carries no `fold` line (`scripts/merge-paper-instance.sh:1654-1660`), helped by `journal_is_legacy` and `journal_legacy_key` (`scripts/merge-paper-instance.sh:1550-1558`). Rule 1: the scan is this PR's own mechanism.
3. **A rollback cannot lock a database inside a deleted deployment directory and reports that as lock contention.** — rollback now skips a folded database that is absent, locking and fingerprinting neither, the way it already skips a missing config (`scripts/merge-paper-instance.sh:1425-1431`).
4. **An out-of-order rollback still reverts a newer merge when that merge was interrupted.** — the interrupted-journal scan is split into `scan_other_journals_interrupted` and every mode runs it, rollback included (`scripts/merge-paper-instance.sh:1665-1676`, `scripts/merge-paper-instance.sh:1709`).
5. **A pruned source channel key is re-routed by any later merge that adds the plain `-paper` key.** — a `<platform>-paper:<id>` key is pruned only when `<platform>-paper` already carries the value, so no later fold can slip a key between it and the bare platform key (`scripts/merge-paper-instance.sh:598-605`). Rule 1: main pruned only the bare `-paper` target; this PR extended the prune to source targets.

### Test edits
1. **`== --diff previews a source alias, its channel key and a redundant one`** (`scripts/test_merge_paper_instance.sh:1289`) — Outdated; ground: finding 5 of this review and the invariant it states, that once a source partition's channel routing is proven a later fold does not change it. It now asserts `--diff` keeps `discord.channels.hyperliquid-paper:eth=C-live` instead of pruning it.

New fixtures, none of them replacing an existing one: stacked applies that share a fold key rolled back newest first over a drop-in that existed before any merge; a repeat apply of one stacked run; a legacy-format journal that owns its drop-in name, with the same-partition and rolled-back legs; a rollback refused by another journal's interrupted apply, with the complete and rolled-back legs; a rollback of the interrupted run itself; a rollback whose folded database sits inside a deleted deployment tree; and the two channel-prune cases (pruned when `<platform>-paper` pins the value, kept and still correct after a later paper fold).

Verified: `go test ./...` in `scheduler/`, the paper handoff shell suite (`TestMergePaperInstance`, `TestUpdateShellScriptSyntax`) and `scripts/test_update_helpers.sh` all pass. The combined-unit service fixture needs root and runs in CI.

---

richkuo:
@claude review

PR #1565, opened 2026-09-16
feat(#1561): navigate the dashboard by paper source [C72, Opus 5, high]

richkuo:
## Plain simple English

One production scheduler can now hold live trading and several paper deployments. The dashboard had no way to look at one of them alone. This change adds a partition selector. Each panel then shows only that partition's strategies, ranking, trade history and risk. Cash flow stays with live, because only live has it.

## Summary

Adds the dashboard source navigation from issue #1561's fifth checklist item. Partitions 1562 built in the runtime now reach the operator surfaces.

- **Query key.** `partition`, taking the runtime's own partition text: `live`, `paper`, `paper:<source id>`. Absent = every partition, the behaviour every caller had before.
- **Invalid key.** Unparseable text returns 400. A well-formed partition this process does not own returns 404, so a stale selector cannot read a neighbour deployment's view.
- **Selector roster.** `/status` gains `partitions`: partition, label, scope and source for every owned partition in the stable order (live, default paper, then sources by id). An unlabelled source reads as `paper:<id>`, never the bare id, which the selector could not tell from the default paper partition.
- **Filtered endpoints.** `/api/strategies`, `/api/strategies/overview`, `/api/leaderboard`, `/api/strategies/dead`, `/api/diagnostics`.
- **Filter before pagination and totals.** `TradeDiagnosticsRowsPageForPartition` resolves the owning file by storage role, never by scope, and counts only that partition's ids, so the total matches the pages the operator can walk. A `strategy` query outside the selected partition returns an empty page.
- **Ownership preserved.** Cash flow stays live-owned: `live_owned: true`, and a paper partition gets `available: false` with a reason rather than the live wallets under a paper label. The close-evaluator catalogue stays shared and is labelled `shared: true`. Correlation and portfolio risk were already keyed by partition in 1562; the front end now reads only the selected key.
- **Front end.** Every dashboard read carries the selection. The control stays hidden while the process owns fewer than two partitions, so an install with a single partition sees no change; a live plus default-paper install owns two and does get the selector. Any 404 of a partitioned read falls that read back to every partition, so sibling reads in one refresh recover together. A reply that lands after the selection changed is dropped rather than drawn under the new selection. A selected strategy the new partition does not own is cleared rather than left showing another partition's detail.
- **Unchanged:** authentication, the loopback bind, and the `mu > strategiesMu` lock order.

### Verification

- `go test ./...` in `scheduler/` passes (464s).
- `go vet ./...`, `gofmt` and `scripts/check_test_budget.py` pass.
- New tests in `scheduler/ui_partition_test.go`: per-endpoint partition selection (table-driven over all/live/paper/source), 400 and 404 refusals, diagnostics total and page scoped to one partition, a cross-partition `strategy` query, cash flow staying live-owned, `UIStrategy` carrying its partition, and the `/status` roster order and label fallback.

Part of #1561. Does not close it: the cutover rehearsal, retiring the folded units and the operator-document sync remain.

---

https://claude.ai/code/session_01B3gxvvkNNPgCSuoxsx7q6Y


richkuo:

richkuo:
Addressed review feedback (claude-fable-5 · 2026-09-16T02:05:25Z) in 1b41291558d20b0fd00e69842306de5207254b69.

All five Recommended Optional findings traced to current code, confirmed, and fixed in this PR. No inline threads were open and every CI check passed on the reviewed head.

### Fixed

1. **With a partition selected, the correlation panel can render the legacy scope's snapshot untagged when the selected partition has no by-scope entry.** — the untagged `resp.correlation` fallback now runs only under the all-partitions selection, so a selected partition with no snapshot of its own falls through to the panel's "No correlation snapshot yet" empty state (`scheduler/static/ui/app.js:2461`). The sibling in `refreshRiskPanel` is guarded the same way and pushes "No portfolio risk reading for this partition yet" instead of the legacy row (`scheduler/static/ui/app.js:2228-2233`). A third path that made the risk-panel fallback reachable is closed at the source: `renderPartitionSelect` now resets the selection to every partition when it hides the control, so a selection the operator can no longer see cannot keep scoping the panels (`scheduler/static/ui/app.js:418-424`).

2. **The risk panel filters the kill-switch row by the selected partition but lists per-strategy circuit-breaker rows from every partition.** — the circuit-breaker and pending-circuit-close rows now filter on the row's own `partition` (`scheduler/static/ui/app.js:2248-2250`). The reviewer's state (2) is handled server-side rather than by guessing on the client: `/status` names a strategy's partition only when the roster still configures it, so a state row with no config row carries no `partition` and `partitionRowVisible` shows it under every selection (`scheduler/server.go:433-439`, `scheduler/server.go:476`, `scheduler/static/ui/app.js:405-410`). Under the auto-protective carve-out a protective alert is never hidden by the filter; the previous code would have filed such a row under `paper`, because `partitionFor` reads a zero config as paper (`scheduler/risk_partition.go:87-92`).

3. **Partition-scoped fetches carry no request generation, so a response from the previous selection can render under the new one.** — `getJSONForPartition` now returns the selection each response was fetched for, and every panel drops its render when that selection is no longer current (`scheduler/static/ui/app.js:375-399`; call sites at `app.js:456`, `2192`, `2336`, `2366`, `2401`, `2499`). Comparing the selection rather than a counter keeps the 404 retry correct: the retry re-reads bare and reports `all`, which is the selection it actually served, so the fallback's own render is never dropped.

4. **When the selected partition is retired, only the first concurrent 404 falls back; sibling requests rethrow and paint one error cycle.** — any 404 of a partitioned read retries the bare URL regardless of the filter's current value, so every sibling inside one `Promise.all` recovers in the same cycle (`scheduler/static/ui/app.js:383-389`). The reset to `all` is applied only by the read whose own selection is still current, so a concurrent selector change is not clobbered.

5. **The README paragraph contradicts itself and the code on when the selector appears.** — the lead sentence now states the two-or-more-partitions condition the code implements and names the live plus default-paper case explicitly (`README.md:331`). The PR description's front-end bullet is corrected the same way. `SKILL.md:1012` gains the new invariants: no legacy-scope fallback under a selection, `partition` named only for a configured strategy, the hidden-selector reset, the any-404 retry, and the stale-response drop.

### Test edits

None. One test added: `TestStatusNamesNoPartitionForAnUnconfiguredStateRow` (`scheduler/ui_partition_test.go:311`) asserts `/status` omits `partition` for a state row the roster does not configure while still naming `live`, `paper` and `paper:btc` for the configured ones.

Verification: `go -C scheduler build ./...`, `go -C scheduler vet ./...` and `go -C scheduler test ./...` all pass (full suite, 459s); `gofmt -l scheduler/` is clean; `node --check scheduler/static/ui/app.js` passes; `uv run --no-sync python scripts/check_test_budget.py` passes at 172 wording-only tests against a baseline of 172.

---

richkuo:
@claude sonnet review

PR #1566, opened 2026-09-16
docs(#1561): operator guidance and backup discovery for several paper sources [C38, Opus 5, high]

richkuo:
## Plain simple English

Three merged changes let one scheduler run several paper deployments. The operator documents still described one paper file and one merge. This change updates them: which files a backup copies and in what order, which settings need a restart, how to fold several deployments, and the journal line after which you must roll back.

## Summary

Closes the operator-document and backup-discovery item on #1561. Documents only; the code it describes is already merged in #1562, #1563 and #1564.

**SKILL.md § Storage Ownership**

- `paper_sources` and `paper_source`: the id rule (`[a-z0-9][a-z0-9_-]{0,31}`, never `live`, `paper` or `primary`) and the four refusals — a live strategy carrying a source, a source no entry declares, a duplicate id, and a nested `portfolio_risk.paper`.
- The **risk partition** (`live`, `paper`, `paper:<id>`) as the runtime identity that owns limits, latch, correlation and state file; one file per partition; the lock and save order primary, paper, then sources by id.
- **Backup and restore** over every retained file with its `-wal` / `-shm` sidecars, restored in that same order. `storage-inspect --json` is named as the discovery command, and `update_resolve_db_exclude` as the updater's copy of the same list, with the warning that a new state path must be added there.
- **Service writable paths**: one drop-in per folded deployment, `StateDirectory=` for a `/var/lib` path and `ReadWritePaths=` for a deploy-tree one.
- **Access mapping**: a folded unit's port and route are retired after proof; a separate paper-testing deployment keeps its own.
- **Journals and rollback order**: one journal per run keyed on the whole fold set, the ownership scan every mode runs, legacy journals that carry no fold line, per-run retained copies, and the newest-first rollback rule.
- A **cutover checklist** naming the recovery boundary at `config begin`, the journal line the apply writes before it retains the live config and installs the staged one: before that line nothing outside the staging area and the journal has changed; from it on the merged config may already be installed even while `config done` is missing, and only `--rollback` or a re-`--apply` returns the deployment. It also states that `VERDICT: APPLIED` is not proof that trading resumed, and that a rollback never restores a database.
- The drift audit's source-aware alias pairing from #1563.
- The restart-required list and the config table gain `paper_sources`, its `db_file`, each strategy's `paper_source`, and each source's `portfolio_risk.max_notional_usd`.

**README.md**

- `--source <id>=<instance>` and combined runs in the folding section, plus the newest-first rollback rule.
- `paper_sources` and `paper_source` rows in the split-state-file table, retitled for source files.
- Reset and restore rows now name every configured state file and the restore order.
- Two troubleshooting rows: which files a backup needs, and a folded unit restarting against a database the merged process owns (exit 79).

### Verification

Every claim was checked against the code before it was written: `paperSourceIDPattern` and `reservedPaperSourceIDs` (`scheduler/config.go:2272`), the four source refusals (`scheduler/config.go:2286`), the restart-required rejections (`scheduler/config_reload.go:1149`), the same-physical-file refusal and `ExitStorageOwnership = 80` (`scheduler/storage_layout.go:281`, `scheduler/exit_codes.go:9`), `update_resolve_db_exclude` enumerating `paper_sources[].db_file` (`scripts/update_helpers.sh:288`), the drift audit reading `paper_source` (`scripts/check-live-paper-config-drift.sh:120`), and the merge tool's own usage text and journal rules.

The recovery boundary was re-checked after the first review: the apply appends `config begin` (`scripts/merge-paper-instance.sh:1960`), retains the live config (1961), installs the staged one over it (1962) and appends `config done` only afterwards (1963), while `restore_from_retained` keys on either marker (`scripts/merge-paper-instance.sh:1579`). Step 5 now names `config begin`.

`go test -run 'TestAgentInfo|TestBootstrap'` passes. No test edits: this changes documents only.

Part of #1561.

---

https://claude.ai/code/session_01B3gxvvkNNPgCSuoxsx7q6Y


richkuo:
@claude review

richkuo:
Addressed review feedback (Claude Opus 4.8 (1M) · 2026-09-16T02:07:29Z) in a3c14ba4.

### Fixed

1. **The cutover checklist names `config done` as the recovery boundary, but the live config is installed one step earlier, so a crash-window journal without `config done` can still need rollback.** — Confirmed against the code and fixed. The apply appends `config begin` (`scripts/merge-paper-instance.sh:1960`), retains the live config (`scripts/merge-paper-instance.sh:1961`), installs the staged config over it (`scripts/merge-paper-instance.sh:1962`) and appends `config done` only afterwards (`scripts/merge-paper-instance.sh:1963`); `restore_from_retained` triggers on `journal_has "config done" || journal_has "config begin"` (`scripts/merge-paper-instance.sh:1579`) and restores the copy taken at 1961. `SKILL.md:853` step 5 now names `config begin` as the recovery boundary, states that a journal carrying `config begin` without `config done` may already hold the merged config, and tells the operator not to start the merged unit or set such a journal aside because `config done` is missing. Both documented recovery commands (`--rollback`, re-`--apply`) already work in that window, and the text says so.

Two claims the fix touches were verified before they were written, neither of them challenged by the review:

- Each drop-in follows the same rule from its own `override begin` line, and the restore keys on it the same way (`scripts/merge-paper-instance.sh:1970`, `scripts/merge-paper-instance.sh:1602`). The suffix is empty for `--paper` and `.<id>` for a source (`scripts/merge-paper-instance.sh:172`), so step 5 writes `override begin.<id>` for a source.
- A crash after `config begin` but before the retain completes leaves no retained copy and an untouched live config; `restore_from_retained` is then a safe no-op, because the missing-copy CRITICAL path is gated on `config done` alone (`scripts/merge-paper-instance.sh:1590`).

The PR body was updated to match: its cutover-checklist bullet and its verification section now name `config begin`, and the `## Plain simple English` block was trimmed under 55 words.

### Test edits

None. This change edits `SKILL.md` only.

### Verification

`go -C scheduler build ./...` and `go -C scheduler test -count=1 -run 'TestAgentInfo|TestBootstrap' .` both pass. No Go, Python or shell source is referenced from `SKILL.md`, so no other suite is in play.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (Claude Sonnet 5 · 2026-09-16T02:28:48Z) in 50d66b2dfbb9379a35bb85a7b9d763ba80e7f4e5.

One finding, confirmed against the code and in scope under rule 1: the row lives in this PR's own diff.

### Fixed
1. **The new "folded unit restarts" troubleshooting row says the live unit exits 79, but the code and this PR's own SKILL.md text say the restarting (folded) unit is the one that fails.** — confirmed and reworded (`README.md:506`). Traced independently: `acquireStateOwnership` is called by each starting scheduler and only the caller that cannot take the lock reaches `os.Exit(ExitSingletonLock)` (`scheduler/main.go:176-181`), with the refusal message "refusing to start so this process can't double-trade against the same state files" (`scheduler/state_store.go:631`) and `ExitSingletonLock = 79` (`scheduler/exit_codes.go:5`). The holder is never signalled, so the live unit that already owns the file keeps trading. The row now states the rule in the direction the code enforces — whichever scheduler starts **second** refuses to start — tells the operator to read the log of the unit that exited and to leave the running one alone, and keeps the folded-unit restart as the usual cause. Stating the rule rather than one direction of it also covers the reverse incident, a live unit started while a folded unit is still running, which the old row could not describe at all.

Checked that the same misattribution appears nowhere else: `grep` over `README.md`, `SKILL.md` and `docs/` finds the one occurrence. `SKILL.md:855`, added by this PR, was already correct, and the pre-existing `SKILL.md:843` paragraph describes the handoff case correctly.

### Test edits
None. This commit changes one line of one document, so no Outdated, Wrong or Obsolete case arose.

Verified: `go -C scheduler build ./...` and `go -C scheduler test -run 'TestAgentInfo|TestBootstrap'` pass. No Go or Python source reads `README.md`, and the change touches no code, so no further suite applies. All four CI checks were passing on the reviewed head, so there was no CI finding.

---

richkuo:
@claude review

